PR #108, opened 2014-06-12
Added sharding support to DynamoDBSessionStateStore to allow for sessions larger than 64k

jeffersonjhunt:


normj:
Since Item size has increased to 400 KB this seems unnecessary.

PR #121, opened 2014-08-08
do not specify the encrypted flag if specifying a snapshot id

xeshxzh:
Otherwise, we get:
Amazon.EC2.AmazonEC2Exception: Parameter encrypted is invalid. You cannot specify the encrypted flag if specifying a snapshot id in a block device mapping. ---> System.Net.WebException: The remote server returned an error: (400) Bad Request.
   at System.Net.HttpWebRequest.GetResponse()
   at Amazon.Runtime.AmazonWebServiceClient.getResponseCallback(IAsyncResult result)
   --- End of inner exception stack trace ---
   at Amazon.Runtime.AmazonWebServiceClient.HandleHttpWebErrorResponse(AsyncResult asyncResult, WebException we)
   at Amazon.Runtime.AmazonWebServiceClient.getResponseCallback(IAsyncResult result)
   at Amazon.Runtime.AmazonWebServiceClient.endOperation[T](IAsyncResult result)

This exception happens on .NET SDK 2.1.1 and above, and it doesn't happen on 2.1.0 and below
Thanks.


PavelSafronov:
This was fixed in version 2.3.1 of the SDK.

PR #145, opened 2014-11-01
Fix download on Linux if filepath contains directory

ermshiperete:
This change fixes a problem on Linux with downloading files.

On Linux we shouldn't replace forward slashes with backslashes because the forward slash serves as directory separator.


normj:
Sorry this got dropped. I have merged in the changes and it will go out with the next release.

normj:
This changed was released today with version 2.3.34 of the SDK.

PR #164, opened 2015-02-20
Adding option to disable ResetItemTimeout calls when session is disabled

djluck:
This change adds the ability to suppress the session "keepalive" message(s) that are sent to dynamo when session is disabled for a request.

For those unaware, even if you explicitly disable sessions for a page/view/handler/etc., the ASP.NET session state module will request that the session is kept alive. This results in 1-2 dynamo db requests per request. By disabling this functionality, consumers of this API would be able to reduce unnecessary traffic to dynamo db.

This does not remove the session keepalive functionality from the DynamoDBSessionStateStore. When obtaining an exclusive lock on a session item, the "Expires" attribute is updated then. Keepalive is only disabled when session has been disabled for a request.

To illustrate how useful this functionality this could be, I'll describe the situation that we are in. We have a request being initially served by a MVC controller with 5-10 http handlers providing content that is loaded in as a script, image or via AJAX. Each handler is issuing a dynamo db request which means for every visit to a page we have 5-10 unnecessary dynamo db API calls. Removing these redundant dynamodb calls would help us reduce our (significant) dynamo db traffic.


normj:
This is great. Thanks for sending us the pull request. It will be pushed to NuGet as part of our next SDK release.

djluck:
Thanks for the quick merge and publication normj, we really appreciate it.

PR #177, opened 2015-03-12
Add IAsyncSearch<T> to DynamoDBContext to support mocking

samritchie:
If you mock IDynamoDBContext for unit testing, you can’t test async scan or query operations as it’s not possible to create an AsyncSearch<T> from client code (the constructor is internal).

I’ve added an IAsyncSearch<T> interface to the .NET 4.5 (task-based async) DynamoDBv2 library, and changed the async query/scan method signatures.


PavelSafronov:
The 4.5 changes look good, but I think IAsyncSearch interface and the corresponding Begin/End calls should also be included in the 3.5 solution. Otherwise this is a strangely-specific change to only support 4.5 mocking.

samritchie:
Okay, I’ve added the interface for 3.5, although I haven’t tried to mock on this version and I don’t know if it works.

PavelSafronov:
I'm sorry, I overlooked something in the original review: you are changing the return types on `DynamoDBContext` and `IDynamoDBContext`, which is a breaking change. Because of that we won't be able to accept this pull request in its current form.

samritchie:
Hmm - the alternative is probably going to be putting explicit testing types in the assembly?

samritchie:
I’ll open a separate PR

PR #181, opened 2015-03-17
Support Mocking Query/Scan operations on DynamoDBContext

samritchie:
Okay, I’ve had another bash at this:

* added a public default constructor on AsyncSearch
* all public members on AsyncSearch are now virtual

This allows mocking frameworks to generate mock AsyncSearch instances to test query & scan operations, without changing the context method signatures.


samritchie:
Any feedback as to what's wrong with this one?

Issue #158, opened 2015-01-05
getpresignedurl + apostrophe (possible bug)


foxontherock:
We are storing our files in S3 exactly like they were on our disk drive. Some of them have special characters, unicode, or anything else. We do not have any trouble with all of them, except one: the apostrophe ('), single quote.

That character is not url-encoded like others do when using GetPreSignedURL function. 

The generated url works fine in IE and Chrome, but not in firefox that re-encode it, and the signature does not match. In the sample below, the trouble is not in the content-dispoisition header, but in the key. I can change my content-disposition filename to fix half of it, but I can't change my key.

Sample (FAIL, single quote) - not working in firefox because FF replace ' with %20 automatically
   Key: ace48a/49277_fred's image.jpg
   url: http://(bucket).s3.amazonaws.com/ace48a/49277_fred's%20image.jpg?AWSAccessKeyId=<Access Key Removed>&Expires=1416405527&Signature=KDahwlD72RbAV1P%2BZxCdMhH%2FngA%3D
Single quote NOT replaced in url.

Sample 2: (WIN, with double quote)
   Key: ace48a/49277_fred"s image.jpg
   url: http://(bucket).s3.amazonaws.com/ace48a/49277_fred%22s%20image.jpg?AWSAccessKeyId=<Access Key Removed>&Expires=1416405464&Signature=2OUOFRWvhdNJSqGtTvnjttDnZlU%3D
Double quote REPLACED by %22 in url.

bootleg224:
Looks like Firefox  35 and onwards started escapaing characters like ' and [ and ].

 https://github.com/aws/aws-sdk-net/blob/6c5b5a7c61ba1dfc17bd2b7e39813b37b77aa9f7/AWSSDK_DotNet35/Amazon.Util/AWSSDKUtils.cs

This method likely needs to be updated

internal static string UrlEncode(int rfcNumber, string data, bool path)

bootleg224:
Looks like DetermineValidPathCharacters should have ' [ and ] removed?

jimfl:
Hi. This is a tricky one. It looks like this is an unfortunate interaction between Firefox uri encoding and the signature verification in the S3 service. I did some tests, and it turns out that if you enable Sigv4 when creating pre-signed URLs work from both browsers. It's not clear that we wouldn't break existing consumers of the API by fixing it in the SDK.

You can enable Sigv4 signing with

[...]

PR #192, opened 2015-03-30
Remove ' [ and ] from base path chars

bootleg224:
Potential fix for this issue: https://github.com/aws/aws-sdk-net/issues/158


bootleg224:
Looks like my fix breaks the RFC convention, but it only impacts URLs of this type: 
image

Not sure what the best solution is.

bootleg224:
I actually don't see an issue with breaking the RFC -- because the [ ] URLs used will never be possible, since the GetPresignedUrl returns an S3 URL, not an IP address.

I can't find any examples of where ' as path seperator

Issue #197, opened 2015-04-10
S3 GetObject request doesn't support third-party S3-compatible storages


nikitozz:
There are third-party storage providers that are compatible with S3 SDK, for example: _Google Cloud Storage_, _Ceph Object Store_, etc. Object download requests issued by AWS .NET SDK always fail with all these providers.
`AmazonS3KmsHandler` evaluates if _SigV4_ should be used for all `GetObject` requests. As a part of evaluation, it converts an endpoint to an `AmazonS3Uri` instance to get a region. Since third-party storage providers don't provide an endpoint URL that starts with _"s3"_, this conversion fails and makes the request to fail.
`AmazonS3KmsHandler` should skip the _SigV4_ check for non-Amazon endpoints.

bpendharkar:
I have been encountering a similar problem where the Third Party Endpoint does not contain 's3' in its URL and a call to 'Get-Object' fails in AmazonS3Uri constructor.
It would be good to have an overloaded constructor with an additional boolean argument as 'isThirdParty' or something similar. Accordingly . the Uri parser would/won't check for a valid S3 URI

steveataws:
Change shipped in SDK version 2.3.32.0. Thanks for the pull request.

PR #200, opened 2015-04-15
Adding support for third-party S3-compatible storage providers in GetObject request

nikitozz:
A fix for [#197](https://github.com/aws/aws-sdk-net/issues/197).
`AmazonS3Uri.IsAmazonS3Endpoint` method is introduced to check if a URL is an Amazon S3 endpoint.
The method is used in `AmazonS3KmsHandler.EvaluateIfSigV4Required` to perform the evaluation only if a `GetObject` request endpoint is an Amazon S3 endpoint, and skip the evaluation for other providers.


steveataws:
Change shipped in SDK version 2.3.32.0. Thanks for the pull request.

PR #222, opened 2015-07-22
Do not use parent directory of "My Documents" folder to locate User directory in DetermineCredentialsFilePath()

atp1122:
Use HOME Environment Variable to locate the User directory in case "My Documents" folder has been moved.

If the user has moved their "My Documents" folder, then its "parent" directory will not be the correct location of the aws credentials.


normj:
Thanks for the pull request. I merged in the changes with some extra checks in case the HOME environment variable isn't set. The change will be published with our next release.

atp1122:
Great, thanks.

Issue #240, opened 2015-09-02
SignatureDoesNotMatch exception when using custom service endpoint that has a path component.


rafd123:
Here's a quick snippet that demonstrates how one can reproduce the issue; note that the service URL has a path component:

    var clientConfig = new AmazonEC2Config
    {
        ServiceURL = "http://somehost/somepath"
    };

    var ec2Client = new AmazonEC2Client(
            new BasicAWSCredentials("access_key_here", "access_key_secret_here"),
            clientConfig);

    var response = ec2Client.DescribeAccountAttributes();

The root cause appears to be that when AWS4Signer goes to canonicalize the request, it strictly uses IRequest.ResourcePath, which is null most of the time; in some cases, it gets hardcoded to "/". As a result, the canonicalized request has the wrong canonical URL which ultimately leads to the request being signed incorrectly.

rafd123:
@PavelSafronov I'm a little confused why this is labeled as a feature request when the issue seems to be a violation of the [Signature Version 4 spec](http://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html):

> The canonical URI is the URI-encoded version of the absolute path component of the URI—everything in the URI from the HTTP host to the question mark character ("?") that begins the query string parameters (if any).

> Normalize URI paths according to RFC 3986 by removing redundant and relative path components. Each path segment must be URI-encoded. If the absolute path is empty, use a forward slash (/). In the example request, nothing follows the host in the URI, so the absolute path is empty.

Additionally, other AWS SDKs (e.g. boto3) handle this case with no issues.

PavelSafronov:
Currently, the .NET SDK does not access any AWS services which use subpaths in their endpoints, so I judged that the lack of this feature was not strictly a bug but rather a feature. That being said, I have no objection to labeling this as a bug if the lack of this functionality is causing issues for you.

steveataws:
PR merged and updated core version 3.1.1.1 released, closing.

PavelSafronov:
@rafd123 it looks like this commit has broken a [fairly specific scenario](https://github.com/aws/aws-sdk-net/issues/259), so I'm looking at reworking the approach we use to access a service with a path component. Can you provide some information on how you are accessing a service with a path component? I'm trying to figure out how to test your specific scenario so I can verify that my alternative approach doesn't end up breaking you.

PavelSafronov:
[...]

PR #242, opened 2015-09-02
ISSUE #240: Fix signing issues for endpoints with a path component

rafd123:
In order to properly preserve the relationship between IRequest.Endpoint and IRequest.ResourcePath, have, AWS3Signer and AWS4Signer canonicalize the URL returned by AmazonServiceClient.ComposeUrl()

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS3Signer.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS3Signer.cs
@@ -129,9 +129,7 @@ namespace Amazon.Runtime.Internal.Auth
             }
             else
             {
-                Uri url = request.Endpoint;
-                if (!string.IsNullOrEmpty(request.ResourcePath))
-                    url = new Uri(request.Endpoint, request.ResourcePath);
+                Uri url = AmazonServiceClient.ComposeUrl(request);
 
                 stringToSign = request.HttpMethod + "\n"
                     + GetCanonicalizedResourcePath(url) + "\n"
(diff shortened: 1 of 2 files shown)


steveataws:
Thanks for the PR. We're running it through our integration test suite and will follow up later.

rafd123:
Thanks, @steveataws 

I have to admit, going into this, I didn't fully understand the intent of IRequest.ResourcePath as it relates to IRequest.Endpoint.

The closer I look at the code, I'm wondering if the signers should be using AmazonServiceClient.ComposeUrl() as a means to resolve the URL that ultimately gets canonicalized.

Do you agree? If so, the PR should be revised.

steveataws:
Hi,

When I ran the integration tests we had a couple of fails in a test which validate some S3 keys we've had issues with in the past. You can find the test class in sdk\test\IntegrationTests\Tests\S3\KeyNameTests.cs called TestKeyNameWithAwkwardChars_AWS4Signing if you want to take a look.

The two keys that triggered fails are: 

ObjectWith!and?\+forgood:measureThis=And&InKeynÄme

and 

ObjectWith\InKeyname

In the first case, the use of Uri.AbsolutePath returns 'ObjectWith!and', terminating on the ? character. In the second case the Uri class is encoding the \ before we get a chance to work with the url in canonicalization and as a result we end up working with a double-encoding (this has been a common problem for use with use of the Uri class, which is why we tend to prefer string manipulation for urls for as long as possible).

Do you want to take another look at the fix in light of these two keys? The rest of the integration test suite ran without any issues.

rafd123:
Will do!

rafd123:
I addressed the problematic keys by grabbing the unescaped path from the composed request URL and feeding it to the original logic of "CanonicalizeResourcePath."

I ran the failing integration test using my account and verified that it now passes.

Let me know! :)

steveataws:
OK, will do. Thanks!

steveataws:
Hi,

I've completed running all of the integration tests against the revised update and all seems good. We'll get started on incorporating the changes.

Thanks for the contribution!

rafd123:
Sweet! This is how I feel about being an official contributor: https://www.youtube.com/watch?v=-7aIf1YnbbU

steveataws:
PR is merged into core version 3.1.1.1.

Issue #302, opened 2016-01-29
'=' in secret key causes truncation by credentials parser


L0G1C:
Hello,

So I have a aws_secret_access_key in a credentials file that ends with the characters "==".

When I put my credentials in AWS Explorer, or 'hard code' them into my AmazonS3Client config, everything works. But when I try to pull from the credentials file, I find that the .net sdk truncates the two "=" operators. 

I decompiled the StoredProfileAWSCredentials.cs I noticed the following:

` private class CredentialsFileParser
    { ... private const string keyValueSeparator = "="; ...}
`

The KeyValueSeparator being set to "=" makes me think that my secret key is being truncated by throwing off the parser?

I end up with a truncated secret key which fails to authenticate with the client.

normj:
I checked with the auth team and they said AWS doesn't vend AWS secret key that have "=" characters. Are you sure you are working with an AWS secret key?

L0G1C:
Oh I see. We are using an "S3 compatible" storage, not actually an AWS s3.

PavelSafronov:
We're looking into updating the credentials parser implementation to support your usecase. Will update this issue when this change is made.

L0G1C:
Do you guys accept pull requests from others if I want to work on the change on my end and send you my solution incase you want to use it?

PavelSafronov:
Absolutely. Feel free to submit a PR and we'll review it.

PavelSafronov:
The PR for this feature has been accepted. Closing.

L0G1C:
Just curious, how once PR has been accepted how long before I see it on the nuget package for the AWS SDK? What's the software deployment cycle like?

PavelSafronov:
The NuGet package should be updated today. I'll comment here when it is.
Mostly, we update the packages when there is a service update, a major bug is found, or if someone requests it.

PavelSafronov:
[AWSSDK.Core NuGet package 3.1.4.4](https://www.nuget.org/packages/AWSSDK.Core/3.1.4.4) contains this change. You should now be able to update the core package in your VS project to the latest version.

L0G1C:
Man, this sucks. I got version 3.1.4.4 and see that the change to the split function is there. But somehow after using StoredProfileAWSCredentials I still wind up with the _wrapped credentials "secret key" value missing the 2 "=" at the end of the string. =/

PavelSafronov:
[...]

PR #306, opened 2016-02-02
Fix for issue #302 - Adding whitespace to keyValueSeparator

L0G1C:
Adding a space before and after the "=" string makes the split call on line 559 match exactly " = ". 
This prevents splitting of secret keys with "=" in the value.

**Note: This assumes that credentials files always match the AWS documentation for credentials files (note the space between "=" and the key values):**
aws_access_key_id = <key>
aws_secret_access_key = <key>

[Issue](https://github.com/aws/aws-sdk-net/issues/302)
[Credential file doc](http://docs.aws.amazon.com/AWSSdkDocsNET/latest/V2/DeveloperGuide/net-dg-config-creds.html)

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/AWSCredentials.cs
+++ b/sdk/src/Core/Amazon.Runtime/AWSCredentials.cs
@@ -556,7 +556,7 @@ namespace Amazon.Runtime
             private static List<string> SplitData(string line)
             {
                 var split = line
-                    .Split(new string[] { keyValueSeparator }, 3, StringSplitOptions.None)
+                    .Split(new string[] { keyValueSeparator }, 2, StringSplitOptions.None)
                     .Select(s => s.Trim())
                     .Where(s => !string.IsNullOrEmpty(s))
                     .ToList();
--- a/sdk/src/Core/Amazon.Runtime/AWSCredentials.cs
+++ b/sdk/src/Core/Amazon.Runtime/AWSCredentials.cs
@@ -504,7 +504,7 @@ namespace Amazon.Runtime
             private const string profileNamePrefix = "[";
             private const string profileNameSuffix = "]";
             private const string dataPrefix = "aws_";
-            private const string keyValueSeparator = "=";
+            private const string keyValueSeparator = " = ";
             private const string accessKeyName = "aws_access_key_id";
             private const string secretKeyName = "aws_secret_access_key";
             private const string tokenName = "aws_session_token";
--- a/sdk/src/Core/Amazon.Runtime/AWSCredentials.cs
+++ b/sdk/src/Core/Amazon.Runtime/AWSCredentials.cs
@@ -504,7 +504,7 @@ namespace Amazon.Runtime
             private const string profileNamePrefix = "[";
             private const string profileNameSuffix = "]";
             private const string dataPrefix = "aws_";
-            private const string keyValueSeparator = " = ";
+            private const string keyValueSeparator = "=";
             private const string accessKeyName = "aws_access_key_id";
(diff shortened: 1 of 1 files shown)


PavelSafronov:
While the credentials file documentation does have spaces, I believe it's a bad idea to make this assumption. A better approach would be to modify [this line](https://github.com/aws/aws-sdk-net/blob/master/sdk/src/Core/Amazon.Runtime/AWSCredentials.cs#L559) and use a [version of String.Split which accepts a count parameter](https://msdn.microsoft.com/en-us/library/1bwe3zdy.aspx), thereby preventing the splitting of the string beyond the first occurrence of "=".

L0G1C:
I agree adding a count could be better. Putting a count of 3 correctly gives the substrings "aws_access_key_id",  "<accesskey>", "aws_secret_access_key". Then everything after the 3rd substring is given, so it is the full  "<secretkey>" code.

L0G1C:
Not sure why the CI build is failing. It's a one line change =x

PavelSafronov:
Count of 3 will strip an equals sign from the end of the access key. For instance, making the call
```"aws_access_key_id = ACCESSKEY=".Split(new string[] { "=" }, 3, StringSplitOptions.None)```
will result in the following array
```{string[3]}
    [0]: "aws_access_key_id "
    [1]: " ACCESSKEY"
    [2]: ""
```

As you can see, count of 3 resulted in 3 entries. What we really need is count of 2, resulting in 2 entries. Effectively, count of 2 tells ```String.Split``` to stop after finding **the first matching separator**. If we change the above code to
```"aws_access_key_id = ACCESSKEY=".Split(new string[] { "=" }, 2, StringSplitOptions.None)```
we will end up with the parts
```{string[2]}
    [0]: "aws_access_key_id "
    [1]: " ACCESSKEY="
```
which is exactly what we are looking for.

L0G1C:
With only a count of 2, though, then nothing is properly split for the secret key.

Example:
```
var text = @"aws_access_key_id = ACCESSKEY
aws_secret_key_id = secretwith==";

var split = text.Split(new string[] { "=" }, 2, StringSplitOptions.None);

Console.WriteLine(split);
```

You end up with 2 parts:
[0]aws_access_key_id  
[1] ACCESSKEY
aws_secret_key_id = secretwith==

PavelSafronov:
The [```SplitData``` method in the SDK](https://github.com/aws/aws-sdk-net/blob/master/sdk/src/Core/Amazon.Runtime/AWSCredentials.cs#L556) operates on a single line of text. If you are testing String.Split in a standalone application, remember to first split your data into lines. For instance, in the following example I first split the text into separate lines, then split the lines at the "=" character to a maximum of 2 parts.
```
var text = @"aws_access_key_id = ACCESSKEY
aws_secret_key_id = secretwith==";

var split = text
    .Split(new string[] { Environment.NewLine }, StringSplitOptions.None)
    .Select(line => line.Split(new string[] { "=" }, 2, StringSplitOptions.None))
    .ToArray();
foreach(var partsSet in split)
{
    Console.WriteLine("Parts set length = " + partsSet.Length);
    Console.WriteLine("Part 1 = " + partsSet[0]);
    Console.WriteLine("Part 2 = " + partsSet[1]);
}
```

L0G1C:
Doh, I see what you mean now, yes. the split is acting on one line at a time. So yes, 2 would be correct count for split.

L0G1C:
CI build has reference errors. Do I have to commit something else in to make sure it builds correctly? This is just a one line change on one file.

PavelSafronov:
I believe the CI build is a temporary failure. I'm going to try merging your PR locally and build that.

PR #325, opened 2016-03-16
SES: SendRawEmailMessage fails to deliver to Bcc when bcc address is set through Destinations property

alex-ross83:
email-2010-12-01.normal: Fixed what looks like incorrect mapping type Destinations, misspelled from Destination, which is the property that actually contains the To:, CC:, and BCC: addresses mentioned in the comment. Without this, any consumer fails to deliver its email to the BCC as the Destination header in the email message does not contain the appropriate recipient type (ToAddresses, CCAddresses and BccAddresses)

SimpleEmail: Added simple integration test to verify that the change to the template fixes the issue.


PavelSafronov:
Unfortunately, the change you are suggesting is a breaking change (an existing property is being removed), as well as a deviation from the service specification.
To achieve what you are attempting, leave the Destinations field empty and populate MailMessage.To and MailMessage.Bcc collections with the destination emails. This is the [officially recommended approach](https://forums.aws.amazon.com/message.jspa?messageID=336807#336807).

alex-ross83:
I understand, I actually ran into the same issue described in the same post and the next post by jyotik in response to JustinC@AWS, and I found the Destination approach, which I tried to fix, as leaving the Destinations field empty and populating the To, CC and Bcc delivers email to all but Bcc.
Looks like I should first fix the original problem and work my way there. Anyway, thanks for the feedback!

PR #341, opened 2016-04-18
Update EC2Metadata to use GetResponseAsync

bryanroscoe:
These are the changes I made in our code to get EC2Metadata working on dot net core.

### DIFF
--- a/sdk/src/Services/EC2/Custom/_bcl/Util/EC2Metadata.cs
+++ b/sdk/src/Services/EC2/Custom/_bcl/Util/EC2Metadata.cs
@@ -392,9 +392,12 @@ namespace Amazon.EC2.Util
                 else
                     request = WebRequest.Create(EC2_METADATA_ROOT + path) as HttpWebRequest;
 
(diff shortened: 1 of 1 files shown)


sdk/src/Services/EC2/Custom/_bcl/Util/EC2Metadata.cs
```diff
@@ -392,9 +392,12 @@ private static List<string> GetItems(string path, int tries, bool slurp)
                 else
                     request = WebRequest.Create(EC2_METADATA_ROOT + path) as HttpWebRequest;
 
-                request.Timeout = (int)TimeSpan.FromSeconds(5).TotalMilliseconds;
+                var responseTask = request.GetResponseAsync();
 
-                using (var response = request.GetResponse())
+                if (!responseTask.Wait((int)TimeSpan.FromSeconds(5).TotalMilliseconds))
+                    throw new WebException("Timedout");
```
bryanroscoe: After taking another look at this, I am not aborting the task on timeout. That cancellation is not trivial since it is not possible to pass in a cancellation token.

PR #357, opened 2016-05-19
Add HintPath for MacOS

Thaina:


PavelSafronov:
You should not be updating the **.Unity.csproj** files manually, as these files are generated during the build process. Closing the pull request.

Take a look at [generator](https://github.com/aws/aws-sdk-net/tree/master/generator) folder, and more specifically the [UnityProjectFile.tt template file](https://github.com/aws/aws-sdk-net/blob/master/generator/ServiceClientGeneratorLib/Generators/ProjectFiles/UnityProjectFile.tt). This is the file you would want to update, so that these changes would be correctly added to the **.Unity.csproj** files during build.

PR #373, opened 2016-06-09
Fixes to target NET Standard 1.3

adamhathcock:
1.5 is a bit too high.  I think everything can target 1.3.

I haven't fixed all the libraries but I thought I'd ask before moving on.  Is this viable?

### DIFF
--- a/sdk/src/Core/project.json
+++ b/sdk/src/Core/project.json
@@ -1,10 +1,10 @@
 {
-  "version": "1.0.0-*",
-  "description": "",
-  "authors": [ "" ],
-  "tags": [ "" ],
-  "projectUrl": "",
-  "licenseUrl": "",
+    "version": "1.0.0-*",
+    "description": "",
+    "authors": [ "" ],
+    "tags": [ "" ],
+    "projectUrl": "",
+    "licenseUrl": "",
 
 
   "dependencies": {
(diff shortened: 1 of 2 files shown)


sdk/src/Core/project.json
```diff
@@ -38,32 +38,32 @@
...
+                "System.Collections": "4.0.11-rc2-24027",
+                "System.Collections.NonGeneric": "4.0.1-rc2-24027",
+                "System.Console": "4.0.0-rc2-24027",
+                "System.Diagnostics.Tools": "4.0.1-rc2-24027",
+                "System.IO.FileSystem": "4.0.1-rc2-24027",
+                "System.Linq": "4.1.0-rc2-24027",
+                "System.Net.Http": "4.0.1-rc2-24027",
+                "System.Net.Requests": "4.0.11-rc2-24027",
+                "System.Reflection.Extensions": "4.0.1-rc2-24027",
+                "System.Reflection.TypeExtensions": "4.1.0-rc2-24027",
```
adamhathcock: Added this

sdk/src/Core/project.json
```diff
@@ -38,32 +38,32 @@
...
-        "System.Console": "4.0.0-rc2-24027",
-        "System.Diagnostics.Tools": "4.0.1-rc2-24027",
-        "System.IO.FileSystem": "4.0.1-rc2-24027",
-        "System.Linq": "4.1.0-rc2-24027",
-        "System.Net.Http": "4.0.1-rc2-24027",
-        "System.Net.Requests": "4.0.11-rc2-24027",
-        "System.Reflection.Extensions": "4.0.1-rc2-24027",
-        "System.Runtime.Extensions": "4.1.0-rc2-24027",
-        "System.Runtime.InteropServices": "4.1.0-rc2-24027",
-        "System.Runtime.InteropServices.PInvoke": "4.0.0-rc2-24027",
```
adamhathcock: Removed this

normj:
I put the .NET Core version of the SDK at 1.5 initially with the assumption we could always go down but going up would be hard. I want to really understand what APIs we are giving up using by going to 1.3.

I'm curious what your use case is for needing 1.3. I Understand there are valid use cases, I just want to understand what your pain point is.

adamhathcock:
Pulling in the dependency as 1.5 makes our apps require 1.5

We don't have everything on .NET Core yet and having to manage Windows instances to install higher versions of the full framework is what we want to avoid.  

I'm getting around this by having libs also target net46 or whatever but I was hoping I didn't have to.  Covering the diffs of multitargetting everything is also painful.

normj:
A status update. We still have several high level abstractions in the SDK that need to be ported over to .NET Core. I want to finish that off before lowering the net standard just to make sure we do not lose any functionality we might need for those abstractions.

adamhathcock:
No worries.  Thanks for the update!

normj:
I pushed out version 3.2.6-beta NuGet that switches support to netstandard 1.3. I didn't use this pull request because it was out of date by the time I could get to it.

Issue #375, opened 2016-06-10
Non AWS/S3 endpoint being forced to use v4 Auth even when v2 is specified


ben-schumacher:
Even after setting AmazonS3Client SignatureVersion="2", non-amazon S3 endpoints are being interpreted as real S3 endpoints and forced to use v4 authentication/signature on GET object.  This is because the regex is encompassing all endpoints that contain "s3".  For example, s3.mycompany.com returns true as a match in the regex expression.  This breaks the code when pointing the SDK to non-amazon backed that doesn't support v4 authentication/signatures.

Code forcing v4 Auth is in \src\services\s3\custom\internal\AmazonS3KmsHandler.cs
`internal static void EvaluateIfSigV4Required(IRequest request)
        {
            // Skip this for S3-compatible storage provider endpoints
            if ((request.OriginalRequest is S3.Model.GetObjectRequest) && (AmazonS3Uri.IsAmazonS3Endpoint(request.Endpoint)))
            {
                var region = new AmazonS3Uri(request.Endpoint.OriginalString).Region;
                if(region != RegionEndpoint.USEast1)
                    request.UseSigV4 = true;
            }
        }`

Regex code is in \src\services\s3\custom\util\AmazonS3Uri.cs

`private const string EndpointPattern = @"^(.+\.)?s3[.-]([a-z0-9-]+)\.";`

normj:
Version 3.1.7.1 of the S3 Nuget package was just released that contains this fix.

PR #379, opened 2016-06-13
Proposed fix for issue #375 to prevent v4 auth from being forced on GET object

ben-schumacher:
Proposed fix for issue #375  to prevent v4 auth from being forced on GET when using non s3 endpoint.  The use of v4 auth was being forced for endpoints that contain s3 and this is not necessarily a real AWS s3 endpoint - like s3.company.com.

### DIFF
--- a/sdk/src/Services/S3/Custom/Util/AmazonS3Uri.cs
+++ b/sdk/src/Services/S3/Custom/Util/AmazonS3Uri.cs
@@ -169,7 +169,7 @@ namespace Amazon.S3.Util
                 throw new ArgumentNullException("uri");
 
             var match = new Regex(EndpointPattern).Match(uri.Host);
-            if (uri.Host.EndsWith("amazonaws.com"))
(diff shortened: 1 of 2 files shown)


steveataws:
Thanks for the PR. As it currently stands, the change won't work for users in the Beijing (cn-north-1) region, as those endpoints end with 'amazonaws.com.cn'. Do you want to revise the change and resubmit, or would you prefer the SDK team to make a similar change to the code and discard this PR?

ben-schumacher:
Hi Steve,
I'd be happy to revise and resubmit.  Let me take a look and incorporate
your suggestion.  Should have something back to you in a couple hours.  I
appreciate the prompt response.
-Ben

ben-schumacher:
Hi Steve,
Code has been update to handle China (Beiging) Region.
Thanks,
Ben

normj:
This was released with version 3.1.7.1 of the S3 NuGet package

Issue #402, opened 2016-06-25
RefreshingAWSCredentials does not refresh credentials for Unity


madangosain:
```
public abstract class RefreshingAWSCredentials : AWSCredentials
```

does not refresh credentials automatically. It consistently gives error for multiple devices over different regions:

AmazonClientException: The retrieved credentials have already expired: Now = 06/23/2016 22:20:36, Credentials expiration = 06/23/2016 22:16:01
AmazonClientException: The retrieved credentials have already expired: Now = 06/23/2016 16:54:28, Credentials expiration = 06/23/2016 15:55:41
AmazonClientException: The retrieved credentials have already expired: Now = 06/23/2016 21:51:26, Credentials expiration = 06/23/2016 20:50:09

I believe that RefreshingAWSCredentials code is not waiting for new credentials to loaded for Unity and assuming it is synchronous. 

Moreover AWSCredentials.cs does not expose _currentState or either CurrentState.ExpirationTime for developer to implement their own auto-refreshing. We are stuck with current issue without any help. I've seen several threads for the same without any solution
https://forums.aws.amazon.com/thread.jspa?threadID=233919
https://forums.aws.amazon.com/thread.jspa?threadID=234211

At the minimum, if you can make currentState public, that will allow us to auto-refresh the credentials ourselves by calling GetCredentialsAsync

madangosain:
This bug is still reproducible and I see that Andreyul's PR is not merged. Do we have any workaround for this?

madangosain:
@karthiksaligrama , any update on this issue? Can you please make currentState public so that developer can refresh the credentials as needed.

mcmorry:
Any update? is several months that this bug is open...

PavelSafronov:
PR #427 replaced all the `DateTime.Now` and `DateTime.UtcNow` uses in the SDK with a "corrected" DateTime value, so this issue should not be occurring with the latest version of the SDK.
Resolving.

Issue #393, opened 2016-06-21
[AWS Credentials] The retrieved credentials have already expired - Wrong device date/time


lukeWW:
Hi, we have encountered a problem with the AWS signature. 
We're using the AWS SDK for Unity 2.1.0.1
In some cases we have users whose devices are failing to retrieve the correct credentials and the error is the following:
"The retrieved credentials have already expired ..."

We realized that the problem is due to the wrong date/time set on that devices. 

How can we avoid that, using a wrong date/time on the device the call won’t fail? 
As a quick fix we replaced all references of DateTime.Now and UtcNow in the code with our class, is there any better solution?

PavelSafronov:
PR #427 replaced all the `DateTime.Now` and `DateTime.UtcNow` uses in the SDK with a "corrected" DateTime value, so this issue should not be occurring with the latest version of the SDK.
Resolving.

ajithrk:
Hi,
This issue is still occurring for the latest aws unity sdk (aws-sdk-unity_3.3.131.0). Please advice.

Exception:
Amazon.Runtime.AmazonClientException: The retrieved credentials have already expired: Now = 08/15/2017 16:25:36, Credentials expiration = 07/31/2017 11:55:38
  at Amazon.Runtime.RefreshingAWSCredentials.UpdateToGeneratedCredentials (Amazon.Runtime.CredentialsRefreshState state) [0x00000] in <filename unknown>:0 
  at Amazon.Runtime.RefreshingAWSCredentials.GetCredentials () [0x00000] in <filename unknown>:0 
  at Amazon.Runtime.Internal.Util.SdkCache+CacheKey.Create (Amazon.Runtime.AmazonServiceClient client, System.Object cacheType) [0x00000] in <filename unknown>:0 
  at Amazon.Runtime.Internal.Util.SdkCache.GetCache[String,TableDescription] (Amazon.Runtime.AmazonServiceClient client, System.Object cacheIdentifier, IEqualityComparer`1 keyComparer) [0x00000] in <filename unknown>:0 
  at Amazon.DynamoDBv2.DocumentModel.Table.LoadTableInfo () [0x00000] in <filename unknown>:0 
  at Amazon.DynamoDBv2.DocumentModel.Table.LoadTable (IAmazonDynamoDB ddbClient, Amazon.DynamoDBv2.DocumentModel.TableConfig config) [0x00000] in <filename unknown>:0 
[...]

PR #404, opened 2016-06-26
Added DateTimeProvider property to RefreshingAWSCredentials.

Andreyul:
Workaround for exception "The retrieved credentials have already expired".
This allows using other DateTime source (NTP) instead of device time.

Usage:
```
    credentials = new CognitoAWSCredentials(poolId, region);
    credentials.DateTimeProvider = GetDateTime;

...

    private DateTime GetDateTime()
    {
        // Reading and writing DateTime is not thread safe
        lock (dateLock)
        {
            return reliableDate;
        }
    }
```

Related issues: https://github.com/aws/aws-sdk-net/issues/393 https://github.com/aws/aws-sdk-net/issues/402

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/AWSCredentials.cs
+++ b/sdk/src/Core/Amazon.Runtime/AWSCredentials.cs
@@ -1145,6 +1145,8 @@ namespace Amazon.Runtime
             }
         }
 
+        public Func<DateTime> DateTimeProvider { get; set; }
+
         #endregion
 
         #region Override methods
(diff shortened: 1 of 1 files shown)


PavelSafronov:
Thank you for submitting this PR.

The approach looks straight-forward, but I'm concerned that it's fairly specific to refreshing credentials. This works fine for this case, but there are other places in the SDK that are impacted by an incorrect device time, such as generating S3 pre-signed urls. If we go with this approach for credentials, we'll end up having to keep adding similar solutions (DateTime delegate) for other use-cases.

An alternative approach we're thinking about is to provide a property ```TimeSpan? ManualClockCorrection``` to the [Amazon.AWSConfigs class](https://github.com/aws/aws-sdk-net/blob/master/sdk/src/Core/AWSConfigs.cs#L70), making it possible for a user to set this to the difference between device clock and "actual" time. This property would then be used to determine the "correct" current time, and will be used everywhere the SDK currently uses ```DateTime.Now``` or ```DateTime.UtcNow```. Would this approach work for you? Do you see an issue with using a static offset and not a delegate?

Andreyul:
I have opened a new PR https://github.com/aws/aws-sdk-net/pull/427 for property `TimeSpan? ManualClockCorrection`.

madangosain:
Delegate is  preferred over Delta as it gives better control to user. I am using third party tool to detect actual time, delegate would simply route it to third party library.

Issue #406, opened 2016-06-28
[Unity] Endless sync lock and no callback caused by network failure during GetIdentityId


pfleeter:
`Dataset.SynchronizeAsync` will call `FireSyncFailureEvent` if it determines that the network status is `NotReachable`.  It makes this determination using Unity's `Application.internetReachability` in `NetworkInfo`.  However, according to Unity's [documentation](http://docs.unity3d.com/ScriptReference/Application-internetReachability.html), this check cannot be used to determine whether the internet is actually reachable.  Assume that it is not actually functional.

In this case, `SynchronizeAsync` will call `SynchronizeHelperAsync` (which locks itself), which will call `SynchornizeInternal` in another thread via the `AsyncExecutor`.  In `SynchornizeInternal`, `CognitoCredentials.GetIdentityId` is called.  Since it cannot communicate with the identity service, it will return a "base stream is null" exception, which is not handled in `SynchornizeInternal`.  The exception is caught by `SafeExecute`, the sync success or failure events are never fired, and the lock on the dataset is never released.

A logged example of this can be seen in [this forum post](https://forums.aws.amazon.com/message.jspa?messageID=721512#721512).

klaytaybai:
Closing due to age, inactivity, PR referenced in the linked forum post

klaytaybai:
Closing due to age, inactivity, PR referenced in the linked forum post

PR #408, opened 2016-06-28
Add try catch block around GetIdentityId during sync.

pfleeter:
Handle exception leading to issue #406

### DIFF
--- a/sdk/src/Services/CognitoSync/Custom/SyncManager/Dataset.cs
+++ b/sdk/src/Services/CognitoSync/Custom/SyncManager/Dataset.cs
@@ -690,15 +690,12 @@ namespace Amazon.CognitoSync.SyncManager
         {
 
             //make sure we have the latest identity id
-#if BCL35 || UNITY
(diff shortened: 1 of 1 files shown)


sdk/src/Services/CognitoSync/Custom/SyncManager/Dataset.cs
```diff
@@ -691,12 +691,15 @@ private void SynchornizeInternal(Runtime.AsyncOptions options)
 
             //make sure we have the latest identity id
 #if BCL35 || UNITY
-            CognitoCredentials.GetIdentityId();
+            try
```
karthiksaligrama: This will cause build to fail on BCL45 and PCL platforms. You need to have the try block above the `#if BCL35 || UNITY` code. also you can get rid of the line 700 & 701.

normj:
Thanks for the pull request. This was released in version 3.3.1.22 of AWSSDK.CogntioSync

Issue #391, opened 2016-06-20
[COGNITO] [SYNC] Unhandled (by SDK) conflict breaks synchronization


sebas86:
Hi, I have one another, much bigger problem with synchronization using Cognito Sync with Unity.

After few synchronizations I can't save any new data in cloud. When I check communication between client app and server I see many _409 Conflict_ responses but in application I gets only synchronization success event (`Dataset.OnSyncSuccess`). Also in server response I get always this same message:

``` json
{
    "message": "Current SyncCount for: SomeKeyName is: 2 not: 1"
}
```

My code handles conflicts just like in provided sample. Conflict code also was tested and it just works.

This bug occurs more often when many unrelated keys is saved in random order and time. After saving all data under one key as simple JSON string this bug occurs less often, but still occurs.

Application restart doesn't solve problem, error disappear only if I clean application data.

Problem occur in Unity Editor and on iOS devices.

pfleeter:
I see this issue very frequently in the wild, as well.  I believe the issue is that the local device sqlite db has a record that is 1) modified, 2) has the same value as the remote record, and 3) has a different sync count.

The conflict detection code in RunSyncOperation ignores the differing SyncCount's:
```
// if conflict, prompt developer/user with callback
List<SyncConflict> conflicts = new List<SyncConflict>();
List<Record> conflictRecords = new List<Record>();
foreach (Record remoteRecord in remoteRecords)
{
    Record localRecord = Local.GetRecord(IdentityId,
                                          DatasetName,
                                          remoteRecord.Key);
    // only when local is changed and its value is different
    if (localRecord != null && localRecord.IsModified
        && !StringUtils.Equals(localRecord.Value, remoteRecord.Value))
    {
        conflicts.Add(new SyncConflict(remoteRecord, localRecord));
        conflictRecords.Add(remoteRecord);
    }
}
```

I also only see this issue on iOS (not on Android), which seems odd since I don't immediately notice any platform specific code related to this.

sebas86:
After latest tests we also gets this same error on Android devices. Problem is almost 100% to reproduce just by setting some new values in Dataset during synchronization – next synchronization should just fail with 409 conflict and block all future synchronizations.

pfleeter:
I have been reading through the SDK's code some more, and my initial comment now seems off base.  It looks like in the case where an identical value is reported from the server with a different sync count than a locally modified record, it should just stomp the local record (resetting the is modified flag and upping the sync count).  

[...]

PR #409, opened 2016-06-28
Improve DataConflictException handling in case of partial local modification

pfleeter:
In the case where only some of the records in a patch are skipped by `ConditionallyPutRecords`, maxPatchSyncCount will not cause the dataset sync count to be reset sufficiently to correct the skipped records when they experience a `DataConflictException` on future sync attempts.  (Especially likely if multiple records are always updated together.)

Another possibility for this might be to have `ConditionallyPutRecords` return a flag for whether it skipped any records, and if it did, to not update the dataset sync count in the `DataConflictException` handler.

Attempts to address issue #391


steveataws:
PR merged, thanks for the contribution.

Issue #360, opened 2016-05-24
Undeleted merge datasets accumulating


pfleeter:
When I go into the identity browser on the AWS console, I see that I have 7 merge data sets (or at least data sets named <datasetname>.<merged identity>.  However, when I use the sqlite3 client on my Mac to look in my local storage, I see only one.  When I sync datasets in the Unity editor, only the one that is in my local storage is handed to the merge callback.

How can I get the other datasets into my local storage, so that they will trigger the merge callback?

normj:
I believe this was fixed with PR https://github.com/aws/aws-sdk-net/pull/414

PR #414, opened 2016-06-30
Make the local merge dataset sync count match the remote sync count

pfleeter:
Once the merge data set is created under the new authenticated cognito id, the sync count is 1 (even though records in that data set are unmodified and probably higher than 1).  Since the local merge dataset does not have a modified sync count, when a synchronize is performed prior to a deletion, the synchronize fails, with a message that the local sync count does not exist.  This modification updates the sync count in the local data store to 1, so that it matches the one on the server.


pfleeter:
I believe this is also related to #360

normj:
Thanks for the pull request. Version 3.3.1.23 of CognitoSync contains this change.

PR #417, opened 2016-07-04
Make AMA event store parsing robust to corrupted / obsolete JSONs

noamgat:
When upgrading from the old aws-sdk-unity to the new aws-sdk-net with unity support, some of the json formats change. This causes the delivery client to not be able to deliver ANY events, because it fails to convert the old jsons to the new Event model.

This renders AMA nonfunctional for users who will upgrade to our new version. The suggested fix catches these errors, writes to the log and discards the event so the rest of the pipeline will be able to continue functioning.

### DIFF
--- a/Unity.README.md
+++ b/Unity.README.md
@@ -74,7 +74,9 @@ You can also do this configuration in a script
 			<namespace fullname="System.Security.Cryptography" preserve="all"/>
    		</assembly>
 
-		<assembly fullname="AWSSDK.Core" preserve="all"/>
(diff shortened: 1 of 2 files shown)


noamgat:
There is a second commit in this pull request as well, about documentation:
According to the unity docs, the assembly preserve="all" mode only works with IL2CPP compilation. So, when using Android Mono2x scripting backend (the default, as Android IL2CPP is considered experimental) it does not put the correct classes inside the apk. The pull request link.xml modification is required for mobile analytics to work on android with mono2x.

Issue #152, opened 2014-12-08
GetObjectRequest hangs randomly.


Stormsys:
**Randomly** my application will hang infinity waiting for a response in the Amazon library for AmazonS3Client.GetOject see the stacktrace below, the object is not being reused.

![image](https://cloud.githubusercontent.com/assets/578241/5337169/6afe61f4-7ebd-11e4-852a-c359a295c842.png)

here is a sample of the code run in that thread:

``` c#
using (var amazonS3Client = CreateS3Client())
{
    var s3PartDownloadRequest = new GetObjectRequest
    {
        BucketName = download.Bucket,
        Key = download.ObjectKey,
        ByteRange = new ByteRange(position, end)
    };

    using (var getObjectResponse = amazonS3Client.GetObject(s3PartDownloadRequest)) //hangs on this line
    {    
        getObjectResponse.WriteResponseStreamToFile(dest);
    }
}
```

saguiitay:
Has anyone been able to resolve this issue? I'm facing it too...

theofanis:
I also experience this sometimes, UploadPartAsync never returns, and actually, it doesn't even upload, since the StreamTransferProgress handler doesn't get anything.

I also provide a CancellationToken, and when the problem occurs, it even ignores the cancellation, so there's actually no way to get this task finished.

kobi:
In case anyone still has this issue, I've found two changes that removed the problem:
* Before you initialize any of ASW SDK's class set `ServicePointManager.DefaultConnectionLimit = 1000;`.  
I set it to 1000, but any big number works. [MSDN says](https://msdn.microsoft.com/en-us/library/system.net.servicepointmanager.defaultconnectionlimit.aspx) the default number is `Int32.MaxValue`, but in fact it is just **2**. 
The AWS SDK for .Net does [use a value of 50](https://github.com/aws/aws-sdk-net/blob/f5c9bce1d6303036718e1b0ad09fe481dd1b52c9/sdk/src/Core/Amazon.Runtime/_bcl/ClientConfig.bcl.cs#L95-L104), but that might not be enough (I also use [SWF](https://aws.amazon.com/swf/), which requires a large number of open connections).
[...]

PR #449, opened 2016-09-01
Close HttpWebRequest's response stream on WebExceptions

lewislabs:
If `httpRequest.GetResponse()` internally throws a `WebException` then there is still a response stream to close. I believe this is causing sockets to remain open and leading to #152

### DIFF
--- a/AWSSDK_DotNet35/Amazon.Runtime/Pipeline/HttpHandler/HttpHandler.cs
+++ b/AWSSDK_DotNet35/Amazon.Runtime/Pipeline/HttpHandler/HttpHandler.cs
@@ -18,6 +18,7 @@ using Amazon.Runtime.Internal.Util;
 using Amazon.Util;
 using System;
 using System.Globalization;
+using System.Net;
 using System.Reflection;
 using System.Text;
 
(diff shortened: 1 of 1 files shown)


gokarnm:
Hi @lewislabs , thanks for the PR! I'll go through the change, test it and get back you.

lewislabs:
Thanks @gokarnm. I think these commits could also go into master. Let me know what you think.

gokarnm:
Hi @lewislabs , the changes look good! I'm curious about how you tested this, specifically [WriteContentToRequestBody()](https://github.com/aws/aws-sdk-net/blob/aws-sdk-net-v2/AWSSDK_DotNet35/Amazon.Runtime/Pipeline/HttpHandler/HttpHandler.cs#L75) throwing an exception leading to CompleteFailedRequest() being called.

lewislabs:
Hi @gokarnm. I haven't tested this explicitly, I'm primarily going on this comment here https://github.com/aws/aws-sdk-net/blob/master/sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/HttpHandler.cs#L98 

If the `httpRequest.GetResponse()` call throws an exception (more than likely a `WebException`) then the `response` variable will be null and will not be disposed. If you don't dispose of the response stream then this will leave a socket open in CLOSE_WAIT state and will keep hold of one of the connections from the ServicePoint's pool. If this happens a few times then you run the risk of using up all the sockets available to the ServicePoint and any subsequent `httpWebRequest.GetResponse()` calls will simply block forever.

Issue #461, opened 2016-09-29
RegionEndpoint for Mumbai removed?


Singh400:
I recently upgrade my libs from `AWSSDK.Core v3.1.0.0` to `AWSSDK.Core v3.3.0.0` and noticed that `RegionEndpoint.APSouth1` (Mumbai) was no longer supported.

Is this correct? As [AWS Regions and Endpoints](http://docs.aws.amazon.com/general/latest/gr/rande.html) still lists Mumbai as an available endpoint for most AWS services. I primarily use SNS and SQS and both those services still list Mumbai as an available endpoint.

If I downgrade `AWSSDK.Core v3.3.0.0` to `AWSSDK.Core v3.1.11.0` then the Mumbai endpoint is available.

Many thanks,
Indy

normj:
Sorry about that, we will get that patch back in right away.

normj:
Version 3.3.0.2 of the AWSSDK.Core is out with this constant added back in. Thanks for letting us know.

PR #462, opened 2016-09-29
Restored Mumbai Region Endpoint

Singh400:
Re: #461


normj:
Thanks for the pull request but I have already made the fix. It is going through our build system now and should be out later this evening.

PR #467, opened 2016-10-12
Fix LruCache NullReferenceException

arieltrevisan:
The method `S3Signer.Sign()`, when using V4, uses the `LruCache` utility.
Found the utility fails to properly link the elements in the LRU list, causing a `NullReferenceException` whenever it tries to get any item in the middle of the list.
Here is the fix, with a unit test to verify it.


vellozzi:
Thanks arieltrevisan.  I also found another problem.  The Remove method isn't cleaning up as it should.  I'll get both fixes out ASAP.

vellozzi:
Done.  Your fix is merged into master.  The other one will go out soon.

arieltrevisan:
Thank you @vellozzi for your quick response and merge! This really helps.

Issue #490, opened 2016-11-17
Unable to convert JSON data of type UInt to DynamoDB type


rheusinkveld:
When using Document.FromJson the SDK throws an exception that is does not support UInt if the Json contains a number like 3221422080.
The Object Persistence Model does support UInt:
http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DotNetSDKHighLevel.html#DotNetDynamoDBContext.SupportedTypes
I believe the Document Model should also support this. I'm working on a pull request for this.

sstevenkang:
Howdy.  Which version of Core are you building against?  Also, could you provide a code snippet that repros this behavior?  Thanks!

wcrooy:
Hi @sstevenkang,

We're using it in .NET 4.5.1  environment with these versions:
```
<package id="AWSSDK.Core" version="3.3.5" targetFramework="net452" />
<package id="AWSSDK.DynamoDBv2" version="3.3.0.3" targetFramework="net452" />
```


```
public class ExampleItem
        {
            public string ExampleStringItem { get; set; }
            public uint ExampleValue { get; set; }
        }

public void InsertIntoDynamoDB()
        {
            var jsonExampleItem = new ExampleItem() {ExampleStringItem = "ExampleString", ExampleValue = 3221422080};

            var jsonExample = new
            {
                userId = Guid.NewGuid().ToString("N"),
                list = new List<ExampleItem> {jsonExampleItem}
            };

            var json = JsonConvert.SerializeObject(jsonExample);
            var jsonDoc = Document.FromJson(json);

            var table = dbContext.GetTargetTable<InputData>(dynamoDbOperationConfig);
            table.PutItem(jsonDoc);
        }
```

rheusinkveld:
Tested the pull request and it works for the scenario above. 
The check appears to have failed because of a build time out. Is it possible to restart the check?

sstevenkang:
Hey Ruben.  Thanks for sending us a PR.  It seems like a proper fix would also handle the `ulong `case.  We have a commit ready with accompanying unit tests.  We'll have that out shortly.

sstevenkang:
The fix 33968c6bdc0ed58d94b6811c470ae994380bc388 should be in AWSSDK.DynamoDB 3.3.1.  Let us know if your problem is resolved.  Thanks!

rheusinkveld:
Thanks! The fix is working for us. Great that you were able to release this so fast!

[...]

PR #493, opened 2016-11-18
Unable to convert JSON data of type UInt to DynamoDB type 

rheusinkveld:
Fix for issue #490


rheusinkveld:
Fixed with change 33968c6

PR #514, opened 2016-12-20
Fixed thread-unsafe access to dictionary

bcuff:
Reading a dictionary while it's being modified can cause an
infinite loop to manifest.

See docs:
https://msdn.microsoft.com/en-us/library/xfhwa508(v=vs.110).aspx#Anchor_10

We discovered this on a few of our servers. CPU would spike for a period of time until the thread was eventually aborted. The profile below shows that a large chunk of time was being spent in this method.

image

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Internal/Util/SdkCache.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Util/SdkCache.cs
@@ -333,7 +333,10 @@ namespace Amazon.Runtime.Internal.Util
         {
             get
             {
-                return Contents.Keys.ToList();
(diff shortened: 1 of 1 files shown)


PavelSafronov:
We've updated AWSSDK.Core to [3.3.7.1](https://www.nuget.org/packages/AWSSDK.Core/3.3.7.1) and AWSSDK.DynamoDBv2 to [3.3.1.4](https://www.nuget.org/packages/AWSSDK.DynamoDBv2/3.3.1.4) with this fix. Thank you for the pull request.

bcuff:
Thanks!

PR #521, opened 2017-01-03
Glacier: Allow setting retrieval tier using ArchiveTransferManager

seaeagle1:
This introduces the Tier option from JobParameters into InitiateArchiveRetrievalOptions so the retrieval speed can be set.


PavelSafronov:
Thanks for the PR.
It looks like InitiateArchiveRetrievalOptions are also consumed in the [async version of ArchiveTransferManager](https://github.com/aws/aws-sdk-net/blob/master/sdk/src/Services/Glacier/Custom/_bcl+coreclr/Transfer/_async/ArchiveTransferManager.async.cs#L237), can you add the matching changes there as well?

seaeagle1:
Sure! See added commit.

PR #531, opened 2017-01-19
Added option in LitJson to SkipNonMembers

Deathwing:
Implemented the handling of original LitJson 0.9.0 for giving the option of skipping non members.


vellozzi:
Can you elaborate on how this change improves the SDK?  If you're using LitJson directly we recommend that you pull in a dependency on that instead; we don't guarantee that we'll always use LitJson as our Json library.

sstevenkang:
Closing due to inactivity.

Issue #430, opened 2016-08-04
Unity SDK Cognito sync count exception mishandled


pfleeter:
If the client attempts to update records with a stale sync count, the exception returned from the service is not handled properly.  

From `Dataset.RunSyncOperation`, `CognitoSyncStorage.PutRecords` is called, which calls `AmazonCognitoSyncClient.UpdateRecords`.  If the sync count of an updated record is stale, the service responds with an error.  `HttpErrorResponseExceptionHandler.HandleException` attempts to un-marshall this error and throw an exception. 

I am seeing that the un-marshalled error code is "InvalidParameterException" and the un-marshalled status code is `BadRequest`.  The message is something of the form "Current SyncCount for: blarg is: 27 not: 26".  

It appears to me that `UpdateRecordsResponseUnmarshaller` should be detecting a `ResourceConflictException`, but it fails to do so because the following logic is trying to match "InvalidParameterException" to "ResourceConflict".

```
if (errorResponse.Code != null && errorResponse.Code.Equals("ResourceConflict"))
{
     return new ResourceConflictException(errorResponse.Message, innerException, errorResponse.Type, errorResponse.Code, errorResponse.RequestId, statusCode);
}
```

 (It also has a near miss at matching to "InvalidParameter" and throwing an `InvalidParameterException`.)

Because this match fails (as do all others in `UpdateRecordsResponseUnmarshaller`), the default case of `AmazonCognitoSyncException` is thrown.

Later, when `CognitoSyncStorage.PutRecords` catches this exception and calls its `HandleException` method, it is unable to generate the `DataConflictException` that `Dataset.RunSyncOperation` is expecting.  It does so, because the following logic fails to match (because the exception is an `AmazonCognitoSyncException` not a `ResourceConflictException` and because the StatusCode is `BadRequest` (400) not `Conflict` (409).

```
else if (ase.GetType() == typeof(ResourceConflictException)
         || ase.StatusCode == System.Net.HttpStatusCode.Conflict)
{
    return new DataConflictException(message);
}
```

pfleeter:
The initial post in this issue was for iOS.  For some reason, we are seeing slightly different behavior on Android.  Instead of `BadRequest` / "InvalidParameterException", Android is seeing `OK` / "ResourceConflictException", which is failing to generate a `DataConflictException` as well.

PR #532, opened 2017-01-20
Handle Cognito Sync Count Exception

pfleeter:
Addresses unhandled mismatch sync count exception per issue #430


aphexyuri:
@soapyigu can we get this looked at as well?

soapyigu:
In my opinion, this pull request is unnecessary as currently this kind of error falls into data storage exception, which is the base class that data conflict exception extends from.

pfleeter:
@soapyigu Per the issue #430, the problem is that the consumer of this exception is specifically expecting a data conflict exception, not a data storage exception.  See https://github.com/aws/aws-sdk-net/blob/master/sdk/src/Services/CognitoSync/Custom/SyncManager/Dataset.cs#L619

normj:
Thanks for the pull request. Version 3.3.1.23 of CognitoSync contains this change.

PR #659, opened 2017-05-24
Updating SQS constants for FIFO-related attributes.

Singh400:
Updating SQS constants for FIFO-related attributes.

## Description
Add constants to represent attributes; `FifoQueue` and `ContentBasedDeduplication`

## Motivation and Context
To keep consistency and to remove magic strings from consumers of the SQS API.

## Testing
~~I didn't.~~

Ran all related SQS tests

N/A

## Types of changes
Not sure if it is a new feature _technically_, but it's the closest category.

## Checklist

## License

### DIFF
--- a/sdk/src/Services/SQS/Custom/Model/GetQueueAttributesResponse.extensions.cs
+++ b/sdk/src/Services/SQS/Custom/Model/GetQueueAttributesResponse.extensions.cs
@@ -175,22 +175,38 @@ namespace Amazon.SQS.Model
         /// <summary>
         /// Whether or not the queue is a First-In-First-Out (FIFO) queue
         /// </summary>
-        public bool? FifoQueue
+        public bool FifoQueue
         {
             get
             {
-                return getAttributeValue(SQSConstants.ATTRIBUTE_FIFO_QUEUE)?.ToUpper().Equals("TRUE");
+                var attributeValue = getAttributeValue(SQSConstants.ATTRIBUTE_FIFO_QUEUE);
+
+                if (attributeValue == null || attributeValue.Trim() == string.Empty)
+                {
+                    return false;
+                }
+
+                bool result;
+                return bool.TryParse(attributeValue, out result);
             }
         }
 
         /// <summary>
         /// Whether or not the First-In-First-Out (FIFO) queue has content based deduplication enabled
         /// </summary>
-        public bool? ContentBasedDeduplication
+        public bool ContentBasedDeduplication
         {
             get
             {
-                return getAttributeValue(SQSConstants.ATTRIBUTE_CONTENT_BASED_DEDUPLICATION)?.ToUpper().Equals("TRUE");
+                var attributeValue = getAttributeValue(SQSConstants.ATTRIBUTE_CONTENT_BASED_DEDUPLICATION);
+
+                if (attributeValue == null || attributeValue.Trim() == string.Empty)
+                {
+                    return false;
+                }
+
+                bool result;
+                return bool.TryParse(attributeValue, out result);
             }
         }
 
--- a/sdk/src/Services/SQS/Custom/Model/GetQueueAttributesResponse.extensions.cs
+++ b/sdk/src/Services/SQS/Custom/Model/GetQueueAttributesResponse.extensions.cs
@@ -13,12 +13,7 @@
  * permissions and limitations under the License.
  */
 using System;
-using System.Collections.Generic;
-using System.Xml.Serialization;
-using System.Text;
-using System.IO;
 
-using Amazon.Runtime;
 using Amazon.SQS.Util;
 using Amazon.Util;
 
@@ -84,7 +79,6 @@ namespace Amazon.SQS.Model
                 return value;
             }
         }
-
 
         /// <summary>
         /// Gets the approximate number of messages from the Attributes collection.
@@ -175,6 +169,28 @@ namespace Amazon.SQS.Model
             get
             {
                 return getAttributeValue(SQSConstants.ATTRIBUTE_POLICY);
+            }
+        }
+
+        /// <summary>
+        /// Whether or not the queue is a First-In-First-Out (FIFO) queue
+        /// </summary>
+        public bool? FifoQueue
+        {
+            get
+            {
+                return getAttributeValue(SQSConstants.ATTRIBUTE_FIFO_QUEUE)?.ToUpper().Equals("TRUE");
+            }
+        }
+
+        /// <summary>
+        /// Whether or not the First-In-First-Out (FIFO) queue has content based deduplication enabled
+        /// </summary>
+        public bool? ContentBasedDeduplication
+        {
+            get
+            {
+                return getAttributeValue(SQSConstants.ATTRIBUTE_CONTENT_BASED_DEDUPLICATION)?.ToUpper().Equals("TRUE");
             }
         }
 
--- a/sdk/src/Services/SQS/Custom/Model/GetQueueAttributesResponse.extensions.cs
+++ b/sdk/src/Services/SQS/Custom/Model/GetQueueAttributesResponse.extensions.cs
@@ -197,15 +197,10 @@ namespace Amazon.SQS.Model
             {
                 var attributeValue = getAttributeValue(SQSConstants.ATTRIBUTE_CONTENT_BASED_DEDUPLICATION);
 
-                if (attributeValue == null || attributeValue.Trim() == string.Empty)
-                {
-                    return null;
-                }
-
                 bool result;
                 if (bool.TryParse(attributeValue, out result))
                     return result;
-                return false;
+                return null;
             }
         }
 
--- a/sdk/src/Services/SQS/Custom/Model/GetQueueAttributesResponse.extensions.cs
+++ b/sdk/src/Services/SQS/Custom/Model/GetQueueAttributesResponse.extensions.cs
@@ -181,13 +181,10 @@ namespace Amazon.SQS.Model
             {
                 var attributeValue = getAttributeValue(SQSConstants.ATTRIBUTE_FIFO_QUEUE);
 
-                if (attributeValue == null || attributeValue.Trim() == string.Empty)
-                {
-                    return false;
-                }
-
                 bool result;
-                return bool.TryParse(attributeValue, out result);
+                if (bool.TryParse(attributeValue, out result))
+                    return result;
+                return false;
             }
         }
 
--- a/sdk/src/Services/SQS/Custom/Model/GetQueueAttributesResponse.extensions.cs
+++ b/sdk/src/Services/SQS/Custom/Model/GetQueueAttributesResponse.extensions.cs
@@ -192,9 +192,9 @@ namespace Amazon.SQS.Model
         }
 
         /// <summary>
-        /// Whether or not the First-In-First-Out (FIFO) queue has content based deduplication enabled
+        /// Whether or not the First-In-First-Out (FIFO) queue has content based deduplication enabled. For non-FIFO queues this will return null.
         /// </summary>
-        public bool ContentBasedDeduplication
+        public bool? ContentBasedDeduplication
         {
             get
             {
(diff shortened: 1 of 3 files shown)


PavelSafronov:
Thanks for the PR!

A few concerns:
* tabbing is inconsistent
* integ tests for these attributes should be added [here](https://github.com/aws/aws-sdk-net/blob/7ddeaa8bb98a26e704df7fca51d6b2c9da05d756/sdk/test/IntegrationTests/Tests/SQS.cs)
* it may be a good idea to add matching `bool?` properties for these attributes in [GetQueueAttributesResponse.extensions.cs](https://github.com/aws/aws-sdk-net/blob/7ddeaa8bb98a26e704df7fca51d6b2c9da05d756/sdk/src/Services/SQS/Custom/Model/GetQueueAttributesResponse.extensions.cs)

Singh400:
@PavelSafronov thanks for picking up those concerns. More than happy to tackle them.

Cheers,
Indy

Singh400:
@PavelSafronov I couldn't see from the Travis log but do the integration tests run against N. Virginia (`us-east-1`)? If so, then I would expect any FIFO-related integration tests to fail as FIFO is only currently available in regions Ohio (`us-east-2`) and Oregon (`us-west-2`).

Thanks,
Indy

PavelSafronov:
Travis does not run integ tests, but our internal build servers do, and these tests can be run manually of course.
The integ tests are indeed configured to run against **us-east-1**: [3.5 config](https://github.com/aws/aws-sdk-net/blob/master/sdk/test/IntegrationTests/Config/App.35.config#L7), [4.5 config](https://github.com/aws/aws-sdk-net/blob/master/sdk/test/IntegrationTests/Config/App.45.config#L7).

In the case where a particular feature is only available in some regions, it makes sense to write a test method like this, where we construct and dispose of the client in the test:
```
[TestMethod]
[TestCategory("SQS")]
public void SQSFIFOTest()
{
	using(var sqsClient = new AmazonSQSClient(RegionEndpoint.USEast2))
	{
		// test
	}
}
```

Singh400:
Any guidance on what to do with a stalled build? I am sure it was not the commit as @PavelSafronov said Travis does not run integration tests.

Raw Travis log: https://api.travis-ci.org/jobs/236472740/log.txt?deansi=true

```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

The build has been terminated
```

Am I able to force another build on Travis?

PavelSafronov:
Hmm, not sure what happened with Travis, but looks like subsequent [build 236671766 passed](https://travis-ci.org/aws/aws-sdk-net/builds/236671766?utm_source=github_status&utm_medium=notification).

PavelSafronov:
Looks like you're continuing making changes to the PR, that's great! Please ping when you've finished and I'll review the changes.

Singh400:
@PavelSafronov ping. I can not see anything else I would like to cleanup/improve. So feel free to review the changes :)

Cheers,
Indy

sdk/src/Services/SQS/Custom/Model/GetQueueAttributesResponse.extensions.cs
```diff
@@ -178,6 +172,28 @@ public string Policy
...
         }
 
+        /// <summary>
+        /// Whether or not the queue is a First-In-First-Out (FIFO) queue
+        /// </summary>
+        public bool? FifoQueue
+        {
+            get
+            {
+                return getAttributeValue(SQSConstants.ATTRIBUTE_FIFO_QUEUE)?.ToUpper().Equals("TRUE");
```
PavelSafronov: `Boolean.TryParse` should be used here. If parsing fails or if the value is null, we return `false`.

sdk/src/Services/SQS/Custom/Model/GetQueueAttributesResponse.extensions.cs
```diff
@@ -178,6 +172,28 @@ public string Policy
             }
         }
 
+        /// <summary>
+        /// Whether or not the queue is a First-In-First-Out (FIFO) queue
+        /// </summary>
+        public bool? FifoQueue
```
PavelSafronov: Initially I suggested having the new properties be `bool?`, but giving it more thought I think simply `bool` is more appropriate for `FifoQueue`, since that property can only have two values: either the queue is FIFO or not.

sdk/src/Services/SQS/Custom/Model/GetQueueAttributesResponse.extensions.cs
```diff
@@ -178,6 +172,28 @@ public string Policy
...
+        {
+            get
+            {
+                return getAttributeValue(SQSConstants.ATTRIBUTE_FIFO_QUEUE)?.ToUpper().Equals("TRUE");
+            }
+        }
+
+        /// <summary>
+        /// Whether or not the First-In-First-Out (FIFO) queue has content based deduplication enabled
```
PavelSafronov: The comment should be clear that this will be `null` if the queue is not FIFO.

sdk/src/Services/SQS/Custom/Model/GetQueueAttributesResponse.extensions.cs
```diff
@@ -178,6 +172,28 @@ public string Policy
...
+        /// </summary>
+        public bool? ContentBasedDeduplication
+        {
+            get
+            {
+                return getAttributeValue(SQSConstants.ATTRIBUTE_CONTENT_BASED_DEDUPLICATION)?.ToUpper().Equals("TRUE");
```
PavelSafronov: Use `Boolean.TryParse` and if `TryParse` returns `false`, return `null` for this property.

sdk/test/IntegrationTests/Tests/SQS.cs
```diff
@@ -265,5 +261,40 @@ private string createQueueTest(string name)
...
             return "fail";
         }
+
+        [TestMethod]
+        [TestCategory("SQS")]
+        public void SQSFIFOTest()
+        {
+            string fifoQueueName = prefix + new Random().Next() + ".fifo";
+
+            var sqsClient = new AmazonSQSClient(RegionEndpoint.USEast2);
```
PavelSafronov: To make sure the client is always disposed, wrap its use in a `using` block:
```
using(var sqsClient = new AmazonSQSClient(RegionEndpoint.USEast2))
{
}
```

This way you don't need to remember the `sqsClient.Dispose()` call and `Dispose` will be called in the event of an exception, like if an `Assert` or the queue creation fails.

The code tests FIFO queues, but for completeness this test should also create a non-FIFO queue and verify that we get expected values for the new properties.

Singh400: Tackled in a8686a9

sdk/src/Services/SQS/Custom/Model/GetQueueAttributesResponse.extensions.cs
Singh400: I wouldn't want to null check this value if I was consuming the API. So I've made it returned false where the value is null. See 0e2b3b0.

sdk/src/Services/SQS/Custom/Model/GetQueueAttributesResponse.extensions.cs
Singh400: Thanks, good idea. See 0e2b3b0.

sdk/src/Services/SQS/Custom/Model/GetQueueAttributesResponse.extensions.cs
Singh400: Thanks, good idea. See 0e2b3b0.

sdk/src/Services/SQS/Custom/Model/GetQueueAttributesResponse.extensions.cs
Singh400: Tackled in 0e2b3b0.

Singh400:
@PavelSafronov Think I have addressed your comments. Let me know if anything needs tidying up :)

sdk/src/Services/SQS/Custom/Model/GetQueueAttributesResponse.extensions.cs
```diff
@@ -178,6 +172,44 @@ public string Policy
...
+            {
+                var attributeValue = getAttributeValue(SQSConstants.ATTRIBUTE_FIFO_QUEUE);
+
+                if (attributeValue == null || attributeValue.Trim() == string.Empty)
+                {
+                    return false;
+                }
+
+                bool result;
+                return bool.TryParse(attributeValue, out result);
```
PavelSafronov: This works with the current service behavior: if the service returns *anything* for this attribute, it will be `true`, otherwise the service throws an exception.
If the service changes behavior in the future and does not throw an exception for non-FIFO queues and does return `true` and `false` values, this code will always return "true".

The safer approach here is to do this:
```
bool result;
if (bool.TryParse(attributeValue, out result))
    return result;
return false;
```

sdk/src/Services/SQS/Custom/Model/GetQueueAttributesResponse.extensions.cs
PavelSafronov: While working with `bool?` is more complicated for the end-user, it is the more correct type to use here, since this attribute has three states: ContentBasedDeduplication can be on or off for a FIFO queue, or undefined for a non-FIFO queue. So I still think `bool?` is appropriate here. (Note that FifoQueue does not have this triple-state issue, hence why `bool` works there.)

sdk/src/Services/SQS/Custom/Model/GetQueueAttributesResponse.extensions.cs
```diff
@@ -178,6 +172,44 @@ public string Policy
...
+            {
+                var attributeValue = getAttributeValue(SQSConstants.ATTRIBUTE_CONTENT_BASED_DEDUPLICATION);
+
+                if (attributeValue == null || attributeValue.Trim() == string.Empty)
+                {
+                    return false;
+                }
+
+                bool result;
+                return bool.TryParse(attributeValue, out result);
```
PavelSafronov: Similar comment about returning `result` from this method.

sdk/src/Services/SQS/Custom/Model/GetQueueAttributesResponse.extensions.cs
Singh400: Ah, I understand where you are coming from. I have made the changes in c30d5bb :)

sdk/src/Services/SQS/Custom/Model/GetQueueAttributesResponse.extensions.cs
Singh400: See c30d5bb.

sdk/src/Services/SQS/Custom/Model/GetQueueAttributesResponse.extensions.cs
Singh400: Ah! My bad. I completely missed that. Thanks for pointing that out. Rectified in	816b607.

Singh400:
@PavelSafronov ping.

sdk/src/Services/SQS/Custom/Model/GetQueueAttributesResponse.extensions.cs
```diff
@@ -181,32 +181,31 @@ public bool FifoQueue
...
                 {
-                    return false;
+                    return null;
                 }
 
                 bool result;
-                return bool.TryParse(attributeValue, out result);
+                if (bool.TryParse(attributeValue, out result))
+                    return result;
+                return false;
```
PavelSafronov: If we were not able to parse the service response, we should be returning `null`.

PR #669, opened 2017-06-07
Expose EncryptionMaterials MaterialsDescription property for use with Client Side Encryption

majic79:
## Description
Added the internal field MaterialsDescription so that EncryptionMaterials can optionally be stored in x-amz-matdesc metadata.

## Motivation and Context
Per aws developer forum thread : https://forums.aws.amazon.com/thread.jspa?messageID=608416
KMS envelope encryption on the client side permits KMS CMK to be used for envelope encryption client side prior to storage in S3. In order to provide transparency for other areas (Athena, EMRFS, HDFS etc,) KMS CMK ID must be stored, however this is not explicitly apparent. In all cases, the required key (x-amz-matdesc) is only populated with "{}" whereas by exposing the materials description dictionary, it can provide the necessary hint for server side decryption using the same KMS key

## Testing
Untested - presumed trivial, could not get SLN to build (VS2015 Update 3), could not test

## Types of changes

## Checklist

## License


majic79:
I did eventually manage to test this. I've added the feature I meant to add originally (stashing materialsDescription as Json in the item MetaData). When the materaisl description is emtpy (null) this still stashes "{}". With MaterialsDescription supplied, the dictionary is parsed and stashed as metadata.

vellozzi:
Although we don't have an ETA, we are working on adding KMS data key encryption functionality to the .NET client.  This PR will conflict with that work.

Storing the KMS CMK ID in the materials description doesn't do anything in itself.  The .NET client doesn't currently encrypt the data key using KMS.  So adding the KMS CMK ID to the metadata won't do anything.

Does that make sense?  Is there something about your use case that I'm missing?

majic79:
The AWS Developer forum thread I linked in the original post (https://forums.aws.amazon.com/thread.jspa?messageID=608416) contained a solution for a customer (e.g. myself) to implement KMS encryption client side. The data we are uploading we are querying with Athena. per this article http://docs.aws.amazon.com/athena/latest/ug/encryption.html, Athena supports three different modes of encryption, the one we're targetting is Client Side Encryption with KMS managed keys. Using the technique documented (our own Encyption Algorithm that leans heavily on KMS) we should be able to encrypt a .csv file and have Athena able to view the data.

With the .Net SDK in it's current state, the provided solution does not work, and as I don't know how AWS is able to do the required decryption, I looked at the Java SDK to try and identify the difference. I can see in there that they stash the kms_cmk_id in meta data, and I tried to duplicate that in my own solution. This doesn't work if I set the metadata in the Put object, as the SDK overwrites it with "{}" regardless of what I try to set it to. My PR attempts to address this to allow the storing of this metadata.

I'm gladdened to hear that there's effort ongoing to address the lack of support in the .Net SDK for KMS encryption - it's a shame there's no ETA, we wanted to do this two weeks ago, and the lack of support for the solution was a bit disheartening, especially given a solution for the problem was documented 2 years ago. Not being familiar with your coding standards (and somewhat lacking on my own time) I'm hesitant to offer to help with this

vellozzi:
The solution from that thread doesn't take into account some requirements, such as compatibility with the Java SDK.  I've commented on that thread to alert others to the shortcomings of that implementation.

majic79:
Indeed, and the SDK as it currently stands explicitly prevents setting at least one of the metadata values which is required. And it's that specific issue this PR is attempting to address. It's not aimed at being a full solution, it's just giving a way to override current behaviour which blocks any custom implementation. The ideal would be a full client side encryption implementation, but as you said, there's no ETA on that

vellozzi:
As a temporary workaround you can set the metadata on the object with a separate API call.  We'll continue to work on the full solution and get that released as soon as possible.

vellozzi:
Client side encryption using KMS keys has been released.  Here are the NuGet packages:
AWSSDK.S3 3.3.11
AWSSDK.Core 3.3.17.9
AWSSDK.KeyManagementService 3.3.3.1

You can read about the changes here:
https://aws.amazon.com/blogs/developer/using-aws-kms-master-keys-with-the-amazons3encryptionclient-in-the-aws-sdk-for-net/

Issue #499, opened 2016-11-30
Asp.Net Core Not picking up profile from appsettings.json


ralsu091:
Hello,

Following: https://aws.amazon.com/blogs/developer/configuring-aws-sdk-with-net-core/
I have setup appsettings.json as the following:
```json
{
  "AWS": {
    "Profile": "profileName",
    "Region": "us-east-1"
  }
}
```

My Startup.cs has the following:
```c#
public Startup(IHostingEnvironment env)
        {
            var builder = new ConfigurationBuilder()
                .SetBasePath(env.ContentRootPath)
                .AddJsonFile("appsettings.json", optional: true, reloadOnChange: true)
                .AddJsonFile($"appsettings.{env.EnvironmentName}.json", optional: true)
                .AddEnvironmentVariables();


            Configuration = builder.Build();

        }

public void ConfigureServices(IServiceCollection services)
        {
            var mapper = ConfigureAutoMapper.Configure();
            services.AddOptions();
            services.Configure<MongoOptions>(Configuration);
            services.AddSingleton<IHttpContextAccessor, HttpContextAccessor>();
            services.AddSingleton<IRepository, Repository>();
            services.AddSingleton(mapper);
            services.AddMvc();
            services.AddDefaultAWSOptions(Configuration.GetAWSOptions());
            services.AddAWSService<IAmazonSimpleEmailService>();

        }
```

When I deploy the website to a Windows Server 2008 R2, the website crashes on startup and the logs show the following:
```c#
Application startup exception: Amazon.Runtime.AmazonServiceException: Unable to find credentials

Exception 1 of 3:
System.ArgumentException: App.config does not contain credentials information. Either add the AWSAccessKey and AWSSecretKey properties or the AWSProfileName property.
   at Amazon.Runtime.StoredProfileAWSCredentials..ctor(String profileName, String profilesLocation)
   at Amazon.Runtime.FallbackCredentialsFactory.<>c.<Reset>b__6_0()
   at Amazon.Runtime.FallbackCredentialsFactory.GetCredentials(Boolean fallbackToAnonymous)

Exception 2 of 3:
System.InvalidOperationException: The environment variables AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY/AWS_SESSION_TOKEN were not set with AWS credentials.
   at Amazon.Runtime.EnvironmentVariablesAWSCredentials.FetchAWSCredentials()
[...]

PR #680, opened 2017-06-25
Support loading AccessKey and SecretKey from appsettings.json

davismj:
Since the "AWS_REGION" environment variable maps to "AWS:Region", the "AWS_ACCESS_KEY_ID" environment variable maps to "AWS:AccessKeyId" and the "AWS_SECRET_ACCESS_KEY" environment variable maps to "AWS:SecretAccessKey", like so:

```
"AWS": {
  "AccessKeyId": "abc",
  "SecretAccessKey": "xyz"
}
```

## Description
You can now add your access key and secret key directly to your appsettings.json instead of loading and registering profiles. The naming conventions have been inferred from the environment variable names. 

## Motivation and Context
Many developers including myself attempted to add their keys to the appsettings.json. This is a bit more straightforward that creating a custom profile, and this is more consistent with how the environment variables work. There are a few issues floating around that carry this sentiment including but not limited to https://github.com/aws/aws-sdk-net/issues/499.

## Types of changes

## Checklist

## License

### DIFF
--- a/extensions/src/AWSSDK.Extensions.NETCore.Setup/ConfigurationExtensions.cs
+++ b/extensions/src/AWSSDK.Extensions.NETCore.Setup/ConfigurationExtensions.cs
@@ -128,6 +128,14 @@ namespace Microsoft.Extensions.Configuration
                 options.Region = RegionEndpoint.GetBySystemName(section["AWSRegion"]);
             }
 
+            if (!string.IsNullOrEmpty(section["AccessKeyId"]) && !string.IsNullOrEmpty(section["SecretAccessKey"]))
+            {
+                options.Credentials = new BasicAWSCredentials(
+                    section["AccessKeyId"],
+                    section["SecretAccessKey"]
+                );
+            }
+
             return options;
         }
     }


sstevenkang:
I am not sure if this usage is something we want to support.  app.config and web.config generally are checked into source control which means your credentials is checked into repo.

davismj:
Yes, but you can't protect developers from their own poor decisions. Fact of the matter is that the preferred method for setting environment-specific variables in .net core is through setting ASPNETCORE_ENVIRONMENT which in turn loads an environment specific appsettings.json. At the moment, the only way for me to achieve the above behavior is to set global environment variables, and that's no good.

sstevenkang:
By providing this functionality, we are implicitly encouraging the poor practice.  Is there a reason to make this change for everyone?

davismj:
Visual Studio and VS Code both provide options for setting environment variables. They are stored in the code, launchsettings.json for VS and tasks.json for VS Code. A user can store secrets in there and they are likely to get committed. 

Are stupid developers going to do stupid things? Yes. Should that mean no one gets to have nice things? If anything, this can be resolved by adding a warning in the documentation not to commit secrets to repositories, even private repositories.

In the mean time, this solves the very issue you're trying to prevent. A user can configure development settings in appsettings.Development.json, a file that would be easily ignored when committing and deploying.

If your vote is still no, there is still a documentation problem. Numerous people including myself have come here opening up issues because they aren't sure why their appsettings.json is getting ignored. There is very little documentation around what keys are accepted. Those settings should be enumerated and a warning could be added informing the developer that the secrets cannot be read from appsettings.json.

normj:
To accept this I'm looking for a compelling reason to support credentials in appsettings files besides it is easier than dealing with profiles. We have seen too many users checkin their credentials and then push to GitHub to add supporting and continuing this happening. We support credentials in web.config for legacy reasons but if I was redoing things I would not add that support. I'm not saying no, but what is the use case you have that requires credentials over a profile.

And yes I totally agree we should do a better job with documentation and error messaging.

gibwar:
I honestly feel like the better approach would be to support the use of reading the credentials from `IConfiguration` (like this) but rather than using the `appsettings.json`, read the "normal" environment keys that work anyway. This enables the support of following Microsoft's best practice recommendation of [Safe storage of app secrets during development](https://docs.microsoft.com/en-us/aspnet/core/security/app-secrets) in a `secret.json` that is nowhere near the repository but also allows the quick and dirty of adding it directly to the `appsettings.json` file as well (not recommended).

This would also bring the configuration extensions more in line with what developers would expect between using the `secrets.json` file and setting the environment variables for production use.

We just ran in to this problem at work since apparently at some point in the past had used the awscli tools and ended up with a default profile and `.aws\credentials` file I wasn't aware of, and deploying with real environment variables worked as well, but new developers ran in to the problem of not being able to find the credentials even though it was added to their `secret.json` file.

normj:
I'm still waiting on a concrete example of why this is needed besides "I don't want to use profiles". Continuing the possibility of developers exposing their AWS credentials via a checkin is too dangerous to do without a good reason.

I know .NET Core has the secret configuration options but it is not on by default and I suspect many developers don't even know how to use it.

@gibwar your problem about picking up default credentials from the `.aws\credentials` file is an issue regardless of this specific question. In the SDK we have logic to search for credentials in a variety of places to make it easy to work in several different environments including locally and then deploying to AWS. The problem we haven't figured out yet is how do we communicate from a library where we got credentials from.

davismj:
Okay. I went with a credentials file instead. Worked great. BUT I CHECKED IN MY CREDENTIALS FILE BY ACCIDENT. MY ENTIRE APP WAS HACKED. WHY DID YOU LET ME DO THIS? SAVE ME FROM MYSELF.

I don't see the difference between these things other than the name of the files. However, the document isn't super clear for the credentials file, at least when I was working on it. Instead, seeing that AWS had an official configuration plugin that read from the appsettings.json file, I tried in vain for a couple hours to get AWS to pick up my credentials from there. It seems other developers have had the same experience.

Other than this, I don't have a better argument for you. Sadly, our experience with AWS (serverless, dynamodb, and s3) was not excellent and we're going to be moving away from S3 in the long run, so it's of little difference to me whether or not you implement this feature. I took the time to create a PR in hopes of saving future developers from the same pain. Feel free to close the PR if you feel it doesn't fit with your vision.

normj:
The credential file is by default under your <user-homey/.aws directory or if you are using the VS Toolkit encrypted under your user AppData along way from your source. It is the rare exception that people override the location of the credential file. I have never heard of anybody accidentally checking in the aws credentials file for any of our AWS SDK or Tools.

I do really appreciate the pull request and the team and myself have been debating about accepting this for awhile. I left it open looking to hear if there was a use case we were missing that warranted continuing the security vulnerability of putting AWS credentials into your source repository. A simple [google search](https://www.google.com/search?q=github+aws+credentials&oq=github+aws+credentials) will show how common it is for developers to checkin their credentials. And when they do check them in they don't realize when they remove them that they are still available in the git history.

I will do some work to the extension library to improve the messaging for how credentials are found and improve the error messaging when they are not found.

Sorry to hear your experience using AWS was not great. I would appreciate hearing what was not great and what we can do to improve the experience going forward.

Understandably we are coming at the issue from different sides and I do appreciate your point of view but I'm going to close the PR in preference of security.

marexv:
Not supporting loading of AccessKey and SecretKey from appsettings.json also complicates deployments to environments other than aws. 

I run all my python apps on elastic beanstalk, but once I saw that I can't use eb console for setting environment variables for my dotnet app (not out of the box at least) and that handling certificates is also difficult I migrated my dotnet apps to Azure app services.

I am sure that there is a way to get this working on azure with profile but I am not sure that I want to introduce new level of complexity to my app (it is another thing that you have to explain to OPS). So I will try to proceed with solution suggested in other threads related to this issue:

```
Environment.SetEnvironmentVariable("AWS_ACCESS_KEY_ID", Configuration["AWS:AccessKey"])
Environment.SetEnvironmentVariable("AWS_SECRET_ACCESS_KEY", Configuration["AWS:SecretKey"])
Environment.SetEnvironmentVariable("AWS_REGION", Configuration["AWS:Region"])
```

Also people keep database connection strings in appsettings.json so I would assume that they understand that this info is sensitive.

If somebody does commit keys to a public repo it is not that hard to resolve this with invalidating these keys and getting new ones. Hopefully developers will never see production keys because OPS will take care of those?

I understand that AWS is a big player and that you can do things your way and have an impact in global security, but as seen from example people will try to workaround the issue and in the end you will just end up loosing some developers along the path.

PR #715, opened 2017-08-03
DynamoDB Unmarshal Performance Improvements

bcuff:
## Description
Improves performance and memory consumption of UnmarshallerContext.TestExpression(...)

## Motivation and Context
Performance profiling shows this as a hotspot in our dynamodb-heavy application.

## Testing
I ran the DynamoDB Integration tests.

image

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Internal/Transform/UnmarshallerContext.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Transform/UnmarshallerContext.cs
@@ -147,8 +147,10 @@ namespace Amazon.Runtime.Internal.Transform
                     startingStackDepth++;
                 }
             }
-            return (startingStackDepth == currentDepth
-                    && currentPath.EndsWith("/" + expression, StringComparison.OrdinalIgnoreCase));
+            return startingStackDepth == currentDepth
+                   && currentPath.EndsWith(expression, StringComparison.OrdinalIgnoreCase)
+                   && currentPath.Length > expression.Length
+                   && currentPath[currentPath.Length - expression.Length - 1] == '/';
         }
 
         #region Abstract members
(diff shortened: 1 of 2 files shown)


PavelSafronov:
We can't take this PR in its current state. [AttributeValueUnmarshaller.cs](https://github.com/aws/aws-sdk-net/blob/master/sdk/src/Services/DynamoDBv2/Generated/Model/Internal/MarshallTransformations/AttributeValueUnmarshaller.cs) is a generated file, so any changes that you make to it will be overwritten next time we generate the SDK, which is almost daily. 

Changes to unmarshallers need to be made in the [code generator](https://github.com/aws/aws-sdk-net/tree/master/generator/ServiceClientGeneratorLib), and these changes will impact other, non-DynamoDB and non-AttributeValue unmarshallers. So that's going to be a rather large change, and much riskier for us to take.

There really isn't a good way to replace a single unmarshaller in the SDK (for instance, we use AttributeValueUnmarshaller in [other places](https://github.com/aws/aws-sdk-net/blob/master/sdk/src/Services/DynamoDBv2/Custom/DocumentModel/Util.cs#L503)), that aspect isn't as modularized as the rest of the SDK. For now, I think your best approach would be to fork this project and make these changes locally.

bcuff:
I see. @PavelSafronov Can you take https://github.com/aws/aws-sdk-net/pull/715/commits/8d89511c382bd737ff73dcf3f0846ffb939675b9 without the AttributeValueUnmarshaller change?

PavelSafronov:
Yes, that is a reasonable and small change. Though I would recommend moving the `EndsWith` call to be the final check, so the less-expensive range and char-equality checks are done first.

bcuff:
ah. good idea. I'll do that an re-submit. thanks.

PR #731, opened 2017-08-23
fixes DynamoDBContext API hanging problem on .NET CORE platform

dirty49374:
## Description
fixes DynamoDBContext API hanging problem on .NET CORE platform

## Motivation and Context
DynamoDBContext API internally calls AsyncRunner.Run() which creates new thread to process the request and wait for thread to join **synchronously** (**blocking an .net worker thread**). Because of this, when enough requests comes in s​imultaneously (more than free .net worker pool threads left), there are no worker threads left to resume after dynamodb http calls finishes and whole application hangs.

## Testing
I've tested it with below codes.
```
    class DynamoDbTest
    {
        public class TestDoc
        {
            public int Key { get; set; }
            public int Value { get; set; }
        }

        private static async Task SaveAsync(int key, DynamoDBContext context)
        {
            for (var value = 1; value <=1000; ++value)
            {
                await context.SaveAsync(new TestDoc {Key = key, Value = value});
                Console.WriteLine($"saved {key},{value}");
            }
        }

        internal static async Task TestAsync()
        {
            var context = new DynamoDBContext(new AmazonDynamoDBClient(RegionEndpoint.APNortheast2));
            await Task.WhenAll(Enumerable.Range(0, 100).Select(x => SaveAsync(x, context)));
        }
    }
```
## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Internal/_async/AsyncRunner.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/_async/AsyncRunner.cs
@@ -35,21 +35,28 @@ namespace Amazon.Runtime.Internal
             Task<T> task = Task.Run<T>(action);
             return task;
 #else
-            return Task<T>.Run(() =>
+            return Task<T>.Run(async () =>
             {
                 Exception exception = null;
                 T result = default(T);
-                Thread thread = new Thread(() =>
+
+                using (var semaphore = new SemaphoreSlim(0))
                 {
-                    try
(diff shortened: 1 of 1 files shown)


normj:
Thanks for the pull request. This was merged in version 3.3.17.7 of AWSSDK.Core. The dependency on AWSSDK.DynamoDbv2 wasn't updated yet. Until it is you will need to specifically add the AWSSDK.Core package to your project.

PR #764, opened 2017-10-04
Adds a "/" to request URI if required

njstoffel:
## Description
This change adds a check when composing a request URL and includes an intervening "/" between the base URL and resource path if there is not one included in either of those values already.

## Motivation and Context
This change is a fix for a bug that arises when an S3 request has a ServiceURL containing a subdirectory specified,  ForcePathStyle set to true, and v4 signing enabled. 

The CanonicalResourcePath is created for the AWS4Signer  by combining the ServiceURL with the ResourcePath (the BucketName and the Key) using CanonicalizeResourcePath at AWSSDKUtils.cs:282. Here it is assumed that there is no trailing  "/" in the ServiceURL or leading "/" in the BucketName/ResourcePath as one is placed in between them during concatenation. 

When the same request is composing the url against which the request will be made, these same values (ServiceURL and ResourcePath beginning with BucketName) are combined at AmazonServiceClient.cs:505. Here it is assumed that an intervening "/" is already present, as the values are concatenated without a "/".

This results in either the CanonicalResoucePath being incorrect by having an extra "/" or the request URL being incorrect by lacking the "/" between the subdirectory of the ServiceURL and the beginning of the BucketName, depending on whether or not the "/" is included when these values are assigned.

## Testing
Perform a request through an AmazonS3Client with configuration values formatted like the following:

Client:
ServiceURL = "example.com/sub/directory"
ForcePathStyle = true
(in AWSConfig) useSignatureVersion4 = true

Request:
BucketName = "bucket"
Key = "resource/path/filename.ext"

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/AmazonServiceClient.cs
+++ b/sdk/src/Core/Amazon.Runtime/AmazonServiceClient.cs
@@ -502,7 +502,10 @@ namespace Amazon.Runtime
                     "by System.Uri and thus cannot be handled by the .NET SDK.", resourcePath));
 
             var parameterizedPath = string.Concat(AWSSDKUtils.UrlEncode(resourcePath, true), sb);
-            var uri = new Uri(url.AbsoluteUri + parameterizedPath);
+            var hasSlash = url.AbsoluteUri.EndsWith("/") || parameterizedPath.StartsWith("/");
+            var uri = hasSlash
+                ? new Uri(url.AbsoluteUri + parameterizedPath)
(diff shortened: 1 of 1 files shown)


normj:
Thanks for the pull request. and the change looks good. I have another Core change I'm finishing up, when I complete that I'll include this change with that Core release.

normj:
Thanks again for the pull request. This went out with version 3.3.21.19 of AWSSDK.Core

Issue #766, opened 2017-10-08
AsyncRunner.CS performance bug affecting DynamoDB calls and client system as a whole


anmarht:
DynamoDB calls end up being executed in https://github.com/aws/aws-sdk-net/blob/master/sdk/src/Core/Amazon.Runtime/Internal/_async/AsyncRunner.cs. Code currently creates a Thread, eventhough it is reserving a thread on the Threadpool, then Joining both threads. This results in a high CPU load affecting the whole system, plus can't achieve highest performance possible.

## Expected Behavior
- CPU usage to align with performance
- High performance

## Current Behavior
Unneeded threads created. CPU goes high. Low performance.

## Possible Solution
I've created the fix on my forked repo:

https://github.com/anmarht/aws-sdk-net/commit/61d002a1c36f7dfedbb6b26af54e10626f7a3d1c

### Some test results:

Below are test results of using UpdateItemAsync to update 10000 items.

|                      | Original AWS SDK    | After applying my fix |
|----------------------|---------------------|-----------------------|
| Updating 10000 items | 48,254 milliseconds | 9,830 milliseconds    |

## Steps to Reproduce (for bugs)

Here are 2 test applications, to show performance with the original code, and after fixing it as well:

https://github.com/anmarht/aws-sdk-bug

## Context
Making a lot of DynamoDB Updates in a heavily multi-threaded application would slow the system and not achieve adequate performance. I've noticed this because when dealing with other services performance was very good (like S3 or SQS), except DynamoDB so i've done my research to find if there's any possible bug. 

## Your Environment
- Used the SDK code directly at https://github.com/aws/aws-sdk-net
- Also used Nugets AWSSDK.DynamoDBv2 v3.3.4.17 and AWSSDK.Core v3.3.17.9
- OS: Windows 10 64bit Core i7 with an SSD
- Microsoft Visual Studio Enterprise 2017 V15.3.3
- .NET: .NET 4.6.2 (or SDK solution AWSSDK.Net45.sln)

[logging]: http://docs.aws.amazon.com/sdk-for-net/v3/developer-guide/net-dg-config-other.html

alfattal:
Congrats on this achievement!

normj:
This was fixed with PR https://github.com/aws/aws-sdk-net/pull/896

PR #896, opened 2018-03-10
DynamoDB - Remove AsyncRunner and blocking on async

slang25:
As reported in #766, the DynamoDB client does a complex version of `Task.Run(...` in addition to `Thread.Join`, which results in terrible thread usage. There have been efforts to simplify and reduce the complexity to `AsyncRunner`, I think it should be removed altogether.

## Description
I removed `AsyncRunner` and worked backwards, with the following strategy:
- If a method calls a synchronous method where an async version exists, make an async copy of that method.
- Pass through `CancellationToken`s.
- If the method is public, make the `CancellationToken` optional, otherwise require it.
- Use `.ConfigureAwait(false)` when awaiting
- If the only async is the last method called, just return that `Task` and don't mark the method `async`.
- Where there is both a sync and an async version of a method, and it has an `isAsync` bool parameter, I remove it and inline the appropriate value, with explicit parameter name so that it remains clear what it's purpose is.

There is a fair bit of duplication in some of the lengthier methods, the synchronous parts could be pulled out into private methods, but I find that approach leads to unnatural method names and an extra layer of indirection.

## Motivation and Context
This has already been identified as a performance issue. It is important that when a method appears to be async, it uses no threads for IO bound work, the current code uses many threads at once for IO bound work and requires attention as people are refusing to adopt this library as a result.

## Testing
Unit tests and integrations tests pass

## Types of changes

Technically, this is a breaking change as there is a virtual method (`ExecuteHelper`) on `BatchGet` and `BatchWrite` that has had a parameter removed, and these are public abstract classes, however these classes don't look like they are something people should be extending so I have additionally marked the methods as `internal`. These methods are also undocumented.
I would say this would warrant a minor version change not major (not sure how that fits into AWSSDK versioning).

## Checklist
- [x] The existing tests cover my changes

## License


slang25:
@normj Could you take a look when you have some time

danielmarbach:
@slang25 great change and great application of the async best practices. Well done 🎆

normj:
I agree it looks good. We will get this pushed out.

slang25:
Awesome. Just a heads up, I've just discovered that I need to do some changes to support the other solutions, so I'll see if I can get those working too before removing `WIP` from the title.

slang25:
@normj This is ready to be looked at now

FireNero:
@normj any ETA of this being merged and released?

InvertedAcceleration:
@normj this is critically important to our company to get this merged and released.

FireNero:
I've also performed some dirty performance testing of this patch and here are some results:

| Requests count | Old implementation                    | With PR fix                         |
|:----------------:|:---------------------------------------:|:-------------------------------------:|
|      1374      | 12516ms spent<br>1397 threads created |  5337ms spent<br>44 threads created |
|      2748      |             25032ms spent             |             9124ms spent            |
|      4580      | 40836ms spent<br>4607 threads created | 14401ms spent<br>34 threads created |

It is a very dirty benchmark, so numbers are not 100% accurate, but improvement is evident.

normj:
Thanks for the performance numbers. Those are exciting numbers. I haven't had a chance to take care of this pull request yet but I will try to get to it as soon as I can.

normj:
Status update, I'm going through this now and hit a snag with the Unity build. I'll see what we need to do to get Unity working again.

slang25:
Thanks @normj, let me know if I can help.

normj:
Another status update. I've gone through the changes and ran through our build system and everything looks good. Given the large change though I'm having another one of our devs double check everything before releasing it. I appreciate your patience and hope to get this out soon.

normj:
Version 3.3.9.0 of the package `AWSSDK.DynamoDBv2` was released with this pull request. Thanks again for the awesome work.

AnthonySteele:
Is possible that similar code occurs in publishing a message to SNS?

normj:
@AnthonySteele The problem here was we wrote these abstractions before async took over everything and we never had the chance to update the high levels to have a pure async only code path. For SNS there is no high level abstraction you are just working with the service client, `AmazonSimpleNotificationService`, which is generated with a sync and async code path. So you shouldn't be having the same problem.

slang25:
Thanks again @normj and your team for making this a priority 🎉

Issue #894, opened 2018-03-07
[S3] PutObject High Memory Usage


Singh400:
Whilst under taking performance work centred around memory, we have discovered that this method:-

https://github.com/aws/aws-sdk-net/blob/9814e245e0cd9c4435a17185e866813869d427da/sdk/src/Services/S3/Generated/_bcl45/AmazonS3Client.cs#L4421

Causes LOH allocations as well as fragmentation. As we are uploading small sized images (many thousand times a day) we can't see why this should be the case. This was revealed by taking dumps of production servers and dumping out the byte arrays. The byte arrays in question contained `;chunk-signature=` with image data afterwards.

Using dotTrace we have been able to narrow down the allocations to `ChunkedUploadWrapperStream` specifically the constructor:-

https://github.com/aws/aws-sdk-net/blob/9814e245e0cd9c4435a17185e866813869d427da/sdk/src/Core/Amazon.Runtime/Internal/Util/ChunkedUploadWrapperStream.cs#L73

For every invocation of the constructor two byte arrays are created each having a minimum size of `131072`:-

https://github.com/aws/aws-sdk-net/blob/9814e245e0cd9c4435a17185e866813869d427da/sdk/src/Core/Amazon.Runtime/Internal/Util/ChunkedUploadWrapperStream.cs#L40

The threshold for objects to go onto the LOH is 85,000 bytes. We believe this is the root cause of the problem we are experiencing.

## Expected Behavior
That uploading small files doesn't cause LOH allocation and further on down the line fragmentation.

## Current Behavior
Every time you upload an image to S3 two byte arrays are created with a minimum size of 131072. This is greater than the 85,000 threshold causing both byte arrays to be allocated directly to the LOH.

## Possible Solution
There are a few possible solutions:-

1. Use [System.Buffers](http://adamsitnik.com/Array-Pool/) to have a pool of byte arrays that can be recycled.
2. Use [Microsoft.IO.RecyclableMemoryStream](http://www.philosophicalgeek.com/2015/02/06/announcing-microsoft-io-recycablememorystream/) to have a pool of MemoryStreams - we have used this to great effect in our own code to reduce LOH allocations by 99.6%.
3. Expose `DefaultChunkSize` so that consumers of the API can set it themselves
4. Lower `DefaultChunkSize` to a number that is by default lower than the LOH threshold. Microsoft use 81920 in `Stream`:- https://referencesource.microsoft.com/#mscorlib/system/io/stream.cs,50

[...]

PR #899, opened 2018-03-13
[S3] Fixing PutObject High Memory Usage

Singh400:
See https://github.com/aws/aws-sdk-net/issues/894 for detailed description

## Description
Previously, when invoking the constructor of `ChunkedUploadWrapperStream` it used to create two byte arrays each with a size of `131,072` (`128*1024`). From a memory aspect this is inefficient as objects over `85,000` bytes go straight into an area of memory known as the Large Object Heap (LOH). This area is never compacted by default (as compaction of the LOH is expensive), but is collected. Over a lifetime of an application it is possible to end up in a situation where the LOH has taken up most of the RAM.

Now, the byte arrays are still created but with a size of `81,920`. The reason for this number can be found here: https://referencesource.microsoft.com/#mscorlib/system/io/stream.cs,50. As `81,920` is under the `85,000` threshold the byte arrays in question go onto the Small Object Heap (SOH). This area of memory is collected **and** compacted by default.

## Motivation and Context
To reduce memory consumption of this call:-
https://github.com/aws/aws-sdk-net/blob/9814e245e0cd9c4435a17185e866813869d427da/sdk/src/Services/S3/Generated/_bcl45/AmazonS3Client.cs#L4421

| |Total | SOH | LOH
------------ | ------------- | ------------- | -------------
Before|114MB|64MB|50MB
After|98MB|98MB|0.4MB

## Testing
See https://github.com/aws/aws-sdk-net/issues/894 for benchmarks within a test harness for pre/post change.

See https://github.com/aws/aws-sdk-net/issues/894 for screenshots

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Internal/Util/ChunkedUploadWrapperStream.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Util/ChunkedUploadWrapperStream.cs
@@ -37,7 +37,7 @@ namespace Amazon.Runtime.Internal.Util
     /// </summary>
     public class ChunkedUploadWrapperStream : WrapperStream
     {
-        public static readonly int DefaultChunkSize = 128*1024;
+        public static readonly int DefaultChunkSize = 81920;
 
         private const string CLRF = "\r\n";
         private const string CHUNK_STRING_TO_SIGN_PREFIX = "AWS4-HMAC-SHA256-PAYLOAD";


Singh400:
@sstevenkang Is there anything I can do here to get this PR merged?

Thanks,
Indy

sstevenkang:
Interesting.  Let me check with the team regarding the history of choosing 128*1024 as the default size.

Singh400:
Hi @sstevenkang,

Any news?

Thanks,
Indy

normj:
Thanks for the pull request. This went out with version 3.3.21.19 of AWSSDK.Core

Singh400:
Thanks for accepting this, really appreciate it. FYI `GetPreSignedURL` offers much better performance within a memory context; https://github.com/aws/aws-sdk-net/issues/894#issuecomment-378058075

PR #908, opened 2018-03-28
Mobile implementation to allow to use a custom Http message handler, like ModernHttpClient. (Core and S3 Service)

Magic73:
The actual mobile implementation doesn't allow to use a custom Http message handler, like ModernHttpClient.
This will patch will affect the S3 service, but can be ported to other projects who internally create HttpClient object.

**Motivation and Context**
I have to use AWS SDK liberary in a mobile project. But I have to integrate Mobile Iron support, so I have to use a specific set of Http Client. In may case I have to use ModernHttpClient library.
Actually I have to use AWS source code directly, but it would be great if this request would be added directly to the AWS code.

**Testing**
S3 example:
var config = new AmazonS3Config();
config.SetHttpMessageHandler(new ModernHttpClient());
:
var s3 = new AmazonS3Client(credentials, config);

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_mobile/HttpRequestMessageFactory.cs
+++ b/sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_mobile/HttpRequestMessageFactory.cs
@@ -163,7 +163,7 @@ namespace Amazon.Runtime
 
         private static HttpClient CreateHttpClient(IClientConfig clientConfig)
         {
-            var httpMessageHandler = new HttpClientHandler();
+            var httpMessageHandler = clientConfig.GetHttpMessageHandler() ?? new HttpClientHandler();
 #if CORECLR
             if (clientConfig.MaxConnectionsPerServer.HasValue)
                 httpMessageHandler.MaxConnectionsPerServer = clientConfig.MaxConnectionsPerServer.Value;
--- a/sdk/src/Core/Amazon.Runtime/_mobile/ClientConfig.cs
+++ b/sdk/src/Core/Amazon.Runtime/_mobile/ClientConfig.cs
@@ -27,6 +27,7 @@ namespace Amazon.Runtime
     /// </summary>    
     public abstract partial class ClientConfig
     {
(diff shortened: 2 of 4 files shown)


normj:
Thanks for the pull request. I want to hold off on taking it right now because it is on our backlog to rework how we cache the HttpClient to better handle things like DNS changes. I worry this pull request might interfere with that work. Once we do that work I can take another look at pulling in this pull request.

Also what was the use case of passing in a Func that creates the HttpClientHandler?

Magic73:
Hi Norm,
it shouldn't interfere, or at least it would be the base... since you have just to add a default HttiClientHandler and let the user to override it with an external implementation.

On mobile, at least in my case, I have to use a specifi HttpClientHandler so that my client which use an MDM (Mobile Device Management) like MobileIron, can depoly the app in its internal enterprise store.

MobileIron, like other MDM, can inspect the app, and override the default http calls in order to allow the app to work in a reserved VPN.

More details: https://www.mobileiron.com/en/products/tunnel

As a dev, I have to just use the HttpClient that they handle, ignoring how the VPN is. They they'll change/update the app in order to do all their stuff under the hood.

Without this change, I'm not able to use the SDK.
right now, I'm using a local version of the SDK.

matteo-prosperi:
You should be able to achieve this by setting an `HttpClientFactory` in the service client configuration.
You will need to create a class that extends `HttpClientFactory` and provides the service clients with an `HttpClient` that is configured according to your needs.
This is an example:
```
class HttpInjectorFactory : HttpClientFactory
{
  private readonly HttpClient InjectedClient;
  public HttpInjectorFactory(HttpClient injectedClient)
  {
    InjectedClient = injectedClient;
  }
  public override HttpClient CreateHttpClient(IClientConfig clientConfig)
    => InjectedClient;
  public override bool DisposeHttpClientsAfterUse(IClientConfig clientConfig)
    => false;
  public override bool UseSDKHttpClientCaching(IClientConfig clientConfig)
    => false;
}
```

matteo-prosperi:
Closing this since it is very old and the SDK should have capability to achieve this now.

PR #915, opened 2018-04-03
Expose Custom Logging

slang25:
## Description
I've moved `InternalLogger` up into the `Amazon.Runtime` namespace, renamed it to `BaseLogger` and made it public. Then I added the ability to pass in an instance into the config, and an enum option to enable it.

## Motivation and Context
Currently on .NET Core there is no nice way to get logs. If we targeted `netstandard2.0` we could use the `TraceSource` stuff as a bridge. We could even use the PLC implementation of `FileLogger`, although I would like to be able to provide my own logger implementation still.

I went down the route of using reflection to interact with `Microsoft.Extensions.Logging.Abstractions` so that you could pass in an instance of a `Logger` (as an untyped `object`) and do it the same way as with `log4net`, but I think that it is too much kung-fu for what should be a first class abstraction.

These APIs have been stable for long enough that they can now be made public IMO.

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/BaseLogger.cs
+++ b/sdk/src/Core/Amazon.Runtime/BaseLogger.cs
@@ -1,0 +1,74 @@
+using System;
+
+namespace Amazon.Runtime
+{
+    /// <summary>
+    /// Abstract logger class, base for any custom/specific loggers.
+    /// </summary>
+    public abstract class BaseLogger
+    {
+        public Type DeclaringType { get; private set; }
+
+        public bool IsEnabled { get; set; }
+
+        public BaseLogger(Type declaringType)
+        {
+            DeclaringType = declaringType;
+            IsEnabled = true;
+        }
+
+        #region Logging methods
+
+        /// <summary>
+        /// Flushes the logger contents.
+        /// </summary>
+        public abstract void Flush();
+
+        /// <summary>
+        /// Simple wrapper around the log4net IsErrorEnabled property.
+        /// </summary>
+        public virtual bool IsErrorEnabled { get { return true; } }
+
+        /// <summary>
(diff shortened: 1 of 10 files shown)


slang25:
This is just to get the discussion started.

Would this be something that could be accepted, or would it require the core team to think it over and design it properly?

slang25:
Hey @ashishdhingra thanks for the message, my previous comment still stands (regardless of the staleness of the PR), this PR is more a starting point for a discussion about if/how you want to support alternative logging.

My need for this is way less than it once was now that we can plug in the OTEL AWS contributor packages to get really good insights into what the SDK is doing.

sdk/src/Core/Amazon.Runtime/BaseLogger.cs
```diff
@@ -0,0 +1,74 @@
+using System;
+
+namespace Amazon.Runtime
+{
+    /// <summary>
+    /// Abstract logger class, base for any custom/specific loggers.
+    /// </summary>
+    public abstract class BaseLogger
```
peterrsongg, line 8: While I understand wanting to expose the `Logger` class, switching it from `Internal` to `Public` means we won't be able to make any backwards incompatible changes since it would now be publicly facing. This is probably why we kept it internal in the first place. I think this is more of a design question for the team which I will bring up.

boblodgett:
Thank you for the PR but this will require team design discussions to support what we envision for logging in the future. We recognize the need to revamp how logging is done in the SDK including providing the functionality for custom logging. Perhaps this could be done by defining abstract logging where users can register a plugin for the logging framework of choice. Because we want to take a holistic look at logging and a pass at the design, I am closing this PR.

PR #919, opened 2018-04-07
Fix common race condition in RefreshingAWSCredentials

slang25:
Again, this PR is to start a discussion. To fix the race conditions that occur when using the async APIs, I have made some tweaks.

This does not completely fix this code, there really needs some mechanism to provide single access to stuff in the async paths, for example use of `SemaphoreSlim`, however this adds questions around does it need to be disposed (and I think the answer is no, maybe) and this class doesn't implement `IDisposable`.

## Description
- Changed `ShouldUpdate` property to `CredentialsEmptyOrExpired()` method, for readability.
- Changed `Task.Run` to `Task.FromResult` to reduce thread wastage.
- Introduced double checked locking, so that concurrent requests don't do unneeded work.
- Optimise for the case where the method returns synchronously, which should be the case where the `GenerateNewCredentialsAsync` method is not overridden (most cases).

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Credentials/RefreshingAWSCredentials.cs
+++ b/sdk/src/Core/Amazon.Runtime/Credentials/RefreshingAWSCredentials.cs
@@ -85,7 +85,7 @@ namespace Amazon.Runtime
             lock (this._refreshLock)
             {
                 // If credentials are expired, update
-                if (ShouldUpdate)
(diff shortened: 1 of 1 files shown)


slang25:
It looks like most of this is now redundant due to the fixes here: https://github.com/aws/aws-sdk-net/commit/66fd44997b51e4ca63b086d9569da5dd4b80d5ac

Closing

Issue #940, opened 2018-04-26
New IConfigurationProvider Extension for SSM Parameter Store


KenHundley:
Hi AWS team,

I have created an implementation of Microsoft.Extensions.Configuration.IConfigurationProvider with SSM Parameter store as the backing data source. This can be added to the array of configuration providers at the startup of an Asp.Net Core app, allowing you to get your configuration values from SSM in addition to json files, environment variables, etc.

This enhancement seems like it would fit very nicely into this project under the "extensions" section, but I wanted to see if this submission is something you would be interested in before I took the time to fork and add all the necessary updates to fit it into this project.

The basic use looks like:
```
.ConfigureAppConfiguration((hostingContext, configBuilder) =>
{
    configBuilder.AddJsonFile("appsettings.json", true, true);
    configBuilder.AddEnvironmentVariables();

    var config = configBuilder.Build();
    configBuilder.AddAmazonSimpleSystemsManagement(source =>
    {
        source.AwsOptions = config.GetAWSOptions();
        source.Prefix = "/app/environment/etc"; //or config["SSM:Prefix"], etc.
        source.ReloadAfter = TimeSpan.FromHours(1); //optional
    });
})
```

Let me know and I will work on this over the weekend.

normj:
Agreed, that would be a great extension to have. It is something we have talked about doing but haven't been able to find time. Do you have a repo I can check out?

KenHundley:
not yet, just local development, but I can can easily upload a copy later tonight when I get home

KenHundley:
Hi @normj,

Check out: [This sample code](https://github.com/KenHundley/Amazon.Extensions.Configuration.AmazonSimpleSystemsManagement)

It's fully functional and the extension classes should ideally only need some cleanup, exception handling, and xml comments updates.

Before you run it you should run the following commands to insert the needed SSM Params:
```
aws ssm put-parameter --name "/myapp/dev/applicationsettings/settingsssmplain" --value "FromSSM-Plain" --type String --overwrite
aws ssm put-parameter --name "/myapp/dev/applicationsettings/settingsssmsecured" --value "FromSSM-Secured" --type SecureString --overwrite
```
If you have input or concerns just let me know.

KenHundley:
Also, I'm pretty sure I can create a version of this for the new Secrects Manager service. I'll look at that over the next few days.

KenHundley:
Hey @normj,

[...]

PR #961, opened 2018-05-23
New Extensions.Configuration.AWSSystemsManager for reading ASP.Net Core configuration values from Systems Manager Parameter Store

KenHundley:
## Description
I created this project as a way for an ASP.Net Core application to read some or all of it's configuration values from the Systems Manager Parameter Store. This is built on top of the Microsoft Configuration Extensions framework and is identical in use to `.AddEnvironmentVariables()`, `.AddJsonFile("path")`, etc. 

The basic configuration values are:
* Path: Path that will be queried (recursively) in Systems Manager
  * Required: maps to the path in `.GetParametersByPathAsync`
* AWSOptions: Lets you specify the credential options used to communicate with Systems Manager
  * Optional: Will attempt to create one for you if not supplied
* ReloadAfter: TimeSpan for how often to reload the parameters from Systems Manager
  * Optional: Used with `IOptionsSnapshot` to re-retrieve values that may change over time
* OnLoadException: Delegate for capturing errors encountered while calling Systems Manager
  * Optional: Lets the application respond to captured exceptions, as well as choosing to ignore them (not throw them)

Sample use (although there are many configurations available)
```csharp
.ConfigureAppConfiguration(builder =>
{
    builder.AddAWSSystemsManager("/myapp/dev/");
})
```

Related to Question: https://github.com/aws/aws-sdk-net/issues/940

## Motivation and Context
Greatly simplifies using Systems Manager Parameter Store as a source for configuration information for an ASP.Net Core application.

## Testing
I have tested this library code by utilizing it in an ASP.Net Core Web API project. I have tested all the configurations and also left it running for multiple days to look for any issues that could be encountered over time. 

I have also added unit tests for all the new code, with one caveat. Since the code relies on `awsOptions.CreateServiceClient<IAmazonSimpleSystemsManagement>()` but there aren't any interfaces on AWSOptions for mocking this call I had to create an internal class that wraps these operations in an injectable way. However, when I was done it worked like a charm and I was able to mock out those dependencies without changing AWSOptions to match

## Types of changes

- [x] New feature (non-breaking change which adds functionality) **(completely new project)**

## Checklist

  **(projects like AWSSDK.Extensions.NETCore.Setup aren't currently documented in the README.md, so I did not add this project, however, I did fully document the code with XML Docs)**
  **(completely new project, all new tests pass)**

## Open Items
* I was not sure how to version this since it's brand new. Normally this would be 1.0.0, but in this case it encompassed in a larger project. I'm hoping you can help resolve this
* I created a nuspec file for this, but you may want to review the text I used. I tried to follow the examples from NETCore.Setup.

## License


KenHundley:
np :)

Feel free to move around, rename, or otherwise restructure anything you see fit, or ask me to adjust. I tried to follow the established practices I saw in other projects, but there were some places where it wasn't clear what you would prefer, so I made a best guess.

KenHundley:
Hey Norm,

Have you had a chance to review this. No rush, I'm just interested in your thoughts/input.

normj:
I haven't had a chance to yet, I have been busy prepping for a conference. I am super interested in this and hope to find time soon to dive in.

KenHundley:
Hey Norm,

Any chance to review yet? Thoughts?

ulfsa:
I love this!

For inter-operability reasons however I would prefer to have the AWS Parameter Store secrets (and non-secrets) available at run time for me in some /secrets/secrets.json file (so much like an appsettings.json file). This way it would be independent of the cloud provider.

The only way I can figure that might work is to have the Dockerfile contain some RUN awscli installation line + producing some secrets.json by echoing the results of aws cli get-parameters calls = quite a nasty implementation.

Thoughts?

reflectivecode:
@ulfsa It sounds like you might be interested in confd with the ssm backend.

ulfsa:
Confd looks interesting!

However it would still not be as independent as "somehow" creating a /secrets/secrets.json containing ssm-stored parameters (rendered in clear text) to be used by the ConfigureAppConfiguration's config.AddJsonFile("/secrets/secrets.json", optional: true).

Still unsure as to how to get it done the easiest way...

normj:
Now that Lambda .NET Core 2.1 is out I have time to take a look. Sorry for radio silence I needed to buckle down and finish off that work.

KenHundley:
@ulfsa I could be misunderstanding but what advantage would that have over this implementation? 

ReloadAfter is optional, so you could just load them at app start and never rely on the Param Store after that point. Wouldn't that be the same as loading them into a file at startup? Plus, for local or dev use just use the AddJsonFile or AddEnvironmentVariables instead and you don't have a reliance on the Param Store either.

I think, with the right configuration, this solves your described use case

KenHundley:
oh, your are wanting to embed the settings into your container? If that's the case then yea this won't solve your problem. You'll need to produce something that will query and embed it into your image at build time.

KenHundley:
@normj Great! I look forward to your input, and I'm very excited to hear about 2.1 support. We are already looking to utilize it on some new projects ;)

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSDK.Extensions.Configuration.AWSSystemsManager.csproj
```diff
@@ -0,0 +1,12 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard1.3</TargetFramework>
```
normj: Let's just start at netstandard2.0. Before 2.0 there are weird dependency references issue it would be nice to avoid and most people are just using netstandard2.0 anyway. It also reduces are test coverage requirements and I can't imagine there are very many people still using ASP.NET Core 1.x and would also be doing major work like changing their configuration

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSDK.Extensions.Configuration.AWSSystemsManager.nuspec
```diff
@@ -0,0 +1,30 @@
...
+    <authors>Amazon Web Services, Ken Hundley</authors>
+    <description>ASP.NET Configuration Extensions for AWS Systems Manager</description>
+    <language>en-US</language>
+    <licenseUrl>http://aws.amazon.com/apache2.0/</licenseUrl>
+    <projectUrl>https://github.com/aws/aws-sdk-net/</projectUrl>
+    <tags>AWS Amazon aws-sdk-v3</tags>
+    <iconUrl>https://sdk-for-net.amazonwebservices.com/images/AWSLogo128x128.png</iconUrl>
+	
+    <dependencies>
+      <group targetFramework="netstandard1.3">
```
normj: Again I think we should start the baseline at 2.0

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSDK.Extensions.Configuration.AWSSystemsManager.nuspec
```diff
@@ -0,0 +1,30 @@
...
+        <dependency id="AWSSDK.Extensions.NETCore.Setup" version="[3.3.0, 3.4)" />
+        <dependency id="AWSSDK.SimpleSystemsManagement" version="[3.3.15, 3.4)" />
+        <dependency id="Microsoft.Extensions.Configuration" version="1.1.0" />
```
normj: Start at 2.0.0

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSystemsManagerExceptionContext.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// <summary>
+        /// The <see cref="T:Microsoft.Extensions.Configuration.AWSSystemsManagerConfigurationProvider" /> that caused the exception.
+        /// </summary>
+        public AWSSystemsManagerConfigurationProvider Provider { get; set; }
+
+        /// <summary>The exception that occured in Load.</summary>
+        public Exception Exception { get; set; }
+
+        /// <summary>If true, the exception will not be rethrown.</summary>
+        public bool Ignore { get; set; }
```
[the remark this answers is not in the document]
KenHundley: I don't mind changing this, but I don't think we should. I actually named all this to be consistent with the other providers like AddJsonFile

```csharp
builder.AddJsonFile(source =>
{
    source.OnLoadException = context =>
    {
        var ex = context.Exception;
        context.Ignore = true;
    };
});
```

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSystemsManagerConfigurationProvider.cs
```diff
@@ -0,0 +1,113 @@
+﻿/*
+ * Copyright 2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
KenHundley: Sure, I'll change this. I copied these blurbs from the AWSSDK.Extensions.NETCore.Setup project which had 2016, so I left them there in case it was that way for a reason.

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSDK.Extensions.Configuration.AWSSystemsManager.csproj
KenHundley: funny, I actually had everything as 2.0 to start, but thought "I bet they will want all this to be backwards compatible as much as possible" :)  
I'll get this updated and adjust any packages or other items as needed

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSystemsManagerConfigurationProvider.cs
```diff
@@ -0,0 +1,113 @@
...
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Amazon.Extensions.NETCore.Setup;
+using Amazon.SimpleSystemsManagement.Model;
+using AWSSDK.Extensions.Configuration.AWSSystemsManager;
+using Microsoft.Extensions.Primitives;
+
+// ReSharper disable once CheckNamespace
+namespace Microsoft.Extensions.Configuration
```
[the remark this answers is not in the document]
KenHundley: Everything else goes in `AWSSDK.Extensions.Configuration.AWSSystemsManager`?
or you prefer something starting with `Amazon.` for consistency with other SDK projects?

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSystemsManagerProcessor.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        public async Task<List<Parameter>> GetParametersByPath(AWSOptions awsOptions, string path)
+        {
+            using (var client = awsOptions.CreateServiceClient<IAmazonSimpleSystemsManagement>())
+            {
+                var parameters = new List<Parameter>();
+                string nextToken = null;
+                do
+                {
+                    var response = await client.GetParametersByPathAsync(new GetParametersByPathRequest { Path = path, Recursive = true, WithDecryption = true, NextToken = nextToken });
```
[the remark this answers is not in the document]
KenHundley: Actually I have some more details that we should consider
>If you specify a parameter hierarchy using forward slashes, the hierarchy can have a maximum of five levels, as shown in this example: /Level-1/Level-2/Level-3/Level-4/Level-5/parameter-name
Also, ":" isn't allowed. Just ".-_"

So if we want to allow collapsed hierarchies or allow depths greater than 5 then we will need to support a different convention. My suggestion is __ since it's consistent with `.AddEnvironmentVariables`
>For hierarchical config values specified in environment variables, a colon (:) may not work on all platforms. Double underscore (__) is supported by all platforms.
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-2.1&tabs=basicconfiguration#configuration-by-environment

However, I'm open to other suggestions.

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,5 @@
+﻿
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("Configuration.AWSSystemsManagerTests")]
```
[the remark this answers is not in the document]
KenHundley: Makes sense. I'll implement this

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSystemsManagerConfigurationProvider.cs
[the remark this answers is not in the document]
KenHundley: nm, I see it's `Amazon.` in the other projects

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSDK.Extensions.Configuration.AWSSystemsManager.csproj
KenHundley: Done

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSDK.Extensions.Configuration.AWSSystemsManager.nuspec
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package> 
+  <metadata> 
+    <id>AWSSDK.Extensions.Configuration.AWSSystemsManager</id>
+    <title>AWSSDK - ASP.NET Configuration Extensions for AWS Systems Manager</title>
+    <version>3.3.0.0</version>
```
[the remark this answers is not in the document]
KenHundley: Done

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSDK.Extensions.Configuration.AWSSystemsManager.nuspec
KenHundley: Done

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSDK.Extensions.Configuration.AWSSystemsManager.nuspec
KenHundley: Done

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSystemsManagerConfigurationProvider.cs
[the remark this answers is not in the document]
KenHundley: Done

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSystemsManagerConfigurationProvider.cs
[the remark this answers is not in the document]
KenHundley: Done

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSystemsManagerExceptionContext.cs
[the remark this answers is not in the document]
KenHundley: Awaiting Input

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSystemsManagerExtensions.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// <summary>
+        /// Adds an <see cref="IConfigurationProvider"/> that reads configuration values from AWS Systems Manager Parameter Store with a specified path.
+        /// </summary>
+        /// <param name="builder">The <see cref="IConfigurationBuilder"/> to add to.</param>
+        /// <param name="awsOptions"><see cref="AWSOptions"/> used to create an AWS Systems Manager Client connection</param>
+        /// <param name="path">The path that variable names must start with. The path will be removed from the variable names.</param>
+        /// <param name="reloadAfter">Initiate reload after TimeSpan</param>
+        /// <param name="onLoadException">Delegate to call on Exception</param>
+        /// <returns>The <see cref="IConfigurationBuilder"/>.</returns>
+        public static IConfigurationBuilder AddAWSSystemsManager(this IConfigurationBuilder builder, AWSOptions awsOptions, string path, TimeSpan? reloadAfter = null, Action<AWSSystemsManagerExceptionContext> onLoadException = null)
```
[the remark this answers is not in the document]
KenHundley: Done

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSystemsManagerProcessor.cs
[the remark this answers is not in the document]
KenHundley: Awaiting Input

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/Properties/AssemblyInfo.cs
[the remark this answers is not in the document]
KenHundley: Done

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSystemsManagerExceptionContext.cs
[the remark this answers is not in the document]
normj: Agreed

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSDK.Extensions.Configuration.AWSSystemsManager.csproj
normj: The 1.x era is a pain so we want to push as much as possible to 2.0.

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSystemsManagerProcessor.cs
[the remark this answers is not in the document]
normj: When I was suggesting this I was thinking if recusive was turned off, which would not be the default, we would only add the parameters at the root to the configuration. So in your example `IOptions<AppSettings>` and SSM with

```
/approot/appsettings/connectionstrings/connection1
/approot/appsettings/connectionstrings/connection2
/approot/appsettings/url
/approot/appsettings/port
```

Only url and port would be added to the `IOptions` object when recursive was set to false. I'm not sure what the use case would be for doing that, I just wondered if we should provide that switch to users. I'm also good with not having it and if there is an ask later we can added.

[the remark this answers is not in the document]
KenHundley: Your case won't actually work as you're are expecting, at least not without a rework that adds some extra settings and complexity. I've tried to type up all the reasoning why a couple times, but each time I deleted it because it started getting too complex and I felt like it would cause more questions than answers. 

I think there are a few ways to approach this:
  1. Play around with it in the code and I think you'll see why
  2. We have an actual conversation somewhere and I can explain (IM or voice)
  3. Just trust me :)

Btw, if you do want to have a conversation at some point just let me know. I can reach out to our TAM to put us into contact with each other. This definitely isn't necessary, but if you would find it beneficial at any point it is an option

KenHundley:
Personally I believe in code in isolation, but not at the expense of needless complexity or burden in maintenance. I think it makes sense to at minimum isolate all this extension code into their own repository. I definitely didn't need all that SDK code just to build this new extension project. 

As far as individual repos per extension, I think that's a better long term strategy assuming there isn't a lot of complexity and maintenance in building and maintaining all those repos and keeping all those solutions up-to-date with the updates.

But, there are a couple other questions. What would this request from me look like in that type of individual repo environment?  Would I have went into a different repo (maybe the SDK) and opened a ticket to fork my code? There wouldn't be an existing repo for me to submit a pull request into. Just some food for thought.

KenHundley:
Hey Norm, I realized today that I needed to update the process around parameter reloads. Previously if there was an error during reload and the params weren't marked optional it would throw an exception, but I feel that we need to ignore all reload errors regardless of the optional flag. That way a subsequent failure to call the param store won't cause an unexpected exception after the app is already loaded and running. 

Also, the travis builds failed for this, but it appears to be unrelated to the code.

normj:
I agree that seems the right behavior for reloads. 

On another note I think I am going to get another repo created "aws-sdk-net-extensions" or something like that and move AWSSDK.Extensions.NETCore.Setup moved to it. Once that is created I'll ask you to move this pull request to that repo. Does that sound okay?

KenHundley:
Hey Norm, checking in. Was the new repo created or still waiting?

normj:
Still waiting on the approvals for the the new repo. Thanks for your patiences.

KenHundley:
I was looking today to see about adding a version of this for Secrects Manager, however, I see that AWS released the ability to query secrets from Secrets Manager directly through the param store using the path "/aws/reference/secretsmanager/{secretname}". 

With this enhancement do you see any need for creating another extension for direct Secrets Manager integration? Seems unnecessary with this now available.

normj:
I agree, it seems unnecessary right now. Secret Manager might add something complying in future that is not exposed through parameter store, but for now I would recommend to keep it simple with parameter store.

vlesierse:
Sorry to drop in a bit late, but I just came across of this just now while looking for this solution.

Looking at the naming shouldn't is be `AWSSDK.Extensions.Configuration.SimpleSystemStateManager` as the naming of the SDK?

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSystemsManagerExtensions.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// <returns>The <see cref="IConfigurationBuilder"/>.</returns>
+        public static IConfigurationBuilder AddAWSSystemsManager(this IConfigurationBuilder builder, string path, bool optional = false, TimeSpan? reloadAfter = null, Action<AWSSystemsManagerExceptionContext> onLoadException = null)
+        {
+            if (string.IsNullOrWhiteSpace(path)) throw new ArgumentNullException(nameof(path));
+            var config = builder.Build();
+            return builder.AddAWSSystemsManager(config.GetAWSOptions(), path, optional, reloadAfter, onLoadException);
+        }
+
+        /// <summary>
+        /// Adds an <see cref="IConfigurationProvider"/> that reads configuration values from environment variables.
```
vlesierse: Incorrect code documentation.

KenHundley: Good catch. I copied the hints from another implementation, but I guess I missed updating this text

KenHundley:
@vlesierse The name was change to it's current state based on the preference of the AWS team. 
https://github.com/aws/aws-sdk-net/issues/940#issuecomment-390763346

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSystemsManagerConfigurationProvider.cs
```diff
@@ -0,0 +1,119 @@
...
+                var parameters = await _awsSystemsManagerProcessor.GetParametersByPathAsync(awsOptions, path);
+
+                Data = ProcessParameters(parameters, path);
+
+                OnReload();
+            }
+            catch (Exception ex)
+            {
+                var optional = this._configurationSource.Optional;
+                if (optional | reload) return;
```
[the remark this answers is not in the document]
KenHundley: I feel like I did this for a reason as I never use a single | in normal use, but tbh I don't recall why. 
I'll take a look later and if it can be changed I will.

KenHundley:
I'm actually thinking that Secrets Manager will need to be separate, but I just hadn't brought it back up yet.

The reason is that Secrets Manager typically returns json, but that json would need to be flattened into key/value pairs to work with ConfigurationManager. Making it closer in behavior to the "AddJsonFile" than this one. Well I guess kind of a hybrid of the two. When I tested calling Secrets Manager through the Param store it just dumped out the json. Also the GetByPrefix API call this is based on doesn't work with Secrets Manager routes. So it would require and additional overload.

I think long term it will end up being cleaner to have one for param store and one for secrets manager. 

I have a functional prototype for Secrets Manager that I was going to cleanup and submit once this PR is complete.

vlesierse:
For `builder.AddAWSSystemsManager(prefix);` I agree. However the package and namespace naming should be either `SystemsManager / SimpleSystemsManager` as the Amazon/AWSSDK at the beginning makes it redundant.

`Amazon.Extensions.Configuration.AWSSystemsManager` -> `Amazon.Extensions.Configuration.SimpleSystemsManager`

KenHundley:
Ah.
I'll let the AWS guys chime in on the namespace since it'll end up being their preference, but I see what you are saying

rcollette:
Agree with @vlesierse on the namespace name.

For places where AWS remains in type names you might consider making it `Aws` according to Microsoft naming conventions
>The PascalCasing convention, used for all identifiers except parameter names, capitalizes the first character of each word (including acronyms over two letters in length)...
A special case is made for two-letter acronyms in which both letters are capitalized

Ex `HttpContext` or even as present in the comments here `AddJsonFile`
Further examples here:
https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions

I like the prefix on type names because the IDE will generally auto-import nicely if the class is unique enough and I can see all things AWS in intellisense just by typing it, but I cannot argue that namespace is the qualifier for uniqueness in a domain.

Looking forward to this and being able to show there is some form of parity .NET Core with Spring Cloud AWS functionality with regards to configuration.

rcollette:
Hmm, except that the rest of the SDK doesn't follow that naming conventions, so squash that idea I guess.

normj:
We were discussing naming today. And were wondering if we should call it Amazon.AspNetCore.Configuration and remove the service being used completely.  That way if we add service configurations coming from multiple services we don't have to create separate libraries for each. What do you guys think of that?

On another note now that PowerShell Lambda is out this has bubbled up to the top of priority for myself and @sstevenkang.

KenHundley:
Couple notes here.
1) I would not suggest putting AspNetCore in the namespace. With Core 2.1 Microsoft has now created a generic version of `IWebHost` called `IHost` that can be used by console apps and non-aspnet based applications. It supports this configuration standard and so this extension would also be supported by those non-asp apps.
2) I was going to submit a Secrets Manager implementation once this PR was complete. I don't think it will be good long term to require a growing list of referenced aws packages to support the different configuration sources that exist and will exist in the future. Right now it would be potentially require "AWSSDK.SimpleSystemsManagement" and "AWSSDK.SecretsManager", but could continue to grow in the future.

Thoughts?

KenHundley:
oh, as a side note that may be of interest. I created a prototype general lambda function that uses the generic IHost to establish the apps runtime and give all the additional features it supports, like all the various configuration options, IOC, logging framework, etc. 

I think there is potential there for it be become a powerful way to initialize a lambda function's startup and get all those features without an hassle.

sstevenkang:
Regarding your notes:

1.  do you have suggestions on what the idiomatic nuget package name and namespace would be?  `Amazon.Extensions.Hosting.Configuration`?
2. Since it's bringing in additional dependencies, I wonder if we should distribute them as separate libraries (e.g. if a developer only wants to use parameter store configuration, why introduce secrets manager dependnecy)?

normj:
I was imagining if Secret Manager support existed I would use it in conjunction with Parameter Store and only use Secret Manager for things like connection strings and parameter store for general settings that I want to be read externally. 

Using secret manager for all settings could be costly since you pay per secret in Secret Manager. The alternative is we could put all of the settings in a JSON document stored as a secret in Secret Manager but that doesn't seem user friendly.

That is why I thought combining them into the same package would keep things convenient and easy to setup and adding the secret manager assembly is pretty small which would only be used if they were actually pulling secrets from Secret Manager. I'm not dead set on this idea I just prefer setup as easy as possible.

Good call on not including AspNetCore into the assembly name. The base package name is `Microsoft.Extensions.Configuration` and we need to communicate this is the configuration provider for AWS services but not that we created a whole new configuration system. I'm not sure what 'Hosting` adds to the name.  We also need to take into account the possibility of some day adding a configuration provider for app.config/web.config so the name should make sure this is .NET Core configuration system and not .NET Framework. My next name to throw into the pool is `Amazon.NETCore.ConfigurationProvider`?

KenHundley:
1. Personally I would make it
`Amazon.Extensions.Configuration.SystemsManager`
`Amazon.Extensions.Configuration.SecretsManager`
etc.
This stays consistent with existing providers
`Microsoft.Extensions.Configuration.Json`
`Microsoft.Extensions.Configuration.EnvironmentVariables`
`Microsoft.Extensions.Configuration.AzureKeyVault`

2. Exactly what I was thinking

KenHundley:
As for Secrets Manager I see a need to support both. 

We have a Secret that rotates an external vendor's credentials every 30 days (vendor requires a password rotation within every 90 days). Right now the goal for our use is that all application configuration will come from Param Store except this Secret needs to come from Secrets Manager since it's rotated with a custom function. 

I looked and when you call Secrets Manager through Param store you get a dump of the Secret's json, AND you are charged for a call to Secrets Manager. So that json would need additional processing to flatten into a usable key/value set, and would complicate this Param Store implementation for Secrets Manager specific functionality. 

Along with the discussion of cost, I think that makes it even more necessary to make it an explicit package that the developer actively decides to pull into their application. They should already understand the cost model of Secrets Manager before they try to consume it in their app. Also, if Secrets are pulled in separately from Param Store configuration they can have their own secret specific expiration timer. Meaning you may want to pull Param Store configs once a day but one secret weekly, another monthly, and another only at app startup (no refresh). It gives you a lot of flexibility and control over cost.

KenHundley:
oh, and one clarification. I was never envisioning lumping all configuration into Secrets Manager. Just that some configuration could come from param store, and other could come from Secrets Manager, just like some configuration can come from a Json file and Environment Variables at the same time.

KenHundley:
Also, I'm not sure if it's clear in the current description, but it's also perfectly valid to pull configuration data from multiple param store entries.

```csharp
.ConfigureAppConfiguration(builder =>
{
    builder.AddEnvironmentVariables();
    builder.AddAWSSystemsManager("/dev/myapp/");
    builder.AddAWSSystemsManager("/dev/shared/");
    builder.AddAWSSystemsManager("/global/");
})
```

all of those will be added to the "pool" of config data collected from all sources

normj:
I'm good with `Amazon.Extensions.Configuration.SystemsManager` and `Amazon.Extensions.Configuration.SecretsManager`. Lets have @sstevenkang chime in as well before doing any renaming.

aidansteele:
@KenHundley Do you have any public links to your experimentation with IHost and Lambda? I’ve been ruminating about the same thing for a while now and would love to take a look at what you’ve done.

KenHundley:
Not publicly since I just did it in the last week, but I don't mind sharing what I have done. It actually turned out to be extremely simple. I'll produce a sample tomorrow and share the link.

normj:
Here is my understanding for secret manager please correct me if I'm wrong as I haven't used secret manager much. 

First secret manager could be containing a plain string or a JSON document. If the secret is a JSON document then we would flatten. If it is a plain string should we take the leaf name of the secret name or the full name?

Getting the secret straight from secret manager seems straight forward other then the choice above about flattening as mention before.

If I am using parameter store and I configure

```csharp
public static IWebHostBuilder CreateWebHostBuilder(string[] args) =>
    WebHost.CreateDefaultBuilder(args)
        .ConfigureAppConfiguration(builder =>
        {
            builder.AddAWSSystemsManager("/aspnetcore/config", false, TimeSpan.FromMinutes(1));
            builder.AddAWSSystemsManager("/aws/reference/secretsmanager/dbconnection");
        })
        .UseStartup<Startup>();
```

Which intends to pull in the a dbconnection secret from secret manager do we want to support this in V1? If so we need to update the code to call GetParameter and not GetParameterByPath for parameter paths starting with `/aws/reference/secretsmanager` and then do decide how to flatten the data. If we don't want to support in V1 then we should check the string prefix and if it starts with `/aws/reference/secretmanager/` throw an error saying secret manager isn't support. Currently the error message is confusing saying you can't use a string starting with `/aws` which will confuse users as that is the documented way for access secrets. This error happens because this extension is calling a method under the covers that doesn't support getting secrets from secret manager.

KenHundley:
What you explained is what I realized when I started playing with Secrets Manager through parameter store, and I felt like this simple thing started getting complicated, "murky", and obfuscated that there was a charge to consume the secret. If you would prefer that we can extend the Systems Manager provider (this code) to allow it to handle those Secret Manager outputs, I would not suggest it on v1. I would definitely prefer that we finalize this with just param store for now and extend for Secrets Manager support later.

That said, I personally feel that long term it would be better suited to separate the into two different packages (System Manager vs Secrets Manager) that can be maintained and enhanced separately. This allows for maximum simplicity and flexibility. However, I also understand that there could be maintenance, certification, or other business related implications that I'm not aware of, and those could make it more complicated for you guys long term.

KenHundley:
One other quick note about the Secrets Manager output. We "could" accept non-json secret data, but you would probably need to pull the value directly from the configuration object at runtime and not use `IOptions` since there wouldn't be any structure to map over the model class. I think it should expect the data to be json since it's the standard for Secrets Manager and the data needs some structure for `IOptions` to work. That said, it should be easy enough to handle both.

KenHundley:
Actually, I know I caused this by mentioning Secrets Manger, but can we finalize this PR just based off straight forward Systems Manager support? The PR is almost 4 months old and I think once the namespacing and class naming is finalized it's pretty much ready for use. 

I can open an issue to discuss Systems Manager support. Would that work?

normj:
Yes, I totally agree of putting off Secret Manager. I just needed to do some research to make sure we weren't getting locked down any paths that we wouldn't be happy in the future. I think the only thing we need to do for V1 about secret manager is just the error check if the prefix starts with `/aws/reference/secretsmanager` we throw a meaningful error message. I can see people expecting this to work in V1 and I want to make it is clear they aren't doing anything wrong the support just isn't there yet.

normj:
Also I want to thank you for patience, I know 4 months is too long for this PR which is an awesome PR.

We are in a agreement about the separate packages for SSM and Secret Manager. Just hold off on any renaming while we confirm with our PM they are happy with the name.

KenHundley:
Sounds good to me :)

KenHundley:
Was the new repo ever created? You had asked that I move this PR once it was available

sstevenkang:
We are working on it :).

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSDK.Extensions.Configuration.AWSSystemsManager.nuspec
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="utf-8"?>
```
sstevenkang: Is there a reason to use nusepc file instead  of creating a package from .csproj?  Or, were you just following the convention of other SDK extension libraries?

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSDK.Extensions.Configuration.AWSSystemsManager.csproj
```diff
@@ -0,0 +1,17 @@
...
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <AssemblyName>AWSSDK.Extensions.Configuration.AWSSystemsManager</AssemblyName>
+    <RootNamespace>Amazon.Extensions.Configuration.AWSSystemsManager</RootNamespace>
+    <ApplicationIcon />
+    <OutputType>Library</OutputType>
+    <StartupObject />
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="AWSSDK.Extensions.NETCore.Setup" Version="3.3.0" />
```
sstevenkang: the versions here are inconsistent with the version ranges in nuspec.

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSDK.Extensions.Configuration.AWSSystemsManager.nuspec
```diff
@@ -0,0 +1,30 @@
...
+<package>
+  <metadata>
+    <id>AWSSDK.Extensions.Configuration.AWSSystemsManager</id>
+    <title>AWSSDK - ASP.NET Configuration Extensions for AWS Systems Manager</title>
+    <version>1.0.0</version>
+    <authors>Amazon Web Services, Ken Hundley</authors>
+    <description>ASP.NET Configuration Extensions for AWS Systems Manager</description>
+    <language>en-US</language>
+    <licenseUrl>http://aws.amazon.com/apache2.0/</licenseUrl>
+    <projectUrl>https://github.com/aws/aws-sdk-net/</projectUrl>
```
sstevenkang: once we create a new repo, we should probably update this.  I'll take a note so we don't forget.

sstevenkang: Per our conversation, we should probably drop ASP.NET from the description :)

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSDK.Extensions.Configuration.AWSSystemsManager.nuspec
KenHundley: convention from existing SDK Projects

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSDK.Extensions.Configuration.AWSSystemsManager.csproj
KenHundley: Not sure what you mean. Both package references fall in the range defined
`<PackageReference Include="AWSSDK.Extensions.NETCore.Setup" Version="3.3.0" />`
`<dependency id="AWSSDK.Extensions.NETCore.Setup" version="[3.3.0, 3.4)" />`
3.3.0 is >= 3.3.0 & < 3.4

`<PackageReference Include="AWSSDK.SimpleSystemsManagement" Version="3.3.15" />`
`<dependency id="AWSSDK.SimpleSystemsManagement" version="[3.3.15, 3.4)" />`
3.3.15 is >= 3.3.15 & < 3.4

I'm I incorrect?

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSDK.Extensions.Configuration.AWSSystemsManager.nuspec
KenHundley: Agreed on the ASP.Net. What would you like instead?

"Host Configuration Extensions for AWS Systems Manager"?

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSDK.Extensions.Configuration.AWSSystemsManager.csproj
sstevenkang: The problem will go away once we remove nuspec and use csproj to drive nuget package generation.  However, I was confused why the nuspec defines a range but the csproj doesn't.  I thought you could express a range of versions as a dependency in csproj?

KenHundley:
@aidansteele 
[Lambda using the Generic Host for startup](https://github.com/KenHundley/LambdaWithIoc)

This is a first attempt version, but I think it's clean, straight forward, and I've tested that it works. It uses Autofac for it's IOC since it's my preference, but if you prefer the .Net Core native IOC just modify the Startup to remove Autofac.

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSDK.Extensions.Configuration.AWSSystemsManager.csproj
KenHundley: I've checked and PackageReference doesn't currently support ranges, but does support wildcards
[package-references-in-project-files](https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files)
[Open Issue for allowing ranges](https://github.com/NuGet/docs.microsoft.com-nuget/issues/869)

KenHundley:
Hey @sstevenkang, I'd like to add the additional exception handling, naming updates, etc. this weekend if possible. If you have some details about what's been settled and is ready for update just let me know.

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSDK.Extensions.Configuration.AWSSystemsManager.nuspec
KenHundley: perfect

swlasse:
Nice work @KenHundley - really looking forward to see this as part of the SDK.

karthickg2007:
This is awesome. 
Definitely looking forward to use this extension.

extensions/src/AWSSDK.Extensions.Configuration.AWSSystemsManager/AWSSystemsManagerConfigurationProvider.cs
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        private static string NormalizeKey(string key)
+        {
+            return key.Replace("/", ConfigurationPath.KeyDelimiter);
+        }
+
+        public static IDictionary<string, string> ProcessParameters(List<Parameter> parameters, string path) =>
+            parameters
+                .Select(parameter => new
```
robichaud: Please consider allowing something like a ```IParameterProcessor``` to be passed into the Provider so that we can customize this. You can make this the default implementation that can be created in an overloaded extension method.

KenHundley: I had previously thought about doing this. I didn't do it in the initial implementation since the Param Store guidelines say to use "/" to organize the parameters, and I didn't want to make the initial version over complicated. However, I realize that users will create their own organization structures. I'll look at making this update over the weekend. I'll probably only add it to the source extension and not make a special extension for this. The inline extensions are for the more straight forward/default uses. Once it gets into the more complicated structures that what the source version is for.

example:
```csharp
.AddSystemsManager(source =>
{
    source.ParameterProcessor = MyCustomProcessor()
}
```

satuday:
this awesome.. cant wait to use this.. literal.. is there an existing separate nuget pkg?

normj:
I wanted to let everybody know we are pushing forward with this great PR but we have moved this to its own repository that we haven't made public yet. I'm gonna close this PR since we aren't tracking it anymore through this PR but once we make the new repo public I'll comment back here.

robichaud:
@normj When can we expect the new repo's availability?

dmak007:
Yes, please. Would be great.

robichaud:
@normj @KenHundley Any update on when this will be available?

normj:
We are slammed with re:Invent happening just around the corner so it will get released to NuGet post re:Invent but hopefully not much later afterwards.

KenHundley:
The new repo is now publicly available, but it's still considered "in development" until its official release sometime after re:Invent. Nuget packages won't be available until that time.

https://github.com/aws/aws-dotnet-extensions-configuration

mduplessis:
@normj @KenHundley Any update on this?

normj:
The library was released today! https://aws.amazon.com/blogs/developer/net-core-configuration-provider-for-aws-systems-manager/

Issue #886, opened 2018-03-02
Extend Amazon.S3.Model.ByteRange so that it supports range that is missing "start" or "end" value


eangelov:
With current implementation of the ByteRange  class only range that have both start and end values can be set: bytes=500-999

According to [w3](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.35) the range can be in the following formats:
bytes=-500
bytes=9500-

sstevenkang:
I didn't quite follow what this means:

> With current implementation of the ByteRange class only range that have both start and end values can be set: bytes=500-999

Could you clarify your question?

eangelov:
The current implementation of the [ByteRange ](https://docs.aws.amazon.com/sdkfornet/v3/apidocs/items/S3/TByteRange.html) class has only 1 constructor with 2 mandatory parameters, making the creation of a range that does not have an "end"( example: bytes=9500- ) impossible

binsenpai:
Also would like to see this implemented. +1 from me.

klaytaybai:
Hi @eangelov, in case you missed it, bisenpai made a pull request for this that was now submitted.

eangelov:
Hi when are the changes made by binsenpai expected to enter the production build ?

PR #999, opened 2018-07-09
Implement support for user-specified byte-ranged in S3's GetObject

binsenpai:
## Description

When byte-ranges are used in web requests, sometimes we get formats such as `0-`, `-500`, or even `500-900`. The AWS S3 SDK for .NET has only supported having both the start of the range and the end of the range when requesting byte ranges. According to [RFC 7233 (Range Requests)](https://tools.ietf.org/html/rfc7233), the start value or end value of the range is not required and can be assumed by the server.

I've changed the `ByteRange.cs` to support a private value called `formattedByteRange`, which is a string that can be set by the user, in the public setter defined at `FormattedByteRange`. The getter of `FormattedByteRange` returns that value, or the formatted byte range based on the longs, if null. I've also added a second constructor which allows for setting the property directly as well. ~~`long`s that are `null` due to the fact that both can be null. When the getter for `FormattedByteRange` is called, it checks if both values are null, and if so, it returns a null `string`.~~ I've also modified `GetObjectRequest`'s `IsSetByteRange` to support checking whether both values are null as well. This prevents an empty header from being added to the request.

## Motivation and Context

I implemented this change because on one of my projects, I serve a file using the `ResponseStream` I get from the AWS SDK for .NET from my ASP.NET Core app. I also checked the request's headers for byte-ranges and would forward those to the AWS SDK for .NET, but the SDK doesn't support an empty starting or ending range, so I was stuck when Google Chrome tried to request ranges `0-`.

This change should provide a solution for, and fix the issue #886 

## Testing

I haven't done any proper testing yet, but my changes affect the way the SDK determines if a `byteRange` has been set or not, ~~and allowed for nullable `long`s in~~ added a constructor and a `formattedByteRange` value to `ByteRange.cs`, ~~which means that existing code *might* break. I'm confused about whether to set this as a breaking change or not, but I will.~~

I would appreciate if someone else added tests to this PR and test my code thoroughly, and if someone could place a PR for the documentation as well. Thanks.

None are available and none are needed.

## Types of changes

## Checklist

## License


gokarnm:
@binsenpai , changing the properties from `long` to `long?` will be a binary breaking change. Given that Range header can have other values which cannot be set using just the Start and End properties, it would be better to add a setter to `FormattedByteRange` so you can directly provide a value for the range header. The `FormattedByteRange` getter can then use Start/End or FormattedByteRange based on which ones were set by user code.

Here are examples of range header that cannot be set using Start and End properties.

https://tools.ietf.org/html/rfc7233#page-5
```
   o  The first and last bytes only (bytes 0 and 9999):

        bytes=0-0,-1

   o  Other valid (but not canonical) specifications of the second 500
      bytes (byte offsets 500-999, inclusive):

        bytes=500-600,601-999
        bytes=500-700,601-999
```

binsenpai:
@gokarnm I understand, I'll go ahead and implement that in a few.

binsenpai:
@gokarnm Alright, I added the changes here. Feel free to have a look. Would it be all-right if I added a check to see if the "bytes=" value was present in the constructor and setter, and append it, if not?

binsenpai:
@normj Any chance we could get a review?

gokarnm:
@binsenpai , thanks for updating the PR! I'll review it and get back to you soon.

sdk/src/Services/S3/Custom/Model/ByteRange.cs
```diff
@@ -61,7 +75,14 @@ public string FormattedByteRange
         {
             get
             {
-                return string.Format(CultureInfo.InvariantCulture, "bytes={0}-{1}", this.Start, this.End);
+                if (this._formattedByteRange != null)
```
gokarnm: Check for empty string too, using string.IsNullOrEmpty.

sdk/src/Services/S3/Custom/Model/GetObjectRequest.cs
```diff
@@ -156,7 +156,7 @@ public ByteRange ByteRange
         // Check to see if ByteRange property is set
         internal bool IsSetByteRange()
         {
-            return this.byteRange != null;
+            return this.byteRange != null && this.byteRange?.FormattedByteRange != null;
```
gokarnm: Nit, `byteRange?` seems unnecessary as null check is already done previously, I'd replace it with `this.byteRange.FormattedByteRange`.

sdk/src/Services/S3/S3.sln
```diff
@@ -128,6 +128,46 @@ Global
 		{66F78F86-68D7-4538-8EA5-A669A08E1C19}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{66F78F86-68D7-4538-8EA5-A669A08E1C19}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{66F78F86-68D7-4538-8EA5-A669A08E1C19}.Release|Any CPU.Build.0 = Release|Any CPU
+		{78A15B4B-1533-4AA1-9E8C-571618A74FDB}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
```
gokarnm: Could you exclude this file from the PR, it is autogenerated.

gokarnm:
@binsenpai , I've added some minor comments on the PR.

binsenpai:
@gokarnm Question, do you want me to exclude the file or exclude that specific line?

gokarnm:
Exclude the `sdk/src/Services/S3/S3.sln` file, as it's autogenerated.

binsenpai:
It's fine, I'll do it; I've got free time on my hands.

gokarnm:
@binsenpai , the PR has been [merged](https://github.com/aws/aws-sdk-net/commits?author=binsenpai).

Issue #1004, opened 2018-07-17
Dotnet Core 2.1 - AmazonSQSClient AmazonServiceException when call ReceiveMessageAsync


actumn:
Hi.
When I updated dotnet core (2.0.6 -> 2.1) and deployed on server, we met some error about credentials in AmazonSQSClient.
I think it is from [dotnet core 2.1 HttpClientHandler update](https://blogs.msdn.microsoft.com/dotnet/2018/04/18/performance-improvements-in-net-core-2-1/).
So I rollback dotnet core version (2.1 -> 2.0.6), then there is no exception anymore.

## Expected Behavior

Whenever we call `Amazon.SQS.AmazonSQSClient.ReceiveMessageAsync(ReceiveMessageRequest request, CancellationToken cancellationToken = default(CancellationToken))`, there should be no exception.

## Current Behavior

The function throws exception very frequently(3 minutes in 6 hour average) like below.
```
Amazon.Runtime.AmazonServiceException: Failed to retrieve credentials from EC2 Instance Metadata Service.
   at Amazon.Runtime.DefaultInstanceProfileAWSCredentials.GetCredentials()
   at Amazon.Runtime.DefaultInstanceProfileAWSCredentials.GetCredentialsAsync()
   at Amazon.Runtime.Internal.CredentialsRetriever.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.Runtime.Internal.RetryHandler.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.Runtime.Internal.RetryHandler.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.Runtime.Internal.CallbackHandler.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.Runtime.Internal.CallbackHandler.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.Runtime.Internal.ErrorCallbackHandler.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.Runtime.Internal.MetricsHandler.InvokeAsync[T](IExecutionContext executionContext)
```
## Context

I want to get [the network performance provided by dotnet core 2.1](https://blogs.msdn.microsoft.com/dotnet/2018/04/18/performance-improvements-in-net-core-2-1/), with AWS features.

## Your Environment

* AWSSDK.Core version used: 3.3.24.2
* AWSSDK.SQS version used: 3.3.3.10
[...]

PR #1010, opened 2018-07-24
Reduce locking contention in DefaultInstanceProfileAWSCredentials

AnthonySteele:
## Description

* The static "Instance" property should use the "double-checked locking" pattern when checking for nulls https://en.wikipedia.org/wiki/Double-checked_locking
* Do the fast assign to get in a good state before the slower logging code in the exception handler (with `Instance.logger.Error` ).
* remove `?` from `Instance?.` - Instance can't be null, the adjacent lines assume that.
* In `RenewCredentials()`, `Instance` is just a heavyweight alias for `this` with added locks and recursion. Removed in favour of implicit `this` which also make it clear that it is never null. 
*  Do not use `ReaderWriterLockSlim` or similar "slim" locks, they use spinwait, which [scales badly if high contention is generated across lots of threads](https://github.com/dotnet/coreclr/issues/12780). This lock was also held for up to 5 seconds, which is a very long time to lock in this context.
* [.NET reference assignment is an atomic operation anyway](https://stackoverflow.com/questions/11745440/what-operations-are-atomic-in-c).  There is no half-way state that needs locking inside a statement of the form `a = b;`.
* When there is a `TaskCanceledException` in `RenewCredentials`, keep the `lastRetrievedCredentials`.

### For future reference elsewhere:
* Double-checked locking is good
* Slim locks held for long or under high concurrency are bad.
* Reference assignment is atomic

### Not addressed in this PR

* `RenewCredentials` calls `EC2InstanceMetadata.IAMSecurityCredentials` which eventually does `AWSSDKUtils.DownloadStringContent` i.e. synchronous http. I am not going to make this whole call stack async as it's not a simple change, and it is not directly related. 
* `AWSSDKUtils.DownloadStringContent` does a `using (var client = new System.Net.Http.HttpClient)` Creating a new client for each request is a technique that can cause socket exhaustion when under high load. `HttpClientFactory` was created to solve this problem.
* `AWSSDKUtils.DownloadStringContent` does async things and bodges them to be synchronous. 

This all happens inside `RenewCredentials` [which happens inside a `lock()` statement](https://stackoverflow.com/questions/7612602/why-cant-i-use-the-await-operator-within-the-body-of-a-lock-statement) in one case.

## Motivation and Context

We think that this is behind one of our production issues with CPU shooting up to 100% under load in ASP.NET core 2.1, and at the same time errors with stack trace
````
Amazon.Runtime.AmazonServiceException: Failed to retrieve credentials from EC2 Instance Metadata Service.
   at Amazon.Runtime.DefaultInstanceProfileAWSCredentials.GetCredentials()
   at Amazon.Runtime.DefaultInstanceProfileAWSCredentials.GetCredentialsAsync()
````

Issues 4953686341 and 5182648891

## Testing

We are considering running load through this code on our test environment this week. And possibility production, to verify that it fixes issues seen.

The line `Instance.logger.Error(e, FailedToGetCredentialsMessage);` in `RenewCredentials` - where does that log message go? An additional verification would be to find this.

## Types of changes

## Checklist

## License

Thanks to @slang25 @mgoodfellow @martincostello for review and shaping this.

### CODE REVIEW

**Comment by slang25** [comment]
@anthonysteele I want to make you aware [of this](https://github.com/aws/aws-sdk-net/issues/1004#issuecomment-407640381) since you are modifying that code. And in general would be a good thing for you to be aware of in case it crops up in usage of JustSaying.

File: sdk/src/Core/Amazon.Runtime/Credentials/_bcl+coreclr/DefaultInstanceProfileAWSCredentials.cs [hunk 1]
```diff
@@ -84,25 +84,14 @@ public override ImmutableCredentials GetCredentials()
...
-            try
-            {
-                refreshLock.EnterReadLock();
-                credentials = lastRetrievedCredentials?.Copy();
-            }
-            finally
-            {
-                refreshLock.ExitReadLock();
-            }
+            var credentials = lastRetrievedCredentials?.Copy();
```
arishlabroo: IMHO, if lastRetrievedCredentials is null, it should call RenewCredentials(null) before returning.
Related  https://github.com/aws/aws-sdk-net/issues/1004#issuecomment-407611176

sdk/src/Core/Amazon.Runtime/Credentials/_bcl+coreclr/DefaultInstanceProfileAWSCredentials.cs
```diff
@@ -131,6 +131,11 @@ private void RenewCredentials(object unused)
 
                 lastRetrievedCredentials = newCredentials;
             }
+            catch (TaskCanceledException e)
```
martincostello: You _may_ want to use `OperationCanceledException` instead, which this derives from. I can't remember the exact scenario where you get the base class instead of the other.

sdk/src/Core/Amazon.Runtime/Credentials/_bcl+coreclr/DefaultInstanceProfileAWSCredentials.cs
AnthonySteele: @arishlabroo I have taken the other option suggested, don't always set the `lastRetrievedCredentials` on an exception. I am now doing

````
            catch (TaskCanceledException e)
            {
                // in this case, keep the lastRetrievedCredentials
            }
            catch (Exception e)
            {
                lastRetrievedCredentials = null;
               ````
````

There is also the option to _never_ set `lastRetrievedCredentials` to null on any exception. Or to try to `RenewCredentials` when it's null in `GetCredentials` as you suggest above.

I'm not quite sure which of these is the best way to avoid nulls.

sstevenkang: Thanks for this PR Anthony.  We have a CR review going within our team that contains some of the changes you've already addressed here.  So, if you are open to it let's merge the changes :)

We were thinking of addressing the thread contention problem by doing lazy fetch of credentials instead of doing the initial fetch at creation of the singleton instance. 

1. We will no longer do `_instance.RenewCredentials(null);`  inside the lock.
2. Start the timer immediately in the constructor `credentialsRetrieverTimer = new Timer(RenewCredentials, null, TimeSpan.Zero, neverTimespan);` 
3. Refactor the credential fetch logic in `RenewCredentials()`
4. In `GetCredentials()` we would do a fallback of directly fetching IMDS credentials using the refactored logic from (3.) if `lastRetrievedCredentials` is `null.  This will also address some cases of https://github.com/aws/aws-sdk-net/issues/1004 

Using the repro application your team provided, we saw only about 2~3 direct fetches from IMDS before clients started using cached credentials.  If the above description is not detailed enough, I can create PR--sometime later today--with the change so we can take a look at our proposed change.  

What do you think about this approach?

AnthonySteele: This sounds great. You are of course welcome to merge this PR or just use parts as suggestions for other PRs.

So the issues that I have seen (and I think #1004) are not just at  thread contention at startup, they happen after the app has been running for a while, and load is elevated. `RenewCredentials` is where I think it goes wrong.

We have not yet had the time to load-test code using this PR, but hope to do that soon.

sstevenkang: I created a PR of our change here as a reference: https://github.com/aws/aws-sdk-net/pull/1017

If you could do a review & roll up the change in to this PR, we'll start the process of merging this PR in.  Before we do the merge, however, we would like to make sure that the change in https://github.com/aws/aws-sdk-net/pull/1017 does indeed alleviate the thread contention problem in your app. environment.

AnthonySteele: I'm not quite clear - do you want me to review #1017 as it stands, or to merge #1017 into #1010 (this PR?)

sstevenkang: Sorry for the confusion!  I would like you to merge #1017 into this PR.

AnthonySteele:
I have updated. 

FYI, [I checked the docs on `?.`](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-conditional-operators) and it "is thread-safe" in the sense that `a?.b` will read `a` exactly once, so there is no way for it to be set to null by another thread halfway through and cause a failure.  This means that the expression `lastRetrievedCredentials?.Copy();` is thread-safe and should not throw. Yes it can return null, but it can do that even if access to this line is inside a lock.

sstevenkang:
Thanks for this contribution @AnthonySteele .  I'll do a bit more testing on our side & release this change asap.

AnthonySteele:
FYI, we have managed to do some load tests on .dlls built from this PR, and while they are not conclusive:
* We did not see the "Failed to retrieve credentials ...at .. GetCredentials()" error logged at all.
* Our api still has limits beyond which it fails, but they seem to be a bit higher. This still occurs when the machine is using < 10% CPU. Failure is now in our code, not AWS SDK, and we have some updates to apply to that.

sstevenkang:
We released a [new version](https://www.nuget.org/packages/AWSSDK.Core/3.3.24.6) of AWSSDK.Core 3.3.24.6 with this change.

AnthonySteele:
PYI, we have been running `AWSSDK.Core 3.3.24.6` for 6 days now, over the weekly busy periods, and the issues mentioned above have not occured at all during that time. This appears to be fixed.

sstevenkang:
Great news! Thanks again for the PR!

PR #1017, opened 2018-07-27
reference PR -- ignore

sstevenkang:
created PR for discussion in https://github.com/aws/aws-sdk-net/pull/1010

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Credentials/_bcl+coreclr/DefaultInstanceProfileAWSCredentials.cs
+++ b/sdk/src/Core/Amazon.Runtime/Credentials/_bcl+coreclr/DefaultInstanceProfileAWSCredentials.cs
@@ -31,7 +31,7 @@ namespace Amazon.Runtime
     internal class DefaultInstanceProfileAWSCredentials : AWSCredentials, IDisposable
     {
         private static object instanceLock = new object();
-        
+
         private Timer credentialsRetrieverTimer;
         private ImmutableCredentials lastRetrievedCredentials;
         private ReaderWriterLockSlim refreshLock;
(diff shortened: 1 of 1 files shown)


slang25:
@AnthonySteele I think the commit message might answer that:

> Previous implementation of retrieving credentials from EC2 IMDS would hammer the service when too many clients were being created in short
> amount of time. We decided to implement a singleton class to address
> that problem. The change introduced a regression where too many client
> creation would create a thread pool bottle-neck.
> 
> The solution here is to create service clients without the initial fetch
> of the credentials--it would just start the timer task and return the
> instance immediately yielding the thread to other tasks.
> 
> When the DefaultInstanceProfileAWSCredentials.GetCredentials() is
> called, we would hit IMDS if no credential has been cached already.  The
> perf is actually not that bad, since the GetStringContentAsync() gets to
> execute, in a stress test, we only hit IMDS 2~3 times before the timer
> task runs and cached credentials is available.

AnthonySteele:
@slang25 What i mean is that the `credentialsRetrieverTimer` is in this PR set up on line 72 with ` new Timer(RenewCredentials, null, TimeSpan.Zero, neverTimespan)` - this creates it in a dormant, disabled state, I think?

It is then activated by line 161: `Instance.credentialsRetrieverTimer.Change(refreshRate, neverTimespan);`
But that's inside `RenewCredentials`  which won't get called.

sstevenkang:
The 3rd parameter `TimeSpan.Zero`  should start the timer immediately. If we wanted in disabled state, we would pass `neverTimespan` like we did previously.

sdk/src/Core/Amazon.Runtime/Credentials/_bcl+coreclr/DefaultInstanceProfileAWSCredentials.cs
```diff
@@ -95,6 +91,22 @@ public override ImmutableCredentials GetCredentials()
...
             }
 
+            // If there's no credentials cached, hit IMDS directly.
+            if (credentials == null)
+            {
+                credentials = FetchCredentials();
+
+                try
+                {
+                    refreshLock.EnterWriteLock();
```
sstevenkang: Since we are doing away with locks, we should remove this.

AnthonySteele:
My mistake, yes the 3rd parameter value `TimeSpan.Zero` should launch it immediately.

PR #1052, opened 2018-08-30
support downloading and uploading large files irrespective of network disconnect

sunchunqiang:
## Description

The changes are to support downloading and uploading large files when the network is unstable. 
The issues fixed in the changes include
1. When the network is disconnected, HttpClient.SendAsyn will throw ObjectDisposedException, the upload api will fail with this kind of exception, instead of retry. The changes will  catch the exception ObjectDisposedException from HttpClient.SendAsyn and re-throw an IOexception  so that upload process will retry.

2. When the network is disconnected, System.Net.WebConnectionStream.Read() may also throw ObjectDisposedException, The changes will re-throw a webexception  so that download process will retry,.

3. When the network is disconnected, the network api may throw a timeout exception, the changes will retry the download process for this kind of exception. 
4. When the network is disconnected, the network api may not throw any exception and not return any error code, the changes will retry the download process for this case. 
5. When the network is disconnected and then restored and the download process is resumed, the download progress passed to  WriteObjectProgressEvent is from 0, instead of the previous download position. The changes will fix the issue
6. When retry times is more than 14, download api will crash, the change will fix the issue
7. Fix a bug in LogForRetry which can lead the application to crash

## Motivation and Context

Wen uploading large files, if the internet is lost for even one second, the sdk will throw this error and stop upload the file. It should get the last part successful sent to the server, and continue sending the file from this point.  Downloading a large file have a similar issue.
The changes are going to fix the issues

## Testing

1. Download or upload a large file with the sdk
2. disconnect the network on the device
3. connect the network after 5 minutes
4. the download/upload should continue 

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Internal/Util/WrapperStream.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Util/WrapperStream.cs
@@ -257,7 +257,19 @@ namespace Amazon.Runtime.Internal.Util
         /// </returns>
         public override int Read(byte[] buffer, int offset, int count)
         {
-            return BaseStream.Read(buffer, offset, count);
+            try
+            {
+                return BaseStream.Read(buffer, offset, count);
+            }
+            catch (ObjectDisposedException e)
+            {
+                string fullname = BaseStream.GetType().FullName;
+                if (fullname == "System.Net.WebConnectionStream")
+                {
+                    throw new System.IO.IOException("HttpClient.SendAsync throws ObjectDisposedException exception due to network connectivity failure.", e);
+                }
+                throw; 
+            }
         }
 
         /// <summary>
--- a/sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_mobile/HttpRequestMessageFactory.cs
+++ b/sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_mobile/HttpRequestMessageFactory.cs
@@ -425,8 +425,17 @@ namespace Amazon.Runtime
         {
             try
             {
-                var responseMessage = await _httpClient.SendAsync(_request, HttpCompletionOption.ResponseHeadersRead, cancellationToken)
-                    .ConfigureAwait(continueOnCapturedContext: false);
+                HttpResponseMessage responseMessage; 
+                try
+                {
+                    responseMessage = await _httpClient.SendAsync(_request, HttpCompletionOption.ResponseHeadersRead, cancellationToken)
+                        .ConfigureAwait(continueOnCapturedContext: false);
+                }
+                catch(ObjectDisposedException e)
+                {
+                    throw new System.IO.IOException("HttpClient.SendAsync throws ObjectDisposedException exception due to network connectivity failure.", e);
+                }
+
 
                 bool disposeClient = !_clientConfig.CacheHttpClient;
                 // If AllowAutoRedirect is set to false, HTTP 3xx responses are returned back as response.
--- a/sdk/src/Services/S3/Custom/Internal/AmazonS3RetryPolicy.cs
+++ b/sdk/src/Services/S3/Custom/Internal/AmazonS3RetryPolicy.cs
@@ -45,6 +45,10 @@ namespace Amazon.S3.Internal
         public AmazonS3RetryPolicy(IClientConfig config) :
             base(config)
         {
+#if !PCL
+            WebExceptionStatusesToRetryOn.Add((WebExceptionStatus)10);   //SecureChannelFailure, disconnected network may cause this exception)
+#endif
+
         }
 
         /// <summary>
@@ -113,6 +117,17 @@ namespace Amazon.S3.Internal
                         // this means that the synchronous check is inconclusive and
                         // the caller needs to check for a bucket/region mismatch
                         return null;
+                    }
+                }
+            }
+            else
+            {
+                WebException webException = exception as WebException;
+                if (exception is WebException || IsInnerException<WebException>(exception, out webException))
+                {
+                    if (this.WebExceptionStatusesToRetryOn.Contains(webException.Status))
+                    {
+                        return true;
                     }
                 }
             }
--- a/sdk/src/Services/S3/Custom/Model/GetObjectResponse.cs
+++ b/sdk/src/Services/S3/Custom/Model/GetObjectResponse.cs
@@ -56,13 +56,24 @@ namespace Amazon.S3.Model
 
         private string bucketName;
         private string key;
-                
+        private long bytesTransferred = 0;
+
         /// <summary>
         /// Flag which returns true if the Expires property has been unmarshalled
         /// from the raw value or set by user code.
         /// </summary>
         private bool isExpiresUnmarshalled;
         internal string RawExpires { get; set; }
+
+        /// <summary>
+        /// Gets and sets the BytesTransferred property. 
+        /// BytesTransferred is the the total number of bytes to be transferred before this retry
+        /// </summary>
+        public long BytesTransferred
+        {
+            get { return this.bytesTransferred; }
+            set { this.bytesTransferred = value; }
+        }
 
         /// <summary>
         /// Gets and sets the BucketName property.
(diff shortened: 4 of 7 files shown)


sdk/src/Services/S3/Custom/Model/GetObjectResponse.cs
```diff
@@ -56,14 +56,25 @@ public partial class GetObjectResponse : StreamResponse
 
         private string bucketName;
         private string key;
-                
+        private long transferredalready = 0;
```
kvasukib: Suggestion: can you rename this variable to "bytesTransferrred" ?

sdk/src/Services/S3/Custom/Model/GetObjectResponse.cs
```diff
@@ -56,14 +56,25 @@ public partial class GetObjectResponse : StreamResponse
...
         /// from the raw value or set by user code.
         /// </summary>
         private bool isExpiresUnmarshalled;
         internal string RawExpires { get; set; }
 
+        /// <summary>
+        /// Gets and sets the TransferredAlreaddy property. 
+        /// TransferredAlreaddy is the the total number of bytes to be transferred before this retry
+        /// </summary>
+        public long TransferredAlready
```
kvasukib: change this to BytesTransferred

sdk/src/Core/Amazon.Runtime/Internal/Util/WrapperStream.cs
```diff
@@ -257,7 +257,24 @@ public override void Flush()
...
+            }
+            catch(Exception e)
+            {
+                if (e is System.ObjectDisposedException )
+                {
+                    string name =  BaseStream.GetType().Name;
+                    string fullname = BaseStream.GetType().FullName; 
+                    if(fullname == "System.Net.WebConnectionStream")
+                    {
+                        throw new System.Net.WebException("ObjectDisposedException exception throw by WebConnectionStream", e, System.Net.WebExceptionStatus.ConnectFailure, null);
```
kvasukib: Add one more phrase: "Network connectivity failure"

sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_mobile/HttpRequestMessageFactory.cs
```diff
@@ -425,8 +425,21 @@ public async System.Threading.Tasks.Task<IWebResponseData> GetResponseAsync(Syst
...
+                try
+                {
+                    responseMessage = await _httpClient.SendAsync(_request, HttpCompletionOption.ResponseHeadersRead, cancellationToken)
+                        .ConfigureAwait(continueOnCapturedContext: false);
+                }
+                catch(Exception e)
+                {
+                    if(e is System.ObjectDisposedException)
+                    {
+                        throw new System.Net.WebException("ObjectDisposedException exception throw by HttpClient", e, System.Net.WebExceptionStatus.ConnectFailure, null);
```
kvasukib: Add one more phrase: "Network connectivity failure"

sdk/src/Services/S3/Custom/Model/GetObjectResponse.cs
```diff
@@ -56,14 +56,25 @@ public partial class GetObjectResponse : StreamResponse
...
+
         /// <summary>
         /// Flag which returns true if the Expires property has been unmarshalled
         /// from the raw value or set by user code.
         /// </summary>
         private bool isExpiresUnmarshalled;
         internal string RawExpires { get; set; }
 
+        /// <summary>
+        /// Gets and sets the TransferredAlreaddy property.
```
sstevenkang: property name is `BytesTransferred` but the docs says it's `TransferredAlreaddy`

sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_mobile/HttpRequestMessageFactory.cs
```diff
@@ -425,8 +425,17 @@ public async System.Threading.Tasks.Task<IWebResponseData> GetResponseAsync(Syst
...
             {
-                var responseMessage = await _httpClient.SendAsync(_request, HttpCompletionOption.ResponseHeadersRead, cancellationToken)
-                    .ConfigureAwait(continueOnCapturedContext: false);
+                HttpResponseMessage responseMessage; 
+                try
+                {
+                    responseMessage = await _httpClient.SendAsync(_request, HttpCompletionOption.ResponseHeadersRead, cancellationToken)
+                        .ConfigureAwait(continueOnCapturedContext: false);
+                }
+                catch(ObjectDisposedException e)
```
sstevenkang: How can you be sure ObjectDisposedException is because of network connectivity failure?  Also, aren't we just re-throwing the exception?  Why not just rethrow ObjectDisposedException with better message instead?

sdk/src/Services/S3/Custom/_bcl+coreclr+pcl/Transfer/Internal/DownloadCommand.cs
```diff
@@ -55,6 +55,8 @@ internal partial class DownloadCommand : BaseCommand
             (WebExceptionStatus)12,
             (WebExceptionStatus)1,
             (WebExceptionStatus)3,
+            (WebExceptionStatus)14, //timeout
```
sstevenkang: I am not sure if this is the correct behavior.  If we retry on timeout with delay, in the worst case scenario, we would be waiting for a very long time before we give up.  Is that the behavior we want?

sdk/src/Services/S3/Custom/Internal/AmazonS3RetryPolicy.cs
```diff
@@ -116,6 +120,17 @@ public partial class AmazonS3RetryPolicy : DefaultRetryPolicy
                     }
                 }
             }
+            else
```
sstevenkang: This is a behavior I think we need a team discussion on.  Before, we didn't retry on any of these webexceptions for S3, but now we are.  Could you get @gokarnm's opinion on this?

sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_mobile/HttpRequestMessageFactory.cs
sunchunqiang: I am not sure ObjectDisposedException  is only due to a network connectivity failure, but we are not able to tell the difference of failure sources here, so I consider ObjectDisposedException is a kind of network connectivity failure. 
The reason I rethrow an IOException is because that our retry policy will handle IOException , but not ObjectDisposedException . We can rethrow ObjectDisposedException with a better message, but then we need add handler for this kind of exception in our retry policy. Which solution do you prefer to?

sdk/src/Core/Amazon.Runtime/Internal/Util/WrapperStream.cs
```diff
@@ -257,7 +257,20 @@ public override void Flush()
...
-            return BaseStream.Read(buffer, offset, count);
+            try
+            {
+                return BaseStream.Read(buffer, offset, count);
+            }
+            catch (ObjectDisposedException e)
+            {
+                string name = BaseStream.GetType().Name;
+                string fullname = BaseStream.GetType().FullName;
+                if (fullname == "System.Net.WebConnectionStream")
```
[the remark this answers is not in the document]
sunchunqiang: AWSSDK.Core does not have reference to System.Net.WebConnectionStream, it's referred by AWSSDK.S3 project, so cannot use the class  System.Net.WebConnectionStream directly here, we can only use its name. 

The code   string name = BaseStream.GetType().Name;   is not necessary. I removed it

sdk/src/Services/S3/Custom/Model/GetObjectResponse.cs
sunchunqiang: Fixed it

jcruzfrontline:
@gokarnm  - Do you have any reservations with this PR?

normj:
Closing for lack of activity and we are moving in a different direction with Xamarin support focusing more on .NET Standard 2.0.

PR #1064, opened 2018-09-07
Concurrent dictionary in HttpRequestFactory

AnthonySteele:
## Description

reduce locking in `HttpRequestMessageFactory`

## Motivation and Context

For discussion.

We know that `ReaderWriterLockSlim` is not great under load, is CPU intensive, and so where possible lighter-weight techniques should be used.

We can use `ConcurrentDictionary` in .NET 4.0 and up instead of a lock around a dictionary. As far as I can tell, the class `HttpRequestMessageFactory` is not used in the .NET 3.5 build so we're OK.

` Interlocked.CompareExchange` can be used to express an atomic operation of "if `_httpClientCache` is equal to null,  set it to the value `CreateHttpClientCache()`"

We are not sure if this code is actually causing issues with us during SNS publish, but it might be. 

## Types of changes

## Checklist

- [x ] My code follows the code style of this project

## License

- [ x] I confirm that this pull request can be released under the Apache 2 license

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_mobile/HttpRequestMessageFactory.cs
+++ b/sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_mobile/HttpRequestMessageFactory.cs
@@ -18,6 +18,7 @@ using Amazon.Runtime.Internal.Transform;
 using Amazon.Runtime.Internal.Util;
 using Amazon.Util;
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
@@ -39,9 +40,7 @@ namespace Amazon.Runtime
     public class HttpRequestMessageFactory : IHttpRequestFactory<HttpContent>
     {
         // This is the global cache of HttpClient for service clients that are using 
-        static readonly ReaderWriterLockSlim _httpClientCacheRWLock = new ReaderWriterLockSlim();
-        static readonly IDictionary<string, HttpClientCache> _httpClientCaches = new Dictionary<string, HttpClientCache>();
-
+        static readonly ConcurrentDictionary<string, HttpClientCache> _httpClientCaches = new ConcurrentDictionary<string, HttpClientCache>();
 
         private HttpClientCache _httpClientCache;
         private IClientConfig _clientConfig;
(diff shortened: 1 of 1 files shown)


slang25:
Another option is to use `Lazy<T>` (see this example in HttpClientFactory: https://github.com/aspnet/HttpClientFactory/blob/master/src/Microsoft.Extensions.Http/DefaultHttpClientFactory.cs#L50) which will guard from more than one thread initializing the same thing. However in this case, it looks like the initialization is cheap enough not to be a concern. Looks good to me 👍

AnthonySteele:
Good point, we can use `Lazy<T>` since: 
* it is in .NET 4.0 and up 
* "By default, all public and protected members of the Lazy<T> class are thread safe and may be used concurrently from multiple threads. "  [docs](https://docs.microsoft.com/en-us/dotnet/api/system.lazy-1?view=netframework-4.7.2) . TIL.

sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_mobile/HttpRequestMessageFactory.cs
```diff
@@ -69,54 +68,17 @@ public IHttpRequest<HttpContent> CreateHttpRequest(Uri requestUri)
...
-                            {
-                                _httpClientCache = CreateHttpClientCache(_clientConfig);
-                            }
-                        }
-                        finally
-                        {
-                            _httpClientCacheRWLock.ExitWriteLock();
-                        }
+                        // atomic operation - if _httpClientCache still equals null, swap in the CreateHttpClientCache 
+                        Interlocked.CompareExchange(ref _httpClientCache, CreateHttpClientCache(_clientConfig), null);
```
martincostello: Does anything need to be done with the created value if the exchange fails as it's already set, or is it OK to just be left to be GC'd?

sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_mobile/HttpRequestMessageFactory.cs
```diff
@@ -39,9 +40,7 @@ namespace Amazon.Runtime
     public class HttpRequestMessageFactory : IHttpRequestFactory<HttpContent>
     {
         // This is the global cache of HttpClient for service clients that are using 
-        static readonly ReaderWriterLockSlim _httpClientCacheRWLock = new ReaderWriterLockSlim();
-        static readonly IDictionary<string, HttpClientCache> _httpClientCaches = new Dictionary<string, HttpClientCache>();
-
+        static readonly ConcurrentDictionary<string, HttpClientCache> _httpClientCaches = new ConcurrentDictionary<string, HttpClientCache>();
```
martincostello: Keys are case-sensitive - is that OK?

AnthonySteele: That's likely more of a potential issue than an actual one. It's not a change in behaviour though.

sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_mobile/HttpRequestMessageFactory.cs
AnthonySteele: I was assuming that it's safe to just GC it. If not, then we need `Lazy<T>`

sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_mobile/HttpRequestMessageFactory.cs
martincostello: Yes you're right, I clearly forgot to properly look at the line that was deleted.

sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_mobile/HttpRequestMessageFactory.cs
martincostello: It's just I can't see what `CreateHttpClientCache()` does from the diff, hence the question.

sstevenkang:
Wow.  Thanks for this PR--this changes seems reasonable at a high level.  Do you have any concrete data on whether this change improves performance through reducing lock contention?

AnthonySteele:
Unfortunately no concrete data yet.  I made this PR "For discussion" so that I didn't forget it, and in case it turned out to be the cause of some CPU spikes and general slowdowns in SNS publishing that we 
 observe under load, which could be possible.  

This week we are updating the NETCore framework and ASP version, to see if [this http fix](https://github.com/aspnet/KestrelHttpServer/issues/2694#issuecomment-420347853) makes  a difference.  If not, then we might have to try and gather that concrete data.

slang25:
@AnthonySteele https://github.com/AnthonySteele/aws-sdk-net/pull/1
This illustrates what this combined with Lazy looks like, it would avoid @martincostello question about the "thrown away" instance if there was one spare instance created in a race.

AnthonySteele:
How many keys are present in the dictionary in an "under load" scenario? I.e. what is the degree of concurrency - just a single-digit number of items, or hundreds?

normj:
Closing since @AnthonySteele says it is no longer necessary and there has been other changes in this area causing conflicts.

AnthonySteele:
I think it's right to close this PR. 

But it could also be useful to take ideas from it, i.e. use code like that shown here, in order to get performance under load.

PR #1089, opened 2018-10-08
save the clientid to the local setting before assign it to class memb…

sunchunqiang:
…er variable so that if the saving operation throw an exception, the class member will not be assigned

## Description

Fix an issue in ClientContext constructor
## Motivation and Context

save the clientid to the local setting before assign it to class member variable so that if saving operation throw an exception, the class member will not be assigned

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/_pcl/ClientContext.pcl.cs
+++ b/sdk/src/Core/Amazon.Runtime/_pcl/ClientContext.pcl.cs
@@ -38,9 +38,10 @@ namespace Amazon.Runtime.Internal
             {
                 _clientID = _appSetting.GetValue(APP_ID_KEY, ApplicationSettingsMode.Local);
                 if (string.IsNullOrEmpty(_clientID))
-                {
-                    _clientID = Guid.NewGuid().ToString();
-                    _appSetting.SetValue(APP_ID_KEY, _clientID, ApplicationSettingsMode.Local);
+                { 
+                    string newClientID = Guid.NewGuid().ToString();
+                    _appSetting.SetValue(APP_ID_KEY, newClientID, ApplicationSettingsMode.Local);
+                    _clientID = newClientID;
                 }
             }
         }


sdk/src/Core/Amazon.Runtime/_pcl/ClientContext.pcl.cs
```diff
@@ -38,9 +38,10 @@ public ClientContext(string appID)
             {
                 _clientID = _appSetting.GetValue(APP_ID_KEY, ApplicationSettingsMode.Local);
                 if (string.IsNullOrEmpty(_clientID))
-                {
-                    _clientID = Guid.NewGuid().ToString();
-                    _appSetting.SetValue(APP_ID_KEY, _clientID, ApplicationSettingsMode.Local);
+                { 
+                    string newClientID = Guid.NewGuid().ToString();
+                    _appSetting.SetValue(APP_ID_KEY, newClientID, ApplicationSettingsMode.Local);
```
kvasukib: Should we check if `SetValue` failed or succeeded and then log an error message when SetValue fails and not store _clientID?

sdk/src/Core/Amazon.Runtime/_pcl/ClientContext.pcl.cs
```diff
@@ -38,9 +39,18 @@ public ClientContext(string appID)
...
+                { 
+                    string newClientID = Guid.NewGuid().ToString();
+                    try
+                    {
+                        _appSetting.SetValue(APP_ID_KEY, newClientID, ApplicationSettingsMode.Local);
+                    }
+                    catch(Exception e)
+                    {
+                        Logger.GetLogger(typeof(ClientContext)).Error(e, "Failed to save client id to local application setting");
+                        throw;
```
sstevenkang: Logging is good, but I think we should also throw the exception with a helpful message.  Instead of simply rethrowing, could we wrap in a new exception and make the original exception an inner exception?

sunchunqiang: done!

sdk/src/Core/Amazon.Runtime/_pcl/ClientContext.pcl.cs
```diff
@@ -38,9 +39,19 @@ public ClientContext(string appID)
...
+                    string newClientID = Guid.NewGuid().ToString();
+                    try
+                    {
+                        _appSetting.SetValue(APP_ID_KEY, newClientID, ApplicationSettingsMode.Local);
+                    }
+                    catch(Exception e)
+                    {
+                        Logger.GetLogger(typeof(ClientContext)).Error(e, "Failed to save client id to local application setting");
+                        Exception exception = new Exception("Failed to save client id to local application setting", e);
+                        throw;
```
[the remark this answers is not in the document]
sunchunqiang: Done

PR #1093, opened 2018-10-11
AWS Credentials from User Environment Variables

travisgosselin:
One of the credential options used automatically as a part of the FallbackCredentialsFactory is to pull the key ID and secret directly out of Environment Variables with the following class: EnvironmentVariablesAWSCredentials. Currently, this class pulls only from "Process" level environment variables. This change adds a check to "User" level variables to see if the key ID and secret are available.

## Motivation and Context
The key motivation for this is mainly for local development purposes using .NET 4.5+. Our team specifies key ID, secret and session token via User level environment variables that is generated and added to the system via many different command line tools available. This is necessary since the role to assume for the applications requires MFA entry on assume (and therefore cannot be assumed directly by the application itself). A temporary session token and values are stored in user level environment variables. When developing in .NET 4.5 against IIS Express all the user level variables are automatically brought into the process and this works fine. However, working against local IIS (with app pool configured as my own user account with the variables) the environment user variables are not loaded into the process (even with the "load user profile" IIS option).

This change would be highly convenient, and also would seem to make sense to pull in User level variables for this. Since IIS Express and .NET Core already automatically bring in the User level variables on start, this should provide more consistency as well.

## Testing
Existing supported tests are passing: EnvironmentalVariablesAWSCredentialsTest
I would like to add additional testing here, but would require further changes to the EnvironmentVariablesAWSCredentials class to mock out the usage of Environment variables (abstract it). I'm open to making these changes if interested.

## Types of changes
The change would be entirely backwards compatible.

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Credentials/_bcl+coreclr/EnvironmentVariablesAWSCredentials.cs
+++ b/sdk/src/Core/Amazon.Runtime/Credentials/_bcl+coreclr/EnvironmentVariablesAWSCredentials.cs
@@ -65,11 +65,11 @@ namespace Amazon.Runtime
         /// <returns></returns>
         public ImmutableCredentials FetchCredentials()
         {
-            string accessKeyId = Environment.GetEnvironmentVariable(ENVIRONMENT_VARIABLE_ACCESSKEY);
-            string secretKey = Environment.GetEnvironmentVariable(ENVIRONMENT_VARIABLE_SECRETKEY);
+            string accessKeyId = GetEnvironmentVariable(ENVIRONMENT_VARIABLE_ACCESSKEY);
+            string secretKey = GetEnvironmentVariable(ENVIRONMENT_VARIABLE_SECRETKEY);
             if (string.IsNullOrEmpty(secretKey))
             {
-                secretKey = Environment.GetEnvironmentVariable(LEGACY_ENVIRONMENT_VARIABLE_SECRETKEY);
+                secretKey = GetEnvironmentVariable(LEGACY_ENVIRONMENT_VARIABLE_SECRETKEY);
                 if (!string.IsNullOrEmpty(secretKey))
                     logger.InfoFormat("AWS secret key found using legacy and non-standard environment variable '{0}', consider updating to the cross-SDK standard variable '{1}'.",
                                       LEGACY_ENVIRONMENT_VARIABLE_SECRETKEY, ENVIRONMENT_VARIABLE_SECRETKEY);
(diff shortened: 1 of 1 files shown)


vellozzi:
I'd add a second environment variable, AWS_CREDENTIALS_IN_USER_ENV.  If that variable is set to true  then get variables from the user's environment, otherwise get them from the process's environment.  This makes the behavior opt-in and ensures that all the values come from the process's environment _OR_ the user's environment, and there's not some mix.

In the meantime I'll discuss this PR with the team.

travisgosselin:
Sure if you think its necessary to add the "opt-in" I can add that in for sure. Let me know what you guys think and if you want to move ahead with that I can:

- Add the opt-in environment variable (which would ideally exist at the "User" level)
- Abstract out the Environment Variable usage to another class for reuse
- Add more unit tests for the enhancements.

I'll wait till I hear back from you. Thanks @vellozzi

travisgosselin:
Hey @vellozzi just wondering if there was any update here? Did you have a chance to talk to the team? Any changes I can start making?

vellozzi:
I'm discussing it with the team.  This is something that we need to standardize across all the AWS SDKs.  So the process will take some time.

It would definitely be an improvement but are you able to work around it for the short term?

vellozzi:
@travisgosselin Have you run a test to see if you can successfully get user environment variables in your IIS use case?  
It seems like if it's not behaving the same way between IIS Express and IIS the user level variables might not even be available.

travisgosselin:
@vellozzi Currently I have implemented a workaround for our teams where they have some custom code that checks the User Environment variables and if it has any, it passes them in as BasicCredentials to the instantiating service, and if none in user environment variables it then uses the normal default CredentialFallback in AWS SDK. 

So to answer your questions:
1) Yes we have workaround in place now, but would like to have this standard for our development. No rush though at the moment on my side.
2) Based on the workaround I have in place on my projects, I can easily pull User environment variables and see the values without issue.

bpayette:
Hi @travisgosselin Sorry for the long silence. I have a question (warning I am not an IIS expert) but shouldn't setting this element in the app pool configuration do what you need?
```xml
 <processModel identityType="ApplicationPoolIdentity" loadUserProfile="true" setProfileEnvironment="true" />
```
`loadUserProfile` is false by default.

bpayette:
Hi @travisgosselin Sorry for the long silence. I have a question (warning I am not an IIS expert) but shouldn't setting this element in the app pool configuration do what you need?
```xml
 <processModel identityType="ApplicationPoolIdentity" loadUserProfile="true" setProfileEnvironment="true" />
```
`loadUserProfile` is false by default.

travisgosselin:
@bpayette Just reviewing this as it was some time ago now. I'd have to dive a bit back into it, but I believe the issue is that it was not pulling User level variables, and only looking at the Process level. I did add a note that even when using the loadUserProfile this was not working as expected: "even with the "load user profile" IIS option".

joshuaflanagan:
I think this change is unnecessary and should be closed. The USER environment variables are set on the process when the process is created by the user.

normj:
I'm an agreement with @joshuaflanagan. I don't understand why we would need to do this. When the .NET process starts all of the System and user environment variables are set for the process. I'm going to close given the length this has been open and I don't see this being needed.

travisgosselin:
Still feels like a gap based on behavioral differences of IIS vs IIS Express on what is expected for consistency. That being said, I'm not sure the responsibility is on AWSSDK To resolve that gap. IIS/Express is not really part of much of my local development workflows anymore and resolving this is not need anymore. Appreciate the time spent in the discussion, even if for not. Thanks for updating and closing.

PR #1108, opened 2018-10-25
DynamoDB: Document should allow assigning of null property values

jens-peterolsson:
DynamoDB DocumentModel

Assigning a Document property with null works, but assigning it from an object with a string property set to null doesn't work. It should generate the same result (a new primitive that will not be included when saving in DynamoDB since empty string values are not allowed).

The added logic applies to strings in DynamoDBEntry, see "this" property in Document for an example of the same logic.

## Testing
See included unit tests.

## Types of changes

## Checklist

## License


normj:
This was merged and released with version 3.3.105.5 of AWSSDK.DynamoDBv2. Thanks for the contribution.

PR #1143, opened 2018-12-12
Make S3 metadata collection keys case-insensitive

Vertigo093i:
## Description
As the title states this PR makes S3 metadata collection keys case-insensitive to be consistent with the headers collection and also with the other comparisons within the class itself.

## Motivation and Context
This makes the SDK compatible with Minio since Minio itself won't be compatible with AWS (minio/minio#5594, minio/minio#6471).

## Testing
Tested with Minio and https://github.com/hotchkj/AspNetCore.DataProtection.Aws project.

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/Model/MetadataCollection.cs
+++ b/sdk/src/Services/S3/Custom/Model/MetadataCollection.cs
@@ -26,7 +26,7 @@ namespace Amazon.S3.Model
     {
         internal const string MetaDataHeaderPrefix = "x-amz-meta-";
             
-        IDictionary<string, string> values = new Dictionary<string, string>();
+        IDictionary<string, string> values = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
 
         /// <summary>
         /// Gets and sets meta data for the object. Meta data names must start with "x-amz-meta-". If the name passeed in as the indexer 


matteo-prosperi:
Because the content of the MetadataCollection is publicly accessible (users can add, read and enumerate metadata), we may break the existing behavior of customer applications by making this change.

Vertigo093i:
Hm, in which a way it may break? Users still can _add, read and enumerate metadata_ without any noticeable impact. This is a widening change so it shouldn't break anything. But this fixes some strange misbehaviors. Moreover, according to HTTP RFC header names (which are essentially metadata keys) a not case sensitive therefore `x-amz-meta-yyy` and `X-Amz-Meta-Yyy` should be treated equally.

Nevertheless, other parts of code within the class already use case-insensitive comparison of keys.

matteo-prosperi:
Within a minor version of the SDK we are extremely cautious to maintain backward compatibility. In case someone tried adding both `x-amz-meta-yyy` and `X-Amz-Meta-Yyy` to the metadata they would have a different experience than today. Also someone who looks searches the collection for `x-amz-meta-yyy` when only `X-Amz-Meta-Yyy` is present would have a different behavior.
I understand that these may not be likely scenarios, but I wouldn't consider even such a small change in behavior "safe".

swlasse:
We have been hit by the exact same issue when integrating Minio.

To my understanding the S3 headers are case-insensitive ( see: https://forums.aws.amazon.com/thread.jspa?threadID=86140 and https://stackoverflow.com/a/9202551/700926 ).

Given this, the case-_sensitive_ behavior of the `Amazon.S3.Model.MetadataCollection` class is inconsistent with the behavior of S3 which is a bit confusing and error prone.

I understand that you have to be careful not introducing breaking changes here, so I can only support that this is getting fixed in the next major version of the SDK. I just wanted to point out that we are more people "out there" getting hit by this.

Vertigo093i:
I have to note that now we use another S3-compatible cloud storage provider which have anoother different way of capitalizing such headers: `X-amz-meta-yyy`.

So please @matteo-prosperi pay more attention to this.

costleya:
Closing issue. We will look at this change as part of a major version bump of the SDK, but there is the possibility of backwards incompatibility, even if it is to conform to spec. Due to this risk, and the fact that it is only affecting 3rd party "s3 compatible solutions," we are not going to make this change at this time.

Thank you for your feedback.

Vertigo093i:
A very bad decision...

Issue #772, opened 2017-10-16
Cannot mock DynamoDBv2 AsyncSearch for unit testing AWS calls


jazpearson:
Would like to be able to properly unit test my code which uses the AWS SDK, and in particular the AsyncSearch class.

## Possible Solution
1. Create an interface for this class
2. Create an empty public constructor may also work

mercury-jin-autodesk:
I have  the same issue with AsyncSearch.

ColWhi:
Any update on this, causing us some pain

assyadh:
Thanks @Supercheez, I'll have a look at the PR.

arifemre61:
> at the PR.

Do you have chance to look at it ?

0leksii:
Hello,

Any update regarding this?

j1mb0jay:
This would be very nice please

leninator:
+ 1 vote for me. It's painful to test.

Elchapanebro:
Frankly, The TDD approach of this lib is pretty shocking, why have you got interfaces that expose internally constructed objects such as `AsyncSearch<T>` and `Table` ? It makes for grim reading when trying to carry out the simplest of mocking in unit tests.

Just adding a public empty constructor would allow a safe non breaking fix for this - albeit not the best, but it should not have taken over a year to come to this conclusion.

kwbauer:
When will we see movement on this issue. Not being able to write proper unit tests is very frustrating.

Just add the empty and public constructor so that we can mock the class.

Elchapanebro:
> When will we see movement on this issue. Not being able to write proper unit tests is very frustrating.
> 
> Just add the empty and public constructor so that we can mock the class.

@kwbauer  - I ended up caving in and wrapping this code in a new class with what should be the appropriate interface in this library. - I means that my unit tests are testing the wrapper, and not the actual call within the AWSSDK which means I'm not 100% happy about it, but it will do. Code snippets below.

```c#
    internal interface IAsyncSearch<T>
    {
        bool IsDone { get; }

        Task<IReadOnlyList<T>> GetNextSetAsync(CancellationToken cancellationToken);

        Task<IReadOnlyList<T>> GetRemainingAsync(CancellationToken cancellationToken);
    }

    internal sealed class AsyncSearchWrapper<T> : IAsyncSearch<T>
    {
        private readonly AsyncSearch<T> asyncSearch;

        public AsyncSearchWrapper(AsyncSearch<T> asyncSearch)
[...]

PR #1187, opened 2019-01-23
Added interfaces for AsyncSearch and AsyncSearch.Async and updated us…

Supercheez:
…ages.

Added interfaces for AsyncSearch and AsyncSearch.Async and updated usages. This means that the return value from IDynamoDBContext async methods can be mocked for unit testing purposes.  

## Description

- Added IAsyncSearch/IAsyncSearch interfaces corresponding to AsyncSearch/AsyncSearch.Async
- Updated IDynamoDBContext.Async to return an interface rather than a concrete class
- Updated implementations of IDynamoDBContext.Async to match IDynamoDBContext.Async

## Motivation and Context

This fix replaces the concrete class(es) returned by the DynamoDbContext with a corresponding interface so that they can be mocked for unit testing purposes.

https://github.com/aws/aws-sdk-net/issues/772

## Types of changes

## Checklist

## License


sdk/src/Services/DynamoDBv2/Custom/DataModel/_async/IAsyncSearch.Async.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace AWSSDK.DynamoDBv2.CoreCLR.Custom.DataModel
+{
+    /// <summary>
+    /// A strongly-typed object for retrieving search results (Query or Scan)
```
ColWhi: Copy paste error

assyadh:
Thanks @Supercheez, I'll have a look at the PR.

sdk/src/Services/DynamoDBv2/Custom/DataModel/_async/Context.Async.cs
```diff
@@ -301,7 +302,7 @@ public Task ExecuteBatchWriteAsync(BatchWrite[] batches, CancellationToken cance
         /// </param>
         /// <param name="operationConfig">Config object which can be used to override that table used.</param>
         /// <returns>AsyncSearch which can be used to retrieve DynamoDB data.</returns>
-        public AsyncSearch<T> ScanAsync<T>(IEnumerable<ScanCondition> conditions, DynamoDBOperationConfig operationConfig = null)
+        public IAsyncSearch<T> ScanAsync<T>(IEnumerable<ScanCondition> conditions, DynamoDBOperationConfig operationConfig = null)
```
assyadh: Changing the return type here is not possible as it would be a breaking change. 
We may have customers with a code base relying on the return type AsyncSearch<T>

Supercheez: Sure, have created a new interface so that consumers can choose whether to have an concrete class or an interface returned.

assyadh:
I don't think duplicating the interfaces and classes is the best way to go. Ultimately if your end goal is to mock AsyncSearch with moq, there might be a better way. 

Let me investigate and I will see if I can come up with something cleaner.

H.

sdk/src/Services/DynamoDBv2/Custom/DataModel/_async/Context.Async.cs
normj: You might want to look into using explicit interface implementation https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/interfaces/explicit-interface-implementation

It should allow you to leave the original method untouched but create a second implementation something like

```csharp
IAsyncSearch<T> IAsyncSearch<T>.ScanAsync<T>(IEnumerable<ScanCondition> conditions, DynamoDBOperationConfig operationConfig = null)
{
  ...
}

```
The new method is only visible when working with a variable declared as the interface type.

storkie:
I've ended up creating wrapper classes and interfaces.
Example here:
https://gist.github.com/storkie/45c32d1ba69e3d1f6364c9ffd7ad857e

Use the DynamoDbContextWrapper/IDynamoDbContextWrapper where you need to use it with Moq for unit testing.
It should be only a few changes to your existing code and it won't break on AWS SDK updates.

```
using (IDynamoDbContextWrapper dynamoDbContext = new DynamoDbContextWrapper())
{
	IAsyncSearchWrapper<MyDataModel> query = dynamoDbContext.FromQueryAsync<MyDataModel>(new QueryOperationConfig
	{
		Filter = queryFilter,
		Limit = 100
	});
	
	
	/*...*/
}
```

For Moq it would be something like this:

```
Mock<IAsyncSearchWrapper<MyDataModel>> asyncSearchWrapper = new Mock<IAsyncSearchWrapper<MyDataModel>>();

asyncSearchWrapper.Setup(x => x.IsDone).Returns(true);
asyncSearchWrapper.Setup(x => x.GetNextSetAsync(It.IsAny<CancellationToken>()))
	.ReturnsAsync(listOfMyDataModel);

Mock<IDynamoDbContextWrapper> dynamoDbContext = new Mock<IDynamoDbContextWrapper>();

dynamoDbContext.Setup(x => x.FromQueryAsync<MyDataModel>(
		It.IsAny<QueryOperationConfig>, null))
	.Returns(asyncSearchWrapper.Object);
```

RamiSh:
The wrapper approach doesn't work for me. Curios if anyone ran into this issue where the mocked method returns a null IAsyncSearchWrapper<T>

kwbauer:
Have a look at the comments for issue #772 (https://github.com/aws/aws-sdk-net/issues/772) for a much simpler solution to the unit testing problem. Just add an empty constructor and make it public.

That is definitely not a breaking change.

Supporting IAsyncEnumerable would be nice but is beyond the scope of the original purpose. That should definitely be a separate issue.

RichieRunner:
When is this going to be merged?

RichieRunner:
@normj  @costleya  please re-review this PR. The community has done the work, and it has been half a year, what else are we waiting for?

dario-griffo-cko:
Any update on this?

dario-griffo-cko:
> I don't think duplicating the interfaces and classes is the best way to go. Ultimately if your end goal is to mock AsyncSearch with moq, there might be a better way.
> 
> Let me investigate and I will see if I can come up with something cleaner.
> 
> H.

Sadly Moq requires public parameterless constructors for objects

Kralizek:
That is not entirely accurate. Moq supports public constructors with parameters, but you'll have to provide values for the parameters yourself.

cbude:
A solution would be appreciated. 
I tried the approach of instead mocking the AmazonDynamoDBClient / IAmazonDynamoDB instead and injecting it in the DynamoDBContext but that is also a dead end since the non-asynchronous methods (i.e. DescribeTable) aren't available in .net core interface.

p-brito:
@Supercheez when will this PR be merged?

p-brito:
@normj @costleya any update on this?

kidchenko-tw:
there is any other option to mock this?

p-brito:
> there is any other option to mock this?

You can perhaps create some wrapper around it, but the point of doing unit tests is lost because you aren't actually testing the real component.

kidchenko-tw:
Everyone is creating a wrapper around it, can we just merge this PR  it contains the `IDynamoDBContextWithInterfaces` that is what we need @normj  any thoughts about?

mbp:
@normj please let us have this merged

andyfurniss4:
I'd also like to see this merged. There doesn't seem to be a nice/easy way to unit test and mock DynamoDB queries due to the QueryAsync method not returning an interface.

> 'Unsupported expression: x => x.GetRemainingAsync(It.IsAny<CancellationToken>())
> Non-overridable members (here: AsyncSearch<Subscription>.GetRemainingAsync) may not be used in setup / verification expressions.'

sdk/src/Services/DynamoDBv2/Custom/DataModel/_async/Context.Async.cs
Kirkaiya: @normj - is someone else picking up this PR?  I was just trying to mock DynamoDBContext with IDynamoDBContext and ran into the same issue (AsyncSearch doesn't implement an interface, and it's constructor is not even public, so it basically makes mocking the context impossible)?  Thanks!  - kirkdavi@

Kirkaiya:
> 
> 
> A solution would be appreciated.
> I tried the approach of instead mocking the AmazonDynamoDBClient / IAmazonDynamoDB instead and injecting it in the DynamoDBContext but that is also a dead end since the non-asynchronous methods (i.e. DescribeTable) aren't available in .net core interface.

I just ran into the exact same thing (got the exception for DescribeTable(), and wasn't able to work around it).

jcmorenoo (approved):

normj:
We [released](https://github.com/aws/aws-sdk-net/pull/2089) a change that allows to mock the AsyncSearch by sub classing AsyncSearch.
```
            var mockDBContext = new Mock<IDynamoDBContext>();
            mockDBContext
                .Setup(x => x.ScanAsync<DataItem>(
                   It.IsAny<IEnumerable<ScanCondition>>(),
                   It.IsAny<DynamoDBOperationConfig>()))
                .Returns(
                   new MockAsyncSearch<DataItem>() // Return mock version of AsyncSearch
                );

            var search = mockDBContext.Object.ScanAsync<DataItem>(new List<ScanCondition>());
            Assert.IsInstanceOfType(search, typeof(MockAsyncSearch<DataItem>));

            var items = await search.GetNextSetAsync();
            Assert.AreEqual(0, items.Count());
        }

        public class DataItem
        {
            public string Id { get; set; }
        }

        public class MockAsyncSearch<T> : AsyncSearch<T>
        {
            public override Task<List<T>> GetNextSetAsync(CancellationToken cancellationToken = default(CancellationToken))
            {
                return Task.FromResult(new List<T>());
            }
        }
```
We debated about an approach like this using explicit interface implementation but using sub classing seemed less impactful.

PR #1206, opened 2019-02-05
Allow setting aws logger through appsettings.json in .net core and respond to runtime configuration changes

Bio2hazard:
## Description

Added the ability to set a logger through appsettings.json ( or any IConfiguration source ) under the `LogTo` property. `LogTo` was chosen as a name to mirror `AWSConfigs.cs`. I am aware that the field is obsolete in `AWSConfigs.cs`, however the related `LogResponse` and `LogMetrics` can already be set on the configuration file, and I thought it would be cleaner to keep them together.

Added an options monitor that listens for configuration changes and will forward any changes made to the `LogTo` property to `AWSConfigs.cs`. The  `InvokeChanged` method can be extended in the future to react and propagate other updated configurations in the future. A lock is used to prevent rapid file updates from colliding.

As the unit test needs to write changes to a JSON file, I've added a helper that creates a temporary file and stores the path to it.

The monitor absolutely the IConfiguration, so the change monitoring only works if either of the new `AddDefaultAWSOptions` overloads is used. 

One open thought or concern is that users could inject their AWSOptions class somewhere and change LogTo, and it would not do anything ( as only the change monitor propagates the updates value to AWSConfigs.cs ). I think the best approach there is to make AWSOptions implement property change notification and making the monitor respond to them. This ensures we only have a single AWSOptions-of-truth ( the singleton in the DI context ) and only respond to changes made to that particular instance.

## Motivation and Context

We would have been able to debug an ongoing issue if we have had the ability to turn on logging during runtime. With this change you can go onto your box/container, update appsettings to specify LogTo and get the data you need.

This is especially important as dotnet tooling for linux is still lacking, and any form of debugging or dumping requires extra software that can't always be installed on production boxes, meaning logging can sometimes be the only way to debug ongoing production issues.

## Testing

Added 3 Tests that ensure correct propagation of the LogTo property and the monitor's ability to respond to file changes.

## Types of changes

## Checklist

I held off on documentation changes for now, as I am not 100% sure how you want to document this new functionality.

## License

### DIFF
--- a/extensions/src/AWSSDK.Extensions.NETCore.Setup/AWSOptions.cs
+++ b/extensions/src/AWSSDK.Extensions.NETCore.Setup/AWSOptions.cs
@@ -47,6 +47,12 @@ namespace Amazon.Extensions.NETCore.Setup
         public RegionEndpoint Region { get; set; }
 
         /// <summary>
+        /// The Logging destination.
+        /// Changes to this setting will propagate within 500 ms.
+        /// </summary>
+        public LoggingOptions LogTo { get; set; }
+
+        /// <summary>
         /// AWS Credentials used for creating service clients. If this is set it overrides the Profile property.
         /// </summary>
         public AWSCredentials Credentials { get; set; }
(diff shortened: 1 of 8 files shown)


afischer-opentext-com:
What needs to be done to move this MR forward?

normj:
The `AWSSDK.Extensions.NETCore.Setup` library has been updated to support configuring logging. Here is an example `appsettings.json` file showing how to configure logging. https://github.com/aws/aws-sdk-net/blob/master/extensions/test/NETCore.SetupTests/TestFiles/EnableLoggingTest.json

Issue #1258, opened 2019-04-12
signature mismatch when user metadata keys contain _ (underscore)


ben-schumacher:
## Expected Behavior
When setting user metadata keys in put object request, an underscore will create a canonicalized header string in the wrong order causing a signature mismatch.

## Current Behavior

Request return a signature mismatch error instead of 200 OK response.

The string comparer is upper-casing the the metadata key names which will incorrectly order the metadata keys with "_" to be lower in the sort.

## Possible Solution

https://github.com/aws/aws-sdk-net/blob/master/sdk/src/Core/Amazon.Runtime/Internal/Auth/S3Signer.cs#L154
Need to change the comparer to InvariantCultureIgnoreCase in that line.

## Steps to Reproduce (for bugs)

public static void Main(string[] args)
        {
            // create the AWS S3 client
            AmazonS3Client s3 = AWSS3Factory.getS3Client();

            // retrieve the object key and object content from the user
            Console.Write("Enter the object key: ");
            string key = Console.ReadLine();
            Console.Write("Enter the object content: ");
            string content = Console.ReadLine();

            // create object request with retrieved input
            PutObjectRequest request = new PutObjectRequest()
            {
                BucketName = AWSS3Factory.S3_BUCKET,
                ContentBody = content,
                Key = key
            };

            request.Metadata.Add("acquisitionprotocol", "http");
            request.Metadata.Add("acquisition_time_ms", "5");

            // create the object in demo bucket
            s3.PutObject(request);
           

            // print out object key/value for validation
            Console.WriteLine(string.Format("Created object {0}/{1} with content: {2}", AWSS3Factory.S3_BUCKET, key, content));
            Console.ReadLine();
        }

## Context

Can't create objects with "_" character in the metadata key.

## Your Environment

* AWSSDK.Core version used: 3.3.100.5
* Service assembly and version used: 
* Operating System and version: Windows 10
* Visual Studio version: Community 2019
* Targeted .NET platform: 4.5

klaytaybai:
@ben-schumacher, thanks for submitting this.  I was able to reproduce it with .NET Framework 4.5.  We'll take a look at your pull request.

ben-schumacher:
@klaytaybai I unfortunately discovered today that my fix now incorrectly sorts:

(-)   45  02/13   55  2D                 HYPHEN, MINUS SIGN
(|)  124  07/12  174  7C                 VERTICAL LINE, VERTICAL BAR
[...]

PR #1259, opened 2019-04-12
Changed OrderBy comparer to handle underscore characters in metadata …

ben-schumacher:
…keys.

## Description

Changed the OrderBy comparer so that when user metadata keys are supplied (that container "_" underscore), the canonical string is sorted correctly and we don't get signature mismatch.

## Motivation and Context

Signature mismatch error.

https://github.com/aws/aws-sdk-net/issues/1258

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [ x] I have read the **README** document

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4Signer.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4Signer.cs
@@ -629,18 +629,22 @@ namespace Amazon.Runtime.Internal.Auth
         /// <param name="requestHeaders">The set of proposed headers for the request</param>
         /// <returns>List of headers that must be included in the signature</returns>
         /// <remarks>For AWS4 signing, all headers are considered viable for inclusion</remarks>
-        protected static IDictionary<string, string> SortAndPruneHeaders(IEnumerable<KeyValuePair<string, string>> requestHeaders)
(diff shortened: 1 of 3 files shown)


sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4Signer.cs
```diff
@@ -629,18 +629,22 @@ internal static string DetermineService(IClientConfig clientConfig)
...
-                if (_headersToIgnoreWhenSigning.Contains(header.Key))
-                {
+                var lowerKey = item.Key.ToLowerInvariant();
+                if (_headersToIgnoreWhenSigning.Contains(item.Key))
                     continue;
-                }
-                sortedHeaders.Add(header.Key, header.Value);
+                preSortedHeaders.Add(lowerKey, item.Key);
+            }
+            var sortedHeaders = new Dictionary<string, string>();
```
normj: I'm missing something with this PR. If this is returning a Dictionary and not a SortedDictionary anymore then it seems like we have lost the sort order since Dictionary doesn't guarantee the enumeration will come back in the same order they were added.

ashishdhingra:
Issue fixed in version 3.5.123.1, with refactored commit https://github.com/aws/aws-sdk-net/commit/1f6a23aa77a0e4a1f6d73cfc813c1672431f48c4

PR #1340, opened 2019-07-09
Update AmazonSecurityTokenServiceConfig.cs

jiveabillion:
the call to FallbackRegionFactory.GetRegionEndpoint(false) throws exceptions when the configuration is not set. This causes problems when the application has no configuration, does not intend to ever have one, and AmazonSecurityTokenServiceClient is instantiated by passing the RegionEndpoint in the constructor. It really should never call FallbackRegionFactory.GetRegionEndpoint(false) if the RegionEndpoint is passed to the constructor. This caused me an entire day of frustration, and required me to set AWSConfig.Region in my application startup.

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/SecurityToken/Generated/AmazonSecurityTokenServiceConfig.cs
+++ b/sdk/src/Services/SecurityToken/Generated/AmazonSecurityTokenServiceConfig.cs
@@ -42,7 +42,15 @@ namespace Amazon.SecurityToken
         public AmazonSecurityTokenServiceConfig()
         {
             this.AuthenticationServiceName = "sts";
-            var region = FallbackRegionFactory.GetRegionEndpoint(false);
+            RegionEndpoint region;
+            try
+            {
+                region = FallbackRegionFactory.GetRegionEndpoint(false);
+            }
+            catch
+            {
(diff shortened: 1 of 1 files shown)


normj:
I don't see how `FallbackRegionFactory.GetRegionEndpoint(false)` would throw an exception even if config is not set. Could you provide a sample of the exception you got?

jiveabillion:
It does it man, just try it without any settings in your app. I wasted a
good few hours trying to figure out why I couldn't just set the region in
the constructor without it throwing an exception. My app doesn't use any
preset credentials or regions, so this api does not appropriately work for
me, but only this class is causing the problem as far as I know

normj:
I have tried it and have not been able to reproduce the exception. I have called the `AmazonSecurityTokenServiceClient` constructor passing in a region as well as directly call `Amazon.Runtime.FallbackRegionFactory.GetRegionEndpoint(false)` which returns null for me. So something is different between your environment and mine so I'm trying to understand what that is.

SrutiG:
Are you able to provide some more information on your environment to help us recreate this issue?

normj:
Closing for lack of activity.

PR #1427, opened 2019-10-11
TryParseAmazonS3Uri(string) should not throw

OlegBoulanov:


SrutiG:
Could you please provide more details in this PR description about your use case and the error message you are getting? We would definitely like to address the issue of a TryParse method throwing an exception.

OlegBoulanov:
https://github.com/OlegBoulanov/aws-sdk-net/commit/e1a59ca501cfce923704b033eb1c619c0ee92666

The problem is that TryParse(string) invokes TryParse(Uri.Parse(string)) internally - which may throw, but TryParse(whatever) should not throw by definition.

OlegBoulanov:
and of course there is a bug... please reject that PR, I will submit another one

OlegBoulanov:
Ok, all set

normj:
This PR was released in the latest version of the SDK

PR #1447, opened 2019-11-07
Add Cognito to IdentityProviderTypeType

mlapaglia:
the `IdentityProviderTypeType` class is missing an entry for `Cognito`
## Description
Entries exist for `Facebook`, `Google`, etc, but not `Cognito` which is a supported Identity Provider Type.

## Motivation and Context
I like using the constants instead of strings, but I can't because they are missing for `Cognito`
image

## Testing
I ran the application and it created a `UserPoolClient` with Cognito enabled.

## Types of changes

- [*] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [*] My code follows the code style of this project
- [*] I have read the **README** document
- [*] All new and existing tests passed

## License

- [*] I confirm that this pull request can be released under the Apache 2 license

### DIFF
--- a/sdk/src/Services/CognitoIdentityProvider/Generated/ServiceEnumerations.cs
+++ b/sdk/src/Services/CognitoIdentityProvider/Generated/ServiceEnumerations.cs
@@ -1079,7 +1079,7 @@ namespace Amazon.CognitoIdentityProvider
         /// <summary>
         /// Constant Cognito for IdentityProviderTypeType
         /// </summary>
-        public static readonly IdentityProviderTypeType Cognito = new IdentityProviderTypeType("Cognito");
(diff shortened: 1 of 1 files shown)


mlapaglia:
done!

normj:
This change is in the generated code for Cognito Identity Provider. I'm not a Cognito expert but if COGNITO is a valid value then we need to get the service to change their model to include that as a valid value. For background purposes here is the service model the Cognito vends the to SDK that we generate from. You can see that is missing the COGNITO enum value. https://github.com/aws/aws-sdk-net/blob/master/generator/ServiceModels/cognito-idp/cognito-idp-2016-04-18.api.json#L3809

Could you open an issue in the Cognito Forums to let them know this missing enum value. Hearing directly from users helps prioritize service teams task list.
https://forums.aws.amazon.com/forum.jspa?forumID=173

PR #1517, opened 2020-02-02
Fix messaging for diagnostic title

sjp:
## Description
When there is an invalid value passed into a service client, for example an S3 bucket name could be given a single name character for an `Amazon.S3.Model.PutObjectRequest.BucketName` property.

This will (correctly) trigger a code analysis diagnostic to be reported in the form of `{ServiceName}1002`, continuing the S3 example, `S31002`.

However, both the title and the description of the error message are slightly incorrect, and do not match the message of the description. I have updated both the title and the description to match the intention of the diagnostic.

## Motivation and Context
By chance I happened to encounter the affected diagnostic and noticed that it could be improved.

## Testing
I tested this by taking the resulting assembly, referencing it locally, and reproducing the result. I could have made larger changes to also verify via unit tests, but this was avoided because the scope of this change is so small.

## Types of changes

## Checklist

## License


normj:
This was merged and released with the latest version of the SDK.

PR #1523, opened 2020-02-09
AWSSDK.Extensions.NETCore.Setup -> 3.5.0-beta

Jacko1394:
## Types of changes

## Checklist

## License

### DIFF
--- a/extensions/src/AWSSDK.Extensions.NETCore.Setup/AWSSDK.Extensions.NETCore.Setup.csproj
+++ b/extensions/src/AWSSDK.Extensions.NETCore.Setup/AWSSDK.Extensions.NETCore.Setup.csproj
@@ -1,10 +1,12 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.3;netstandard2.0</TargetFrameworks>
+    <LangVersion>8.0</LangVersion>
+    <Nullable>disable</Nullable>
+    <Version>3.5.0-beta</Version>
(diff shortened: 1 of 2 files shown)


normj:
Thanks for the PR getting this library updated for the new 3.5 version of the SDK.

All that really needs to be done to make this library work for the 3.5.0-beta is change the `AWSSDK.Extensions.NETCore.Setup.nuspec` to say `..., 3.5)` instead of `..., 3.4)` for the 2 places `AWSSDK.Core` is called out.

I don't really want to update the versions of the dependencies because I'm sure we still have plenty of users that are still using .NET Core 2.X and we should work in the default versions of those dependencies that shipped with .NET Core.

We are still on the fence with our AWS SDK for .NET version 3.5 whether to remove support for netstandard1.3. 

You can make the changes to the PR but if you want we can take care of the changes as well as this is important to making sure we have a good 3.5 experience.

Jacko1394:
i'd recommend not worrying about 1.3 and just move on forward to `netstandard2.0`

Jacko1394:
Screen Shot 2020-02-28 at 11 33 31 am

normj:
Version 3.3.101.0 of AWSSDK.Extensions.NETCore.Setup was released that makes the package compatible with 3.5.0-beta. I agree we can probably remove netstandard1.3 I just need to spend sometime to collect the data and prove it. Also NuGet can sometimes be pretty slow with updating the metrics. It is not uncommon to see a delay in days for the metrics to be updated.

normj:
Closed as this was already taken care of with version 3.3.101.0 of the package

PR #1545, opened 2020-03-04
Make Metadata Collection case insensitive

### LINKED ISSUE #927: Header keys need to be considered as case insensitive
AWS SDK .NET needs to consider header keys as case insensitive as HTTP RFC says: http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2

```
Each header field consists of a name followed by a colon (":") and the 
field value. Field names are case-insensitive.
```

It would be great of this can be fixed as done in AWS SDK java: https://github.com/aws/aws-sdk-java/pull/1016

rluetzner:
## Description

The metadata collection has been changed to use a case insensitive Dictionary.

## Motivation and Context

I was unable to use the Amazon S3 Encryption Client class with a MinIO S3 server.

Due to the Go Language implementation of HTTP response headers keys are always appended in camelcase. This prevents usage of the Encryption Client, as the expected headers were not detected correctly, even though they were set by the MinIO server.

The GetObject request caused a misleading exception, as the Encryption Client's fallback causes it to look for an instruction file instead.

Additionally this change ensures compatibility with RFC 2616:

Each header field consists of a name followed by a colon (":") and the
field value. Field names are case-insensitive.

https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2

This fixes #927 .

## Testing

Manual tests with MinIO S3 server.

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/Model/MetadataCollection.cs
+++ b/sdk/src/Services/S3/Custom/Model/MetadataCollection.cs
@@ -26,7 +26,7 @@ namespace Amazon.S3.Model
     {
         internal const string MetaDataHeaderPrefix = "x-amz-meta-";
             
-        IDictionary<string, string> values = new Dictionary<string, string>();
+        IDictionary<string, string> values = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
 
         /// <summary>
         /// Gets and sets meta data for the object. Meta data names must start with "x-amz-meta-". If the name passeed in as the indexer 


diegofrata:
Apologies, I did not notice this PR and ended up raising a new one. The only difference is one test case was added on mine. https://github.com/aws/aws-sdk-net/pull/1616

I disagree with this being classified as a breaking change. Anyone using this in production must be using lowercase attributes -- otherwise, their code would not work at all.

rluetzner:
Closed, as this exactly matches the changes in #1616 which contains an additional unit test. Thanks @diegofrata.

rluetzner:
Just for completeness, I considered this a breaking change as this implementation in the SDK seems to work correctly when using Amazon S3 Object Storage. At least it did when I tried it last. 

I only ever had issues when trying to use this SDK in combination with another non Amazon S3 Object Store (MinIO). I considered the possibility that this incompatibility to other S3 Object Store providers might be intended.

diegofrata:
@rluetzner Makes total sense. The reason I made the point is to perhaps sway the maintainers that this is a safe PR to merge as it can't really break any code -- any code depending on case-sensitive lookups is already broken as a roundtrip to S3 will essentially lowercase anything that isn't lowercased. I am observing this behavior in my own code using S3 directly. Having said that, it baffles me that such a simple PR has been sitting unreviewed for over 2 months! I guess everyone at AWS must be doing Amazon Prime deliveries these days... :)

Issue #1547, opened 2020-03-04
S3 Client does not handle 302 and 303 correctly


talibsharif:
S3 Client does not handle 302 and 303 correctly

## Expected Behavior
- The client should follow the redirect.

## Current Behavior
- The client throws an Exception with an Unexpected HTTP Status Code

## Possible Solution
- Fix the Client Code to act the same has HTTP 307

talibsharif:
Here is the Pull Request Addressing this:
https://github.com/aws/aws-sdk-net/pull/1548

klaytaybai:
Thanks for the PR.

talibsharif:
hi @klaytaybai - can you give some guidance on the timelines for getting such a fix into a release version? This will help us know whether to wait for update or just compiled and use a forked build (would prefer not to).

ashishdhingra:
Hi @talibsharif,

Good afternoon.

Looks like the PR #1548 was not approved by @normj. Please let me know if we could close this issue.

Thanks,
Ashish

PR #1548, opened 2020-03-04
- Add support for honoring 302 and 303 redirects

talibsharif:
- Add Support for honoring 302 and 303 as long as Location Header is present (Same behavior as 307).

- This current client implementation expects to be only talking to S3 Directly which only returns 307, but this can change if the Network topology adds a Caching Proxy in the middle, which can at times return 302 or 303.
- The Unit tests are added in the same pattern as existing tests, which is that they are independent and self contained (thus look repetitive).

Add Support for honoring 302 and 303 as long as Location Header is present (Same behavior as 307).

## Description
- This current client implementation expects to be only talking to S3 Directly which only returns 307, but this can change if the Network topology adds a Caching Proxy in the middle, which can at times return 302 or 303.
- The Unit tests are added in the same pattern as existing tests, which is that they are independent and self contained (thus look repetitive).

## Motivation and Context
- Allow S3 Client to work in Network Topology that can have a Proxy in the middle

## Testing
- Unit Tests

## Types of changes

## Checklist

- [X ] My code follows the code style of this project

## License

[issues]: https://github.com/aws/aws-sdk-net/issues/1547

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/RedirectHandler.cs
+++ b/sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/RedirectHandler.cs
@@ -106,7 +106,10 @@ namespace Amazon.Runtime.Internal
             if (response.StatusCode >= HttpStatusCode.Ambiguous &&
                 response.StatusCode < HttpStatusCode.BadRequest)
             {
-                if (response.StatusCode == HttpStatusCode.TemporaryRedirect &&
+                // HTTP 302, 303,and 307 are all valid Redirect Mechanisms
+                if ((response.StatusCode == HttpStatusCode.Redirect ||
+                     response.StatusCode == HttpStatusCode.RedirectMethod ||
+                     response.StatusCode == HttpStatusCode.TemporaryRedirect ) &&
                     response.IsHeaderPresent(HeaderKeys.LocationHeader))
                 {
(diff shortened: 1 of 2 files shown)


talibsharif:
Linked Issue: https://github.com/aws/aws-sdk-net/issues/1547

talibsharif:
hi @klaytaybai - can you give some guidance on the timelines for getting such a fix into a release version? This will help us know whether to wait for update or just compiled and use a forked build (would prefer not to).

klaytaybai:
Hoping for early next week

normj:
Sorry but after reviewing the PR and discussing with other AWS SDK teams we can't accept this PR. Our main concerns are if you look at this change outside of your proxy use case it would break signing as well as potentially send a request to another AWS region. With the AWS SDKs we avoid making cross region calls as much as possible. The main exception being S3's legacy behavior with us-east-1 which we have been trying to move away from. The other problem is it would make the .NET SDK inconsistent with all of the other SDK in a pretty fundamental way and that is another thing we are trying to fix is the AWS SDKs is all of their inconsistences.

I think what we need is a way for you to be able to extend/replace the SDK's redirect behavior because you know your network topology and have the information to make the correct decisison for doing the redirect. If we do it in the SDK we are making that change for all users without knowing the network topology as you said.

PR #1552, opened 2020-03-09
Fix performance issue: Check file existence before reading it.

Vfialkin:
## Description

Fix performance issue: Check file existence before reading it. Throwing exceptions is expensive. 

The current approach uses exceptions to control the flow with empty try catches and no logs, which in some cases results in performance downgrade. Added file validation before the file read. It might still happen due to concurrency that file got removed after the validation but it should cover 99% of the normal scenarios.

## Motivation and Context

We found a degradation in performance when a profile file didn't exist. 
## Testing

Manual, unit tests don't need to be updated
## Types of changes

## Checklist

## License


Vfialkin:
Got it,  thanks for the quick review! Committed a fix for null ref.

Vfialkin:
Hi, I just want to confirm if there is anything else I need to do to make it mergeable?

Vfialkin:
Still not merged :)

boblodgett:
Thanks for the PR. We have manually merged the PR and released version 3.3.106.0 of AWSSDK.Core with this change.

PR #1572, opened 2020-04-08
Adds ToJson as an extension method for IEnumerable<Document>.

andyhopp:
## Description
The current SDK does not expose a way to serialize a list of Document objects returned from operations such as Query or Scan. This change introduces a new extension method that wraps the existing JSON serialization methods inside an array start/end.

## Motivation and Context
Implementers of Serverless Application Model Lambdas need a way to populate the body of the response as a string. If the intent is to return an array of Document objects, they would either have to resort to string concentation + <Document>.ToJson() or other such workarounds. This will eliminate the need for any such customer to write that code.

## Testing
Implemented in a test .NET Core Lambda, ran against populated/empty tables

This change does change the access modifier of JsonUtils.WriteJson from private to internal in order to re-use that code. The fact that it is an extension method means no modification of the Document.cs code was necessary.

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/AWSSDK.NetStandard.sln
+++ b/sdk/AWSSDK.NetStandard.sln
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.26730.12
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.29926.136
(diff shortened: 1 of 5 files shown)


normj:
Version 3.3.106 of AWSSDK.DynamoDBv2 has been released with this PR

PR #1658, opened 2020-07-11
[FIX] Unity Readme: Incorrect XML and NullReferenceException

### LINKED ISSUE #1635: Is there a long-term support for the AWS/Unity integration?
(This is a cross-post from the samples repo, since I could not get anything working in Unity based on the available docs :-/)

We have a Unity app that currently uses a native web view to upload files to S3. To speed up this, we'd love to integrate the AWS SDK into Unity and do this completely natively.

What puzzles me a bit is that there's no "official AWS package" in the Unity asset store, and I can see issues here describing major problems (like not being able to run the samples in Unity 2019). And when I tried them out, I could not even make them work, just ran into one weird exception after the other. I would consider myself as a quite experienced Unity dev.

So I am wondering if there's any kind of official (long-term) support for this integration. It won't help me if it kinda works now, but breaks later and no one cares for the numerous bugs... It is also very confusing that the link to the AWS Mobile SDK for Unity here 

https://docs.aws.amazon.com/mobile-sdk/index.html

actually links to Amplify, while the other links below still refer to the Unity SDK.

The same goes for the "Download AWS SDK for Unity" download link that shows up on the "Getting Started with Cognito page" below. That link points to Amplify, but with no info whatsoever about where to get the actual SDK. This left me with the gut-feeling that the whole Unity support has been dropped with the intention to replace it by Amplify at a later point of time. Correct me if I am wrong.

image

On the same page, there is this link, which does not really point to Unity-specific info:

image

I am particularly getting nervous, because Unity itself is quite well-known for introducing breaking changes in their updates. So, including a package that is not well-maintained itself and kept up-to-date regularly imposes a pretty high upgrade risk on your app.

boxseto:
Update for proper workable format for Unity by fixing example XML.

I have come across the same issus as [#1635](https://github.com/aws/aws-sdk-net/issues/1635), and the error is fixed after modifying the format of the two XML files, which are shown wrongly in the example.

## Description
Changed README for the proper usage of `link.xml` and `awsconfig.xml` file for use in Unity.

## Motivation and Context
Fixed #1635 issue by providing an updated context.

## Testing
My build is using unity v2019.4.2f1, with the latest build of aws .net sdk

## Types of changes

## Checklist

## License

### DIFF
--- a/Unity.README.md
+++ b/Unity.README.md
@@ -32,14 +32,14 @@ There are only a few fundamentals that are helpful to know when developing again
 * To enable logging you need to create a config file called awsconfig.xml in a `Resources` directory add add the following
 
 		<?xml version="1.0" encoding="utf-8"?>
-		<aws 
+		<aws region="ap-northeast-1" correctForClockSkew="true"> 
 			<logging
 	    		logTo="UnityLogger"
 	    		logResponses="Always"
 	    		logMetrics="true"
 	    		logMetricsFormat="JSON" />
 			/>
-		/>
+		</aws>
 	
 You can also do this configuration in a script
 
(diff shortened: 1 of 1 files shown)


normj:
We now use .NET Standard 2.0 for Unity support. This README file is no longer applicable.

Issue #1476, opened 2019-12-05
DynamoDb .NET threadpool starvation


KereyWatters:
When describing tables, there are a few levels of locks which force single-threaded operation
https://github.com/aws/aws-sdk-net/blob/b691e46e57a3e24477e6a5fa2e849da44db7002f/sdk/src/Core/Amazon.Runtime/Internal/Util/SdkCache.cs#L421 https://github.com/aws/aws-sdk-net/blob/b691e46e57a3e24477e6a5fa2e849da44db7002f/sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextInternal.cs#L125 https://github.com/aws/aws-sdk-net/blob/b691e46e57a3e24477e6a5fa2e849da44db7002f/sdk/src/Services/DynamoDBv2/Custom/DataModel/InternalModel.cs#L608

And inside those locks we synchronously call an async method
https://github.com/aws/aws-sdk-net/blob/189abfb1d51c5300e68c741c536f4144d5f2c4fd/sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_mobile/HttpRequestMessageFactory.cs#L499

This is a problem as it creates a new task which may not be able to be run by the ThreadPool for quite some time. Also, because the locks in these three places do not pre-check their condition before locking - they make the problem worse by consuming more threads than necessary.

## Expected Behavior

We run 100 requests, and after the initial describe table call, the cache is loaded, and the program quickly completes the 100 requests

## Current Behavior

Cold start - no caches are filled.
We run 100 dynamodb LoadAsync requests, creating 100 tasks.
The threadpool begins processing the tasks, starting with threads equal to the number of cores we have to work with, say, 4.
The first thread gets through the locks and begins a synchronous call of an async http method, creating our 101th task.
Deadlock. All 4 threads are either stuck at a lock, or waiting for the 101st task to continue
The threadpool adds one more thread to the pool each second to help handle load.
97 (101-4) seconds go by, and the 101st thread picks up the http response task, resolving the deadlock
All the threads enter the lock, use the cache and complete shortly after.

## Possible Solution

To prevent build-up on the locks (and greatly improve performance) all three locks above should be using the double-checked locking pattern to remove unnecessary calls to Monitor.Enter.
https://en.wikipedia.org/wiki/Double-checked_locking#Usage_in_C#

[...]

PR #1678, opened 2020-08-14
Load async casues deadlocks/thread starvation

joro550:
## Description
I was seeing an issue this week with dynamodb and the load async method after quite a lot of investigation (thanks to some tools) I found that the library is doing some sync operations over async, which causes som thread starvation issues and eventually deadlocks, there is a good issue that has been raised here: https://github.com/aws/aws-sdk-net/issues/1476

I thought I'd start adding async methods to some of the base libraries, I have no resolved the issue as of yet, but thought the sooner the better with this change

## Motivation and Context
https://github.com/aws/aws-sdk-net/issues/1476

Fixing scaling issues with the dynamodb sdk

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Internal/Util/SdkCache.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Util/SdkCache.cs
@@ -23,6 +23,9 @@ using Amazon.Util;
 using Amazon.Runtime.Internal.Util;
 using ThirdParty.Json.LitJson;
 using System.Threading;
+#if AWS_ASYNC_API
+using System.Threading.Tasks;
+#endif
 
 namespace Amazon.Runtime.Internal.Util
 {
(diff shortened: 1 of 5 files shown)


acraven:
@joro550 We've been seeing some performance problems that can be attributed to the synchronous `DescribeTable` call. While this PR may not cover the part of the API we're using having this PR approved and merged would allow us to progress.

joro550:
Hey @acraven sadly I don't have permission to actually merge this PR, I am not a member of the AWS team - I am but an employee at a company who uses AWS in a high load environment and trust me we too were having pretty big performance issues because of this particular piece of functionality and it seems that there are a few other issues been raised in the repo for this that are getting absolutely 0 traction with the client team.

I have completely given up with them (especially now that they had a bot close my PR) - what we learned INSTEAD is that the describe table will not run if it all ready has the information cached, so when you first boot your application if you run:

```
IDynamoDBContext context; // just here to show you which interface this belongs to
context.GetTargetTable<TableName>();
```

it _should_ load the table into memory and not do the extra describe call under high load, looking at the code it seems to just be a dictionary that is assigned against the DynamoDBContext so I think if I were you I would make that class a singleton (so there is only one per application) and test that out to see if that helps the performance.

Just to give you a bit of an update - it did seem to improve the performance (or stability) for us but now we have issues with Assuming roles where it seems to do it under 5ms for 100+ calls but then shoots up to 200ms for 2 calls then back down 🤷‍♀️ I'll never fully understand why

PR #1701, opened 2020-09-10
SSO Credentials support

awschristou:
## Description

This change adds support for AWS SSO related Credentials Profiles to resolve credentials.

I implemented this in a similar manner to MFA credentials support.
In this change, SSO Support was not implemented for .NET Framework 3.5, or netstandard 1.3.

## Motivation and Context

This change opens the door for applications being built with the AWS SDK for .NET to work with SSO based credentials.

## Testing

Tests have been added with the code

I also did the following locally to verify the changes:
* added an AWS SSO based profile to my shared credentials file
* make a throwaway test locally that loads this profile, and resolves its credentials (both sync and async)
* verified that the SSO login flow is followed, and credentials are returned
* verified that cached SSO tokens are also honored

## Types of changes

## Checklist

## License


sdk/src/Services/SSOOIDC/Custom/Internal/_bcl45+netstandard/CoreAmazonSSOOIDC.cs
```diff
@@ -0,0 +1,286 @@
...
+                try
+                {
+                    var response = await client.CreateTokenAsync(createTokenRequest);
+
+                    // If we reach here, the user has completed the SSO Login authorization.
+                    return response;
+                }
+                catch (AuthorizationPendingException e)
+                {
+                    // Service is still waiting for user to complete authorization.
```
gsweetwood, line 251: Technically we are supposed to check if (current time + interval) is later than the previously calculated expiration time.

awschristou: Yup, that happens down on line 277

gsweetwood: I thought this action was limited to when the error returned was AuthorizationPendingException.  Here, it is being done on every iteration.

sdk/src/Services/SSOOIDC/Custom/Internal/_bcl45+netstandard/CoreAmazonSSOOIDC.cs
```diff
@@ -0,0 +1,286 @@
...
+                }
+                catch (TimeoutException e)
+                {
+                    // Spec: If the call times out then the tool should double its polling interval and then retry.
+                    intervalSec *= 2;
+                }
+                catch (Exception e)
+                {
+                    logger.Error(e, "Unexpected exception while polling for SSO Token.");
+                    Console.WriteLine($"Poll Exception: {e.Message}");
```
normj, line 273: Don't write to the console

sdk/src/Services/SSOOIDC/Custom/Internal/_bcl45+netstandard/CoreAmazonSSOOIDC.cs
```diff
@@ -0,0 +1,286 @@
...
+                    throw;
+                }
+
+                if (DateTime.UtcNow.AddSeconds(intervalSec) > deviceCodeExpiration)
```
normj, line 217: Should this be using the clock skewed corrected time?

normj (changes requested):

sdk/src/Core/Amazon.Runtime/Credentials/Internal/_bcl45+netstandard/SsoTokenUtils.cs
```diff
@@ -0,0 +1,108 @@
...
+        {
+            if (token == null)
+            {
+                return true;
+            }
+
+            // Spec: When checking to see if a cached token is expired, the SDK MAY treat the token as expired if it will
+            // expire within 15 minutes of the expiresAt time. The SDK MAY choose a longer or shorter expiration window.
+#pragma warning disable CR1003 // Do not use DateTime.Now or DateTime.UtcNow, use AWSSDKUtils.CorrectedNow or AWSSDKUtils.CorrectedUtcNow
+            return token.ExpiresAt < DateTime.UtcNow.AddMinutes(TokenExpirationBufferMinutes);
```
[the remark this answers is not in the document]
awschristou: I don't believe so, the tests in "SsoTokenTest" should cover this.

eg: If it is 1pm now, the expiration boundary would be 1:15pm. A token that expires at 1:10pm would return true (expired) and a token that expires at 1:20pm would return false (not expired).

Would this read any better if it was phrased as "Now > (expiresAt - 15)" ? Open to suggestions.

sdk/src/Core/Amazon.Runtime/Credentials/Internal/_bcl45+netstandard/SsoTokenUtils.cs
```diff
@@ -0,0 +1,108 @@
...
+            return ToJson(token);
+        }
+
+        #endregion
+
+        /// <summary>
+        /// Serializes the SSO Token to JSON
+        /// </summary>
+        /// <param name="token">Token to serialize</param>
+        public static string ToJson(SsoToken token)
```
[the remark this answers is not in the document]
awschristou: My intention was to maintain a separation of json concerns from the token class. The extension method is a convenient way for calling code to access it in a more convenient manner.

sdk/src/Core/Amazon.Runtime/Credentials/Internal/_bcl45+netstandard/SsoTokenCache.cs
```diff
@@ -0,0 +1,164 @@
...
+
+        /// <summary>
+        /// Caches an SSO Token to disk.
+        /// </summary>
+        /// <param name="token">SSO Token to cache</param>
+        public bool TrySave(SsoToken token)
+        {
+            try
+            {
+                if (token.StartUrl != StartUrl)
```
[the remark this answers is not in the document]
awschristou: Since the spec requires the file to be loaded when the access token is needed, multiple credentials objects making reference to the same starturl would access the same file. When the cache reports a non-expired token, things would proceed smoothly for multiple credentials objects. You could get a scenario where multiple credentials objects are each told the token is expired, and each would attempt to perform the SSO flow, then write back a token. The only concern here would be if those writes occurred at the same time. Thoughts?

sdk/src/Services/SSOOIDC/Custom/Internal/_bcl45+netstandard/CoreAmazonSSOOIDC.cs
```diff
@@ -0,0 +1,286 @@
...
+
+#if BCL
+        public static GetSsoTokenResponse GetSsoToken(IAmazonSSOOIDC client, GetSsoTokenRequest request)
+        {
+            return GetSsoToken(client, request, new GetSsoTokenContext());
+        }
+
+        public static GetSsoTokenResponse GetSsoToken(IAmazonSSOOIDC client, GetSsoTokenRequest request, IGetSsoTokenContext context)
+        {
+            var registerClientRequest = new RegisterClientRequest()
```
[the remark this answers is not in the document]
awschristou: Caching the client id/secret to disk is a recommendation, but not a requirement, in the spec. I figured that could be implemented at a later time. The calls that follow RegisterClient must be performed each time, as the resulting token may not be retained in memory.

sdk/src/Services/SSOOIDC/Custom/Internal/_bcl45+netstandard/CoreAmazonSSOOIDC.cs
```diff
@@ -0,0 +1,286 @@
...
+                }
+                catch (TimeoutException e)
+                {
+                    // Spec: If the call times out then the tool should double its polling interval and then retry.
+                    intervalSec *= 2;
+                }
+                catch (Exception e)
+                {
+                    logger.Error(e, "Unexpected exception while polling for SSO Token.");
+                    Console.WriteLine($"Poll Exception: {e.Message}");
```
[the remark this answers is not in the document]
awschristou: Thanks, I forgot to remove those

sdk/src/Core/Amazon.Runtime/Credentials/Internal/_bcl45+netstandard/SsoTokenUtils.cs
```diff
@@ -0,0 +1,108 @@
...
+        public static bool IsExpired(this SsoToken token)
+        {
+            if (token == null)
+            {
+                return true;
+            }
+
+            // Spec: When checking to see if a cached token is expired, the SDK MAY treat the token as expired if it will
+            // expire within 15 minutes of the expiresAt time. The SDK MAY choose a longer or shorter expiration window.
+#pragma warning disable CR1003 // Do not use DateTime.Now or DateTime.UtcNow, use AWSSDKUtils.CorrectedNow or AWSSDKUtils.CorrectedUtcNow
```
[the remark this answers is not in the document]
awschristou: I found working with the current time confusing in the SDK. `CorrectedUtcNow` is marked as obsolete, and its successor `CorrectClockSkew.GetCorrectedUtcNowForEndpoint` involves endpoints, which I wasn't sure was relevant to this PR's changes.

sdk/src/Core/Amazon.Runtime/Credentials/_bcl45+netstandard/SSOAWSCredentials.cs
```diff
@@ -0,0 +1,344 @@
...
+            StartUrl = startUrl;
+            Options = options;
+        }
+
+#if BCL
+        protected override CredentialsRefreshState GenerateNewCredentials()
+        {
+            ValidateCredentialsInputs();
+
+            var oidcClient = CreateClient<ICoreAmazonSSOOIDC>(
```
normj, line 110: Should the service clients be member variables so they don't have to be recreated all of the time.

sdk/src/Core/Amazon.Runtime/Credentials/_bcl45+netstandard/SSOAWSCredentials.cs
```diff
@@ -0,0 +1,344 @@
...
+            _logger.InfoFormat("New SSO credentials created that expire at {0}", credentials.Expiration.ToString("yyyy-MM-ddTHH:mm:ss.fffffffK", CultureInfo.InvariantCulture));
+            return new CredentialsRefreshState(credentials, credentials.Expiration);
+        }
+#endif
+
+        protected override async Task<CredentialsRefreshState> GenerateNewCredentialsAsync()
+        {
+            ValidateCredentialsInputs();
+
+            var oidcClient = CreateClient<ICoreAmazonSSOOIDC>(
```
normj, line 136: Same comment as before about whether these should be member variables

sdk/src/Core/Amazon.Runtime/Credentials/_bcl45+netstandard/SSOAWSCredentials.cs
```diff
@@ -0,0 +1,344 @@
...
+            {
+                throw new NotSupportedException("Unable to get credentials from SSO");
+            }
+
+            _logger.InfoFormat("New SSO credentials created that expire at {0}", credentials.Expiration.ToString("yyyy-MM-ddTHH:mm:ss.fffffffK", CultureInfo.InvariantCulture));
+            return new CredentialsRefreshState(credentials, credentials.Expiration);
+        }
+#endif
+
+        protected override async Task<CredentialsRefreshState> GenerateNewCredentialsAsync()
```
normj, line 132: Do we need to do anything to prevent multiple SSO flows from happening at the same time for the same SSO options? We a version of this in the DefaultInstanceProfileAWSCredentials which uses a singleton and a timer to refresh credentials. Not saying we should do the same design but I don't want the SDK to suddenly cause a dozen pop up boxes to relogin.

normj (changes requested):

sdk/src/Core/Amazon.Runtime/Credentials/Internal/_bcl45+netstandard/SsoTokenCache.cs
```diff
@@ -0,0 +1,164 @@
...
+                Directory.CreateDirectory(Path.GetDirectoryName(_cachePath));
+                File.WriteAllText(_cachePath, json);
+
+                return true;
+            }
+#pragma warning disable CA1031 // Do not catch general exception types
+            catch (Exception e)
+#pragma warning restore CA1031 // Do not catch general exception types
+            {
+                _logger.Error(e, "Error saving SSO Token Cache");
```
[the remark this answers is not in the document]
awschristou: I've added the consequences but left the error so that the exception is logged

sdk/src/Core/Amazon.Runtime/Credentials/Internal/_bcl45+netstandard/SsoTokenCache.cs
```diff
@@ -0,0 +1,164 @@
...
+        private readonly string _cachePath;
+
+        /// <summary>
+        /// Class constructor
+        /// Caches SSO Tokens to disk under (user profile)/.aws/sso/cache/
+        /// </summary>
+        /// <param name="startUrl">The configured sso_start_url for the profile that credentials are being resolved for.</param>
+        public SsoTokenCache(string startUrl)
+            : this(
+                Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), ".aws", "sso", "cache"),
```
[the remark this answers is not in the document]
awschristou: The spec prohibits keeping the token in cache, so I will set this class to do no caching when it is unable to resolve a UserProfile

sdk/src/Core/Amazon.Runtime/Credentials/_bcl45+netstandard/SSOAWSCredentials.cs
```diff
@@ -0,0 +1,339 @@
...
+        public string RoleName { get; private set; }
+
+        /// <summary>
+        /// The main URL for users to login to their SSO directory.
+        /// Provided by the SSO service via the web console.
+        /// </summary>
+        public string StartUrl { get; private set; }
+
+        /// <summary>
+        /// Options to be used in the SSO flow to resolve credentials.
```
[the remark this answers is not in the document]
awschristou: Please check the latest version of the file. It should look like this:
```
        /// <summary>
        /// Options to be used in the SSO flow to resolve credentials.
        /// Developers wishing to support AWS SSO would want to provide the following:
        /// - <seealso cref="SSOAWSCredentialsOptions.ClientName"/>
        /// - <seealso cref="SSOAWSCredentialsOptions.SsoVerificationCallback"/>
        /// </summary>
```

sdk/src/Core/Amazon.Runtime/Credentials/_bcl45+netstandard/SSOAWSCredentials.cs
awschristou: I don't have enough insight on the SDK to make an appropriate call here. I did take inspiration from classes like `AssumeRoleAWSCredentials`, which creates an STS client within `GenerateNewCredentials`

sdk/src/Core/Amazon.Runtime/Credentials/_bcl45+netstandard/SSOAWSCredentials.cs
awschristou: I'm also not familiar with SDK implementation practices in this space, however that sounds like a responsibility of the registered `SsoVerificationCallback` to perform any debouncing if necessary. I would imagine that the calls made to the sso/oidc services would also return throttling based errors in extreme cases.

sdk/src/Core/Amazon.Runtime/Credentials/_bcl45+netstandard/SSOAWSCredentials.cs
normj: Okay, lets stay consistent for now.

sdk/src/Core/Amazon.Runtime/Credentials/_bcl45+netstandard/SSOAWSCredentials.cs
normj: Lets check with other SDK implementations.

normj:
I just have the one last question about multiple requests triggering the SSO flow as the same time. Lets ask the Python team what they are doing so we are consistent with them.

sdk/src/Core/Amazon.Runtime/CredentialManagement/CredentialProfileOptions.cs
```diff
@@ -111,6 +112,27 @@ public class CredentialProfileOptions
         /// </summary>
         public string WebIdentityTokenFile { get; set; }
 
+        /// <summary>
+        /// The AWS account ID that temporary AWS credentials will be resolved for using AWS SSO.
+        /// </summary>
+        public string SsoAccountId { get; set; }
```
ashovlin, line 118: Do these need to be inside a target platform check? Similiar question down on lines 206/215.

ashovlin (commented):
Overall question - why the .NET Framework 3.5 restriction? Fine with .NET Standard 1.3 given https://aws.amazon.com/blogs/developer/net-standard-1-3-in-aws-sdk-for-net-is-now-in-maintenance-mode/, but we don't have any current plans to deprecate net35. Discussed briefly with Norm and we're ultimately okay with it, but just curious.

sdk/src/Core/Amazon.Runtime/CredentialManagement/CredentialProfileOptions.cs
awschristou: I opted for simpler complexity in this case, with the line of thought that these would be dormant fields otherwise -- the SDK would not identify the credential as an SSO profile type in those platforms.

awschristou:
>   why the .NET Framework 3.5 restriction

Some necessary SDK or Crypto calls were not available (I can't recall what it was, but I'll remember to document why next time)

sdk/src/Core/Amazon.Runtime/CredentialManagement/CredentialProfileOptions.cs
ashovlin: Got it, fine for me.

sdk/src/Core/Amazon.Runtime/Credentials/_bcl45+netstandard/SSOAWSCredentials.cs
awschristou: This appears to be handled/guarded against in the [parent class (RefreshingAWSCredentials::GetCredentials)](https://github.com/aws/aws-sdk-net/blob/master/sdk/src/Core/Amazon.Runtime/Credentials/RefreshingAWSCredentials.cs#L121), which wraps the call in a semaphore (prevents concurrent queries), and checks whether the first set of credentials has expired.

sdk/src/Core/Amazon.Runtime/CredentialManagement/CredentialProfileOptions.cs
awschristou: The `EnsureCredentialProfileDetectorSetup` test in `sdk\test\Services\SecurityToken\UnitTests\Custom\StaticCheckTests.cs` says I'm wrong about this. I am adding more platform checks.

sdk/src/Core/Amazon.Runtime/Credentials/_bcl45+netstandard/SSOAWSCredentials.cs
normj: This will work for a single instance of SSOAWSCredentials. If multiple SSOAWSCredentials are created for the same SSO configuration/profile you could get multiple logins. 

I'm fine with documenting for now best practices are to create a single instance. We might need to revisit this as we get customer feedback.

ashovlin (approved):

normj (approved):

normj:
This has been squashed merged and released in version 3.5.2 of AWSSDK.Core

Issue #1700, opened 2020-09-10
General performance problems with the client


Rodrigo-Andrade:
The SDK is  now the bottleneck in several of our systems, the dotnet ecosystem kept getting faster, fleshing out this issue more and more.

We are writing some custom implementations of the clients because the performance of the SDK is not acceptable for some of our workloads, greatly increasing the cost to use AWS services.

I propose that performance tests be included in your development pipeline. 

Here is a sample trace from the S3 client as an exemple of the problems that permeate the SDK:

![image](https://user-images.githubusercontent.com/1174384/92727615-c7498a80-f345-11ea-8614-1f5260c66ed5.png)

You can see that 20% of the CPU time spent sending this get request is at:

```csharp
public static bool IsAmazonS3Endpoint(Uri uri)
{
      Match match = !(uri == (Uri) null) ? new Regex("^(.+\\.)?s3[.-]([a-z0-9-]+)\\.").Match(uri.Host) : throw new ArgumentNullException(nameof (uri));
      return (uri.Host.EndsWith("amazonaws.com", StringComparison.OrdinalIgnoreCase) || uri.Host.EndsWith("amazonaws.com.cn", StringComparison.OrdinalIgnoreCase)) && match.Success;
}
```
Every request runs this code.

ppittle:
Thinking about this, few things to try:

- Create `Regex` as a constant field in `AmazonS3Uri`
- Use `RegexOptions.Compilex` or `Regex.CompileToAssembly` ([Best Practicies](https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices))
- Decorate `IsAmazonS3Endpoint` with [`MethodImplOptions.AggressiveInlining`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.methodimploptions?view=netcore-3.1)

indy-singh:
We've run into similar issues, and found that using pre-signed urls have alleviated a lot of the issues in case that helps @Rodrigo-Andrade 

Regards,
Indy

ppittle:
Performance improvements are available in `[AWSSDK.S3 3.5.19](https://www.nuget.org/packages/AWSSDK.S3/3.5.1.9)`

PR #1705, opened 2020-09-12
Performance improvement in S3 endpoint validation logic

joshongithub:
## Description
The logic for an AmazonS3Uri.IsAmazonS3Endpoint method overload was updated to improve performance.  

## Motivation and Context
The current logic does not take advantage of static and compile options available with Regex usage. Code was updated by creating a static property for the Regex match logic, using the compile option. Additionally, the order of options was changed so that that the Regex logic would not be called unless a valid URI was provided. This improves performance by short circuiting the evaluation.

The performance impact was was reported in Issue #1700. 

## Testing
I created a local test client that tested calls ranging from 5 to 100,000 within an execution, and observed over a 90% performance improvement when using a mix of 70% Amazon vs 30% non-Amazon URI calls. 

I executed the related Unit Tests provided in the solution(AWSSDK.IntegrationTests.S3.Net35, AWSSDK.IntegrationTests.S3.Net45,
AWSSDK.UnitTestUtilities.Net35, AWSSDK.UnitTestUtilities.Net45, AWSSDK.UnitTests.S3.Net35, AWSSDK.UnitTests.S3.Net45). All passed, except for:
- S3CapacityManagerIntegrationTest - Exception "Amazon.Runtime.AmazonUnmarshallingException: Error unmarshalling response back from AWS. Response Body: System.InvalidOperationException: Sequence contains no elements" (this exception was also thrown with the original code)
- EncryptionInteropTest : 9 tests skipped
- Replicationtests : 2 tests skipped
- RequesterPaysTests : 8 tests skipped

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/Util/AmazonS3Uri.cs
+++ b/sdk/src/Services/S3/Custom/Util/AmazonS3Uri.cs
@@ -34,6 +34,10 @@ namespace Amazon.S3.Util
     {
         private const string EndpointPattern = @"^(.+\.)?s3[.-]([a-z0-9-]+)\.";
 
+        private static Regex endpointRegexMatch;
+
+        private static Regex EndpointRegexMatch => endpointRegexMatch ?? (endpointRegexMatch = new Regex(EndpointPattern, RegexOptions.Compiled));
+
         /// <summary>
         /// True if the URI contains the bucket in the path, false if it contains the bucket in the authority.
         /// </summary>
@@ -221,16 +225,14 @@ namespace Amazon.S3.Util
             if (uri == null)
                 throw new ArgumentNullException("uri");
 
-            var match = new Regex(EndpointPattern).Match(uri.Host);
             if (uri.Host.EndsWith("amazonaws.com", StringComparison.OrdinalIgnoreCase) || uri.Host.EndsWith("amazonaws.com.cn", StringComparison.OrdinalIgnoreCase))
             {
-                return match.Success;
+                return EndpointRegexMatch.Match(uri.Host).Success;
             }
             else
             {
                 return false;
             }
-
         }
 
         /// <summary>
--- a/sdk/src/Services/S3/Custom/Util/AmazonS3Uri.cs
+++ b/sdk/src/Services/S3/Custom/Util/AmazonS3Uri.cs
@@ -32,7 +32,7 @@ namespace Amazon.S3.Util
     /// </summary>
     public class AmazonS3Uri
     {
-        private static Regex EndpointRegexMatch = new Regex(@"^(.+\.)?s3[.-]([a-z0-9-]+)\.", RegexOptions.Compiled);
+        private static readonly Regex EndpointRegexMatch = new Regex(@"^(.+\.)?s3[.-]([a-z0-9-]+)\.", RegexOptions.Compiled);
 
         /// <summary>
         /// True if the URI contains the bucket in the path, false if it contains the bucket in the authority.
--- a/sdk/src/Services/S3/Custom/Util/AmazonS3Uri.cs
+++ b/sdk/src/Services/S3/Custom/Util/AmazonS3Uri.cs
@@ -195,7 +195,9 @@ namespace Amazon.S3.Util
                 }
             }
             // intentionally suppress all exceptions, because this is a try method
+#pragma warning disable CA1031 // Do not catch general exception types
             catch { }
+#pragma warning restore CA1031 // Do not catch general exception types
 
             amazonS3Uri = null;
             return false;
--- a/sdk/src/Services/S3/Custom/Util/AmazonS3Uri.cs
+++ b/sdk/src/Services/S3/Custom/Util/AmazonS3Uri.cs
@@ -32,11 +32,9 @@ namespace Amazon.S3.Util
     /// </summary>
     public class AmazonS3Uri
     {
-        private const string EndpointPattern = @"^(.+\.)?s3[.-]([a-z0-9-]+)\.";
-
         private static Regex endpointRegexMatch;
 
-        private static Regex EndpointRegexMatch => endpointRegexMatch ?? (endpointRegexMatch = new Regex(EndpointPattern, RegexOptions.Compiled));
+        private static Regex EndpointRegexMatch => endpointRegexMatch ?? (endpointRegexMatch = new Regex(@"^(.+\.)?s3[.-]([a-z0-9-]+)\.", RegexOptions.Compiled));
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


sdk/src/Services/S3/Custom/Util/AmazonS3Uri.cs
```diff
@@ -34,6 +34,10 @@ public class AmazonS3Uri
     {
         private const string EndpointPattern = @"^(.+\.)?s3[.-]([a-z0-9-]+)\.";
 
+        private static Regex endpointRegexMatch;
+
+        private static Regex EndpointRegexMatch => endpointRegexMatch ?? (endpointRegexMatch = new Regex(EndpointPattern, RegexOptions.Compiled));
```
ppittle, line 39: Why are you lazily initializing the `EndpointRegexMatch` property vs eagerly initializing the field?

ppittle, line 39: Also update L80 to use `EndpongRegexMatch`

ppittle (changes requested):

ppittle:
I think you should also update rework [`AmazonS3KmsHandler.EvaluateIfSigV4Required`](https://github.com/aws/aws-sdk-net/blob/master/sdk/src/Services/S3/Custom/Internal/AmazonS3KmsHandler.cs#L80) to re-use `AmazonS3Uri.TryParseAmazonS3Uri` rather than duplicating that logic.  

Then we need to find a strategy so that `AmazonS3Uri.TryParseAmazonS3Uri` doesn't duplicate checking `EndpointRegexMatch`.

```csharp
public class AmazonS3KmsHandler
{
    // try and reduce duplicate input checking
    internal static void EvaluateIfSigV4Required(IRequest request)
    {
            // Skip this for S3-compatible storage provider endpoints
            if (request.OriginalRequest is S3.Model.GetObjectRequest && 
                AmazonS3Uri.TryParseAmazonS3Uri(request.Endpoint, out var amazonS3Uri))  &&
                amazonS3Uri.Region != RegionEndpoint.USEast1)     
           {  
                 request.UseSigV4 = true;
           }
     }
}
```

sdk/src/Services/S3/Custom/Util/AmazonS3Uri.cs
ppittle, line 39: Can you remove `const string EndpointPattern`  and just put the string literal inside the Regex's constructor.  This will help cut down field clutter and be a quick sanity check to make sure that regex pattern isn't referenced anywhere else.

joshongithub: For consumers of the library that aren't using the functionality, using lazy loading avoids incurring the performance cost of the compiled Regex. Here is blog post covering the topic: http://blog.regexhero.net/2012/03/quick-tip-about-using.html

joshongithub: Good catch. I'll update.

joshongithub: Will update EvaluateIfSigV4Required with your updated logic and remove the const.

joshongithub:
Committed updates mentioned above.

sdk/src/Services/S3/Custom/Util/AmazonS3Uri.cs
ppittle: Per my other comment about reusing `EndpointRegexMatch` in the Constructor, is there still a perf benefit?  Ie, is there a case where a consumer could use a static method that doesn't end up calling the instance constructor?

joshongithub: EndpointRegexMatch is an expression-bodied declared property, and the lazy loading execution is a benefit because the RegexOptions.Compile option complies at runtime. This means that a consumer of the class that doesn't call the public method IsAmazonS3Endpoint, which calls the EndpointRegexMatch when an Amazon URL, wouldn't take the performance hit.

boblodgett (approved):

sdk/src/Services/S3/Custom/Util/AmazonS3Uri.cs
```diff
@@ -188,16 +188,19 @@ public static bool TryParseAmazonS3Uri(string uri, out AmazonS3Uri amazonS3Uri)
...
-                amazonS3Uri = null;
-                return false;
+                if (IsAmazonS3Endpoint(uri))
+                {
+                    amazonS3Uri = new AmazonS3Uri(uri);
+                    return true;
+                }
             }
+            // intentionally suppress all exceptions, because this is a try method
+            catch { }
```
joshongithub, line 198: @ppittle - This a causes a Warning CA1031, which in VS2019 defaults to treating as an error:
https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1031?view=vs-2019

ppittle: suppressed

ppittle (approved):

SrutiG (approved):

ppittle:
- [x] Completed a Dry Run

ppittle:
Released in [`AWSSDK.S3 3.5.19`](https://www.nuget.org/packages/AWSSDK.S3/)

sdk/src/Services/S3/Custom/Util/AmazonS3Uri.cs
```diff
@@ -221,16 +226,14 @@ public static bool IsAmazonS3Endpoint(Uri uri)
             if (uri == null)
                 throw new ArgumentNullException("uri");
 
-            var match = new Regex(EndpointPattern).Match(uri.Host);
             if (uri.Host.EndsWith("amazonaws.com", StringComparison.OrdinalIgnoreCase) || uri.Host.EndsWith("amazonaws.com.cn", StringComparison.OrdinalIgnoreCase))
             {
-                return match.Success;
+                return EndpointRegexMatch.Match(uri.Host).Success;
```
jnyrup, line 231: This could also be written as
```
EndpointRegexMatch.IsMatch(uri.Host);
```

Without knowing the performance differences, it seems [`IsMatch()`](https://source.dot.net/#System.Text.RegularExpressions/System/Text/RegularExpressions/Regex.Match.cs,5a9d5ededeb92966,references) takes some `quick` path where `.Match()` does not.

ppittle: Thanks for the tip.  This PR has already merged and shipped, but I'll save your suggestion next time I'm working with Regular Expressions.

PR #1780, opened 2021-01-26
📦 Include symbols (snupkg's) when creating nuget packages.

### LINKED ISSUE #1777: Will any of the AWSSDK NuGet packages have Source Link enabled?
👋🏻 G'Day AWS team,

I was just doing some debugging with some AWS code (trying to load in some config values using the `AWSSDK.Extensions.NETCore.Setup` and noticed that there's no Source Link available for this package (and so I'm poorly assuming, other AWSSDK nugets also ...).

e.g.
image

So - will you allow Source Link to be enabled for this (and other) AWS packages? If yes, will you accept a PR from the community for this functionality?

If you're unsure [what Source Link is](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/sourcelink), it enables developers to _easily_ step into methods/code from NuGet packages, just like they are part of our own project/code-base.

PureKrome:
## Description
As discussed in #1777 this PR is one of potentially many to come (not necessarily from me, though).

To help developers debug their applications, if a NuGet package (which is being consumed by the developers project) has [Source link enabled](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/sourcelink), then a developer could optionally "step into" a method that is provided by the 3rd party NuGet package.

So in the case of this AWS .NET Sdk, a developer could step into some of the public methods provided in the AWS SDK NuGet's!

To enable this, I had up update the `.csproj` file for the relevant project which will be `dotnet pack` into a NuGet package.

As of the time of me creating the code changes today, there was 260+ projects in the .NET Standard solution! Egads! That's a lot. As such:

- I've only started with the S3 project (in this PR).
- the S3 project takes a dependency on the AWS SDK "Core" project, so I've also enabled that.
- I'm not planning on enabling Source link for all of the other projects because that's too time consuming (for me) and I don't use 99% of them.

- ⚠️ This PR is only tackling .NET Standard. I'm not bothering with .NET35 or .NET45 as these are (more or less) obsolete so I'm spending _my_ energy there. Of course, others can!
- 🚀 If this PR is successful, I'll also do 2x more for SNS and SQS (more services I use).

## Motivation and Context
- Helps resolve #1777 
- Helps developers debug their own code when they consume AWS SDK services and might need to step into the AWS code.

## Testing
- Testing requires `dotnet pack`'ing to occur. 
- In the `sdk\services\s3` folder, type: `dotnet pack -c RELEASE -o C:\temp\Nuget /p:ContinuousIntegrationBuild=true /p:version=100.2.3.4   .\AWSSDK.S3.NetStandard.csproj`

- `/p:ContinuousIntegrationBuild=true` == deterministic builds, I believe.
- `/p:version=100.2.3.4` is just me making a version of the NuGet so it is to find in a NuGet package manager or on the hard disk.

Before (the current vesrion on NuGet.org):

image

And after running the `dotnet pack` command:

image

NOTE: similar info should also apply to the `AWSSDK.Core` NuGet package.

## Types of changes

## Checklist

## License

---

### ⚠️ ⚠️ ⚠️  AWS SDK + CI / CD

I couldn't find any documentation about how this solution/repo handles CI/CD.

- What system does it use? GH Actions? Travis? AppVeyor?
- How does CI/CD kick off? Does a PR kick this off?

As such, I couldn't find where to update the code where/how `dotnet pack` is ran. THIS STEP IS IMPORTANT and requires a possible modification (as listed above in this readme). Specifically the `/p:ContinuousIntegrationBuild=true`

### DIFF
--- a/sdk/src/Core/AWSSDK.Core.NetStandard.csproj
+++ b/sdk/src/Core/AWSSDK.Core.NetStandard.csproj
@@ -23,7 +23,15 @@
     <NoWarn>$(NoWarn);CS1591</NoWarn>        
 
     <NetStandardImplicitPackageVersion Condition="'$(TargetFramework)' == 'netstandard1.3'">1.6.0</NetStandardImplicitPackageVersion>
+      
+    <!-- Sourcelink -->
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
+    <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
+    <IncludeSymbols>true</IncludeSymbols>
+    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
   </PropertyGroup>
+    
     <!-- Async Enumerable Compatibility -->
   <PropertyGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <LangVersion>8.0</LangVersion>
@@ -81,6 +89,10 @@
     <PackageReference Include="System.Security.Cryptography.Csp" Version="4.0.0" />
     <PackageReference Include="System.Threading.Thread" Version="4.0.0" />
     <PackageReference Include="Microsoft.CSharp" Version="4.0.1" />
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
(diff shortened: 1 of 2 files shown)


slang25:
At the moment snupkgs won't be necessary as the pdbs are already shipped in the nupkg. The other changes (deterministic & SourceLink) would be great! I'd be happy to help create PRs once an approach is agreed 🙂

PureKrome:
> as the pdbs are already shipped in the nupkg

@slang25 so why would MS offer the option of creating snupkgs? what is the advantage of having snupkgs vs pdbs, as you've said above? Becuase you're doing something now, that might not be best practice?

slang25:
I agree regarding best practice, this might be a good time to re-evaluate.

What I mean is it's redundant right now to create snupkgs, as they exist as a mechanism to retrieve the pdbs, which are still today being shipping directly in the main package. So they should either be removed from the main package, or we remove snupkgs from this change.

Personally, I favor pdbs shipped in the main package as they are today (it's a better acquisition experience IMO, you already have them), however snupkgs are the .NET teams recommendation 🤷  (I'm writing a long blog post to clarify). This is really a choice for the aws sdk team here to decide.

PureKrome:
@slang25 Heya Stewart 👋🏻 

>  (I'm writing a long blog post to clarify)

Did you ever get around to writing this blog post?

dferretti:
:wave: still interested in this addition!

normj:
Closing this PR. This is something we would like to do sometime but the work is a lot more involved then just this PR which manually updates a copy of the hundreds of projects we have. There is also lot of work that has to be done to our internal build system. Like I said this is something we would like to do but this is something that we have to track as an epic across all of the affected pieces.

PR #1838, opened 2021-04-26
TryParse should handle all exceptions that could arise from parsing

### LINKED ISSUE #1837: Arn.TryParse throws an exception rather than returning false
## Description
Arn.TryParse with and invalid accountId will throw an exception rather than returning false

## Reproduction Steps
Arn.TryParse("arn:aws:iam::123456789012XXX:user/testuser", out Arn arn)

slang25:
Fixes #1837

The current code doesn't catch `AmazonClientException` which can occur if the account id is not well formed: https://github.com/aws/aws-sdk-net/blob/92cdca1f9e331826a46502b1706999af0b8e6880/sdk/src/Core/Arn.cs#L61

### DIFF
--- a/sdk/src/Core/Arn.cs
+++ b/sdk/src/Core/Arn.cs
@@ -98,7 +98,9 @@ namespace Amazon
                     return true;
                 }
             }
+#pragma warning disable CA1031 // Do not catch general exception types
             catch (Exception) { }
+#pragma warning restore CA1031 // Do not catch general exception types
             arn = null;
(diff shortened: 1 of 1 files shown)


normj (approved):

boblodgett (approved):

ppittle (approved):

normj:
PR is released as part of version 3.7.0.17 of AWSSDK.Core. Thanks for the PR.

Issue #1821, opened 2021-03-30
AWSCredentialsFactory.TryGetAWSCredentials() returns invalid SSOAWSCredentials


RamonEbdonSS:
## Description
My company has an SSO setup in AWS (auth via Azure) to allow users to obtain temporary role credentials. Having just updated one of our .NET AWS applications to AWSSDK v3.7.x (via NuGET) from v3.3.x, I was no longer able to run the app locally to access AWS using my credentials (which can be successfully obtained via AWS CLI v2).

The app was previously (using AWSSDK v3.3.x) implicitly detecting my credentials in both ~/.aws/credentials and local Environment vars when creating an `AmazonS3Client()`, but this is now failing. According to the [sequence of credentials detection](https://docs.aws.amazon.com/sdk-for-net/latest/developer-guide/creds-assign.html), it should be fine, and it does appear as if the SDK is finding my shared credentials file, since it's returning the correct type, but the app is throwing an exception when trying to use the returned `AWSCredentials` object.

Following the basic guide to [access credentials and profiles](https://docs.aws.amazon.com/sdk-for-net/latest/developer-guide/creds-locate.html), creating a `new SharedCredentialsFile()` successfully locates and loads my credentials file, and the return `out AWSCredentials` is returning an `SSOAWSCredentials` object, but the object is invalid because two mandatory properties in the `(SSOAWSCredentialsOptions)Options` property are null: `ClientName` and `SsoVerificationCallback`.

When attempting to use the generated credentials in an `AmazonS3Client()`, exceptions are returned saying either:
```
    Value cannot be null. Parameter name: Options property cannot be empty: ClientName
```
Or:
```
    Value cannot be null. Parameter name: Options property cannot be empty: SsoVerificationCallback
```

But, if I manually set those two properties, it all works:
```
    ((SSOAWSCredentials)awsCredentials).Options.ClientName = "SSO";
    ((SSOAWSCredentials)awsCredentials).Options.SsoVerificationCallback = new Action<SsoVerificationArguments>(delegate (SsoVerificationArguments ssoArgs) {
        System.Diagnostics.Process.Start(new System.Diagnostics.ProcessStartInfo() {
            FileName = ssoArgs.VerificationUriComplete,
            UseShellExecute = true
        });
    });
```
`ClientName` appears to accept (and work with) any random string. It doesnt appear to be relevant to any credentials Profile names.
[...]

PR #1850, opened 2021-05-07
Delay SSO options validation until they are used

KenHundley:
## Description
Currently if you are using a SSO profile and you do not configure a `ClientName` or `SsoVerificationCallback` in your application it will always fail to execute, even if these values aren't currently necessary because of a cached SSO token.

This change allows the use of SSO Profiles without needing to add SSO specific code into the project. You can use applications like the AWS CLI v2 to authenticate and cache an SSO auth token that can then be used by any application pointing to that profile, until that token expires. This does not remove any current validations, but simply delays the validation until the values are actually needed. 

## Motivation and Context

This is related to the issue: #1821 

## Testing

I executed the Unit Tests in the solution and they all passed

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Credentials/_bcl45+netstandard/SSOAWSCredentials.cs
+++ b/sdk/src/Core/Amazon.Runtime/Credentials/_bcl45+netstandard/SSOAWSCredentials.cs
@@ -156,16 +156,6 @@ namespace Amazon.Runtime
 
         private void ValidateCredentialsInputs()
         {
-            if (string.IsNullOrEmpty(Options.ClientName))
-            {
-                throw new ArgumentNullException($"Options property cannot be empty: {nameof(Options.ClientName)}");
-            }
-
-            if (Options.SsoVerificationCallback == null)
-            {
-                throw new ArgumentNullException($"Options property cannot be empty: {nameof(Options.SsoVerificationCallback)}");
-            }
-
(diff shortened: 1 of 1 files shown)


KenHundley:
Quick update. I did some additional testing today and this also removed the need to install the SSOOIDC package when using an SSO profile. Since you can do the SSO signin outside of the app, like with the `aws sso login` cli command, the code does not need the SSOOIDC project when using a cached token.

normj (approved):

boyersnet:
Thank you for this @KenHundley!  We are eagerly awaiting the merge and distribution of this update to allow us to proceed with our SSO adoption.

ashovlin:
This has been merged manually via 537a9815b7d5de8d6e32dc6244e0f772e1bad6a3 and released today in [version 3.7.36.1](https://github.com/aws/aws-sdk-net/releases/tag/3.7.36.1) of the SDK ([version 3.7.0.23 of AWSSDK.Core](https://www.nuget.org/packages/AWSSDK.Core/3.7.0.23)).

Manually marking this as closed (I don't think we preserved the commit hash when testing and building this internally, so GitHub won't autoclose).

boyersnet:
Thank you all!  I was able to get my .net core test project to work with AWS CLI v2 sso login.  I did however have to keep the AWSSDK.SSOOIDC package reference.

PR #1871, opened 2021-06-14
Use correct table name in DynamoDB exceptions

krlxgrn:
## Description
When throwing an exception because the number of hash or range keys is incorrect, the code currently uses  `storageConfig.TableName` to add the table name to the exception.

If the table name is overridden using `DynamoDBFlatConfig.OverrideTableName` or `DynamoDBFlatConfig.TableNamePrefix` this means an incorrect table name is used in the exception.

I have updated it to use `GetTableName(...)` method instead.

I also added the table name to the exception when checking the range key in the `MakeKey(...)` method to make it consistent with the other exceptions.

## Motivation and Context

I am using the high level driver without specifying a table name using the `[DynamoDBTable]` attribute because the table names are different per environment so are in config. Instead I provide the table name using the `OverrideTableName` property.
This meant the table name defaulted to the name of the type of the class I was persisting which was completely different to the name of the table I was using and caused some confusion.

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextInternal.cs
+++ b/sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextInternal.cs
@@ -760,9 +760,16 @@ namespace Amazon.DynamoDBv2.DataModel
                 throw new ArgumentNullException("hashKey");
 
             if (storageConfig.HashKeyPropertyNames.Count != 1)
-                throw new InvalidOperationException("Must have one hash key defined for the table " + storageConfig.TableName);
+            {
+                var tableName = GetTableName(storageConfig.TableName, currentConfig);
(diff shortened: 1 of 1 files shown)


normj (approved):

philasmar (approved):

boblodgett (approved):

ashovlin:
Thank you for your PR. I have this staged internally for an upcoming manual release of the SDK (TBD, as opposed to the daily release with service updates).

ashovlin:
Thanks again for the contribution! This has been included in 3.7.0.45 of AWSSDK.DynamoDBv2.

https://www.nuget.org/packages/AWSSDK.DynamoDBv2/3.7.0.45

Issue #1943, opened 2021-11-24
AWS4Signer produces incorrect signature for reserved characters


jarruda:
## Description
[AWS4Signer](https://github.com/aws/aws-sdk-net/blob/master/sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4Signer.cs#L30) produces an incorrect signature for URIs containing reserved characters.

The root cause is [AWSSDKUtils.CanonicalizeResourcePath](https://github.com/aws/aws-sdk-net/blob/db5c9aec40d2d1f9b8f97d34e95a9b8108c39558/sdk/src/Core/Amazon.Util/AWSSDKUtils.cs#L331) incorrectly encoding path components when `detectPreEncode` is true -- which `AWS4Signer` sets for all non-S3 URIs.

I'm aware `AWS4Signer` is an internal class, but unlike other AWS SDKs it doesn't appear that a publicly-vended SigV4 signer is part of the .NET SDK, so I have attempted to use it.  Nonetheless, the root cause is a utility function in the public portion of the SDK.  Please correct me if this is not the case and there is a publicly vended signing utility.

## Reproduction Steps

Prepare a request containing a reserved character such as `:` and attempt to have that request signed by AWS4Signer.  An easier way to isolate the issue is to directly use `AWSSDKUtils.CanonicalizeResourcePath` in the same way it does.  That is, `detectPreEncode` must be set to `true`.

For an example path `/:`, or `/%3A` once encoded to a legal URI:

```cs
var encodedPathComponent = AWSSDKUtils.UrlEncode(":", false);
var path = $"/{encodedPathComponent}";
var canonicalPath = AWSSDKUtils.CanonicalizeResourcePath(new Uri("https://execute-api.us-east-1.amazonaws.com"), path, true);

// path = "/%3A"
// canonicalPath = "/%25253A" -- should be "%253A"     
```

[...]

PR #1962, opened 2021-12-24
Fix the error when object key has special character

vincywindy:
set true to igonre the special char where path property is specified,

Fix https://github.com/aws/aws-sdk-net/issues/1943

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
+++ b/sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
@@ -366,7 +366,7 @@ namespace Amazon.Util
                 }
             }
 
-            var canonicalizedResourcePath = AWSSDKUtils.JoinResourcePathSegments(encodedSegments, false);
+            var canonicalizedResourcePath = AWSSDKUtils.JoinResourcePathSegments(encodedSegments, true);
 
             // Get the logger each time (it's cached) because we shouldn't store it in a static variable.
             Logger.GetLogger(typeof(AWSSDKUtils)).DebugFormat("{0} encoded {1}{2} for canonicalization: {3}",


Edric-Li:
@vincywindy Has this Pull Request been merged and published?

vincywindy:
> @vincywindy此拉取请求是否已合并并发布？

No

Edric-Li:
Why isn't this merged?

vincywindy:
> 为什么没有合并？

I don't know,Is there something I'm not doing？

Edric-Li:
Is anyone still maintaining it? If there is no problem with this PR, can it be merged and released a new version as soon as possible?

normj:
@Edric-Li and all sorry this issue hasn't gotten attention. As small as the PR looks it changes a very low level area and these encoding changes are always tricky to change so we don't fix one issue and cause another. We have it on our queue to get to but there are many competing priorties.

sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
```diff
@@ -366,7 +366,7 @@ public static string CanonicalizeResourcePath(Uri endpoint, string resourcePath,
                 }
             }
 
-            var canonicalizedResourcePath = AWSSDKUtils.JoinResourcePathSegments(encodedSegments, false);
+            var canonicalizedResourcePath = AWSSDKUtils.JoinResourcePathSegments(encodedSegments, true);
```
peterrsongg, line 369: Thanks for opening up a PR. The issue in question #1943 is complaining that we are encoding special characters twice. In the example listed (`"/%3A"`), `CanonicalizeResourcePath` returns `"/%25253A"` while the user is saying it should return `"%253A`. Changing this value from false to true doesn't fix the issue, it just completely ignores uri encoding altogether.

Whether or not this is a bug or by design requires more investigation but this PR needs tests and the solution does not fix the underlying issue so I'm going to go ahead and close the PR. I will leave the issue open as we investigate this further.

PR #1966, opened 2022-01-03
Add target framework for net6.0

jon-armen:
Adding target framework for net6.0

## Description
Adding target framework for net6.0

## Motivation and Context
Add direct support for net6.0 - Implements #1967

## Testing
All existing tests should cover, updated test projects to include dotnet6.0 target framework

## Types of changes

## Checklist

## License


sdk/src/Core/Amazon.Runtime/AmazonServiceClient.cs
```diff
@@ -31,6 +31,7 @@ namespace Amazon.Runtime
 {
     public abstract class AmazonServiceClient : IDisposable
     {
+        private object _lock = new object();
```
sharwell, line 34:
```suggestion
        private readonly object _lock = new object();
```

sdk/src/Core/Amazon.Runtime/AmazonWebServiceRequest.cs
```diff
@@ -23,21 +23,22 @@ namespace Amazon.Runtime
     /// </summary>
     public abstract partial class AmazonWebServiceRequest : Amazon.Runtime.Internal.IAmazonWebServiceRequest
     {
+        private object _lock = new object();
```
sharwell, line 26:
```suggestion
        private readonly object _lock = new object();
```

sdk/src/Core/Amazon.Runtime/Internal/Auth/AbstractAWSSigner.cs
```diff
@@ -26,14 +26,15 @@ public enum ClientProtocol { QueryStringProtocol, RestProtocol, Unknown }
 
     public abstract class AbstractAWSSigner
     {
+        private object _lock = new object();
```
sharwell, line 29:
```suggestion
        private readonly object _lock = new object();
```

sdk/src/Core/Amazon.Runtime/AmazonServiceClient.cs
```diff
@@ -55,14 +56,14 @@ protected virtual bool SupportResponseLogging
         {
             add
             {
-                lock (this)
+                lock (_lock)
```
sharwell, line 59: 💡 This lock would not be necessary if the `mBeforeMarshallingEvent` field was removed and the event declaration simplified to:

```csharp
internal event PreRequestEventHandler BeforeMarshallingEvent;
```

extensions/test/NETCore.SetupTests/NETCore.SetupTests.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFrameworks>netcoreapp3.1;net6.0</TargetFrameworks>
```
sharwell, line 4: 💡 The first framework listed for `TargetFrameworks` controls the default view of a document opened inside Visual Studio. In the current form, documents open by default under a netcoreapp3.1 view. If you reverse them, they open by default under a net6.0 view. There are advantages and disadvantages to both approaches so I'll let you decide the one you want.

sdk/src/Services/AppRegistry/Properties/AssemblyInfo.cs
```diff
@@ -15,6 +15,8 @@
 [assembly: AssemblyDescription("The Amazon Web Services SDK for .NET (NetStandard 2.0) - AWS Service Catalog App Registry. AWS Service Catalog AppRegistry provides a repository of your applications, their resources, and the application metadata that you use within your enterprise.")]
 #elif NETCOREAPP3_1
 [assembly: AssemblyDescription("The Amazon Web Services SDK for .NET (.NET Core 3.1) - AWS Service Catalog App Registry. AWS Service Catalog AppRegistry provides a repository of your applications, their resources, and the application metadata that you use within your enterprise.")]
+#elif NET6_0
+[assembly: AssemblyDescription("The Amazon Web Services SDK for .NET (6.0) - AWS Service Catalog App Registry. AWS Service Catalog AppRegistry provides a repository of your applications, their resources, and the application metadata that you use within your enterprise.")]
```
sharwell, line 19: 💡 These would likely be simplified tremendously if a Directory.Build.props/Directory.Build.targets file was used to generate them from a template across all projects. Would be a follow-up and not part of this pull request.

sdk/src/Services/CloudWatchEvidently/AWSSDK.CloudWatchEvidently.NetStandard.csproj
```diff
@@ -1,10 +1,11 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
    <RunAnalyzersDuringBuild Condition="'$(RunAnalyzersDuringBuild)'==''">true</RunAnalyzersDuringBuild>
-    <TargetFrameworks>netstandard2.0;netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netcoreapp3.1;net6.0</TargetFrameworks>
     <DefineConstants>$(DefineConstants);NETSTANDARD;AWS_ASYNC_API</DefineConstants>
     <DefineConstants Condition="'$(TargetFramework)' == 'netstandard2.0'">$(DefineConstants);NETSTANDARD20;AWS_ASYNC_ENUMERABLES_API</DefineConstants>
```
sharwell, line 6: 💡 `NETSTANDARD2_0` is defined by the SDK if you'd like to remove the custom one here in favor of the standard one.

sharwell, line 6: 💡 A Directory.Build.props shared for the solution would likely simplify the definition of `AWS_ASYNC_ENUMERABLES_API`.

sdk/src/Services/CognitoIdentity/Custom/CognitoAWSCredentials.cs
```diff
@@ -31,6 +31,7 @@ namespace Amazon.CognitoIdentity
     public partial class CognitoAWSCredentials : RefreshingAWSCredentials
     {
         #region Private members
+        private object _lock = new object();
```
sharwell, line 34:
```suggestion
        private readonly object _lock = new object();
```

extensions/test/NETCore.SetupTests/NETCore.SetupTests.csproj
jon-armen: Good callout, I had never noticed that.  I'd leave it up to the maintainers of this repo to main the determination on how they would like to have it organized.

sdk/src/Services/AppRegistry/Properties/AssemblyInfo.cs
jon-armen: Another great callout.  Agree w/ not adding it to the context of this PR, and it would make this much easier.

sdk/src/Core/Amazon.Runtime/AmazonServiceClient.cs
```diff
@@ -344,13 +270,11 @@ protected static TResponse EndInvoke<TResponse>(IAsyncResult result)
...
         {
-            //if (request == null)
-            //    return;
-            if (mBeforeMarshallingEvent == null)
+            if (BeforeMarshallingEvent == null)
                 return;
 
             PreRequestEventArgs args = PreRequestEventArgs.Create(executionContext.RequestContext.OriginalRequest);
-            mBeforeMarshallingEvent(this, args);
+            BeforeMarshallingEvent(this, args);
```
sharwell, line 277: 💡 The recommended event invocation pattern is:

```csharp
BeforeMarshallingEvent?.Invoke(this, PreRequestEventArgs.Create(executionContext.RequestContext.OriginalRequest));
```

Note that `?.` triggers an early escape when null; i.e. `PreRequestEventArgs.Create` will only be called for this expression if one or more event handlers is registered.

sharwell: 📝 This cannot be directly applied to all cases, but instead shows one example to consider in a future PR.

jon-armen: I've updated the one case where we can apply this.

normj:
We have an internal branch we have have been working on adding .NET 6 as a TFM. We aren't in a big rush to do it till we actually start using some .NET 6 APIs because it will significantly increase are already very long build time. I'm closing this PR because taking a new TFM is actually a lot of coordination with our build components outside of the scope of just the SDK code and generator.

Long term this is something we are interesting doing especially as we want make the SDK more trim safe.

PR #1983, opened 2022-02-22
DynamoDBv2 Allow lists and maps to be mapped even if null is set to false

### LINKED ISSUE #1980: DynamoDB Stream - How to Deserialize DynamoDb stream JSON back to my original object? - Serializing and the Deserialising AttributeMap, List of Objects is Lost
## The Question

I have data in DynamoDB. I know the shape of it is a certain object, because it put it in there. The object in this case is `ObjectWithListObjects`. It contains a list of objects with a couple of properties on it:

    public class ObjectWithListObjects
    {
        public List<Translation> BrandFamilyTranslation { get; set; }
    }
 
    public class Translation
    {
        public string LanguageCode { get; set; }
        public string Value { get; set; }
    }

I'm subscribing a lambda to the DynamoDB stream. My lambda handler has a `DynamoDBEvent` property defined so that I can access the `NewImage` data of the stream.

It makes sense to me that I should be able to get from the DynamoDB stream object `NewImage` property back to my own object `ObjectWithListsObjects` again in my lambda, because I know the shape of the data that I put in. I figure that the .NET library is able to do essentially the same thing when using the object model, take the data from DynamoDB and turn it into an object so I wanted to do the same with the stream data.

I wrote a helper that looked promising at first but now I've found that the List (`L`) data is being lost, `Document.FromAttributeMap` interprets the list as an `AlwaysSendList` with the data in there...

image

And then the `AnySendList` is then unable to be processed by the `dynamoDbContext.FromDocument<T>` call and is instead seen as a `DynamoDBNull` which can't be converted to my `List<Translation>`

image

I've created an NUnit test to show the issue in full from creating the object as an Attribute Map (`Dictionary<string,AttributeValue>`) which is the same shape as the `NewImage` property received in the DynamoDB stream, then I serialize it and then attempt to return it back to the object shape using `Document.FromAttributeMap` and then `context.FromDocument<T>` from the DynamoDB library.

     [Fact]
    public void WhyDoesntThisWork()
    {
        AmazonDynamoDBConfig clientConfig = new() { RegionEndpoint = RegionEndpoint.EUWest1 };
        DynamoDBContext dynamoDbContext = new(new AmazonDynamoDBClient(clientConfig));
 
        var initialAttributeMap = new Dictionary<string, AttributeValue>
        {
            {
                "BrandFamilyTranslation", new AttributeValue
                {
                    IsLSet = true,
                    L = new List<AttributeValue>
                    {
                        new AttributeValue
                        {
                            IsMSet = true,
                            M = new Dictionary<string, AttributeValue>
                            {
                                {"LanguageCode", new AttributeValue("en-GB")},
                                {"Value", new AttributeValue("Cookmaster 100")}
                            }
                        }
                    }
                }
            }
        };
 
        var serialised = JsonSerializer.Serialize(initialAttributeMap);
        var attsFromJson = JsonSerializer.Deserialize<Dictionary<string, AttributeValue>>(serialised);
        var dynamoDocument = Document.FromAttributeMap(attsFromJson);
        var mappedObject = dynamoDbContext.FromDocument<ObjectWithListObjects>(dynamoDocument);

        Assert.Single(mappedObject.BrandFamilyTranslation);
        Assert.Equal("Cookmaster 100", mappedObject.BrandFamilyTranslation.ElementAtOrDefault(0)?.Value);
        Assert.Equal("en-GB", mappedObject.BrandFamilyTranslation.ElementAtOrDefault(0)?.LanguageCode);
    } 

    public class ObjectWithListObjects
    {
        public List<Translation> BrandFamilyTranslation { get; set; }
    }
 
    public class Translation
    {
        public string LanguageCode { get; set; }
        public string Value { get; set; }
    }

In this example I have created the object using the object initializer, but I have also tried to create it by deserializing from the exact JSON that is received by my lambda function and that does the same thing. I'm happy to share the exact JSON if this would help.

If I'm completely going off off-piste here using the code I have written, is there an SDK supported way to convert the DynamoDB stream `NewImage` or `OldImage` straight to an object again because I know the shape that I put it in as?

### Environment

* SDK Version: AWSSDK.DynamoDBv2 
* Package Version: 3.7.0 
* OS Info: Windows 10 
* Build Environment: Rider 
* Targeted .NET Platform: net6.0

---

This is a :question: general question

Hazy87:
This change is required to be able to deserialize maps and lists from json when NULL exists in the json. To fix #1980 

## Description
We are moving the order in which attributes types decided so that if NULL is set to false then map and lists are not skipped over.

## Motivation and Context
We want to use the sdk to deserialize messages from the dynamo stream in a lambda.

## Testing

Added 2 tests. One for lists and one for maps.

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/DynamoDBv2/Custom/DocumentModel/Document.cs
+++ b/sdk/src/Services/DynamoDBv2/Custom/DocumentModel/Document.cs
@@ -578,10 +578,6 @@ namespace Amazon.DynamoDBv2.DocumentModel
             if (TryToDynamoDBBool(attributeValue, out ddbBool))
                 return ddbBool;
 
-            DynamoDBNull ddbNull;
-            if (TryToDynamoDBNull(attributeValue, out ddbNull))
-                return ddbNull;
-
             DynamoDBList ddbList;
             if (TryToDynamoDBList(attributeValue, out ddbList))
                 return ddbList;
@@ -589,6 +585,10 @@ namespace Amazon.DynamoDBv2.DocumentModel
             Document document;
             if (TryToDocument(attributeValue, out document))
                 return document;
+
+            DynamoDBNull ddbNull;
+            if (TryToDynamoDBNull(attributeValue, out ddbNull))
+                return ddbNull;
 
             return null;
         }
--- a/sdk/test/Services/DynamoDBv2/UnitTests/Custom/DynamoDBTests.cs
+++ b/sdk/test/Services/DynamoDBv2/UnitTests/Custom/DynamoDBTests.cs
@@ -9,7 +9,7 @@ using Amazon.Auth.AccessControlPolicy;
 using Amazon.Auth.AccessControlPolicy.ActionIdentifiers;
 using Amazon.DynamoDBv2.DocumentModel;
 using Amazon.DynamoDBv2.DataModel;
-
+using Amazon.DynamoDBv2.Model;
 using ThirdParty.Json.LitJson;
 
 namespace AWSSDK_DotNet35.UnitTests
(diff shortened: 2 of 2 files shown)


philasmar (approved):

coultonluke:
Does approved mean it will go on the next build?

Thanks for your help. ☺

normj:
@coultonluke We do two types of releases. There is the daily release for service updates and then we do periodic releases of miscellaneous PRs and core changes. The second type of release happen every 1 to 4 weeks. Assuming there are no surprises this PR would go out in that next PR release.

ashovlin:
This was released today in [AWSSDK.DynamoDBv2 3.7.3.24](https://www.nuget.org/packages/AWSSDK.DynamoDBv2/3.7.3.24) as commit https://github.com/aws/aws-sdk-net/commit/e34690317e6028eb20bf4ee4ee9b057a57246f4c.

Thank you again for your contribution @Hazy87 !

PR #1988, opened 2022-03-07
Consolidate getting profile name

rahulpnath:
Consolidate getting the profile name into one method

## Description
Added a function to resolve the profile name and make this consistent across all the places that require a profile name. 

## Motivation and Context
Some of the places where the profile name is retrieved it does not look at some of the expected configurations (like `AWSConfigs.AWSProfileName`)

## Testing
Added test cases for the scenario where the environment variable can be null or empty and it returns the default profile name.

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Credentials/FallbackCredentialsFactory.cs
+++ b/sdk/src/Core/Amazon.Runtime/Credentials/FallbackCredentialsFactory.cs
@@ -59,7 +59,7 @@ namespace Amazon.Runtime
             };
         }
 
-        public static string GetProfileName()
+        internal static string GetProfileName()
         {
             var profileName = AWSConfigs.AWSProfileName ?? 
                               Environment.GetEnvironmentVariable(AWS_PROFILE_ENVIRONMENT_VARIABLE);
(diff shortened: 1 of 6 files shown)


sdk/src/Core/Amazon.Runtime/Credentials/FallbackCredentialsFactory.cs
```diff
@@ -59,10 +59,22 @@ public static void Reset(IWebProxy proxy)
             };
         }
 
-        private static AWSCredentials GetAWSCredentials(ICredentialProfileSource source)
+        public static string GetProfileName()
```
normj, line 62: Does this need to be `public`? Since it is only being used in Core it could be `internal`.

normj (changes requested):

sdk/src/Core/Amazon.Runtime/Credentials/FallbackCredentialsFactory.cs
rahulpnath: Since it was a calculated property made it public, in case it was required outside of Core. But happy to make it internal for now. Pushed up the changes.

normj (approved):

ppittle (approved):

ppittle:
FYI - I ran the PR against our full automated smoke test suite and a test failed.  Investigating.

ppittle:
Resolved test failure and have this staged internally for an upcoming manual release of the SDK (TBD, as opposed to the daily release with service updates).

philasmar:
This was released today and is available in AWSSDK.Core 3.7.10 https://www.nuget.org/packages/AWSSDK.Core/3.7.10.
Thank you for the contribution!

Issue #900, opened 2018-03-15
Modify CloudSearch Search to use POST instead of GET


joel-cass:
Any search request using AmazonCloudSearchDomain.SearchAsync(searchRequest) or AmazonCloudSearchDomain.Search(searchRequest) that has a large search string will end up with a 413 Full Head message being returned from the AWS service.

This exception is not handled gracefully by the Marshaller and will return a Null Pointer Exception that obfuscates the actual cause of the issue. I don't care much for this; it would be better that the actual issue  above is addressed.

## Expected Behavior
Searching for a long string should return results (create a search longer than 4000 characters for example).

The same search can be run in the console with no problems as it is using post

## Current Behavior
Instead it returns
``[NullReferenceException: Object reference not set to an instance of an object.]
   Amazon.Runtime.Internal.Transform.JsonErrorResponseUnmarshaller.Unmarshall(JsonUnmarshallerContext context) +874
   Amazon.CloudSearchDomain.Model.Internal.MarshallTransformations.SearchResponseUnmarshaller.UnmarshallException(JsonUnmarshallerContext context, Exception innerException, HttpStatusCode statusCode) +40
   Amazon.Runtime.Internal.Transform.JsonResponseUnmarshaller.UnmarshallException(UnmarshallerContext input, Exception innerException, HttpStatusCode statusCode) +61
   Amazon.Runtime.Internal.HttpErrorResponseExceptionHandler.HandleException(IExecutionContext executionContext, HttpErrorResponseException exception) +282

[AmazonUnmarshallingException: Error unmarshalling response back from AWS. ]
   Amazon.Runtime.Internal.HttpErrorResponseExceptionHandler.HandleException(IExecutionContext executionContext, HttpErrorResponseException exception) +1011
   Amazon.Runtime.Internal.ErrorHandler.ProcessException(IExecutionContext executionContext, Exception exception) +275``

And using Fiddler we see the response from the server:
``HTTP/1.1 413 FULL head
Content-Length: 0
Connection: keep-alive``

## Possible Solution
SearchRequestMarshaller has hard-coded the method of GET for requests.

https://github.com/aws/aws-sdk-net/blob/1a5187e0dcddadba10a6b90fd827f2e485af0f9c/sdk/src/Services/CloudSearchDomain/Generated/Model/Internal/MarshallTransformations/SearchRequestMarshaller.cs#L58

Change it to POST, and the problem will probably go away.

## Steps to Reproduce (for bugs)

[...]

PR #2014, opened 2022-04-30
Re: [aws/aws-sdk-net] Modify CloudSearch Search to use POST instead of GET (#900)

joel-cass:
## Description

Please see ticket https://github.com/aws/aws-sdk-net/issues/900

Changing GET to POST allows large requests to be processed. This has been addressed in java equivalent code according to comments on the ticket.

## Motivation and Context

Its been 4 years and no useful response from AWS. I am hoping this PR expedites the process

Issue 900 https://github.com/aws/aws-sdk-net/issues/900

## Testing

Testing method also in ticket

Refer to ticket 900

## Types of changes

## Checklist

## License

- [x ] I confirm that this pull request can be released under the Apache 2 license

### DIFF
--- a/sdk/src/Services/CloudSearchDomain/Generated/Model/Internal/MarshallTransformations/SearchRequestMarshaller.cs
+++ b/sdk/src/Services/CloudSearchDomain/Generated/Model/Internal/MarshallTransformations/SearchRequestMarshaller.cs
@@ -56,7 +56,7 @@ namespace Amazon.CloudSearchDomain.Model.Internal.MarshallTransformations
         {
             IRequest request = new DefaultRequest(publicRequest, "Amazon.CloudSearchDomain");
             request.Headers[Amazon.Util.HeaderKeys.XAmzApiVersion] = "2013-01-01";
-            request.HttpMethod = "GET";
+            request.HttpMethod = "POST";
 
             request.AddSubResource("format", "sdk");
             request.AddSubResource("pretty", "true");


ashovlin:
Thanks for opening a PR.

As pointed out [in the issue discussion](https://github.com/aws/aws-sdk-net/issues/900#issuecomment-373773434), this file is generated from the service model, which specifies "GET" [here](https://github.com/aws/aws-sdk-net/blob/master/generator/ServiceModels/cloudsearchdomain/cloudsearchdomain-2013-01-01.normal.json#L18). Unfortunately the current implementation of this change would get overwritten during the SDK release when we run the generator.

Let me:
1. Ask around internally with how other AWS SDKs are handling this. In addition to Java, it [looks like Ruby has done some work here](https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-cloudsearchdomain/lib/aws-sdk-cloudsearchdomain/plugins/switch_to_post.rb)
2. Think about if and how we'd want to approach this to avoid clashing with the generated code. We do have [customization hooks in the generator](https://github.com/aws/aws-sdk-net/blob/master/generator/ServiceModels/cloudsearchdomain/cloudsearchdomain.customizations.json) that let us plug in functions that can modify requests before or after the generated code.

boblodgett:
I am closing this PR because the change will not stick. It would be overwritten by the generator in the next release. We will need to have the service team change the model or add a generator customization to fix this problem.

I have also requested ticket https://github.com/aws/aws-sdk-net/issues/900 in our backlog be priority re-evaluated by the team.

Issue #1939, opened 2021-11-22
HashSet duplicate key exception when setting up region endpoints


ckuehl:
This is a continuation of #986.

After upgrading to .Net 6 on Linux, we've begun seeing the following exception on some of the machines in our fleet upon startup:

```
System.ArgumentException: An item with the same key has already been added. Key: System.Collections.Generic.HashSet`1[System.String]
at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
at Amazon.Internal.RegionEndpointV3.ServiceMap.Add(String serviceName, Endpoint endpoint, HashSet`1 variants)
at Amazon.Internal.RegionEndpointV3.AddVariantEndpointsToServiceMap(JsonData mergedEndpoint, String regionName, String serviceName, Dictionary`2 mergedVariants)
at Amazon.Internal.RegionEndpointV3.AddServiceToMap(JsonData service, String serviceName)
at Amazon.Internal.RegionEndpointV3.ParseAllServices()
at Amazon.Internal.RegionEndpointV3.GetEndpointForService(String serviceName, HashSet`1 variants)
at Amazon.Internal.RegionEndpointV3.GetEndpointForService(String serviceName, Boolean dualStack)
at Amazon.Runtime.ClientConfig.GetUrl(RegionEndpoint regionEndpoint, String regionEndpointServiceName, Boolean useHttp, Boolean useDualStack)
at Amazon.Runtime.ClientConfig.DetermineServiceURL()
at Amazon.Runtime.Internal.DefaultRetryPolicy..ctor(IClientConfig config)
at Amazon.Runtime.AmazonServiceClient.BuildRuntimePipeline()
```

We're using version 3.7.1 of AWSSDK.Extensions.NETCore.Setup and the latest version of the service-specific SDKs.

Let me know if you need any additional details regarding our environment.

---

This is a :bug: bug-report

ckuehl:
We've also seen:

```
System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
  ?, in bool Enumerator.MoveNext()
  ?, in void RegionEndpointV3.MergeJsonData(JsonData target, JsonData source, Dictionary<HashSet<string>, JsonData> variants)
  ?, in void RegionEndpointV3.AddServiceToMap(JsonData service, string serviceName)
  ?, in void RegionEndpointV3.ParseAllServices()
[...]

PR #2107, opened 2022-09-15
Fix for Issue#1939 - Added multi-thread handling to ServiceMap

EyalMaozCP:
## Description
Fix for [Issue#1939](https://github.com/aws/aws-sdk-net/issues/1939)

## Motivation and Context

We can not upgrade our AWSSDK.Core and therefore cant use any other service SDK... We get a lot of those exception because of large scale. This issue is really critical for our fluent work.
## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/RegionEndpoint/RegionEndpointProviderV3.cs
+++ b/sdk/src/Core/RegionEndpoint/RegionEndpointProviderV3.cs
@@ -21,12 +21,14 @@ using System.Linq;
 using System.Text.RegularExpressions;
 using System.Threading;
 using ThirdParty.Json.LitJson;
+using System.Collections.Concurrent;
 
 namespace Amazon.Internal
 {
     public class RegionEndpointV3 : IRegionEndpoint
     {
         private ServiceMap _serviceMap = new ServiceMap();
+        private readonly object _serviceMapLock = new object();
         public string RegionName { get; private set; }
         public string DisplayName { get; private set; }
         public string PartitionName
@@ -92,7 +94,7 @@ namespace Amazon.Internal
         {
             RegionEndpoint.Endpoint endpointObject = null;
 
-            lock (_serviceMap)
+            lock (_serviceMapLock)
             {
                 if (!_servicesLoaded)
                 {
@@ -239,7 +241,10 @@ namespace Amazon.Internal
             var signatureOverride = DetermineSignatureOverride(mergedEndpoint, serviceName);
 
             RegionEndpoint.Endpoint endpoint = new RegionEndpoint.Endpoint(hostname, authRegion, signatureOverride, dnsSuffix, deprecated);
-            _serviceMap.Add(serviceName, endpoint);
+            lock(_serviceMapLock)
+            {
+                _serviceMap.Add(serviceName, endpoint);
+            }
 
         }
 
@@ -280,7 +285,10 @@ namespace Amazon.Internal
                                      .Replace("{region}", regionName)
                                      .Replace("{dnsSuffix}", variantDnsSuffix);
 
-                _serviceMap.Add(serviceName, new RegionEndpoint.Endpoint(variantHostname, authRegion, signatureOverride, variantDnsSuffix, deprecated), tagsKey);
+                lock(_serviceMapLock)
+                {
+                    _serviceMap.Add(serviceName, new RegionEndpoint.Endpoint(variantHostname, authRegion, signatureOverride, variantDnsSuffix, deprecated), tagsKey);
+                }
             }
         }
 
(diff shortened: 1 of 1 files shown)


sdk/src/Core/RegionEndpoint/RegionEndpointProviderV3.cs
```diff
@@ -92,7 +94,7 @@ public RegionEndpoint.Endpoint GetEndpointForService(string serviceName, HashSet
         {
             RegionEndpoint.Endpoint endpointObject = null;
 
-            lock (_serviceMap)
+            lock (_serviceMapLock)
```
EyalMaozCP, line 97: The reason for this page is the [lock documentation](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/statements/lock): _

> Avoid using the same lock object instance for different shared resources, as it might result in deadlock or lock contention.

_

sdk/src/Core/RegionEndpoint/RegionEndpointProviderV3.cs
```diff
@@ -348,12 +356,12 @@ class ServiceMap
             /// <summary>
             /// Stores the plain endpoints for each service in the current region
             /// </summary>
-            private Dictionary<string, RegionEndpoint.Endpoint> _serviceMap = new Dictionary<string, RegionEndpoint.Endpoint>();
+            private ConcurrentDictionary<string, RegionEndpoint.Endpoint> _serviceMap = new ConcurrentDictionary<string, RegionEndpoint.Endpoint>();
```
ashovlin, line 359: Thanks for taking a stab at this! We made an attempt internally but haven't been able to figure it out yet.

Unfortunately we cannot rely on `ConcurrentDictionary` entirely yet as its not available in the .NET Framework 3.5 target of the SDK.
* Generally we prefer to avoid branching with preprocessor directives to reduce complexity, though given the reports in #1939 appear to focus on .NET 6 maybe it'd be acceptable if net35 isn't impacted as severely. Might need @normj to weigh in here.

Our thinking was that we'd need to focus on the `_serviceMap` lock up on line 97.

EyalMaozCP: Hi, Thank you for considering my changes. I read what you wrote and I'm sorry about the branch name, hope it didn't create too much mess. Regarding the `ConcurrentDictionary `, I now removed any use of it and just added `lock` instead. 
I do think the critical change is in line 97 as you mentioned but I added this behavior also in the class itself just to be on the safe side. I can change this how ever you like just lmk.

EyalMaozCP: Oh and BTW, our product uses a lot of AWS .net SDK packages and when we upgraded the AWSSDK.Core (and deployed to production) we got a lot of exception (not in stage environment) and we saw those exception from all around the SDK's referring to the same source, for example:
image

sdk/src/Core/RegionEndpoint/RegionEndpointProviderV3.cs
```diff
@@ -364,34 +374,46 @@ public void Add(string serviceName, RegionEndpoint.Endpoint endpoint, HashSet<st
...
-                        return false;
-                    }
-                    else
+                    lock (_variantMapLock)
                     {
-                        return _variantMap[serviceName].TryGetValue(variants, out endpoint);
+                        if (!_variantMap.ContainsKey(serviceName))
+                        {
+                            endpoint = default(RegionEndpoint.Endpoint);
+                            return false;
```
EyalMaozCP, line 411: Verified its ok to `return ` inside of `lock` statement:
https://stackoverflow.com/questions/266681/should-a-return-statement-be-inside-or-outside-a-lock

sdk/src/Core/RegionEndpoint/RegionEndpointProviderV3.cs
```diff
@@ -364,34 +374,51 @@ public void Add(string serviceName, RegionEndpoint.Endpoint endpoint, HashSet<st
...
                 {
-                    _serviceMap.Add(serviceName, endpoint);
+                    lock (_serviceMapLock)
+                    {
+                        _serviceMap.Add(serviceName, endpoint);
+                    }
                 }
                 else
                 {
-                    if (!_variantMap.ContainsKey(serviceName) || _variantMap[serviceName] == null)
```
EyalMaozCP, line 371: If '_variantMap[serviceName] == null' we will get exception of KeyAlreadyExist. That why I spitted the if statement to two.

ashovlin:
Hey @EyalMaozCP, as we discussed outside of GitHub last week we had an internal fix in progress for this same issue, so may no longer need this PR.

That was recently released as [AWSSDK.Core 3.7.13.11](https://www.nuget.org/packages/AWSSDK.Core/3.7.13.11): https://github.com/aws/aws-sdk-net/commit/c6c55eadd97e546c42df2ca2bcb13a5c51f54322.

Please let us know if you're still seeing issues upon upgrading, thanks.

EyalMaozCP:
Hey @ashovlin , Thank you very much, we will deploy the new SDK version probably next week (we have holiday this days) will update how it went.

EyalMaozCP:
Hey @ashovlin, we completed our AWS SDK core upgrade in production, and so far it looks good. No errors 💯 
Thank you so much for your help. I will now close this PR.

PR #2549, opened 2023-02-26
Align the default of HttpClientCacheSize to 1 across all the platforms

### LINKED ISSUE #1929: What is the reason for keeping HttpClientCacheSize to number Environment.ProcessorCount on Linux
## The Question
Currently we are investigating slowness of S3 access on **Linux deployment**, upon investigation we found number of cache HttpClient varies between Windows & Linux    [relevant section of the code is ClientConfig.cs](https://github.com/aws/aws-sdk-net/blob/b5d440ba2064956c32a0ced3804b34d151eb3123/sdk/src/Core/Amazon.Runtime/ClientConfig.cs#L732)

What is the reason of having different number for cached HttpClients ?
From .NET Core 2.1 onwards HttpClient uses SocketsHttpHandler on both Windows & Linux  and their should not be any behavior changes. 

Here is the sample scenario to show the difference
On 10 CPU systems make 10 S3 request to **same object** on the **same bucket.**

- **On Windows** - First request takes time, and all subsequent request are fast
- **On Linux** - All request request takes equal amount of time which includes connection establish cost 

So should we set the `AmazonS3Config.HttpClientCacheSize` to 1 on Linux ? , any recommendation ?

### Environment

* SDK Version: 3.7.3.12
* Package Version: Awssdk.s3 ver: 3.7.3.12, Awssdk.core ver:3.7.3.19
* OS Info: Windows 10 , Centos 8 
* Build Environment :Terminal `dotnet` 
* Targeted .NET Platform:  .NET 3.1

---

This is a :question: general question

danielmarbach:
Fixes #1929

## Description

See #1929

## Motivation and Context

See #1929

## Testing

The underlying assumptions are no longer true since years. Changing the default here should be reasonably covered by all existing tests accessing the http pipeline. 

## Types of changes

Arguably this could also be seen as a breaking change, but honestly, I think this should be considered a bug since the assumptions documented in the XML doc are no longer true and the behavior on non windows platform is harmful/detrimental for the throughput. 

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/ClientConfig.cs
+++ b/sdk/src/Core/Amazon.Runtime/ClientConfig.cs
@@ -923,7 +923,11 @@ namespace Amazon.Runtime
                     return _httpClientCacheSize.Value;
                 }
 
+#if NETCOREAPP3_1_OR_GREATER
                 return 1;
+#else
+                return RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? 1 : Environment.ProcessorCount;
+#endif
             }
             set => _httpClientCacheSize = value;
         }
--- a/sdk/src/Core/Amazon.Runtime/ClientConfig.cs
+++ b/sdk/src/Core/Amazon.Runtime/ClientConfig.cs
@@ -910,9 +910,8 @@ namespace Amazon.Runtime
         /// <summary>
         /// If CacheHttpClient is set to true then HttpClientCacheSize controls the number of HttpClients cached.
         /// <para>
-        /// On Windows the default value is 1 since the underlying native implementation does not have throttling constraints
-        /// like the non Windows Curl based implementation. For non Windows based platforms the default is the value return from 
-        /// System.Environment.ProcessorCount.
+        /// The default value is 1 which is suitable for Windows and for all other platforms that have HttpClient
+        /// implementations using <see cref="System.Net.Http.SocketsHttpHandler"/> (.NET Core 2.1 and higher).
         /// </para>
         /// </summary>
         public int HttpClientCacheSize
@@ -924,7 +923,7 @@ namespace Amazon.Runtime
                     return _httpClientCacheSize.Value;
                 }
 
-                return RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? 1 : Environment.ProcessorCount;
+                return 1;
             }
             set => _httpClientCacheSize = value;
         }


peterrsongg:
Hello thanks for opening a pull request. This request seems valid but we want to do some benchmarking ourselves and sanity checks before approving. Please provide some time for us to go over this.

sdk/src/Core/Amazon.Runtime/ClientConfig.cs
```diff
@@ -924,7 +923,7 @@ public int HttpClientCacheSize
                     return _httpClientCacheSize.Value;
                 }
 
-                return RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? 1 : Environment.ProcessorCount;
```
peterrsongg, line 927: Just to be safe can we add a preprocessor directive like 
`#IF NETCOREAPP3_1_OR_GREATER`

danielmarbach: How does that help given that you are not multi targeting your builds and NETCOREAPP3.1 is out of support?

peterrsongg: Even though it's out of support, there are still customers using older versions of netcore, so it's just a precaution

danielmarbach: In case my guess I was right I pushed an update. Otherwise just let me know and I will do another small change

peterrsongg: Yes that was what I was asking for thank you. Once it gets 2 approvers we can merge it in.

peterrsongg (approved):

danielmarbach:
Thanks @peterrsongg !

normj (approved):

danielmarbach:
I have one question. How do you compile the SDK? I'm asking because as far as I understand the directives have an effect only foe the TFMs you compile to. Are you still compiling for netcore31? Because if you aren't then this directive will not solve much

muhammad-othman:
Thanks @danielmarbach for your contribution
We've released it today in version [3.7.106](https://www.nuget.org/packages/AWSSDK.Core/3.7.106)

danielmarbach:
@muhammad-othman thanks!

Did you see my question here?

https://github.com/aws/aws-sdk-net/pull/2549#issuecomment-1456997503

peterrsongg:
> I have one question. How do you compile the SDK? I'm asking because as far as I understand the directives have an effect only foe the TFMs you compile to. Are you still compiling for netcore31? Because if you aren't then this directive will not solve much

Yes, we do compile for netcore31 still. It is one of the target frameworks

danielmarbach:
> I have one question. How do you compile the SDK? I'm asking because as far as I understand the directives have an effect only for the TFMs you compile to. Are you still compiling for netcore31? Because if you aren't then this directive will not solve much

This is what I have been trying to highlight the entire time. Based on my understanding, the guard added there has no effect because the SDK is not cross compiled into multiple TFMs that get added to the nuget. Therefore, the compiler directive has no effect.

madhub:
> > I have one question. How do you compile the SDK? I'm asking because as far as I understand the directives have an effect only for the TFMs you compile to. Are you still compiling for netcore31? Because if you aren't then this directive will not solve much
> 
> This is what I have been trying to highlight the entire time. Based on my understanding, the guard added there has no effect because the SDK is not cross compiled into multiple TFMs that get added to the nuget. Therefore, the compiler directive has no effect. 
> 
> See also the other ping https://github.com/aws/aws-sdk-net/pull/2549#issuecomment-1474143250

Does it mean this changes only effective if the target is netcore3.1 ?

boblodgett:
Per the code it would need to be netcoreapp3.1 or later: `NETCOREAPP3_1_OR_GREATER`. Since your code is targeting net6.0 I would expect nuget to pick the packaged netcoreapp3.1 binaries.

madhub:
But it not working, is it bug ?

danielmarbach:
I think this is due to not using any TFM specific APIs within that ifdef. I'm not an expert but I think it then uses the netstandard TFM

normj:
Sorry the problem was a silly one. Our local builds all compiled correctly but our build server, which is in the process of being replaced, only has .NET Core 3.1 SDK installed. The `OR_GREATER` preprocessors were added for .NET SDK 5. I'm going to add the older `NETCOREAPP3_1` as well as `NETCOREAPP3_1_OR_GREATER` to satisfy our build server. @danielmarbach Since you own this issue do you want to reopen it.

madhub:
> Sorry the problem was a silly one. Our local builds all compiled correctly but our build server, which is in the process of being replaced, only has .NET Core 3.1 SDK installed. The `OR_GREATER` preprocessors were added for .NET SDK 5. I'm going to add the older `NETCOREAPP3_1` as well as `NETCOREAPP3_1_OR_GREATER` to satisfy our build server. @danielmarbach Since you own this issue do you want to reopen it.

Just curious,   aren't there any tests to check this value is 1 in .net core 3.1 or higher on all OS platforms?.

normj:
>> Sorry the problem was a silly one. Our local builds all compiled correctly but our build server, which is in the process of being replaced, only has .NET Core 3.1 SDK installed. The `OR_GREATER` preprocessors were added for .NET SDK 5. I'm going to add the older `NETCOREAPP3_1` as well as `NETCOREAPP3_1_OR_GREATER` to satisfy our build server. @danielmarbach Since you own this issue do you want to reopen it.

>Just curious,   aren't there any tests to check this value is 1 in .net core 3.1 or higher on all OS platforms?.

That is part of the new build system I was alluding to.

danielmarbach:
@normj will you give a headsup once this is fixed?

normj:
@danielmarbach Yes, the change is complete and is queued for release in the next non-service update release. Those happen usually once every 1 or 2 weeks. I have a note on our tracker for who ever does that release to chime back here when the release happens.

normj:
Version [3.7.106.15](https://www.nuget.org/packages/AWSSDK.Core/3.7.106.15) of Core has been released which fixes our build issues with the http size still not being set t 1.

PR #2558, opened 2023-03-03
DynamoDB Fix TryConvert to not throw when there is no converter to be compliant with the signature of Try

danielmarbach:
## Description

I was quite surprised that the TryConvert method throws when there is no converter. That is unexpected when you look at the TryConvert signature. There is also nothing in the XML doc that would hint at that.

When you also look at methods that use the `TryConvert` method like

```
        public DynamoDBEntry ConvertToEntry<TInput>(TInput value)
        {
            DynamoDBEntry entry;
            if (TryConvertToEntry<TInput>(value, out entry))
                return entry;

            throw new InvalidOperationException(string.Format(CultureInfo.InvariantCulture,
                "Unable to convert [{0}] of type {1} to DynamoDBEntry", value, value.GetType().FullName));
        }
```

you can spot that a missing converter would bomb during the evaluation on the if even though I would expect the throw statement to kick into place.

## Types of changes

## Checklist

## License


danielmarbach:
From my perspective, I don't require this change anymore. Feel free to pull in or close.

PR #2565, opened 2023-03-07
DynamoDB Expose ability to map types

danielmarbach:
## Description

As discussed with @normj yesterday, currently, there is no way to use `ToDocument` or `FromDocument` without it doing table description and validation. This means the conversion logic is inaccessible for anyone not following the assumptions of the dynamodb context.

This PR introduces convert methods that allow mapping arbitrary types without triggering table description or validation.

## Motivation and Context

Without this behavior, we would have to go through JSON -> Document -> Attribute maps or completely reimplement the mapping logic which is already part of the SDK ourselves.

Alternatives considered:

- Adding methods to the interface directly (would be a larger breaking change)
- Publicly exposing the `ConvertOnly` flag on `DynamoDBOperationConfig` (feels off because that type can also be used to configure other methods like BatchWrite)
- Extracting the mapping into some static helper class outside the context (felt like a huge effort and potentially introducing regression because of the state access on the context)

## Testing

I have tried to write a test to verify no describe calls are being done but failed due to various constraints. I would need to use strict mocks but unfortunately the GetCache method safely casts to AmazonServiceClient and when that is null bypasses client access

image

furthermore setting up mocks to actually pass through that code would require something like

```csharp
            var dynamodbClientMock = new Mock<AmazonServiceClient>(MockBehavior.Strict);
            var client = dynamodbClientMock.As<IAmazonDynamoDB>();
```

which doesn't work because the config property of AmazonServiceClient is not virtual plus there is not protected empty constructor on AmazonServiceClient. 

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs
+++ b/sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs
@@ -157,6 +157,11 @@ namespace Amazon.DynamoDBv2.DataModel
 
         // Checks if the IndexName is set on the config
         internal bool IsIndexOperation { get { return !string.IsNullOrEmpty(IndexName); } }
+
+        /// <summary>
+        /// Instructs the context to not attempt to do any describe table or validation of the data model being mapped.
+        /// </summary>
+        internal bool? ConvertOnly { get; set; }
     }
 
     /// <summary>
@@ -283,7 +288,8 @@ namespace Amazon.DynamoDBv2.DataModel
             IndexName = null,
             ConditionalOperator = ConditionalOperatorValues.And,
             Conversion = null,
-            IsEmptyStringValueEnabled = null
+            IsEmptyStringValueEnabled = null,
+            ConvertOnly = null
         };
         private static DynamoDBContextConfig _emptyContextConfig = new DynamoDBContextConfig
         {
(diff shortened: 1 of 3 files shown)


danielmarbach:
FYI for us, having a non-generic type-based overload would actually be ideal. That got me thinking about an alternative implementation that would expose [SerializeToDocument](https://github.com/aws/aws-sdk-net/blob/87f5dba301796a389496ccf2dc7f10e2ea74b3cc/sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextInternal.cs#L679) and [DeserializeFromDocument](https://github.com/aws/aws-sdk-net/blob/87f5dba301796a389496ccf2dc7f10e2ea74b3cc/sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextInternal.cs#L669) both as a generic as well as a non-generic version. The upside of that approach would be that it is not necessary to have a `DynamoDBOperationConfig` flag. The downside would be that if we don't expose this functionality over the interface, testability might suffer. Given that the context implementation is a public type already, it might be a reasonable tradeoff between exposing the functionality vs having to extend the interface which could be considered a larger breaking change (I deliberately tried to avoid such breaking changes in the current implementation too to make it more likely to get accepted).

normj:
I don't have any feedback yet but just wanted to acknowledge I'm going through it and debating on it.

danielmarbach:
@normj was the debate with yourself intense? ;)

Issue #1796, opened 2021-02-12
Improve dependency injection and client configuration


rbx-rchall:
I'm trying to use an S3 client as long as I don't have to configure anything that is not available on the appsettings.json file everything is ok, but let's say I would like to configure the `ForcePathStyle`. I haven't been able to find a good solution for this nor documentation about it. 

Note that I could go and wrap the injected client on another class that would configure the client but this is not that pretty. 

## Describe the Feature
Ideally we have the extension method that injects the client provide a delegate with the client configuration or something similar.

```
services.AddAWSService<IAmazonS3ClientA>(options1, (clientConfig) => {
// Where clientConfig is AmazonS3Config
// and then we can things like 
clientConfig.ForcePathStyle = true;
});
```

## Solution
Not sure if something like this would be acceptable but this PR draft provides a better idea of what I consider a potential solution to this problem: https://github.com/aws/aws-sdk-net/pull/1798

Note that we might also follow a different pattern injecting a ConfigurationBuilder and provide a DefaultConfigurationBuilder generic on the Configuration type that can read params from `appsettings.json` file but the above solution also solves the problem if you are using the `options1.CreateServiceClient` in the event you are having to inject multiple clients of the same service with different aws options.

## How to use
```
var options1 = Configuration.GetAWSOptions("AWS_Profile_A");
var client = options1.CreateServiceClient<IAmazonS3>(configureClient:
                    (clientConfig) =>
                    {
                        if (clientConfig is AmazonS3Config config) config.ForcePathStyle = true;
                    });

```
or 

```
services.AddAWSService<IAmazonS3>(options1, configureClient:
                (clientConfig) =>
                {
                    if (clientConfig is AmazonS3Config config) config.ForcePathStyle = true;
                });
```

This is a :rocket: Feature Request

ashishdhingra:
Needs review with the team as this is getting lot of traction.

TylerReid:
@ashishdhingra I implemented this in the way that I described and opened [a PR](https://github.com/aws/aws-sdk-net/pull/2571) for consideration.

madhub:
@ashishdhingra  why this PR is stuck in review since 2021. Pls pull this change its useful feature

ashishdhingra:
> @ashishdhingra why this PR is stuck in review since 2021. Pls pull this change its useful feature

@madhub Thanks for the follow up. The PR is in team's radar and should be reviewed soon.

cadilhac:
[...]

PR #2571, opened 2023-03-21
Added overloads to GetAWSOptions that allow creating service specific client configs

TylerReid:
This implements a solution to https://github.com/aws/aws-sdk-net/issues/1796 where client specific configuration can't be set using the .net core configuration extensions. 

## Description
The main user facing change is an added overload to `GetAWSOptions` that takes a generic parameter of the type of aws config you are wanting to use. This allows settings such as `ForcePathStyle` in a derived class from `ClientConfig` to be set.

## Motivation and Context
This allows greater flexibility in configuring aws clients purely from `IConfiguration` sources. This resolves  https://github.com/aws/aws-sdk-net/issues/1796

## Testing
Testing so far has been done through the unit tests for AWSOption and client configs.

## Types of changes

## Checklist

## License


extensions/src/AWSSDK.Extensions.NETCore.Setup/ClientFactory.cs
```diff
@@ -219,39 +219,37 @@ private static ClientConfig CreateConfig(Type serviceInterfaceType, AWSOptions o
             }
 
             var defaultConfig = options.DefaultClientConfig;
-            if (options.IsDefaultClientConfigSet)
```
TylerReid, line 222: This is the one use of the internal `IsDefaultClientConfigSet` which checks to see if the backing field of `DefaultClientConfig` is null. If you read `DefaultClientConfig` like is done above it [sets the value](https://github.com/aws/aws-sdk-net/blob/master/extensions/src/AWSSDK.Extensions.NETCore.Setup/AWSOptions.cs#L79) so this is always true.

I went ahead and deleted the property in order to not have this gotcha sitting around.

extensions/src/AWSSDK.Extensions.NETCore.Setup/ClientFactory.cs
```diff
@@ -219,39 +219,37 @@ private static ClientConfig CreateConfig(Type serviceInterfaceType, AWSOptions o
...
-            {
-                var emptyArray = new object[0];
-                var singleArray = new object[1];
+            var emptyArray = new object[0];
+            var singleArray = new object[1];
 
-                var clientConfigTypeInfo = typeof(ClientConfig).GetTypeInfo();
-                foreach (var property in clientConfigTypeInfo.DeclaredProperties)
+            var clientConfigTypeInfo = options.DefaultClientConfig.GetType();
+            var properties = clientConfigTypeInfo.GetProperties(BindingFlags.Public | BindingFlags.Instance);
```
TylerReid, line 226: I changed the uses of `GetTypeInfo` to be `GetType` + `GetProperties` with binding flags in order to get inherited properties as well as ones on the declaring type.

This is also probably good to do because `GetTypeInfo` is possibly going to be deprecated as a result of this https://github.com/dotnet/runtime/issues/53217

madhub:
why this PR is stuck in review since 2021. Pls pull this change its useful feature

extensions/src/AWSSDK.Extensions.NETCore.Setup/ClientFactory.cs
peterrsongg: Since options is gauranteed to have a value I agree that it's fine to use `GetType() `here

peterrsongg (approved):
I've reviewed these changes and everything works as expected. Thanks for the contribution! Will await @normj feedback before releasing.

peterrsongg:
I went ahead and resolved the merge conflicts as well.

normj (approved):

peterrsongg:
This was merged in NETCoreSetup v 3.7.7

cadilhac:
I can't find these new signatures in sdk v4...

PR #2593, opened 2023-04-27
Native AOT Part1: Fix up ServiceFactory and InternalConfiguration trim warnings

normj:
## Description
This PR gets Core's trim warnings down from 19 to 15 warnings.

It removes use of the Activator and TypeConverter which were used to create only a small subset of data types. Instead of relying on reflection in Activator and TypeConverter replace them with explicit conditionals.

## Motivation and Context
Attempting to make progress getting the SDK trim compatible.

## Testing
This is just for a feature branch not into `main` so have relied on local testing.

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Internal/InternalConfiguration.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/InternalConfiguration.cs
@@ -13,15 +13,12 @@
  * permissions and limitations under the License.
  */
 using System;
-using System.Globalization;
 
 using Amazon.Runtime.Internal.Util;
 using System.Collections.Generic;
-using Amazon.Util;
 #if BCL || NETSTANDARD
 using Amazon.Runtime.CredentialManagement;
 #endif
-using System.ComponentModel;
 
 namespace Amazon.Runtime.Internal
 {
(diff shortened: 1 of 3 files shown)


philasmar (approved):

sdk/src/Core/Amazon.Util/Internal/PlatformServices/ServiceFactory.cs
```diff
@@ -92,5 +92,32 @@ private static Type GetServiceType<T>()
...
+            {
+                service = new ApplicationInfo();
+            }
+            else if (serviceType == typeof(IEnvironmentInfo))
+            {
+                service = new EnvironmentInfo();
+            }
+            else
+            {
+                throw new ArgumentException($"Unknown service {typeof(T).FullName}");
```
Beau-Gosse-dev, line 117: This doesn't build for me. 

`Error: The type or namespace name 'T' could not be found`

Should it be this instead?

```suggestion
                throw new ArgumentException($"Unknown service {serviceType.FullName}");
```

sdk/src/Core/Amazon.Runtime/Internal/InternalConfiguration.cs
```diff
@@ -130,17 +130,31 @@ private bool TryGetEnvironmentVariable(string environmentVariableName, out strin
                 return null;
             }
 
-            var converter = TypeDescriptor.GetConverter(typeof(T?));
```
Beau-Gosse-dev, line 133: nitpick: You can remove `using System.ComponentModel;` now.

normj: Removed all of the unused namespaces.

Beau-Gosse-dev (approved):

sdk/src/Core/Amazon.Util/Internal/PlatformServices/ServiceFactory.cs
normj: Fixed

PR #2607, opened 2023-05-17
Use Array.Empty<byte>() instead of byte[0]

danielmarbach:
## Description

Use Array.Empty<byte>() instead of byte[0]

## Motivation and Context

It is not necessary to allocate `byte[0]`

## Testing

Passes tests locally

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Internal/Transform/UnmarshallerContext.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Transform/UnmarshallerContext.cs
@@ -13,6 +13,7 @@
  * permissions and limitations under the License.
  */
 
+using Amazon.Util.Internal;
 using Amazon.Runtime.Internal.Util;
 using System;
 using System.Collections.Generic;
@@ -64,7 +65,7 @@ namespace Amazon.Runtime.Internal.Transform
             }
             else
             {
-                return new byte[0];
+                return ArrayEx.Empty<byte>();
             }
         }
 
--- a/sdk/src/Core/Amazon.Runtime/Internal/Util/ChunkedUploadWrapperStream.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Util/ChunkedUploadWrapperStream.cs
@@ -17,6 +17,7 @@ using System.Globalization;
 using System.IO;
 using System.Text;
 using Amazon.Util;
+using Amazon.Util.Internal;
 using Amazon.Runtime.Internal.Auth;
 using System.Security.Cryptography;
 using System.Collections.Generic;
@@ -366,7 +367,7 @@ namespace Amazon.Runtime.Internal.Util
             try
             {
                 var header = Encoding.UTF8.GetBytes(chunkHeader.ToString());
-                var trailer = new byte[0];
+                var trailer = ArrayEx.Empty<byte>();
 
                 // Append a trailing CRLF unless this is the final data chunk and there are trailing headers 
                 if (!(isFinalDataChunk && _trailingHeaders?.Count > 0))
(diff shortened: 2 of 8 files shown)


sdk/src/Core/Amazon.Runtime/Internal/Transform/UnmarshallerContext.cs
```diff
@@ -64,7 +64,7 @@ public byte[] GetResponseBodyBytes()
             }
             else
             {
-                return new byte[0];
+                return Array.Empty<byte>();
```
peterrsongg, line 67: Since we have to compile for NET Framework 3.5 and NET Framework 4.5 can we add preprocessor directives 
`#if NETCOREAPP3_1_OR_GREATER`  around the `Array.Empty` calls? 

`
#if NETCOREAPP3_1_OR_GREATER
return Array.Empty<byte>();
#else
return new byte[0];
`

danielmarbach: Why not simply shim Array.Empty?

danielmarbach: With that we can declare it once and use it consistsantly everywhere

normj: Creating a shim for .NET Framework works as well.

danielmarbach: Ok will do. Thanks for catching those. I usually only open the netstandard solution because everything is quite heavy weight.

danielmarbach: Pushed. Let me know what you think

sdk/src/Core/_bcl/ArrayEx.cs
```diff
@@ -0,0 +1,40 @@
...
+ * 
+ *  http://aws.amazon.com/apache2.0
+ * 
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+namespace System
```
normj, line 16: Since these aren't extension methods It doesn't feel right to put these Shims in `System`. Can you put the shims in `Amazon.Util/Internal/_bcl` and `Amazon.Util/Internal/_netstandard` respecitively and set the namespace to `Amazon.Util.Internal`?

normj (changes requested):
One small change, other then that I'm good with the approach.

sdk/src/Core/_bcl/ArrayEx.cs
danielmarbach: As long as you are ok if Big Integer in bouncy castle namespace gets then a util reference. That was the reason why I put it into system because that is an import that is available everywhere including the third party folder. Plus it is really just a shim API for Array

normj: I'm okay with the bouncy castle change. You are making a change no matter what to switch to `ArrayEx` I would rather it be explicit then it looks like it came from `System`.

danielmarbach: Ok will change

danielmarbach: Done. I have undone the BigInteger changes

normj:
Changes look good. I'll send the PR through our full dry run build and assuming that doesn't have any issues I'll approve it.

normj (approved):

peterrsongg (approved):

normj:
PR is queued up to go out with next non-service update release. Those types of releases usually happen about once every other week.

danielmarbach:
Out of curiosity. Was this included already?

danielmarbach:
Awesome. There is no rush. I was just curious when I was tracking some of my outstanding PRs.

normj:
PR was rebased, merged and released as part of version [3.7.104](https://www.nuget.org/packages/AWSSDK.DynamoDBv2/3.7.104) of the DynamoDB package

PR #2613, opened 2023-05-23
[NativeAOT Progress] Deprecate TypeFactory and remove its uses in the SDK

normj:
## Description
This PR is to the staging `feature/net6` branch where also trying to address Native AOT trimming warnings.

The SDK abstracted away type information through `Amazon.Util.Internal.TypeFactory` and `Amazon.Util.Internal.ITypeInfo`. This was necessary when the SDK supported .NET Standard 1.3 and the `System.Type` type in .NET Standard 1.3 did not have all of the same properties and methods as .NET Framework version did. With .NET Standard 2.0 `System.Type` is now the same, at least for our purposes, between .NET Standard 2.0+ and .NET Framework.

The abstraction was making it difficult to apply the AOT trimming attributes because they can only be set on `System.Type` not our abstraction. Since it is not needed in our SDK because we dropped .NET Standard 1.3 support in 3.7 the usage has been completely removed from the SDK. The `TypeFactory` was left in the SDK for backwards compatibility if a user ends up with a new core but an old service package that relied on `TypeFactory`. To confirm I removed the usage of `TypeFactory` I first deleted `TypeFactory` and then fixed all compile errors in the SDK. Then added back `TypeFactory` with the `Obsolete` attribute.

## License

- [ X] I confirm that this pull request can be released under the Apache 2 license

### DIFF
--- a/sdk/src/Core/Amazon.Util/Internal/InternalSDKUtils.cs
+++ b/sdk/src/Core/Amazon.Util/Internal/InternalSDKUtils.cs
@@ -114,20 +114,20 @@ namespace Amazon.Util.Internal
             if (propertyValues == null || propertyValues.Count == 0)
                 return;
 
-            var targetTypeInfo = TypeFactory.GetTypeInfo(target.GetType());
+            var targetType = target.GetType();
             
             foreach(var kvp in propertyValues)
             {
-                var property = targetTypeInfo.GetProperty(kvp.Key);
+                var property = targetType.GetProperty(kvp.Key);
                 if (property == null)
-                    throw new ArgumentException(string.Format(CultureInfo.InvariantCulture, "Unable to find property {0} on type {1}.", kvp.Key, targetTypeInfo.FullName));
+                    throw new ArgumentException(string.Format(CultureInfo.InvariantCulture, "Unable to find property {0} on type {1}.", kvp.Key, targetType.FullName));
 
                 try
                 {
-                    var propertyTypeInfo = TypeFactory.GetTypeInfo(property.PropertyType);
+                    var propertyTypeInfo = property.PropertyType;
                     if (propertyTypeInfo.IsEnum)
                     {
-                        var enumValue = Enum.Parse(property.PropertyType, kvp.Value.ToString(), true);
+                        var enumValue = Enum.Parse(propertyTypeInfo, kvp.Value.ToString(), true);
                         property.SetValue(target, enumValue, null);
                     }
                     else
@@ -137,7 +137,7 @@ namespace Amazon.Util.Internal
                 }
                 catch(Exception e)
                 {
-                    throw new ArgumentException(string.Format(CultureInfo.InvariantCulture, "Unable to set property {0} on type {1}: {2}", kvp.Key, targetTypeInfo.FullName, e.Message));
+                    throw new ArgumentException(string.Format(CultureInfo.InvariantCulture, "Unable to set property {0} on type {1}: {2}", kvp.Key, targetType.FullName, e.Message));
                 }
             }
         }
--- a/sdk/src/Core/Amazon.Util/Internal/TypeWrapper.cs
+++ b/sdk/src/Core/Amazon.Util/Internal/TypeWrapper.cs
@@ -20,12 +20,14 @@
  */
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Reflection;
 using System.Text;
 
 namespace Amazon.Util.Internal
 {
+    [Obsolete("The TypeFactory abstraction should not be used. This was needed when the SDK targeted .NET Standard 1.3 which was removed as part of version 3.7. Type information should be accessed directly from the System.Type class.")]
     public interface ITypeInfo
     {
         Type BaseType { get; }
@@ -68,11 +70,7 @@ namespace Amazon.Util.Internal
 
         object EnumToObject(object value);
 
-        ITypeInfo EnumGetUnderlyingType();
-
         object CreateInstance();
-
-        ITypeInfo GetElementType();
 
         bool IsType(Type type);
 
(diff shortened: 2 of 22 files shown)


sdk/src/Services/CognitoIdentity/Custom/CognitoAWSCredentials.cs
```diff
@@ -443,14 +444,14 @@ private bool ShouldRetry(AmazonCognitoIdentityException e)
         {
             add
             {
-                lock (this)
+                lock (identityChangeEventLock)
```
boblodgett, line 447: Why does this lock exist at all since it is only setting and getting a single member variable and usage doesn't seem to be protected with the same lock elsewhere. Is there someplace else where `mIdentityChangedEvent` is used that should be locked with `identityChangeEventLock`?

sdk/src/Core/Amazon.Util/Internal/TypeWrapper.cs
```diff
@@ -90,22 +88,38 @@ public interface ITypeInfo
 
     }
 
+    [Obsolete("The TypeFactory abstraction should not be used. This was needed when the SDK targeted .NET Standard 1.5 which was removed as part of version 3.7. Type information should be accessed directly from the System.Type class.")]
```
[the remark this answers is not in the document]
normj: Fixed

sdk/src/Services/CognitoIdentity/Custom/CognitoAWSCredentials.cs
normj: It is not setting but adding a value to a collection via the `+=` and `-=` operators. If memory services me correct we had an issue a long time ago with adding handlers to events for Unity or Xamarin. So this was probably done in defensive of that issue. It quite likely isn't needed anymore but this is not a high throughput area of code and I don't want make that change now.

boblodgett (approved):

sdk/src/Services/DynamoDBv2/Custom/DataModel/Utils.cs
```diff
@@ -55,13 +57,12 @@ internal static class Utils
         };
 
         public static readonly IEnumerable<Type> PrimitiveTypes = new HashSet<Type>(primitiveTypesArray);
-        private static readonly HashSet<ITypeInfo> PrimitiveTypeInfos = new HashSet<ITypeInfo>(primitiveTypesArray
-            .Select(p => TypeFactory.GetTypeInfo(p)));
+        private static readonly HashSet<Type> PrimitiveTypeInfos = new HashSet<Type>(primitiveTypesArray
+            .Select(p => p));
```
Beau-Gosse-dev, line 61: Can you remove this Select?

sdk/src/Services/DynamoDBv2/Custom/DataModel/Utils.cs
```diff
@@ -55,13 +57,12 @@ internal static class Utils
         };
 
         public static readonly IEnumerable<Type> PrimitiveTypes = new HashSet<Type>(primitiveTypesArray);
-        private static readonly HashSet<ITypeInfo> PrimitiveTypeInfos = new HashSet<ITypeInfo>(primitiveTypesArray
-            .Select(p => TypeFactory.GetTypeInfo(p)));
+        private static readonly HashSet<Type> PrimitiveTypeInfos = new HashSet<Type>(primitiveTypesArray
```
Beau-Gosse-dev, line 60: I think you can rename `PrimitiveTypeInfos`.

sdk/src/Services/DynamoDBv2/Custom/DataModel/Utils.cs
```diff
@@ -362,19 +359,16 @@ public static bool IsReadWrite(MemberInfo member)
...
 
-            foreach (var inter in targetTypeWrapper.GetInterfaces())
+            foreach (var inter in targetType.GetInterfaces())
             {
-                var interWrapper = TypeFactory.GetTypeInfo(inter);
-                if (object.Equals(interWrapper, interfaceTypeWrapper))
+                if (object.Equals(inter, interfaceType))
                     return true;
-                if (interfaceTypeWrapper.IsGenericTypeDefinition && interWrapper.IsGenericType)
+                if (targetType.IsGenericTypeDefinition && inter.IsGenericType)
```
Beau-Gosse-dev, line 369: Shouldn't `targetType.IsGenericTypeDefinition` be `interfaceType.IsGenericTypeDefinition`?

sdk/src/Services/DynamoDBv2/Custom/DataModel/InternalModel.cs
```diff
@@ -230,7 +230,7 @@ internal class StorageConfig
         public List<PropertyStorage> Properties { get; private set; }
 
         // target type
-        public ITypeInfo TargetTypeInfo { get; private set; }
+        public Type TargetTypeInfo { get; private set; }
```
Beau-Gosse-dev, line 233: `TargetTypeInfo` could be renamed

Beau-Gosse-dev (changes requested):

sdk/src/Core/Amazon.Util/Internal/InternalSDKUtils.cs
```diff
@@ -114,17 +114,17 @@ public static void ApplyValues(object target, IDictionary<string, object> proper
...
-                    throw new ArgumentException(string.Format(CultureInfo.InvariantCulture, "Unable to find property {0} on type {1}.", kvp.Key, targetTypeInfo.FullName));
+                    throw new ArgumentException(string.Format(CultureInfo.InvariantCulture, "Unable to find property {0} on type {1}.", kvp.Key, targetType.FullName));
 
                 try
                 {
-                    var propertyTypeInfo = TypeFactory.GetTypeInfo(property.PropertyType);
+                    var propertyTypeInfo = property.PropertyType;
                     if (propertyTypeInfo.IsEnum)
                     {
                         var enumValue = Enum.Parse(property.PropertyType, kvp.Value.ToString(), true);
```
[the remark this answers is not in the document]
normj: Done

sdk/src/Services/DynamoDBv2/Custom/DataModel/Utils.cs
normj: Done

sdk/src/Services/DynamoDBv2/Custom/DataModel/Utils.cs
normj: I actually just removed `PrimitiveTypeInfos`. It was only used once doing `Any` operation that the other `PrimitiveTypes` field works just fine for.

sdk/src/Services/DynamoDBv2/Custom/DataModel/Utils.cs
normj: Great catch, fixed.

sdk/src/Services/DynamoDBv2/Custom/DataModel/InternalModel.cs
normj: Done

sdk/src/Services/DynamoDBv2/Custom/Conversion/DynamoDBEntryConversion.cs
```diff
@@ -1,576 +1,569 @@
...
-        #region Private members
-
-        private ConverterCache ConverterCache = new ConverterCache();
-        private ConversionSchema OriginalConversion;
-
-        private void AddConverters(string suffix)
-        {
-            var typedConverterTypeInfo = TypeFactory.GetTypeInfo(typeof(Converter));
-            var assembly = TypeFactory.GetTypeInfo(typeof(DynamoDBEntryConversion)).Assembly;
-#if NETSTANDARD
```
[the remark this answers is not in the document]
normj: This code was written for .NET Standard 1.3 when we couldn't do `var allTypes = assembly.GetTypes();` like we could in .NET Framework. Now that our minimum bar is .NET Standard 2.0 we can remove this code and have both .NET Framework and .NET Standard/Core code paths use the same `var allTypes = assembly.GetTypes();` code path.

PR #2626, opened 2023-06-09
AOT: Add utility json deserialization method 

normj:
To support AOT we need to remove the usage inside the SDK of LitJson's JsonMapper which is not AOT safe. This PR adds a new `Amazon.Util.Internal.JsonSerializerHelper.Deserialize` common utility for handling deserialization. There are 2 different implementations of the method. .NET 6+ it uses System.Text.Json along with a `JsonSerializerContext` that uses a source generator to generate all of the deserialization code at compile removing reflection. For older .NET targets continue to use `JsonMapper`.

The goal is through the SDK you can replace the calls for `JsonMapper` with `JsonSerializerHelper.Deserialize` after defining your own `JsonSerializerContext`. So the code compiles with the same line of code regardless of targets stub classes were added for targets less then .NET 6 for the missing `System.Text.Json` types. Those types are not used for anything other then making a compatible method signature.

Initially I was trying to do some work to make sure the System.Text.Json usage was handling things case insensitively. That was proving to be challenging to get work in an abstract way that also didn't break for .NET Framework. Then I discovered that `JsonMapper` was already doing things case sensitive so I simplified the code and got rid of handling case insensitive names.

### DIFF
--- a/sdk/src/Core/Amazon.Util/EC2InstanceMetadata.cs
+++ b/sdk/src/Core/Amazon.Util/EC2InstanceMetadata.cs
@@ -26,6 +26,7 @@ using System.Globalization;
 using Amazon.Runtime.Internal.Util;
 using AWSSDK.Runtime.Internal.Util;
 using Amazon.Runtime.Internal;
+using Amazon.Util.Internal;
 
 namespace Amazon.Util
 {
@@ -348,7 +349,7 @@ namespace Amazon.Util
                 IAMInstanceProfileMetadata info;
                 try
                 {
-                    info = JsonMapper.ToObject<IAMInstanceProfileMetadata>(json);
+                    info = JsonSerializerHelper.Deserialize<IAMInstanceProfileMetadata>(json, EC2InstanceMetadataJsonSerializerContexts.Default);
                 }
                 catch
                 {
@@ -380,7 +381,7 @@ namespace Amazon.Util
                     var json = GetData("/iam/security-credentials/" + item);
                     try
                     {
-                        var cred = JsonMapper.ToObject<IAMSecurityCredentialMetadata>(json);
+                        var cred = JsonSerializerHelper.Deserialize<IAMSecurityCredentialMetadata>(json, EC2InstanceMetadataJsonSerializerContexts.Default);
                         creds[item] = cred;
                     }
                     catch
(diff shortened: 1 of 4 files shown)


sdk/src/Core/Amazon.Util/EC2InstanceMetadata.cs
```diff
@@ -26,6 +26,8 @@
 using Amazon.Runtime.Internal.Util;
 using AWSSDK.Runtime.Internal.Util;
 using Amazon.Runtime.Internal;
+using Amazon.Util.Internal;
+using System.Text.Json.Serialization.Metadata;
```
dscpinheiro, line 30: Is `NET6_0_OR_GREATER` not required here? You added it in the new class (https://github.com/aws/aws-sdk-net/blob/dd0fc7a34fd787d804b6403db45a651dc028e44a/sdk/src/Core/Amazon.Util/Internal/JsonSerializerContexts.cs#L5-L9)

muhammad-othman (commented):
Can you also switch to the helper on `public static RegionEndpoint Region` getter and remove the using of `ThirdParty.Json.LitJson;`. 
Also I don't see why do we need `using System.Text.Json.Serialization.Metadata;`

sdk/src/Core/Amazon.Util/EC2InstanceMetadata.cs
normj: I forgot to recompile framework which was failing because of the `using System.Text.Json.Serialization.Metadata;`. I have now removed the unnecessary `using`.

normj:
> Can you also switch to the helper on `public static RegionEndpoint Region` getter and remove the using of `ThirdParty.Json.LitJson;`. Also I don't see why do we need `using System.Text.Json.Serialization.Metadata;`

The `using System.Text.Json.Serialization.Metadata;` in `EC2InstanceMetadata.cs` as an accident after repeated refactoring attempts. I have removed it.

This PR is to make sure we are happy with this pattern of abstracting the Json serialization. Once we are it will be used to replace the direct usages of JsonMapper throughout the SDK like your call out for RegionEndpoint.

muhammad-othman (approved):

dscpinheiro (approved):

boblodgett (approved):

PR #2982, opened 2023-06-23
Change constructor accessibility in order to allow mocking

Cussa:
Allows mocking the BatchGet and BatchWrite classes (in a way)

## Description
Today we are not able to mock the BatchGet and BatchWrite classes. A possible workaround is to create fake classes that inherits from these classes and mock the returns. However, as these classes have a `internal` only constructor, this is not possible. Having them as `protected internal` would allow us to have this behaviour.

## Motivation and Context
We want to use Moq in order to Unit test our code.

## Types of changes

## Checklist

## License


peterrsongg (dismissed):
The change itself looks good, but we still need to run it through a full dry run. I'd also like to discuss with others on the team if there could be any repercussions to making it protected internal. Approving for now

peterrsongg:
Dismissing my approval because in[ this commit ](https://github.com/aws/aws-sdk-net/commit/c2f870f107fdcee9b90b06b401838c9f4393078f) the way we made `AsyncSearch` mockable is by creating a parameterless `protected` constructor, then marking the methods as virtual so you can provide custom implementations of those methods for mocking. Will need further discussion on which implementation we want to go with this time.

Cussa:
That would be another possible solution for sure. We use that to make the AsyncSearch already. I just tried to make the minimum amount of changes possible. I am happy with any solution that allow us to mock it and be able to test our implementation. Thanks for keep this in consideration!

normj:
Hi @Cussa, thanks for the PR. I would like to have a consistent experience for mocking these awkward objects and use the same approach mentioned in the commit @peterrsongg called out. If we follow that approach I see no problem pushing the PR through.

PR #2991, opened 2023-07-06
AWS SDK Compatibility: Check for `http_proxy` & `https_proxy` environment variables

### LINKED ISSUE #1977: .NET SDK should support environment variables http_proxy and https_proxy (as AWS CLI does)
I want AWS .NET SDK to fetch proxy settings from environment variables `http_proxy` and `https_proxy` in the same way as AWS CLI does, see <https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-proxy.html>.

### Is your Feature Request related to a problem?

I'm always frustrated that I don't get consistent behaviour of AWS toolchain in proxy handling, esp. proxies which require user/password authorization. Currently I need to specify `https_proxy` variable to get AWS CLI working, and then I need to craft an `app.config` file for _every_ .NET application as described in <https://docs.aws.amazon.com/sdk-for-net/v3/developer-guide/net-dg-config-ref.html#net-dg-config-ref-elements-proxy> .
Situation gets even worse when I need to distribute my apps to other users or organizations, which need to learn the AWS-specific format of the `aws` section in the `app.config` and adjust it to their conditions. A simple and straight environment variable would solve the issue.

Here is the way I have discovered the `http_proxy` and `https_proxy` env. variables don't work with AWS .NET SDK:

- installed a Telerik Fiddler Classic on my machine
- enabled the Fiddler's rule `Require Proxy Authentication` (implies user=1, password=1)
- set the `https_proxy=http://1:1@127.0.0.1:8888`

My .NET Framework-based (not .NET Core) app didn't work (AWS .NET SDK 3.7.5.9).  
When I created an `app.config` following <https://docs.aws.amazon.com/sdk-for-net/v3/developer-guide/net-dg-config-ref.html#net-dg-config-ref-elements-proxy>  it did work, but I don't think the config file should be necessary.

---

This is a :rocket: Feature Request

Mythra:
## Description

this brings us in line with many other AWS SDKs which check for these environment variables to modify client behavior. it does not change behavior if you've already been setting your web proxy explicitly, or in properties (the same as the java (and other) sdks).

we do introduce this with two new methods in ClientConfig to not break any existing callers using the methods that are public. while all of our callers have been updated to properly check web proxy, and then if that's unset check the protocol being used and optionally call `GetHttpProxy`/`GetHttpsProxy` as needed. If we're okay with breaking ever IT MAY be worthwhile to break `GetWebProxy` by taking in the protocol, and then just having the one method to prevent any accidental mis-use, and to ensure everyone is using those environment variables.

it's also important to note:
<https://github.com/dotnet/runtime/issues/31113> meaning a user will just get an error if they try to proxy to something that is using `HTTPS`. i figured just letting it error on construction is probably the 'safest' option. letting the user know that their setting isn't being respected, as opposed to just silently 'ignoring' it. this is also a decision we might want to change.

## Motivation and Context

fixes #1977 

## Testing

- I've manually ensured unit tests pass locally.
- I've confirmed that running integration tests connect through a proxy configured through an environment variable

## Types of changes

## Checklist

## License


dscpinheiro:
Thanks for the PR, we're discussing internally what should be done for the other SDKs that don't support these environment variables (https://github.com/aws/aws-sdk/issues/538#issuecomment-1551622222).

dscpinheiro:
Thanks a lot for the PR, we're discussing internally what should be done for the other SDKs that don't support these environment variables (https://github.com/aws/aws-sdk/issues/538#issuecomment-1551622222).

sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_bcl/HttpWebRequestFactory.cs
```diff
@@ -474,6 +474,14 @@ public virtual void ConfigureRequest(IRequestContext requestContext)
                 requestContext.Metrics.AddProperty(Metric.ProxyPort, requestContext.ClientConfig.ProxyPort);
                 _request.Proxy = proxy;
             }
+            else if (_request.RequestUri.Scheme == Uri.UriSchemeHttp)
+            {
```
dscpinheiro, line 478: AWS service endpoints are usually `HTTPS` (I took a quick look at https://docs.aws.amazon.com/general/latest/gr/aws-service-information.html and a lot of services don't even have an `HTTP` option).

With this approach of checking the scheme, wouldn't it be possible for a customer to set the `http_proxy` variable, which then it'd be ignored when trying to connect to `https://<service>.<region>.amazonaws.com`?

sdk/src/Core/Amazon.Runtime/_bcl/ClientConfig.cs
```diff
@@ -133,6 +133,34 @@ public WebProxy GetWebProxy()
             return proxy;
         }
 
+        /// <summary>
+        /// Returns a WebProxy instance to use for HTTPS connections if an
+        /// explicit web proxy hasn't been configured.
+        /// </summary>
+        public WebProxy GetHttpsProxy()
+        {
+            var httpsProxy = Environment.GetEnvironmentVariable("https_proxy");
```
dscpinheiro, line 142: Was the proxy used during your test configured to use credentials? For example, `http://1:1@127.0.0.1:8888`
During my tests it didn't look like the credentials were automatically parsed from the environment variable when using the .NET Framework.

The SDK still supports `net35` and `net45`, so this is something we should fix.

sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_bcl/HttpWebRequestFactory.cs
Mythra: Hey! Yeah this is a point of confusion, and poor cross-sdk behavior I've run into. The original idea behind these two environment variable (as far as I can tell from the community at large) is that the scheme of what you want to connect too matters. And as for AWS you're right `http_proxy` would basically never be used (unless you're maybe talking to like dynamodb local or something similar).

You could in theory still connect to a proxy being served over HTTPS when trying to reach an HTTP URL, and you can in theory (and may want too) connect to an Proxy over plain HTTP when trying to connect to a HTTPS site.

  - See wget mentioning which should be used depending on the connection type *not the proxies connection type): https://www.gnu.org/software/wget/manual/html_node/Proxies.html
  - Curl's environment variables: https://everything.curl.dev/usingcurl/proxies/env
  - Java does the same with properties/variables: https://docs.oracle.com/javase/8/docs/technotes/guides/net/proxies.html

The reason you'd want a distinction between the two is because proxying to an HTTPS URL requires handling http `CONNECT` which forces a new connection, and is in general not the most widely supported things (it also means you're probably able to do less actual rewriting of traffic unless you have a cert on the clients machine or what have you, so you may have different behavior).

---

That being said some SDKs in AWS do what this PR does where the environment variable is properly based off of the scheme, some of them fallback to http proxy even on https urls, and some don't support it. I'm going for what is probably most universally "correct" here, and what _most_ of the SDKs do, connect depending on the scheme of the request. If you want me to fallback though like some of the other SDKs do I can change this behavior.

sdk/src/Core/Amazon.Runtime/_bcl/ClientConfig.cs
Mythra: Hey! I did use credentials, but this was awhile ago, and I don't think I tested on many dotNet versions. Do you know where the credentials are getting dropped perchance? I can try and take a look at parsing credentials there, and associating them.

sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_bcl/HttpWebRequestFactory.cs
dscpinheiro: Got it, thanks a lot for the explanation!

I understand it's frustrating that the SDKs are not consistent at the moment, but if we had to pick a behavior to "copy", I'd go with following what the CLI does (it was the original request and it's a very common setup for our customers to have both the .NET SDK and CLI in their environments).

Mythra: No problem at all! So the CLI is doing what we're doing here. It uses `http_proxy` for HTTP scheme requests, and `https_proxy` for HTTPS scheme requests. There is no fallback within the CLI. So we should be good to go!

sdk/src/Core/Amazon.Runtime/_bcl/ClientConfig.cs
Mythra: Got it thanks! Let me try with that exact setup, and see what's going on, and make a change & test for it. Appreciate the callout.

Mythra: Hey! So this did indeed slip past, I think what happened is I originally tested my proxy very early on in debug mode, which was allowing no credentials through for debugging purposes at the time.

I've gone ahead and updated it to actually parse out the credentials, and it does now seem to pass them on successfully to the proxy! So marking this as resolved.

dscpinheiro (approved):

sdk/src/Core/Amazon.Runtime/_bcl/IClientConfig.bcl.cs
```diff
@@ -70,5 +70,16 @@ public partial interface IClientConfig
...
+        /// Returns a WebProxy instance to use for HTTPS connections if an
+        /// explicit web proxy hasn't been configured.
+        /// </summary>
+        WebProxy GetHttpsProxy();
+
+        /// <summary>
+        /// Returns a WebProxy instance to use for HTTP connections if an
+        /// explicit web proxy hasn't been configured.
+        /// </summary>
+        WebProxy GetHttpProxy();
```
boblodgett, line 83: Is there a reason these shouldn't return IWebProxy and move all the common GetHttpsProxy/GetHttpProxy code to the non _bcl/netstandard specific partial classes? I think there may have been breaking change concerns but wouldn't it only be breaking if GetWebProxy was changed? I haven't traced all the code so it is possible this must return the utils type of WebProxy.

Mythra: Hey @boblodgett , frankly I'm not sure; I believe you are correct that in theory it could be done, but I haven't done all that tracing myself (and I'm certainly far from being an expert on the aws dotnet sdk).

If we did want to do this I think that'd be fine -- but I'd suggest maybe we do it in another follow up PR that can be rolled out separately? Keep this change focused on just the environment variables (so not only do we get our rollout which we desperately need! but the chance of it blowing up anyone goes down until we can do all that tracing).

Presumably it was done this way for a reason, and if we want to change it that's not really my call, but I'd want to do it as it's own distinct change. I wanna have as close to the original as possible.

sdk/src/Core/Amazon.Runtime/_netstandard/ClientConfig.cs
```diff
@@ -37,6 +37,25 @@ private static RegionEndpoint GetDefaultRegionEndpoint()
...
+                var asUri = new Uri(value);
+                var proxy = new Amazon.Runtime.Internal.Util.WebProxy(asUri);
+                if (!string.IsNullOrEmpty(asUri.UserInfo)) {
+                    var userAndPass = asUri.UserInfo.Split(':');
+                    proxy.Credentials = new NetworkCredential(
+                        userAndPass[0],
+                        userAndPass.Length > 1 ? userAndPass[1] : string.Empty
+                    );
+                }
+                return proxy;
```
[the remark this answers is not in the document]
Mythra: Ah I had missed that! I'll rename it!

sdk/src/Core/Amazon.Runtime/_netstandard/ClientConfig.cs
```diff
@@ -46,6 +65,34 @@ public IWebProxy GetWebProxy()
             return proxy;
         }
 
+        /// <summary>
+        /// Returns a WebProxy instance to use for HTTPS connections if an
+        /// explicit web proxy hasn't been configured.
+        /// </summary>
```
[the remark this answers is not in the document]
Mythra: Hey! You're right we probably want to flip these if proxy is set return that, otherwise return `http_proxy` -- I'll go ahead and do that.

---

As for testing I'd totally agree that'd be great, unfortunately I'm not aware of anyway to mock out environment variables in the testing frameworks for only one unit tests. I'd be happy to add them (i'd like more tests in other areas tbf, having to just set the environment variable and running the full integration test suite is unfortunate as your golden pass), but I did search for quite awhile and wasn't able to find a way to do that. Did you have a particular testing strategy in mind?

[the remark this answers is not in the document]
boblodgett: If we would have designed the SDK from the start to allow the environment variables to be mocked it would have been better and that is something we should keep in mind for the future but in the current state we have used this pattern in the tests:

```
var savedEnvVar = Environment.GetEnvironmentVariable("env_var_name");
try
{
   Environment.SetEnvironmentVariable("env_var_name", "some value");
   //Do the test code here

}
finally
{
   Environment.SetEnvironmentVariable("env_var_name", savedEnvVar);
}
```

It saves the value whatever it may be, sets it to the value needed for the test in the try, and then in the finally restores the environment variable to whatever it was before. There are examples of this in this file: https://github.com/aws/aws-sdk-net-staging/blob/6bac6ac7750f63cef6dbf3158b5d32cc9cbde737/sdk/test/UnitTests/Custom/Runtime/ServiceSpecificEndpointTests.cs#L158

sdk/src/Core/Amazon.Runtime/_bcl/IClientConfig.bcl.cs
boblodgett: We wouldn't be able to change it in a follow-up for the new methods because it would change the return type from WebProxy to IWebProxy which would then be a breaking change. @normj what direction do you think we should go with this?

sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_bcl/HttpWebRequestFactory.cs
```diff
@@ -474,6 +474,14 @@ public virtual void ConfigureRequest(IRequestContext requestContext)
...
+                _request.Proxy = requestContext.ClientConfig.GetHttpProxy();
+            }
+            else if (_request.RequestUri.Scheme == Uri.UriSchemeHttps)
```
normj, line 481: What is the point of putting in the code for HTTPS proxy with .NET Framework when .NET Framework doesn't support it. Users would get errors `The ServicePointManager does not support proxies with the https scheme.'`.

What I'm really worried is existing users running on machines that happened to have `HTTPS_PROXY` env var set and were being successfully ignored by their .NET Framework applications now start failing because we now attempt using `HTTPS_PROXY` on a platform that doesn't support it.

I think for .NET Framework we should only look for the `HTTP_PROXY` environment variable. But then that makes this kind of pointless since it is highly unlikely we would be going against an HTTP endpoint. Should we support the `ALL_PROXY` environment variable that would all using env var to configure an HTTP proxy for an HTTPS request?

normj (changes requested):

sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_bcl/HttpWebRequestFactory.cs
Mythra: To be clear -- `HTTPS_PROXY` the environment variable doesn't have to connect to a proxy being served over HTTPS (what dotNET doesn't support until the release that just happened). Because while the .NET framework doesn't support connecting to a proxy being served over HTTPS, you can have a environment variable like: `export https_proxy="http://user:pass@localhost:8556/proxy"`, which uses HTTP to then establish an HTTPS tunnel through the proxy which is fully supported in all versions of dotnet (in fact it's what I'm trying to do!) I do also want to mention in dotnet 8 and later you should be able to connect to a proxy being served over HTTPS.

I think if users were setting `HTTPS_PROXY`, and it wasn't going through the proxy that'd be more unexpected, again this is what other SDKs do, it's not like we're adding unexpected behavior here. It's not really a variable people just set randomly in my experience? And again it does still have a chance of working! It'll work in all cases if they're using dotNET 8, or later.

I'm against adding support for ALL_PROXY as that is not widely supported by the other SDKs like `http_proxy`, and `https_proxy` are. I don't want to have to juggle yet another environment variable that no one else is using.

normj: That is a good point I missed that HTTPS_PROXY doesn't have to point to a proxy using HTTPS.

sdk/src/Core/Amazon.Runtime/_netstandard/ClientConfig.cs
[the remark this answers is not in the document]
Mythra: I've gone ahead and added these tests, and swapped the logic to be explicit first like the comment mentioned. Resolving for now. Feel free to re-open if there's more here!

sdk/src/Core/Amazon.Runtime/_netstandard/ClientConfig.cs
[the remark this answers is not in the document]
Mythra: Renamed! Resolving for now, feel free to re-open if need be!

sdk/src/Core/Amazon.Runtime/_bcl/IClientConfig.bcl.cs
Mythra: Not sure if we still want to change this, but i've pushed it up without this change (keeping the same patterns we've had for now), if we do wanna change this just lemme know.

normj (changes requested):

sdk/src/Core/Amazon.Runtime/_bcl/IClientConfig.bcl.cs
Mythra: Unfortunately, I don't think we can do it the way you've specified here. In BCL we end up passing the result of `GetHttpsProxy`/`GetHttpProxy` to
`AmazonSecurityTokenServiceClient` which ends up passing the proxy to `SAMLAuthenticationController` which has the following block of code:

```c#
#if NETSTANDARD
        public IWebProxy ProxySettings { get; private set; }
#else
        public WebProxy ProxySettings { get; private set; }
#endif
```

And a `IWebProxy` can't be assigned to a `WebProxy` if we're not using NETSTANDARD (this also shows up in quite a few other places like `IAuthenticationController` -- which the SAML controller passes the proxy to as well). And I think changing this makes it a breaking change if we want to return `IWebProxy`, and change all the argument types of the underlying values being passed around.

I do also want to mention that BCL doesn't have: `Amazon.Runtime.Internal.Util.WebProxy` defined, so the types between BCL, and not would end up being two different `WebProxy`'s. Overall I think we'd end up in a very similar place, and it'd make this change significantly larger/more likely to break as we have to go change a whole bunch of types.

normj: Thanks for the reminder about the different implementations of `WebProxy`. I remember now the original target of the SDK to support .NET Core was .NET Standard 1.3 which didn't have `WebProxy` just the interface `IWebProxy`. That triggered the difference in signature return types. We can't do anything about that till we do a major version bump. You can still do the request change moving the new methods into the base partial with some conditional if checks.

```
#if BCL
        private static WebProxy GetWebProxyWithCredentials(string value)
#else
        private static Amazon.Runtime.Internal.Util.WebProxy? GetWebProxyWithCredentials(string value)
#endif
        {
            if (!string.IsNullOrEmpty(value))
            {
                var asUri = new Uri(value);
#if BCL
                var parsedProxy =  new WebProxy(asUri);
#else
                var parsedProxy =  new Amazon.Runtime.Internal.Util.WebProxy(asUri);
#endif
                if (!string.IsNullOrEmpty(asUri.UserInfo))
                {
                    var userAndPass = asUri.UserInfo.Split(':');
                    parsedProxy.Credentials = new NetworkCredential(
                        userAndPass[0],
                        userAndPass.Length > 1 ? userAndPass[1] : string.Empty
                    );
                }
                return parsedProxy;
            }

            return null;
        }

        /// <summary>
        /// Returns a WebProxy instance to use for HTTPS connections if an
        /// explicit web proxy hasn't been configured.
        /// </summary>
#if BCL
        public WebProxy GetHttpsProxy()
#else
        public IWebProxy GetHttpsProxy()
#endif
        {
            var explicitProxy = GetWebProxy();
            if (explicitProxy != null)
            {
                return explicitProxy;
            }
            return ClientConfig.GetWebProxyWithCredentials(Environment.GetEnvironmentVariable("https_proxy"));
        }

        /// <summary>
        /// Returns a WebProxy instance to use for HTTP connections if an
        /// explicit web proxy hasn't been configured.
        /// </summary>
#if BCL
        public WebProxy GetHttpProxy()
#else
        public IWebProxy GetHttpProxy()
#endif
        {
            var explicitProxy = GetWebProxy();
            if (explicitProxy != null)
            {
                return explicitProxy;
            }
            return ClientConfig.GetWebProxyWithCredentials(Environment.GetEnvironmentVariable("http_proxy"));
        }
```

Since we want essentially the same behavior for both .NET Framework and .NET Core it is worth putting the logic in the shared location with the specific conditional checks as opposed to forking the logic into separate platform files where the implementation could drift over time.

ashishdhingra:
@Mythra Could you please have a look at review comments and help address these in order to proceed further.

Mythra:
Hey @ashishdhingra yep, I saw the comments and will get to it, but it's also the holidays and I'm otherwise preoccupied at the moment. I'll get to it as soon as I can as it is really important I've been watching it like a hawk. Thanks!

sdk/src/Core/Amazon.Runtime/_bcl/IClientConfig.bcl.cs
Mythra: I've gone ahead, and pushed up this change! Marking as resolved for now.

Mythra:
I've come back from holidays, and made the changes!

normj (approved):
Looks good, thanks for making the change and the PR!

boblodgett (approved):

dscpinheiro:
Thanks again for the contribution, we really appreciate it!

This will be included in the next SDK release (which should happen tomorrow).

PR #3001, opened 2023-07-13
Add DynamoDBContextConfig.DisableFetchingTableMetadata to skip the internal DescribeTable call if the index structure is provided statically

ashovlin:
## Description
This is a proposal to _partially_ address https://github.com/aws/aws-sdk-net/issues/1476 if using the object persistence model with a fully annotated class.

Currently both of our higher-level DynamoDB programming models rely on knowledge of the table's index structure which is cached inside the SDK via a `DescribeTable` call.
* For the object mapper, this is done lazily when the first DynamoDB operation is invoked on `DynamoDBContext`
* For the document model, this is more explicit via `Table.LoadTable`

In versions of .NET that only have an async HTTP client, this `DescribeTable` is made via the sync-over-async antipattern becayse we've always carried forward the public sync APIs for the high-level programming models. This is leading to threadpool exhaustion and deadlocks in some scenarios.

As @normj discussed in https://github.com/aws/aws-sdk-net/issues/1476#issuecomment-567221982, one path is to go async-only, but we'd have to do that in a new major version since we currently expose public sync APIs.

Another idea is to remove the need for the `DescribeTable` call if the table structure is provided statically in advance. This PR explores this idea for the object-persistence model.

1. Adds a new option `AWSConfigsDynamoDB.Context.DisableFetchingTableMetadata` which defaults to false.
2. Adds a new option `DynamoDBContextConfig.DisableFetchingTableMetadata`, which defaults to false.
3. If this is set to `true`, when populating the internal SDK cache of table metadata for a given table we only rely on the information from:
    1. The [`DynamoDBAttribute` attributes](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DeclarativeTagsList.html) applied to the .NET class that is mapped to to the table
    2. The global [`AWSConfigsDynamoDB`](https://docs.aws.amazon.com/sdkfornet/v3/apidocs/items/Amazon/TAWSConfigsDynamoDB.html) object.

The SDK fills the cache using that static data, then proceeds as if it has made the `DescribeTable` call. 

## Testing
* Existing DynamoDB test cases pass
* Added new automated tests that verify the following scenarios:
  *  DynamoDB operations can be invoked successfully based on the table info supplied only via class attributes.
  *  DynamoDB operations can be invoked successfully based on the table info supplied via a combination of class attributes and the app config.
  *  DynamoDB operations can be invoked successfully based on the table info supplied via attributes on the parent class.
  *  DynamoDB operations can be invoked successfully based on a Datetime attribute as the hash key that is stored as epoch.
  *  The table info is cached using the actual table name and the key/index info is generated correctly based on DynamoDB attributes.
  * Setting the global `AWSConfigsDynamoDB.Context.DisableFetchingTableMetadata` to true propogates to the local `DynamoDBContextConfig`.

There is a separate backog item (DOTNET-7140) to investigate failure modes when the table info provided via class attributes does not match the actual table.

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/AWSConfigs.cs
+++ b/sdk/src/Core/AWSConfigs.cs
@@ -45,7 +45,7 @@ namespace Amazon
     ///   &lt;proxy host="localhost" port="8888" username="1" password="1" /&gt;
     ///   
     ///   &lt;dynamoDB&gt;
-    ///     &lt;dynamoDBContext tableNamePrefix="Prod-" metadataCachingMode="Default"&gt;
+    ///     &lt;dynamoDBContext tableNamePrefix="Prod-" metadataCachingMode="Default" disableFetchingTableMetadata="false"&gt;
     /// 
     ///       &lt;tableAliases&gt;
     ///         &lt;alias fromTable="FakeTable" toTable="People" /&gt;
--- a/sdk/src/Services/DynamoDBv2/Custom/AWSConfigs.DynamoDB.cs
+++ b/sdk/src/Services/DynamoDBv2/Custom/AWSConfigs.DynamoDB.cs
@@ -2,8 +2,6 @@
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.IO;
-using System.Linq;
-using System.Text;
 
 using System.Xml.Linq;
 
@@ -15,6 +13,8 @@ using System.Configuration;
 using Amazon.Util;
 using Amazon.Util.Internal;
 using Amazon.DynamoDBv2;
+using Amazon.DynamoDBv2.Model;
+using Amazon.DynamoDBv2.DataModel;
 
 namespace Amazon
 {
@@ -46,7 +46,7 @@ namespace Amazon
                 // configure the context config to the default.
                 if (Context == null)
                 {
-                    Context = new DynamoDBContextConfig();
+                    Context = new Util.DynamoDBContextConfig();
                     ConversionSchema = ConversionSchema.V1;;
                 }
             }
@@ -92,12 +92,12 @@ namespace Amazon
         /// <summary>
         /// Settings for DynamoDBContext.
         /// </summary>
-        public static DynamoDBContextConfig Context { get; private set; }
+        public static Util.DynamoDBContextConfig Context { get; private set; }
 
 #if BCL
         internal static void Configure(DynamoDBSection section)
         {
-            Context = new DynamoDBContextConfig();
+            Context = new Util.DynamoDBContextConfig();
             ConversionSchema = ConversionSchema.V1;
 
             if (section != null && section.ElementInformation.IsPresent)
@@ -156,6 +156,17 @@ namespace Amazon.Util
         public MetadataCachingMode? MetadataCachingMode { get; set; }
 
         /// <summary>
+        /// If true disables fetching table metadata automatically from DynamoDB. Table metadata must be 
+        /// defined by <see cref="DynamoDBAttribute"/> attributes and/or in <see cref = "AWSConfigsDynamoDB"/>.
+        /// </summary>
+        /// <remarks>
+        /// Setting this to true can avoid latency and potential deadlocks due to the internal  <see cref="DescribeTableRequest"/> call that is used to populate the SDK's cache of 
+        /// table metadata. It requires that the table's index schema be fully described via the above methods, otherwise exceptions may be thrown and/or 
+        /// the results of certain DynamoDB operations may change. It is recommended to test your application prior to setting it to true in production code.
+        /// </remarks>
+        public bool? DisableFetchingTableMetadata { get; set; }
+
+        /// <summary>
         /// Adds a TableAlias to the TableAliases property.
         /// An exception is thrown if there is already a TableAlias with the same FromTable configured.
         /// </summary>
@@ -190,6 +201,7 @@ namespace Amazon.Util
             {
                 TableNamePrefix = section.TableNamePrefix;
                 MetadataCachingMode = section.MetadataCachingMode;
+                DisableFetchingTableMetadata = section.DisableFetchingTableMetadata;
 
                 InternalSDKUtils.FillDictionary(section.TypeMappings.Items, t => t.Type, t => new TypeMapping(t), TypeMappings);
                 InternalSDKUtils.FillDictionary(section.TableAliases.Items, t => t.FromTable, t => t.ToTable, TableAliases);
@@ -408,6 +420,7 @@ namespace Amazon.Util
         private const string tableAliasesKey = "tableAliases";
         private const string mappingsKey = "mappings";
         private const string metadataCachingModeKey = "metadataCachingMode";
+        private const string disableFetchingTableMetadataKey = "disableFetchingTableMetadata";
 
         [ConfigurationProperty(tableNamePrefixKey)]
         public string TableNamePrefix
@@ -435,6 +448,13 @@ namespace Amazon.Util
         {
             get { return (MetadataCachingMode?)this[metadataCachingModeKey]; }
             set { this[metadataCachingModeKey] = value; }
+        }
+
+        [ConfigurationProperty(disableFetchingTableMetadataKey)]
+        public bool? DisableFetchingTableMetadata
+        {
+            get { return (bool?)this[disableFetchingTableMetadataKey]; }
+            set { this[disableFetchingTableMetadataKey] = value; }
         }
 
     }
--- a/sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs
+++ b/sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs
@@ -61,6 +61,7 @@ namespace Amazon.DynamoDBv2.DataModel
             TableNamePrefix = AWSConfigsDynamoDB.Context.TableNamePrefix;
             Conversion = DynamoDBEntryConversion.CurrentConversion;
             MetadataCachingMode = AWSConfigsDynamoDB.Context.MetadataCachingMode;
+            DisableFetchingTableMetadata = AWSConfigsDynamoDB.Context.DisableFetchingTableMetadata;
         }
 
         /// <summary>
@@ -117,6 +118,17 @@ namespace Amazon.DynamoDBv2.DataModel
         /// .NET and DynamoDB types happens.
         /// </summary>
         public DynamoDBEntryConversion Conversion { get; set; }
+
+        /// <summary>
+        /// If true disables fetching table metadata automatically from DynamoDB. Table metadata must be 
+        /// defined by<see cref="DynamoDBAttribute"/> attributes and/or in <see cref = "AWSConfigsDynamoDB"/>.
+        /// </summary>
+        /// <remarks>
+        /// Setting this to true can avoid latency and potential deadlocks due to the internal  <see cref="DescribeTableRequest"/> call that is used to populate the SDK's cache of 
+        /// table metadata. It requires that the table's index schema be fully described via the above methods, otherwise exceptions may be thrown and/or 
+        /// the results of certain DynamoDB operations may change. It is recommended to test your application prior to setting it to true in production code.
+        /// </remarks>
+        public bool? DisableFetchingTableMetadata { get; set; }
     }
 
     /// <summary>
@@ -320,6 +332,8 @@ namespace Amazon.DynamoDBv2.DataModel
             bool consistentRead = operationConfig.ConsistentRead ?? contextConfig.ConsistentRead ?? false;
             bool skipVersionCheck = operationConfig.SkipVersionCheck ?? contextConfig.SkipVersionCheck ?? false;
             bool ignoreNullValues = operationConfig.IgnoreNullValues ?? contextConfig.IgnoreNullValues ?? false;
+            bool disableFetchingTableMetadata = contextConfig.DisableFetchingTableMetadata ?? false;
+
             bool isEmptyStringValueEnabled = operationConfig.IsEmptyStringValueEnabled ?? contextConfig.IsEmptyStringValueEnabled ?? false;
             string overrideTableName =
                 !string.IsNullOrEmpty(operationConfig.OverrideTableName) ? operationConfig.OverrideTableName : string.Empty;
@@ -346,6 +360,7 @@ namespace Amazon.DynamoDBv2.DataModel
             ConditionalOperator = conditionalOperator;
             Conversion = conversion;
             MetadataCachingMode = metadataCachingMode;
+            DisableFetchingTableMetadata = disableFetchingTableMetadata;
 
             State = new OperationState();
         }
@@ -434,6 +449,9 @@ namespace Amazon.DynamoDBv2.DataModel
         /// .NET and DynamoDB types happens.
         /// </summary>
         public DynamoDBEntryConversion Conversion { get; set; }
+
+        /// <inheritdoc cref="DynamoDBContextConfig.DisableFetchingTableMetadata"/>
+        public bool DisableFetchingTableMetadata { get; set; }
 
         // Checks if the IndexName is set on the config
         internal bool IsIndexOperation { get { return !string.IsNullOrEmpty(IndexName); } }
--- a/sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextInternal.cs
+++ b/sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextInternal.cs
@@ -136,7 +136,7 @@ namespace Amazon.DynamoDBv2.DataModel
                 throw new ArgumentNullException("flatConfig");
 
             string tableName = GetTableName(storageConfig.TableName, flatConfig);
-            var unconfiguredTable = GetUnconfiguredTable(tableName);
+            var unconfiguredTable = GetUnconfiguredTable(tableName, flatConfig.DisableFetchingTableMetadata);
             ValidateConfigAgainstTable(storageConfig, unconfiguredTable);
 
             var tableConfig = new TableConfig(tableName, flatConfig.Conversion, consumer,
@@ -149,7 +149,7 @@ namespace Amazon.DynamoDBv2.DataModel
         // Retrieves Config-less Table from cache or constructs it on cache-miss
         // This Table should not be used for data operations.
         // To use for data operations, Copy with a TableConfig first.
-        internal Table GetUnconfiguredTable(string tableName)
+        internal Table GetUnconfiguredTable(string tableName, bool disableFetchingTableMetadata = false)
         {
             Table table;
             
@@ -174,12 +174,18 @@ namespace Amazon.DynamoDBv2.DataModel
             {
                 _readerWriterLockSlim.EnterWriteLock();
                 
-                // Check to see if another thread go the write lock before this thread and filled the cache.
+                // Check to see if another thread got the write lock before this thread and filled the cache.
                 if (tablesMap.TryGetValue(tableName, out table))
                 {
                     return table;
                 }
 
+                
+                if (disableFetchingTableMetadata)
+                {
+                    return null;
+                }
+                
                 var emptyConfig = new TableConfig(tableName, conversion: null, consumer: Table.DynamoDBConsumer.DataModel,
                     storeAsEpoch: null, isEmptyStringValueEnabled: false, metadataCachingMode: Config.MetadataCachingMode);
                 table = Table.LoadTable(Client, emptyConfig);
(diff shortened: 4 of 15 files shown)


sdk/src/Services/DynamoDBv2/Custom/DocumentModel/Table.cs
```diff
@@ -413,6 +428,155 @@ public static Table LoadTable(IAmazonDynamoDB ddbClient, TableConfig config)
...
+        {
+            // Get the converter that would be used for this property
+            flatConfig.Conversion.TryGetConverter(property.MemberType, out Converter converter);
+
+            // Converters can only convert an actual value, we can't just look up the resulting type 
+            // so create a default value and attempt to convert it
+            object hypotheticalValue;
+            if (property.MemberType == typeof(string))
+            {
+                hypotheticalValue = ""; // Activator.CreateInstance throws for string due to no default constructor
```
vitaliypenkov, line 447:
```suggestion
                hypotheticalValue = string.Empty; // Activator.CreateInstance throws for string due to no default constructor
```

sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs
```diff
@@ -104,6 +104,17 @@ public DynamoDBContextConfig()
...
+        /// <summary>
+        /// Skip the internal DescribeTable call that caches table metadata. Table metadata must be 
+        /// defined by <see cref="DataModel.DynamoDBAttribute"/> attributes and/or in <see cref="AWSConfigsDynamoDB"/>.
+        /// </summary>
+        /// <remarks>
+        /// Setting this to true can avoid latency and potential deadlocks due to the internal DescribeTable call that is used to populate the SDK's cache of 
+        /// table medata. It requires that the table's index schema be fully described via the above methods, otherwise exceptions may be thrown and/or 
+        /// the results of certain DynamoDB operations may change. It is recommended to test your application prior to setting it to true in production code.
+        /// </remarks>
+        public bool? SkipDescribeTableLoad { get; set; }
```
ashovlin, line 117: Curious what other reviewers think, but I'm still not too attached to `SkipDescribeTableLoad`. It might be too from-our-perspective/internal, since some users may not realize that that call is happening.

That said, not I don't have a better alternative.

sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextInternal.cs
```diff
@@ -179,6 +187,11 @@ internal Table GetUnconfiguredTable(string tableName)
                     return table;
                 }
 
+                if (skipDescribeTableLoad)
```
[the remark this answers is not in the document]
96malhar: Fixed the type on line 184. 

We acquire the read lock to just retrieve the table metadata from the `tablesMap` cache. It does not involve making any service calls like `DescribeTable`. Hence, we do not have this check in the read lock.

sdk/src/Services/DynamoDBv2/Custom/DocumentModel/Table.cs
```diff
@@ -413,6 +428,160 @@ public static Table LoadTable(IAmazonDynamoDB ddbClient, TableConfig config)
...
+            // Hash Key
+            // 
+            foreach (var hashKeyPropertyName in itemStorageConfig.HashKeyPropertyNames)
+            {
+                var property = itemStorageConfig.GetPropertyStorage(hashKeyPropertyName);
+                var primitiveType = GetPrimitiveEntryTypeForProperty(property, flatConfig);
+
+                table.HashKeys.Add(property.AttributeName);
+
+                table.Keys.Add(property.AttributeName, new KeyDescription
```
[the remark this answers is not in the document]
96malhar: It is invalid to have multiple properties with the same `AttributeName`

sdk/src/Services/DynamoDBv2/Properties/AssemblyInfo.cs
```diff
@@ -9,8 +9,10 @@
 [assembly: AssemblyTitle("AWSSDK.DynamoDBv2")]
 #if BCL35
 [assembly: AssemblyDescription("The Amazon Web Services SDK for .NET (3.5) - Amazon DynamoDB. Amazon DynamoDB is a fast and flexible NoSQL database service for all applications that need consistent, single-digit millisecond latency at any scale.")]
+[assembly: InternalsVisibleTo("AWSSDK.UnitTests.DynamoDBv2.Net35")]
```
ashovlin, line 12: I believe this file is generated so worried that this will be wiped out.
https://github.com/aws/aws-sdk-net/blob/master/generator/ServiceClientGeneratorLib/Generators/SourceFiles/AssemblyInfo.tt

What test case is this unlocking? I do generally fall in line with the _"test `public` surface area"_`, though not absolutely and acknowledge that it can be challenging/contentious.

ashovlin: Plus locally this is failing for me anyway with
```
Severity	Code	Description	Project	File	Line	Suppression State
Error	CS1726	Friend assembly reference 'AWSSDK.UnitTests.DynamoDBv2.Net35' is invalid. Strong-name signed assemblies must specify a public key in their InternalsVisibleTo declarations.	AWSSDK.DynamoDBv2.Net35	C:\dev\AWSDotNetPublic\sdk\src\Services\DynamoDBv2\Properties\AssemblyInfo.cs	12	Active
```

sdk/test/Services/DynamoDBv2/UnitTests/Custom/DynamoDBTests.cs
```diff
@@ -328,6 +328,30 @@ public void TestVersionAttributeRename()
...
+            var dynamoDBContext = new DynamoDBContext(dynamoDBClient.Object, dynamoDBContextConfig);
+            var dynamoDBFlatConfig = new DynamoDBFlatConfig(null, dynamoDBContextConfig);
+
+            var itemStorageConfig = dynamoDBContext.StorageConfigCache.GetConfig<Employee>(dynamoDBFlatConfig);
+
+            var table = dynamoDBContext.GetUnconfiguredTable("DotnetTest-EmployeeDetails", skipDescribeTableLoad: true);
+            Assert.IsNotNull(table);
+            Assert.AreEqual("DotnetTest-EmployeeDetails", table.TableName);
+
+            table = dynamoDBContext.GetTargetTable(itemStorageConfig, dynamoDBFlatConfig);
```
ashovlin, line 350: 1. Related to the above comment about friend assemblies, could we just use `dynamoDBContext.GetTargetTable<Employee>();` since that's already public? It's unclear from the test name why we want to test the internal methods, comments may help.
2. Can we assert the full contents of the `Table` object we get out? Don't think we're doing that explicitly anywhere (though the integ tests are making use of it).

sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs
96malhar: Deferring to @normj for the naming scheme

sdk/src/Services/DynamoDBv2/Properties/AssemblyInfo.cs
96malhar: I was using the to called the internal `GetUnconfiguredTable` and verify that that table info is cached under the actual table name. However, I have removed this and instead resorted to asserting on the table metadata that is returned via the public `GetTargetTable`.

sdk/test/Services/DynamoDBv2/UnitTests/Custom/DynamoDBTests.cs
96malhar: Done

philasmar (approved):

sdk/test/Services/DynamoDBv2/IntegrationTests/Config/35/App.config
```diff
@@ -36,6 +36,12 @@
             <property name="InternalId" ignore="true" />
             <property name="CurrentStatus" attribute="Status" converter="AWSSDK_DotNet.IntegrationTests.Tests.DynamoDB.DynamoDBTests+StatusConverter, AWSSDK.IntegrationTests.DynamoDBv2.Net35" />
           </map>
+          <map type="AWSSDK_DotNet.IntegrationTests.Tests.DynamoDB.DynamoDBTests+PartiallyAnnotatedEmployee, AWSSDK.IntegrationTests.DynamoDBv2.Net35" targetTable="HashRangeTable">
+            <property name="ManagerName" attribute="Manager" />
+            <property name="CompanyName" attribute="Company" />
+            <property name="InternalId" ignore="true" />
+            <property name="CurrentStatus" attribute="Status" converter="AWSSDK_DotNet.IntegrationTests.Tests.DynamoDB.DynamoDBTests+StatusConverter, AWSSDK.IntegrationTests.DynamoDBv2.Net45" />
```
ashovlin, line 43: A lot of .NET Framework 3.5 tests are failing for me with the following, I think related to the `Net45` reference here.

```
Test method AWSSDK_DotNet.IntegrationTests.Tests.CacheTests.ChangingTableTest threw exception: 
System.TypeInitializationException: The type initializer for 'Amazon.AWSConfigsDynamoDB' threw an exception. ---> System.Configuration.ConfigurationErrorsException: The value of the property 'converter' cannot be parsed. The error is: The type 'AWSSDK_DotNet.IntegrationTests.Tests.DynamoDB.DynamoDBTests+StatusConverter, AWSSDK.IntegrationTests.DynamoDBv2.Net45' cannot be resolved. Please verify the spelling is correct or that the full type name is provided.
```

sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs
```diff
@@ -305,6 +316,8 @@ public DynamoDBFlatConfig(DynamoDBOperationConfig operationConfig, DynamoDBConte
             bool consistentRead = operationConfig.ConsistentRead ?? contextConfig.ConsistentRead ?? false;
             bool skipVersionCheck = operationConfig.SkipVersionCheck ?? contextConfig.SkipVersionCheck ?? false;
             bool ignoreNullValues = operationConfig.IgnoreNullValues ?? contextConfig.IgnoreNullValues ?? false;
+            bool skipDescribeTableLoad = contextConfig.SkipDescribeTableLoad ?? false;
```
ashovlin, line 319: I didn't think about this until after we did it over in https://github.com/aws/aws-sdk-net/pull/3025/files#diff-31f37a1ab8a8a2088f053530284e3276ff31440c8cb042c62a11286006d89cc4, but could we add this to `DynamoDBContextConfig` so users can set it globally?

sdk/src/Services/DynamoDBv2/Custom/DocumentModel/Table.cs
```diff
@@ -125,6 +125,21 @@ private static DynamoDBEntryType GetType(string attributeType)
...
+            switch (primitiveEntryType)
+            {
+                case DynamoDBEntryType.String:
+                    return ScalarAttributeType.S;
+                case DynamoDBEntryType.Numeric:
+                    return ScalarAttributeType.N;
+                case DynamoDBEntryType.Binary:
+                    return ScalarAttributeType.B;
+                default:
+                    throw new ArgumentOutOfRangeException($"{primitiveEntryType} is not a known DynamoDB {nameof(ScalarAttributeType)}");
```
ashovlin, line 139: I believe the single `string` signature is expecting the name of the invalid parameter, not a message.

We probably want something like:

```suggestion
                    throw new ArgumentOutOfRangeException(nameof(primitiveEntryType), $"{primitiveEntryType} is not a known DynamoDB {nameof(ScalarAttributeType)}");
```

sdk/test/Services/DynamoDBv2/IntegrationTests/DataModelTests.cs
```diff
@@ -73,6 +73,92 @@ public void TestContextWithEmptyStringDisabled()
...
+        [TestCategory("DynamoDBv2")]
+        public void TestContext_DescribeTableDisabled_WithFullAttributes()
+        {
+            TableCache.Clear();
+            CleanupTables();
+            TableCache.Clear();
+
+            CreateContext(DynamoDBEntryConversion.V2, true, true);
+
+            TestHashRangeObjects<AnnotatedEmployee>();
```
ashovlin, line 90: Are any other  "sub"-tests from `TestContextWithEmptyStringEnabled` relevant here?

I had only grabbed this one during the prototype since it looked like a good place to start, but I didn't think about the rest.

sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextInternal.cs
```diff
@@ -136,7 +136,15 @@ private Table GetTargetTableInternal<T>(DynamoDBOperationConfig operationConfig)
...
             string tableName = GetTableName(storageConfig.TableName, flatConfig);
-            var unconfiguredTable = GetUnconfiguredTable(tableName);
+            var unconfiguredTable = GetUnconfiguredTable(tableName, flatConfig.SkipDescribeTableLoad);
+
+            if (unconfiguredTable == null && flatConfig.SkipDescribeTableLoad)
+            {
+                // In the case where we want to avoid the DescribeTable call, but failed to find a cached Table
+                // check the original table name in case a prefix has been applied or overriden
+                // This is making the assumption that storageConfig.TableName has the same structure as tableName
+                unconfiguredTable = GetUnconfiguredTable(storageConfig.TableName, flatConfig.SkipDescribeTableLoad);
```
ashovlin, line 146: I think I wrote this when prototyping, but I'm still struggling to understand why. 
1. What user code will hit this code path?
2. Commenting out didn't cause any tests to fail, so could you add a test guarding that this fallback logic remains in place?

sdk/test/Services/DynamoDBv2/IntegrationTests/Config/35/App.config
96malhar: Fixed

sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs
96malhar: Added this to the global `DynamoDBContextConfig`. Updated the PR description to reflect this change and also added a unit test.

sdk/src/Services/DynamoDBv2/Custom/DocumentModel/Table.cs
96malhar: Fixed

sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextInternal.cs
96malhar: I removed this branching since this code path will never be hit. This is because we are always caching the table info using the actual table name irrespective of whether `SkipDescribeTableLoad` is enabled or disabled.

sdk/test/Services/DynamoDBv2/IntegrationTests/DataModelTests.cs
96malhar: I took a look at these tests and did not come across a use case that will potentially be affected via the addition of the `SkipDescribeTableLoad` flag.

normj (changes requested):

sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextInternal.cs
```diff
@@ -196,6 +202,39 @@ internal Table GetUnconfiguredTable(string tableName)
...
+        internal Table StoreUnconfiguredTable(string tableName, Table tableToStore)
+        {
+            Table table;
+
+            try
+            {
+                _readerWriterLockSlim.EnterWriteLock();
+
+                // Check to see if the cache was filled since this was invoked
+                if (tablesMap.TryGetValue(tableName, out table))
```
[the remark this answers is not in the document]
96malhar: Fixed

sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs
96malhar: Renamed the property to `DisableFetchingTableMetadata`

sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextInternal.cs
```diff
@@ -196,6 +202,49 @@ internal Table GetUnconfiguredTable(string tableName)
...
+                {
+                    _readerWriterLockSlim.ExitReadLock();
+                }
+            }
+
+            try
+            {
+                _readerWriterLockSlim.EnterWriteLock();
+
+                tablesMap.Add(tableName, tableToStore);
```
normj, line 235: Now that you have the write lock you need to double check that another thread didn't get the lock before you and add the table to `tablesMap`.

sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs
```diff
@@ -117,6 +118,17 @@ public DynamoDBContextConfig()
         /// .NET and DynamoDB types happens.
         /// </summary>
         public DynamoDBEntryConversion Conversion { get; set; }
+
+        /// <summary>
+        /// Skip the internal <see cref="DescribeTableRequest"/> call that caches table metadata. Table metadata must be
```
normj, line 123: Replace with 
```
If true disables fetching table metadata automatically from DynamoDB. Table metadata must be 
defined by <see cref="DynamoDBAttribute"/> attributes and/or in <see cref="AWSConfigsDynamoDB"/>.
```

sdk/src/Services/DynamoDBv2/Custom/AWSConfigs.DynamoDB.cs
```diff
@@ -155,6 +155,17 @@ public partial class DynamoDBContextConfig
         /// </remarks>
         public MetadataCachingMode? MetadataCachingMode { get; set; }
 
+        /// <summary>
+        /// Skip the internal <see cref="DescribeTableRequest"/> call that caches table metadata. Table metadata must be
```
normj, line 159: Replace with 
```
If true disables fetching table metadata automatically from DynamoDB. Table metadata must be 
defined by <see cref="DynamoDBAttribute"/> attributes and/or in <see cref="AWSConfigsDynamoDB"/>.
```

normj (changes requested):

sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextInternal.cs
96malhar: Done

sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs
96malhar: Done

sdk/src/Services/DynamoDBv2/Custom/AWSConfigs.DynamoDB.cs
96malhar: Done

normj (approved):

sdk/src/Services/DynamoDBv2/Custom/AWSConfigs.DynamoDB.cs
```diff
@@ -155,6 +155,17 @@ public partial class DynamoDBContextConfig
         /// </remarks>
         public MetadataCachingMode? MetadataCachingMode { get; set; }
 
+        /// <summary>
+        /// If true disables fetching table metadata automatically from DynamoDB. Table metadata must be 
+        /// defined by<see cref="DynamoDBAttribute"/> attributes and/or in <see cref = "AWSConfigsDynamoDB"/>.
```
ashovlin, line 160: nit: Missing space between "by" and "DynamoDBAttribute"

ashovlin:
I can't formally approve since I originally created this, but LGTM. Two small nits related to docs (one I forgot to publish last week), but neither are dealbreakers.

sdk/src/Services/DynamoDBv2/Custom/AWSConfigs.DynamoDB.cs
96malhar: Fixed

sdk/src/Core/AWSConfigs.cs
```diff
@@ -45,7 +45,7 @@ namespace Amazon
     ///   &lt;proxy host="localhost" port="8888" username="1" password="1" /&gt;
     ///   
     ///   &lt;dynamoDB&gt;
-    ///     &lt;dynamoDBContext tableNamePrefix="Prod-" metadataCachingMode="Default"&gt;
+    ///     &lt;dynamoDBContext tableNamePrefix="Prod-" metadataCachingMode="Default" skipDescribeTableLoad="true"&gt;
```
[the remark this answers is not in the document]
96malhar: Done

96malhar:
Squashed and merged into the feature branch `feature/ddb-skip-describe-table`

PR #3010, opened 2023-07-28
Fix remaining Core AOT JSON warnings

normj:
## Description
Fix remaining JSON AOT warnings in Core. This involved moving direct usage of LitJson to the new JsonSerializerHelper abstraction that switches between LitJson and System.Text.Json.

### Decisions made
* Mark `ClientContext` as incompatible for AOT due to its JSON usage over `object`. Could possibly rework the class but I don't think the class is even really used anymore. It is in an internal namespace and not referenced anywhere. There are some features in MobileAnalytics that allow you to set a client context JSON. In theory a user could still be using the class even though it is in an internal namespace to generate the JSON. The usage has to be really low so marking incompatible and wait to see if we get request to make it work.
* Mark `Document` as incompatible for AOT. This is a special service data type used for services that work with unstructured data. It is not used very much and the class's design is based on reflection over `object`. I'm willing for the first version of the SDK with AOT support to not have `Document` support and it later when there is demand.
* The `SettingsCollection` class used a `Dictionary<string, Dictionary<string, object>>` as its backing store. The use of `object` makes it impossible to work for AOT when it comes to JSON serialization. The class actually never used anything other than strings so I change the backing store to use `Dictionary<string, Dictionary<string, string>>`. The class is really old predating shared credentials file and I think we just used object for future extensibility but since we really don't even want to be using this settings system anymore, we don't need the extensibility.

## Motivation and Context
Effort to make the SDK compatible with Native AOT

## Testing
Added new JSON unit tests

## License

- [ x] I confirm that this pull request can be released under the Apache 2 license

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Internal/Settings/PersistenceManager.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Settings/PersistenceManager.cs
@@ -18,6 +18,7 @@
  *  AWS SDK for .NET
  */
 using Amazon.Runtime.Internal.Util;
+using Amazon.Util.Internal;
 using System;
 using System.Collections.Generic;
 using System.Globalization;
@@ -253,10 +254,18 @@ namespace Amazon.Runtime.Internal.Settings
                         content = reader.ReadToEnd();
                     }
 
-                    var settings = JsonMapper.ToObject<Dictionary<string, Dictionary<string, object>>>(content);
+                    Dictionary<string, Dictionary<string, string>> settings;
+                    if(!string.IsNullOrEmpty(content))
+                    {
+                        settings = JsonSerializerHelper.Deserialize<Dictionary<string, Dictionary<string, string>>>(content, DictionaryStringDictionaryStringJsonSerializerContexts.Default);
 
-                    if (settings == null)
-                        settings = new Dictionary<string, Dictionary<string, object>>();
+                        if (settings == null)
+                            settings = new Dictionary<string, Dictionary<string, string>>();
+                    }
+                    else
+                    {
+                        settings = new Dictionary<string, Dictionary<string, string>>();
+                    }
 
                     DecryptAnyEncryptedValues(settings);
 
@@ -275,7 +284,7 @@ namespace Amazon.Runtime.Internal.Settings
             }
         }
 
-        static void DecryptAnyEncryptedValues(Dictionary<string, Dictionary<string, object>> settings)
+        static void DecryptAnyEncryptedValues(Dictionary<string, Dictionary<string, string>> settings)
         {
             foreach (var kvp in settings)
             {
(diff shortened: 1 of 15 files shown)


sdk/src/Core/Amazon.Runtime/Internal/Settings/PersistenceManager.cs
```diff
@@ -253,10 +254,18 @@ SettingsCollection loadSettingsType(string type)
...
-                    var settings = JsonMapper.ToObject<Dictionary<string, Dictionary<string, object>>>(content);
+                    Dictionary<string, Dictionary<string, string>> settings;
+                    if(!string.IsNullOrEmpty(content))
+                    {
+                        settings = JsonSerializerHelper.Deserialize<Dictionary<string, Dictionary<string, string>>>(content, DictionaryStringDictionaryStringJsonSerializerContexts.Default);
 
-                    if (settings == null)
-                        settings = new Dictionary<string, Dictionary<string, object>>();
+                        if (settings == null)
+                            settings = new Dictionary<string, Dictionary<string, string>>();
```
boblodgett, line 263: nit: this could be done outside the if and scratch the else completely to remove the duplicate code.

boblodgett (approved):

sdk/src/Core/Amazon.Util/Internal/JsonSerializerHelper.cs
```diff
@@ -29,6 +48,53 @@ public partial class EC2InstanceMetadataJsonSerializerContexts : JsonSerializerC
...
+        public DictionaryStringDictionaryStringJsonSerializerContexts(JsonSerializerOptions defaultOptions)
+            : base(defaultOptions)
+        {
+        }
+#endif
+    }
+
+    [JsonSerializable(typeof(List<string>))]
+    public partial class ListStringJsonSerializerContexts : JsonSerializerContext
+    {
```
dscpinheiro, line 75: Any reason why only the `DictionaryString*` classes have a constructor with `JsonSerializerOptions`? Are those not required for the other contexts?

---

A separate question: Are we going to have to create context classes for every type that'll be serialized?

normj: It was the only one used with JSON pretty print or `WriteIndented` used when serializing. Most of our JSON use in Core is actually deserialization so there is no need for pretty print. If we find we need more with pretty print we can add constructors.

For your second question yes we always need a context class so we can have the source generated JSON marshalling code. I debated about having one context with a lot of `JsonSerializable` for each of the types. I tried something similar for the Lambda repository and ran into issues when the types referenced in the `JsonSerializable` attribute had overlapping subtypes which caused code generation issues.

dscpinheiro (approved):

Issue #1945, opened 2021-11-29
Failed to retrieve credentials from EC2 Instance Metadata Service is misleading 


jamiuaz:
## Description
When make [GetCredentials()](https://github.com/aws/aws-sdk-net/blob/3810f3340ed99b404d19b5b78136beebde3d6ba6/sdk/src/Core/Amazon.Runtime/Credentials/DefaultInstanceProfileAWSCredentials.cs#L81) while there is already existing request with [TryEnterWriteLock](https://github.com/aws/aws-sdk-net/blob/3810f3340ed99b404d19b5b78136beebde3d6ba6/sdk/src/Core/Amazon.Runtime/Credentials/DefaultInstanceProfileAWSCredentials.cs#L105), it fails with `Failed to retrieve credentials from EC2 Instance Metadata Service` instead should be something like `There is request to FetchCredentials in progress `

### Environment

* **SDK Version:** "AWSSDK.RDS" Version="3.7.6.8"
* **Package Version:** [AWSSDK.RDS](https://www.nuget.org/packages/AWSSDK.RDS/3.7.6.8)
* **OS Info:** EC2 - Microsoft Windows Server 2019 Datacenter 
* **Build Environment:**  Visual Studio 2019 
* **Targeted .NET Platform:**  .NET CORE 3.1

## Resolution

---

This is a :bug: bug-report

aaoswal:
Hi @jamiuaz,
Could you please share a sample code and reproduction steps? It would help analyze the issue faster.

Thanks,
Anish.

jamiuaz:
Hi @aaoswal 

To reproduce this 

1. I created a .net core console app.
2. Simulate the GetCredentials with TryEnterWriteLock.
3. Make concurrent call to GetCredentials with 2 threads: first processing delayed than second thread.

Code
`
using System;
using System.Threading;
using System.Threading.Tasks;

namespace IssueReproduce
{
    class DefaultInstanceProfileAWSCredentialsIssue
    {
        private static readonly TimeSpan credentialsLockTimeout = TimeSpan.FromSeconds(20);
        public ReaderWriterLockSlim credentialsLock = new ReaderWriterLockSlim(); // Lock to control getting credentials across multiple threads.

        public string GetCredentials(int delay)
        {
            string result = string.Empty;

            // If there's no credentials cached, hit IMDS directly. Try to acquire write lock.
            if (credentialsLock.TryEnterWriteLock(credentialsLockTimeout))
            {
                try
[...]

PR #3024, opened 2023-08-09
Add thread safety to FallbackCredentialsFactory

alitas:
## Description
This PR adds locking around the cached credentials and refrains from saving temporary credential state in a static variable to avoid race conditions.

## Motivation and Context
If called from multiple threads, the cached credentials in `FallbackCredentialsFactory` can be set to `null` temporarily for failing credential generators, and this can override successfully retrieved credentials from the other threads, causing false negative errors in retrieving credentials. 

Similarly, if the factory cache is reset while getting credentials, the `GetCredentials` method might throw an exception even when it has an actual credential.

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Credentials/FallbackCredentialsFactory.cs
+++ b/sdk/src/Core/Amazon.Runtime/Credentials/FallbackCredentialsFactory.cs
@@ -20,12 +20,16 @@ using System.Globalization;
 using System.IO;
 using System.Net;
 using System.Security;
+using System.Threading;
 
 namespace Amazon.Runtime
 {
     // Credentials fallback mechanism
     public static class FallbackCredentialsFactory
     {
+        // Lock to control caching credentials across multiple threads.
+        private static ReaderWriterLockSlim cachedCredentialsLock = new ReaderWriterLockSlim();
+    
         internal const string AWS_PROFILE_ENVIRONMENT_VARIABLE = "AWS_PROFILE";
         internal const string DefaultProfileName = "default";
 
@@ -45,18 +49,26 @@ namespace Amazon.Runtime
 
         public static void Reset(IWebProxy proxy)
         {
-            cachedCredentials = null;
-            CredentialsGenerators = new List<CredentialsGenerator>
+            try
             {
+                cachedCredentialsLock.EnterWriteLock();
+                cachedCredentials = null;
+                CredentialsGenerators = new List<CredentialsGenerator>
+                {
 #if BCL
                 () => new AppConfigAWSCredentials(),            // Test explicit keys/profile name first.
 #endif
-                () => AssumeRoleWithWebIdentityCredentials.FromEnvironmentVariables(),
-                // Attempt to load the default profile.  It could be Basic, Session, AssumeRole, or SAML.
-                () => GetAWSCredentials(credentialProfileChain),
-                () => new EnvironmentVariablesAWSCredentials(), // Look for credentials set in environment vars.
-                () => ECSEC2CredentialsWrapper(proxy),      // either get ECS credentials or instance profile credentials
-            };
+                    () => AssumeRoleWithWebIdentityCredentials.FromEnvironmentVariables(),
+                    // Attempt to load the default profile.  It could be Basic, Session, AssumeRole, or SAML.
+                    () => GetAWSCredentials(credentialProfileChain),
+                    () => new EnvironmentVariablesAWSCredentials(), // Look for credentials set in environment vars.
+                    () => ECSEC2CredentialsWrapper(proxy), // either get ECS credentials or instance profile credentials
+                };
+            }
+            finally
+            {
+                cachedCredentialsLock.ExitWriteLock();
+            }
         }
 
         internal static string GetProfileName()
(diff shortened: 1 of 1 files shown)


djepson-rms:
Is this resolving the underlying issue causing this https://github.com/aws/aws-sdk-net/issues/1945 ?

alitas:
> Is this resolving the underlying issue causing this https://github.com/aws/aws-sdk-net/issues/1945 ?

That issue was resolved by https://github.com/aws/aws-sdk-net/pull/2624. However this is resolving a really similar issue.

sdk/src/Core/Amazon.Runtime/Credentials/FallbackCredentialsFactory.cs
```diff
@@ -20,12 +20,16 @@
...
 using System.Security;
+using System.Threading;
 
 namespace Amazon.Runtime
 {
     // Credentials fallback mechanism
     public static class FallbackCredentialsFactory
     {
+        // Lock to control caching credentials across multiple threads.
+        private static readonly object CredentialsLock = new object();
```
peterrsongg, line 31: Can you update this implementation to use `ReaderWriterLockSlim`? It is the recommended approach by dotnet for locking.

sdk/src/Core/Amazon.Runtime/Credentials/FallbackCredentialsFactory.cs
```diff
@@ -147,15 +158,21 @@ public static AWSCredentials GetCredentials(bool fallbackToAnonymous)
...
                     errors.Add(e);
                 }
 
-                if (cachedCredentials != null)
+                if (credentials != null)
                     break;
             }
 
-            if (cachedCredentials == null)
+            lock (CredentialsLock)
```
peterrsongg, line 170: Same thing entering write lock with `ReaderWriterLockSlim` here

sdk/src/Core/Amazon.Runtime/Credentials/FallbackCredentialsFactory.cs
```diff
@@ -124,16 +131,22 @@ public static AWSCredentials GetCredentials()
...
+            cachedCredentialsLock.ExitReadLock();
+            
+            if (credentials != null)
+            {
+                return credentials;
+            }
+            
             List<Exception> errors = new List<Exception>();
 
             foreach (CredentialsGenerator generator in CredentialsGenerators)
```
normj, line 145: This whole loop should be avoided running in multiple threads. Only one thread should be hitting EC2 instance metadata or parsing through profile config files to determine credentials. 

I suggest:
* Move the `cachedCredentialsLock.EnterWriteLock()` above this for loop.
* Start a `try` block
* Once in the block check again to see if `cachedCredentials` has a value if so return. It could have a value if this is the second thread through the write block and the first thread obtain the credentials. Here is an another example in the SDK doing the lock double check https://github.com/aws/aws-sdk-net/blob/9a993b556bd6a9ad2e06ce9799bbda1b99a56b69/sdk/src/Core/RegionEndpoint/RegionEndpointProviderV3.cs#L470
* Run the `foreach`.
* After the foreach `cachedCredentials = credentials`
* Close the `try` block and add a `finally` block that calls `cachedCredentialsLock.ExitWriteLock()` to exit the write lock

I think actually if you do put the whole `foreach` in the write lock you can get rid of the temporary `credentials` variable and simplify the logic since access to the `cachedCredentials` variable would now be controlled through the reader writer lock.

normj (changes requested):

sdk/src/Core/Amazon.Runtime/Credentials/FallbackCredentialsFactory.cs
```diff
@@ -124,61 +136,84 @@ public static AWSCredentials GetCredentials()
...
         {
-            if (cachedCredentials != null)
-                return cachedCredentials;
-
-            List<Exception> errors = new List<Exception>();
-
-            foreach (CredentialsGenerator generator in CredentialsGenerators)
+            try
             {
-                try
```
peterrsongg, line 134: I don't think it is necessary to have a read lock here, It's not a problem if multiple threads are trying to read the credentials. So you can just leave it as it was before like:

```
if(cachedCredentials != null)
  return cachedCredentials;
```

sdk/src/Core/Amazon.Runtime/Credentials/FallbackCredentialsFactory.cs
```diff
@@ -124,61 +136,84 @@ public static AWSCredentials GetCredentials()
...
-                // he expects to use the credentials obtained by running the process. Therefore the exception is
-                // surfaced to the user.
-                catch (ProcessAWSCredentialException)
+            }
+            finally
+            {
+                cachedCredentialsLock.ExitReadLock();
+            }
+            
+            try
```
peterrsongg, line 152: I don't think the topmost `try` block here is needed but I'll let @normj chime in. When norm said "Here is an another example in the SDK doing the lock double check" I think he meant something like:

```
if(cachedCredentials != null)
{
  return cachedCredentials
}
List<Exception> errors = new List<Exception>();
  try{
    cachedCredentials.EnterWriteLock();
    if (cachedCredentials != null)
    {
        return cachedCredentials;
    }
     foreach (CredentialsGenerator generator in CredentialsGenerators)
     {
       try
       {
          cachedCredentials = generator();
        }
        .... skipping logic here     
     }
  }
  finally
   {
     cachedCredentialsLock.ExitWriteLock();
   }
return cachedCredentials;

```

sdk/src/Core/Amazon.Runtime/Credentials/FallbackCredentialsFactory.cs
```diff
@@ -124,61 +136,84 @@ public static AWSCredentials GetCredentials()
...
+                            writer.WriteLine();
+                        }
+
+                        throw new AmazonServiceException(writer.ToString());
+                    }
                 }
-            }
 
-            return cachedCredentials;
+                return cachedCredentials;
```
peterrsongg, line 211: we should exit the lock first then return cached credentials

sdk/src/Core/Amazon.Runtime/Credentials/FallbackCredentialsFactory.cs
peterrsongg: I meant to comment on line 141

alitas: The problem is if a thread calls reset between the null check and return, the method will return null if we don't have a read lock

sdk/src/Core/Amazon.Runtime/Credentials/FallbackCredentialsFactory.cs
alitas: This would've worked if we never set the cached credentials to null, both during the loop and via the reset method. Even then, there are JIT optimizations that might cause issues, hence the suggestion to use of a readwritelockslim in the first place.

sdk/src/Core/Amazon.Runtime/Credentials/FallbackCredentialsFactory.cs
alitas: The finally block will ensure that. Otherwise we can return a null value here as it can be overwritten.

peterrsongg: Ah I misread the scope here, I thought we were returning outside the try block

normj (commented):
Locking code looks good. Thanks!

sdk/src/Core/Amazon.Runtime/Credentials/FallbackCredentialsFactory.cs
peterrsongg: Okay, in that case I don't see any issues. Approving

peterrsongg (approved):

alitas:
> Did a full dry run and it passed. Do you mind squashing your commits before we do the release? It will go out in our next manual release.

Squashed and ready to go.

normj (approved):

normj:
PR has been released as part of version 3.7.201.9 of AWSSDK.Core

PR #3038, opened 2023-08-23
AOT Work: Mark PaginatedResourceFactory for BCL only

normj:
The `PaginatedResourceFactory` is an old internal paginator implementation only used in the S3 File IO abstraction which is only available for .NET Framework. This class uses a lot of reflection that is not compatible with AOT compilation. This PR makes the class only available in .NET Framework to remove the AOT trim warnings.

The service client paginators are not affected by this because they are generated without any reflection and do not use the `PaginatedResourceFactory` class. 

I confirmed the fix by ensuring the .NET Standard and .NET Framework solutions still compile after removing `PaginatedResourceFactory` from the .NET Standard/Core targets.

### DIFF
--- a/sdk/src/Core/Amazon.Util/PaginatedResourceFactory.cs
+++ b/sdk/src/Core/Amazon.Util/PaginatedResourceFactory.cs
@@ -27,6 +27,7 @@ using System.Globalization;
 
 using Amazon.Util.Internal;
 
+#if BCL
 namespace Amazon.Util
 {
     public static class PaginatedResourceFactory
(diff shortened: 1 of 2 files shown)


dscpinheiro (approved):

muhammad-othman (commented):
there one occurrence of this class in the `Global.Suppressions.cs` file, should we also wrap it with `#if BCL` ?
I'm not sure if it will cause any issue but I've noticed that there is another suppression that is wrapped with this condition
image

normj:
@muhammad-othman I added the `#if` in the global suppressions file as well.

boblodgett (approved):

muhammad-othman (approved):

PR #3042, opened 2023-09-05
Add a ValueStringBuilder to the Core utilities

danielmarbach:
## Description

This PR is meant to get the discussion started to get references to `System.Memory` and `System.Buffers` in plus adding a first small change like the ValueStringBuilder that allows building string based on a pooled char array.

## Motivation and Context

Before I go down the path of measuring the before and after I would like to get the discussion started here. Happy to provide some initial benchmarks if that is the thing that would prevent discussions from happening. 

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
+++ b/sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
@@ -20,6 +20,7 @@
 
 using Amazon.Runtime.Internal.Util;
 using System;
+using System.Buffers;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
@@ -128,7 +129,7 @@ namespace Amazon.Util
         {
             const string basePathCharacters = "/:'()!*[]$";
 
-            var sb = new StringBuilder();
+            var sb = new ValueStringBuilder(basePathCharacters.Length * 2);
             foreach (var c in basePathCharacters)
             {
                 var escaped = Uri.EscapeUriString(c.ToString());
@@ -221,13 +222,14 @@ namespace Amazon.Util
          */
         internal static string CalculateStringToSignV2(ParameterCollection parameterCollection, string serviceUrl)
         {
-            StringBuilder data = new StringBuilder("POST\n", 512);
+            var data = new ValueStringBuilder(512);
+            data.Append("POST\n");
             var sortedParameters = parameterCollection.GetSortedParametersList();
-            Uri endpoint = new Uri(serviceUrl);
+            var endpoint = new Uri(serviceUrl);
 
             data.Append(endpoint.Host);
             data.Append("\n");
-            string uri = endpoint.AbsolutePath;
+            var uri = endpoint.AbsolutePath;
             if (uri == null || uri.Length == 0)
             {
                 uri = "/";
@@ -237,17 +239,14 @@ namespace Amazon.Util
             data.Append("\n");
             foreach (KeyValuePair<string, string> pair in sortedParameters)
             {
-                if (pair.Value != null)
-                {
-                    data.Append(AWSSDKUtils.UrlEncode(pair.Key, false));
-                    data.Append("=");
-                    data.Append(AWSSDKUtils.UrlEncode(pair.Value, false));
-                    data.Append("&");
-                }
-            }
+                if (pair.Value == null) continue;
 
-            string result = data.ToString();
-            return result.Remove(result.Length - 1);
+                data.Append(AWSSDKUtils.UrlEncode(pair.Key, false));
+                data.Append("=");
+                data.Append(AWSSDKUtils.UrlEncode(pair.Value, false));
+                data.Append("&");
+            }
+            return data.ToString(0, data.Length - 1);
         }
 
         /**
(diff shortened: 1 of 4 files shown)


sdk/src/Core/Amazon.Util/Internal/_netstandard/ValueStringBuilder.cs
```diff
@@ -0,0 +1,337 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
danielmarbach, line 1: This file is partially adopted from the ValueStringBuilder in the .NET runtime repository. What attribution would you like to see or require?

sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
```diff
@@ -128,7 +129,7 @@ private static string DetermineValidPathCharacters()
         {
             const string basePathCharacters = "/:'()!*[]$";
 
-            var sb = new StringBuilder();
+            var sb = new ValueStringBuilder(basePathCharacters.Length * 2);
```
danielmarbach, line 132: For each of these usages it might make sense to help me to find reasonable defaults.

danielmarbach:
Ran a quick benchmark

image

```chsarp
[SimpleJob]
[MemoryDiagnoser]
public class BuilderVsValueBuilder
{

    [GlobalSetup]
    public void Setup()
    {
        parameterCollection = new Dictionary<string, string?>
        {
            { "SomeHeader1", "SomeValue" },
            { "SomeHeader2", "SomeValue" },
            { "SomeHeader3", "SomeValue" },
            { "SomeHeader4", "SomeValue" },
            { "SomeHeader5", "SomeValue" },
            { "SomeHeader6", "SomeValue" },
            { "SomeHeader7", "SomeValue" },
        };
    }
    
    [Benchmark(Baseline = true)]
    public string StringBuilder()
    {
        return GetParametersAsString(parameterCollection);
    }
    
    [Benchmark]
    public string ValueBuilder()
    {
        return GetParametersAsStringNew(parameterCollection);
    }
}
```

danielmarbach:
I guess ultimately one could achieve even more improvements by doing also structural and algorithmic changes across the utilities, but as this demonstrates, introducing the value string builder also leads to significant reductions of memory allocations and a noticeable throughput improvement

normj (commented):
I think using the newer APIs from `System.Memory` and `System.Buffers` is a great idea but we have to use them conditionally for the targets that support. In our case .NET Core 3.1 and greater. It is a big pain for us to add more dependencies and how we have to distribute them. Keep in my NuGet may be the most common way to acquire the SDK and would solve the dependency challenges we have to support distributing the SDKs in channels that don't have access to NuGet.

Also we still support .NET Framework 3.5 and 4.5 so even if a library as a .NET Standard 2.0 package it doesn't help those platforms.

normj:
> Ran a quick benchmark
> 
> image
> 
> ```
> [SimpleJob]
> [MemoryDiagnoser]
> public class BuilderVsValueBuilder
> {
> 
>     [GlobalSetup]
>     public void Setup()
>     {
>         parameterCollection = new Dictionary<string, string?>
>         {
>             { "SomeHeader1", "SomeValue" },
>             { "SomeHeader2", "SomeValue" },
>             { "SomeHeader3", "SomeValue" },
>             { "SomeHeader4", "SomeValue" },
>             { "SomeHeader5", "SomeValue" },
>             { "SomeHeader6", "SomeValue" },
>             { "SomeHeader7", "SomeValue" },
>         };
>     }
>     
>     [Benchmark(Baseline = true)]
>     public string StringBuilder()
>     {
>         return GetParametersAsString(parameterCollection);
>     }
>     
>     [Benchmark]
>     public string ValueBuilder()
>     {
>         return GetParametersAsStringNew(parameterCollection);
>     }
> }
> ```

Impressive reduction in allocations which is an area the SDK really needs improvement on.

danielmarbach:
@normj I finally had time to look at this a bit more and I think in order to safely introduce these changes and adhering to

> I think using the newer APIs from System.Memory and System.Buffers is a great idea but we have to use them conditionally for the targets that support. In our case .NET Core 3.1 and greater. It is a big pain for us to add more dependencies and how we have to distribute them. Keep in my NuGet may be the most common way to acquire the SDK and would solve the dependency challenges but we have to support distributing the SDKs in channels that don't have access to NuGet.
> Also we still support .NET Framework 3.5 and 4.5 so even if a library as a .NET Standard 2.0 package it doesn't help those platforms.

I would need to probably freeze the .NET 3.5/4.5 versions of the Utils and start evolving the Netstandard version only. This would come with the drawbacks of your team having to fix bugs in both versions until you drop the NET3.5 and NET4.5 target or upgrade NET4.5 to NET4.8 target. 

I have tried to think about solutions like partial methods, introducing interface abstractions for the stringbuilder, pooling and return to pool hooks and came to the conclusion it would make the code unnecessary complex to maintain.

What are your thoughts? Would you be OK if I'd go down the route of duplicating the relevant Utils methods?

danielmarbach:
@normj any thoughts on the above?

danielmarbach:
The SDK has announced the deprecation of the NET 3.5 targets
I guess I can resurrect this at some point since the complexity decreased now

normj:
@danielmarbach There is a `v4-development` branch where the removal of .NET Framework 3.5 and updating 4.6.2 is happening. I have a story that we will hopefully get done soon to add to the 4.6.2 target the .NET Standard 2.0 libraries likes System.Memory and System.Buffer. I can ping you when we get that done and then that branch is where I would recommend targeting any performance PRs.

normj:
@danielmarbach the `v4-develoment` branch has been updated to have `System.Memory`, `System.Buffer` and `System.Text.Json`. I'm going to close this PR but if you want to do some perf improvements, which would be great, that is branch you should target.

andriusrazinskas:
Any info on when I could expect V4 being released as stable version?

PR #3044, opened 2023-09-06
Mark the following features as not supported for Native AOT

normj:
Mark the features in the SDK that use unbounded reflection that do not support trimming with the `RequiresUnreferencedCode` attribute. This will tell users of the SDK who are compiling for Native AOT with trimming that they should not use these APIs. This doesn't mean we will never support them for Native AOT but these features require significant work and potentially redesign to get to work in Native AOT and in first phase of SDK AOT support we want to focus on identifying the areas that should be avoid when using Native AOT.

## DynamoDB high level libraries
These libraries use a lot of reflection over user's types that doesn't work in Native AOT. We would probably need to develop some sort of source generator approach like was done with JSON. Also there is a code doing full assembly type scans looking for converters that isn't supported in a trimmed environment.

## CognitioSync service client
There is a pipeline handler injected into the service client that uses reflection over all of the incoming requests to look for the identity from the incoming request object and set it on the `CognitoAWSCredentials`.

## Glacier high level library
It does some JSON deserialization to a `Dictionary<string, object>` and having the `object` as the value in the collection makes it not work for Native AOT trimming.

## AWSSDK.Extensions.NETCore.Setup
This is a tough one to not support but it does a lot of reflection to copy values from a generic service config to a the actual service config. Even worse users register via the service client interface and then library uses string manipulation from the interface name to service client name and then load the service client type via that string name. This will need to be completely reworked to fit in a Native AOT trimmed environment.

### DIFF
--- a/extensions/src/AWSSDK.Extensions.NETCore.Setup/AWSSDK.Extensions.NETCore.Setup.csproj
+++ b/extensions/src/AWSSDK.Extensions.NETCore.Setup/AWSSDK.Extensions.NETCore.Setup.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netcoreapp3.1;net6.0</TargetFrameworks>
     <AssemblyName>AWSSDK.Extensions.NETCore.Setup</AssemblyName>
     <PackageId>AWSSDK.Extensions.NETCore.Setup</PackageId>
     <GenerateAssemblyTitleAttribute>false</GenerateAssemblyTitleAttribute>
@@ -13,6 +13,10 @@
     <GenerateAssemblyVersionAttribute>false</GenerateAssemblyVersionAttribute>
     <GenerateAssemblyFileVersionAttribute>false</GenerateAssemblyFileVersionAttribute>
   </PropertyGroup>
+	
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net6.0'">
+    <EnableTrimAnalyzer>true</EnableTrimAnalyzer>
+  </PropertyGroup>	
 
   <Choose>
     <When Condition=" '$(AWSKeyFile)' == '' ">
--- a/sdk/src/Services/CognitoSync/Custom/AmazonCognitoSyncClient.Extensions.cs
+++ b/sdk/src/Services/CognitoSync/Custom/AmazonCognitoSyncClient.Extensions.cs
@@ -1,0 +1,24 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ * 
+ *  http://aws.amazon.com/apache2.0
+ * 
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+namespace Amazon.CognitoSync
+{
+#if NET6_0_OR_GREATER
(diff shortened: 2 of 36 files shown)


sdk/src/Services/DynamoDBv2/Custom/Internal/InternalConstants.cs
```diff
@@ -0,0 +1,22 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+namespace Amazon.DynamoDBv2.Custom.Internal
+{
+    internal static class InternalConstants
+    {
+        internal const string RequiresUnreferencedCodeMessage = "The Amazon DynamonDB high level libraries in the DocumentModel and DataModel namespace do not supported for Native AOT.";
```
dscpinheiro, line 20: Nit: `do not supported`. 

Did you mean `do not support Native AOT`? Or should it be `have not been updated to support Native AOT` (like the other messages)?

normj: Fixed

dscpinheiro (approved):

extensions/src/AWSSDK.Extensions.NETCore.Setup/AWSSDK.Extensions.NETCore.Setup.csproj
```diff
@@ -13,6 +13,10 @@
     <GenerateAssemblyVersionAttribute>false</GenerateAssemblyVersionAttribute>
     <GenerateAssemblyFileVersionAttribute>false</GenerateAssemblyFileVersionAttribute>
   </PropertyGroup>
+	
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net6.0'">
+    <EnableTrimAnalyzer>true</EnableTrimAnalyzer>
```
ashovlin, line 18: Should we set `<WarningsAsErrors>IL2026</WarningsAsErrors>` anywhere we're adding this? 

We're annotating everything today, but as long as these are warnings I think it'd be easy for somebody to add new functionality that is not compatible and not think about this.

sdk/src/Services/DynamoDBv2/Custom/Internal/InternalConstants.cs
```diff
@@ -0,0 +1,22 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+namespace Amazon.DynamoDBv2.Custom.Internal
+{
+    internal static class InternalConstants
+    {
+        internal const string RequiresUnreferencedCodeMessage = "The Amazon DynamonDB high level libraries in the DocumentModel and DataModel namespace have not been updated to support Native AOT.";
```
ashovlin, line 20: Misspelling in "DynamoDB"

extensions/src/AWSSDK.Extensions.NETCore.Setup/AWSSDK.Extensions.NETCore.Setup.csproj
normj: Yes once all of these warnings have been addressed. I still have a few more in Core that need to be addressed. The whole dynamic loading of assemblies like STS.

sdk/src/Services/CognitoSync/Custom/AmazonCognitoSyncClient.Extensions.cs
```diff
@@ -0,0 +1,24 @@
...
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+namespace Amazon.CognitoSync
+{
+#if NET6_0_OR_GREATER
+    [System.Diagnostics.CodeAnalysis.RequiresUnreferencedCode("Amazon CognitoSync has not been updated to support Native AOT compilations.")]
```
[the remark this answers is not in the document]
normj: Space added

sdk/src/Services/DynamoDBv2/Custom/Internal/InternalConstants.cs
normj: Fixed

ashovlin (approved):

PR #3045, opened 2023-09-08
[WIP] Add runtime dependency registry to avoid assembly reflection load of types

normj:
There are a few use cases like `AssumeRole` where the SDK relies on runtime dependencies where the dependency is load via assembly load and type search via class names. This is done in the SDK either to solve circular dependency issues or not wanting to include a dependency for all scenarios when only used in a small subset of features. This approach does not work for Native AOT with trimming enable because the runtime dependency would be trimmed out of the Native AOT application.

The solution this PR is to create a registry Native AOT users can use to explicit register instances of the runtime dependency the SDK needs. The PR is WIP using the registry to address 2 of the instances where runtime dependencies are used. The goal is get consensus of the approach before using for all of the other scenarios.

## Example
Lets say a user is using a credential provide that requires the STS service client. Before the STS service client is used the user would call.

```
Amazon.RuntimeDependencyRegistry.Instance.RegisterSecuirtyTokenServiceClient(new AmazonSecurityTokenService());
```

Then when the SDK needs it needs an STS client it will check the `RuntimeDependencyRegistry` for an instance. If one exist use it otherwise fallback on the usual reflection based method. If the Native AOT application has trimming enabled, which is the default, the reflection method will fail. A `MissingRuntimeDependencyException` exception will be thrown which will inform the user how to fix the problem. This is similar to the non Native AOT scenario when the user doesn't add the dependency to the runtime dependency on their application.

## Changes

* In `ServiceClientHelpers` which does the reflection work mark the methods with the `RequiresUnreferencedCode` attribute.
* In `RuntimeDependencyRegistry` add specific public register methods for each runtime dependency potential needed. This was done as opposed of a generic method because it avoided users having to understand how the dependency is hashed. It also provides a place to provide specific documentation for the scenario to help users.
* In the areas where `ServiceClientHelpers` is used add a check first for the dependency in the `RuntimeDependencyRegistry`. If an instance exist then use it. If an instance does not exist fallback to using the `ServiceClientHelpers` in case the user has trimming turned off for the SDK forcing all of the SDK into the Native AOT executable. The fallback to `ServiceClientHelpers` also include warning suppressions for AOT warnings since we are trying to handle the situation via the `RuntimeDependencyRegistry`
* If the fallback to `ServiceClientHelpers` throws an error and the environment is a Native AOT environment then throw a new `MissingRuntimeDependencyException` exception.

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Internal/ServiceClientHelpers.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/ServiceClientHelpers.cs
@@ -28,7 +28,13 @@ namespace Amazon.Runtime.Internal
         public const string KMS_ASSEMBLY_NAME = "AWSSDK.KeyManagementService";
         public const string KMS_SERVICE_CLASS_NAME = "Amazon.KeyManagementService.AmazonKeyManagementServiceClient";
 
+
+#if NET6_0_OR_GREATER
+        public static TClient CreateServiceFromAnother<[System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers(System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.PublicConstructors)] TClient, 
+                                    [System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers(System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.PublicConstructors)] TConfig>(AmazonServiceClient originalServiceClient)
+#else
         public static TClient CreateServiceFromAnother<TClient, TConfig>(AmazonServiceClient originalServiceClient)
+#endif
             where TConfig : ClientConfig, new ()
             where TClient : AmazonServiceClient
         {
@@ -48,6 +54,9 @@ namespace Amazon.Runtime.Internal
             return newServiceClient;
         }
 
+#if NET6_0_OR_GREATER
+        [System.Diagnostics.CodeAnalysis.RequiresUnreferencedCode("Using ServiceClientHelper to dynamically load dependency is not supported for Native AOT. SDK calling code should use Amazon.RuntimeDependencyRegistry to use explicitly provided runtime dependencies.")]
+#endif
         public static TClient CreateServiceFromAssembly<TClient>(string assemblyName, string serviceClientClassName,
             RegionEndpoint region)
             where TClient : class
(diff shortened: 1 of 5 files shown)


sdk/src/Core/RuntimeDependencyRegistry.cs
```diff
@@ -0,0 +1,166 @@
...
+    /// <summary>
+    /// Their are some features of the AWS .NET SDK that at runtime load dependencies from assemblies. The most common
+    /// example is using credential profiles that require the service client from the AWSSDK.SecurityToken package.
+    /// Normally these runtime dependencies are resolved by using reflection to load the assembly and the type,
+    /// In Native AOT with trimming turned on that will not worked because the dependency would be trimmed out.
+    /// 
+    /// The RuntimeDependencyRegistry is used to work around this limitation by allowing users to explicitly register
+    /// the runtime dependency instance into the SDK removing any use of reflection.
+    /// </summary>
+    public class RuntimeDependencyRegistry : IDisposable
```
boblodgett, line 19: Does this registry have a flaw in that there is only one static instance of the object of that type for the entire app? So the service instance you register would be locked into whatever region/settings of the one instance with no way to stop conflicts.

sdk/src/Core/RuntimeDependencyRegistry.cs
```diff
@@ -0,0 +1,166 @@
...
+
+    /// <summary>
+    /// This exception is thrown when the SDK is used as part of an application compiled for Native AOT with trimming turned on.
+    /// In this environment assemblies can not be dynamically loaded and the usual ServiceClientHelpers operations will fail.
+    /// </summary>
+    public class MissingRuntimeDependencyException : AmazonClientException
+    {
+
+        public MissingRuntimeDependencyException(string packageName, string className, string registerMethod)
+            : base("Operation failed because of missing runtime dependency. In Native AOT builds runtime dependencies can not be " +
```
boblodgett, line 141: `because of missing` => `because of a missing`

normj: Fixed

sdk/src/Core/Amazon.Runtime/Internal/ServiceClientHelpers.cs
```diff
@@ -48,6 +54,9 @@ public static class ServiceClientHelpers
             return newServiceClient;
         }
 
+#if NET6_0_OR_GREATER
+        [System.Diagnostics.CodeAnalysis.RequiresUnreferencedCode("Using ServiceClientHelper to dynamically load dependency is not supported for Native AOT. SDK calling code should use Amazon.RuntimeDependencyRegistry to use explicitly provided runtime dependencies.")]
```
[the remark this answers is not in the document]
normj: Fixed

sdk/src/Core/RuntimeDependencyRegistry.cs
```diff
@@ -0,0 +1,157 @@
...
+using Amazon.Runtime.Internal;
+using AWSSDK.Runtime.Internal.Util;
+using System;
+using System.Collections.Generic;
+using System.Threading;
+
+namespace Amazon
+{
+    /// <summary>
+    /// Their are some features of the AWS .NET SDK that at runtime load dependencies from assemblies. The most common
```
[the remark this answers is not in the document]
normj: Fixed

normj:
Closed in favor of new PR https://github.com/aws/aws-sdk-net/pull/3048

PR #3048, opened 2023-09-13
Add runtime dependency registry to avoid assembly reflection load of types

normj:
There are a few use cases like `AssumeRole` where the SDK relies on runtime dependencies where the dependency is load via assembly load and type search via class names. This is done in the SDK either to solve circular dependency issues or not wanting to include a dependency for all scenarios when only used in a small subset of features. This approach does not work for Native AOT with trimming enable because the runtime dependency would be trimmed out of the Native AOT application.

The solution this PR is to create a registry Native AOT users can use to explicit register instances of the runtime dependency the SDK needs. SDK users can either provide an instance of the runtime dependency or factory method.

## Example
Here is an example of registing a single STS client to be used for Core assume role type actions

```
GlobalRuntimeDependencyRegistry.Instance.RegisterSecurityTokenServiceClient(new Amazon.SecurityToken.AmazonSecurityTokenServiceClient());

var config = new AmazonDynamoDBConfig { Profile = new Profile("assume-role-test")};
var ddbClient = new AmazonDynamoDBClient(config);
var response = await ddbClient.ListTablesAsync();
Console.WriteLine("Table Count: " + response.TableNames.Count);
```

In this example factory method is registered for STS. The factory method is called every time Core needs an STS client. The context provides information about what the STS client is going to be used for. In the case of STS the context's `SecurityTokenServiceClientContextData` property is set with a `Region` property set telling the factory method what region it expects the returned STS client to be configured for.
```
GlobalRuntimeDependencyRegistry.Instance.RegisterSecurityTokenServiceClient((context) =>
{
    Console.WriteLine("Context Type: " + context.Type);
    Console.WriteLine($"Creating STS client for {context.SecurityTokenServiceClientContextData.Action} using region {context.SecurityTokenServiceClientContextData.Region}");
    return new Amazon.SecurityToken.AmazonSecurityTokenServiceClient(context.SecurityTokenServiceClientContextData.Region);
});

var config = new AmazonDynamoDBConfig { Profile = new Profile("assume-role-test")};
var ddbClient = new AmazonDynamoDBClient(config);
var response = await ddbClient.ListTablesAsync();
Console.WriteLine("Table Count: " + response.TableNames.Count);
```

## Errors

If an application is compiled for Native AOT and the runtime dependency is not registered a `MissingRuntimeDependencyException` is thrown. The exception will inform the user that due to Native AOT that need to register their runtime dependency and inform the user what type to register and with what method.

In the code where ever we use `GlobalRuntimeDependencyRegistry` to get an instance if an instance is not found we still fallback to the reflection method. Most likely this will fail and then we will throw the `MissingRuntimeDependencyException`. We try the reflection method in case the SDK user has excluded the SDK from trimming. That is possibly what they are doing today and if we removed the logic we would introduce a breaking change.

## Testing

* Compiled the full .NET 3.5, 4.5 and .NET Standard solutions confirmed no errors.
* Added new unit tests for the registry
* Build sample apps confirming expected behavior both with and without registering the dependency in AOT applications.

## Changes

* In `ServiceClientHelpers` which does the reflection work mark the methods with the `RequiresUnreferencedCode` attribute.
* In `GlobalRuntimeDependencyRegistry` add specific public register methods for each runtime dependency potential needed. This was done as opposed of a generic method because it avoided users having to understand how the dependency is hashed. It also provides a place to provide specific documentation for the scenario to help users.
* In the areas where `ServiceClientHelpers` is used add a check first for the dependency in the `GlobalRuntimeDependencyRegistry`. If an instance exist then use it. If an instance does not exist fallback to using the reflection method in case the user has trimming turned off for the SDK forcing all of the SDK into the Native AOT executable. The fallback to `ServiceClientHelpers` also include warning suppressions for AOT warnings since we are trying to handle the situation via the `GlobalRuntimeDependencyRegistry`
* If the fallback to `ServiceClientHelpers` throws an error and the environment is a Native AOT environment then throw a new `MissingRuntimeDependencyException` exception.


sdk/src/Core/Amazon.Runtime/Credentials/AssumeRoleAWSCredentials.cs
```diff
@@ -87,34 +89,49 @@ public AssumeRoleAWSCredentials(AWSCredentials sourceCredentials, string roleArn
...
 
+#if NET6_0_OR_GREATER
+        [System.Diagnostics.CodeAnalysis.UnconditionalSuppressMessage("ReflectionAnalysis", "IL2026", 
+            Justification = "Reflection code is only used as a fallback in case the SDK was not trimmed. Trimmed scenarios should register dependencies with Amazon.RuntimeDependencyRegistry.GlobalRuntimeDependencyRegistry")]
+#endif
         protected override CredentialsRefreshState GenerateNewCredentials()
         {
             var region = FallbackRegionFactory.GetRegionEndpoint() ?? DefaultSTSClientRegion;
-            ICoreAmazonSTS coreSTSClient;
+            ICoreAmazonSTS coreSTSClient = GlobalRuntimeDependencyRegistry.Instance.GetInstance<ICoreAmazonSTS>(ServiceClientHelpers.STS_ASSEMBLY_NAME, ServiceClientHelpers.STS_SERVICE_CLASS_NAME,
```
boblodgett, line 99: This still does not support the Options.ProxySettings when coming from the global runtime unless a dev sets them for the obvious reason the instance is already constructed. How is a dev going to know to set these values when they may have been using Options before? Documentation? Prior comment: https://github.com/aws/aws-sdk-net/pull/3045#discussion_r1320364108

Would `Options` need to be passed in the `SecurityTokenServiceClientContext`?

sdk/src/Core/Amazon.Runtime/Credentials/AssumeRoleAWSCredentials.cs
```diff
@@ -87,34 +89,49 @@ public AssumeRoleAWSCredentials(AWSCredentials sourceCredentials, string roleArn
...
+            if (coreSTSClient == null)
             {
-                var stsConfig = ServiceClientHelpers.CreateServiceConfig(ServiceClientHelpers.STS_ASSEMBLY_NAME, ServiceClientHelpers.STS_SERVICE_CONFIG_NAME);
-                stsConfig.RegionEndpoint = region;
+                try
+                {
+                    var stsConfig = ServiceClientHelpers.CreateServiceConfig(ServiceClientHelpers.STS_ASSEMBLY_NAME, ServiceClientHelpers.STS_SERVICE_CONFIG_NAME);
+                    stsConfig.RegionEndpoint = region;
+
+                    if (Options != null && Options.ProxySettings != null)
```
boblodgett, line 109: nit: in another codeblock below this is done as:

```
if (_options?.ProxySettings != null)
```

and as:

```
if (Options.ProxySettings != null)
```

It could be done the same here.

sdk/src/Core/Amazon.Runtime/Credentials/FederatedAWSCredentials.cs
```diff
@@ -189,28 +195,37 @@ private CredentialsRefreshState Authenticate(ICredentials userCredential)
                 region = DefaultSTSClientRegion;
             }
 
-            ICoreAmazonSTS coreSTSClient = null;
-            try
+            ICoreAmazonSTS coreSTSClient = GlobalRuntimeDependencyRegistry.Instance.GetInstance<ICoreAmazonSTS>(ServiceClientHelpers.STS_ASSEMBLY_NAME, ServiceClientHelpers.STS_SERVICE_CLASS_NAME,
+                new CreateInstanceContext(new SecurityTokenServiceClientContext(SecurityTokenServiceClientContext.ActionContext.FederatedAWSCredentials, region)));
```
boblodgett, line 199: Below it is using `AnonymousAWSCredentials`. Should there be `AnonymousAWSCredentials` under *.ActionContext. ?

sdk/src/Core/Amazon.Runtime/Internal/_bcl45+netstandard/SSOServiceClientHelpers.cs
```diff
@@ -21,14 +27,25 @@ public static class SSOServiceClientHelpers
 #endif
         )
         {
-            return CreateClient<ICoreAmazonSSOOIDC>(
-                region,
-                ServiceClientHelpers.SSO_OIDC_SERVICE_CLASS_NAME,
-                ServiceClientHelpers.SSO_OIDC_SERVICE_CONFIG_NAME,
-                ServiceClientHelpers.SSO_OIDC_ASSEMBLY_NAME,
-                proxySettings);
+            ICoreAmazonSSOOIDC coreSSO = GlobalRuntimeDependencyRegistry.Instance.GetInstance<ICoreAmazonSSOOIDC>(ServiceClientHelpers.SSO_OIDC_ASSEMBLY_NAME, ServiceClientHelpers.SSO_OIDC_SERVICE_CLASS_NAME, new CreateInstanceContext(new SSOOIDCClientContext(region)));
```
boblodgett, line 30: Does proxySettings need to be passed to the context here?

sdk/src/Core/RuntimeDependencies/CreateInstanceContext.cs
```diff
@@ -0,0 +1,276 @@
...
+    /// <summary>
+    /// Context for the factory method to create the SigV4aCrtSigner runtime dependency.
+    /// </summary>
+    public class SigV4aCrtSignerContext
+    {
+        /// <summary>
+        /// Create an instance of the context
+        /// </summary>
+        /// <param name="payload"></param>
+        public SigV4aCrtSignerContext(bool payload)
```
boblodgett, line 221: should be `signPayload` not `payload`. The `param` in the comment needs filling out and corrected.

sdk/src/Core/RuntimeDependencies/CreateInstanceContext.cs
```diff
@@ -0,0 +1,276 @@
...
+        {
+            Payload = payload;
+        }
+
+        public bool Payload { get; }
```
boblodgett, line 226: Should be `SignPayload`.

sdk/src/Core/Amazon.Runtime/Credentials/AssumeRoleAWSCredentials.cs
normj: I added the proxy settings to the context. I don't think adding the whole `Options` class should be done. There is different type of options for each possible action and the other properties are more for making the assume call not constructing the STS client.

sdk/src/Core/Amazon.Runtime/Credentials/AssumeRoleAWSCredentials.cs
normj: Done

sdk/src/Core/Amazon.Runtime/Credentials/FederatedAWSCredentials.cs
normj: I added a comment on the `ActionContext` enum that for `FederatedAWSCredentials` the user can use `AnonymousAWSCredentials`. I don't think we need to put an actual instance of `AnonymousAWSCredentials` on the context object and would probably cause confusion for the other use causes.

sdk/src/Core/Amazon.Runtime/Internal/_bcl45+netstandard/SSOServiceClientHelpers.cs
normj: Done

sdk/src/Core/RuntimeDependencies/CreateInstanceContext.cs
normj: Done

sdk/src/Core/RuntimeDependencies/CreateInstanceContext.cs
normj: Done and added NDoc for the property.

boblodgett (approved):

boblodgett (approved):

sdk/src/Core/RuntimeDependencies/CreateInstanceContext.cs
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+
+    /// <summary>
+    /// Context for the factory method to create the SecurityTokenServiceClient runtime dependency.
+    /// </summary>
+    public class SecurityTokenServiceClientContext
+    {
+        /// <summary>
+        /// The possibly actions the the SecurityTokenServiceClient will be created for.
```
peterrsongg, line 166: nit: `The possible actions the` ...

sdk/test/UnitTests/Custom/RuntimeDependencyRegistryTests.cs
```diff
@@ -0,0 +1,214 @@
...
+using Amazon.Runtime.Internal.Auth;
+using Amazon.KeyManagementService;
+using Amazon.SSO;
+using Amazon.SSOOIDC;
+
+namespace AWSSDK.UnitTests
+{
+    [TestClass]
+    public class RuntimeDependencyRegistryTests
+    {
```
peterrsongg, line 24: nit: not necessary to fix since they are just test cases, but there are lot of instances where `registered` is spelled `registerd` and `instance` is spelled `Instanace`

peterrsongg (approved):

sdk/src/Core/RuntimeDependencies/CreateInstanceContext.cs
normj: Fixed

PR #3056, opened 2023-09-22
Optimize memory allocations when HMAC signing from a string

stevenaw:
Use pooled buffers in HMAC signing paths for netcoreapp3.1 to avoid duplicating payloads in memory. Memory requirements for the function are now consistent regardless of payload size which at a 5KB payload size shows an 85% improvement in total memory consumption. I hadn't found any existing tests covering the HMAC behaviour though I added one now in this PR.

## Description
Use the ArrayPool<T> in HMAC signing paths to avoid duplicating payloads in memory on netcoreapp3.1 builds. This requires no change in dependencies. ArrayPool<T> has also been used internally in a private helper method to avoid internal allocations.

## Motivation and Context
Reducing memory overhead and avoiding duplicating payloads in memory - especially large ones - can improve scalability by reducing pressure on the garbage collector.

## Testing
I added a unit test before making any changes to ensure the output didn't change. I also ran benchmarks. They show a flat, consistent memory profile of 920B regardless of payload size. This yields 85% savings on a payload of 5KB

```
| Method | PayloadSize | Mean       | Ratio | Gen0   | Allocated | Alloc Ratio |
|------- |------------ |-----------:|------:|-------:|----------:|------------:|
| Old    | 50          |   880.5 ns |  1.00 | 0.1593 |    1000 B |        1.00 |
| New    | 50          |   860.1 ns |  0.98 | 0.1459 |     920 B |        0.92 |
|        |             |            |       |        |           |             |
| Old    | 500         | 1,091.2 ns |  1.00 | 0.2308 |    1448 B |        1.00 |
| New    | 500         | 1,097.4 ns |  1.01 | 0.1450 |     920 B |        0.64 |
|        |             |            |       |        |           |             |
| Old    | 5000        | 3,341.4 ns |  1.00 | 0.9460 |    5944 B |        1.00 |
| New    | 5000        | 3,098.7 ns |  0.93 | 0.1450 |     920 B |        0.15 |
```

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/AWSSDK.Core.NetStandard.csproj
+++ b/sdk/src/Core/AWSSDK.Core.NetStandard.csproj
@@ -74,5 +74,6 @@
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="1.1.0" />
+    <PackageReference Include="System.Buffers" Version="4.3.0" />
   </ItemGroup>
 </Project>
--- a/sdk/src/Core/AWSSDK.Core.NetStandard.csproj
+++ b/sdk/src/Core/AWSSDK.Core.NetStandard.csproj
@@ -74,6 +74,5 @@
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="1.1.0" />
-    <PackageReference Include="System.Buffers" Version="4.3.0" />
   </ItemGroup>
 </Project>
--- a/sdk/src/Core/Amazon.Util/CryptoUtil.cs
+++ b/sdk/src/Core/Amazon.Util/CryptoUtil.cs
@@ -16,6 +16,8 @@ using Amazon.Runtime;
 using Amazon.Runtime.Internal.Util;
 using Amazon.Util.Internal;
 using AWSSDK.Runtime.Internal.Util;
+using System;
+using System.Buffers;
 using System.IO;
 using System.Security.Cryptography;
 using System.Text;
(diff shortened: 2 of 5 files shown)


normj (changes requested):
Happy to take on PRs for improving the performance of the SDK. It is really challenging for us to take on new dependencies like `System.Buffers` due to other requirements we have for distributing the SDK outside of the NuGet ecosystem. Also the code you are changing is also used in builds for .NET Framework 3.5 and 4.5 which can't use the .NET Standard System.Buffers package.

If you can modify the PR to use the `NETCOREAPP3_1_OR_GREATER` build constant to restrict the usage of System.Buffers to only SDK build targets 3.1 or later. This file was also modified to use System.Buffers conditional https://github.com/aws/aws-sdk-net/blob/33334593b0fd471530220a649e1df2554512fcc9/sdk/src/Services/S3/Custom/Transfer/Internal/_async/MultipartUploadCommand.async.cs#L27

stevenaw:
Thanks for the prompt review @normj . I hadn't realized the source file had been shared with net35 and net45 builds but it makes sense as you've explained it.

I've pushed another round of changes to remove the added dependency and generally minimize changes on non-netcoreapp3.1 paths. I've had a hard time testing the changes across all runtimes locally, but a simple test project targeting `net462;netcoreapp3.1` seemed to behave alright. I've also rebased from latest master, looks like I had been about a year out of date there. As a result, latest changes are force-pushed but still done as separate commits to make post-review changes easier to see.

I'll update the PR description to represent current state

normj (approved):
Thanks for the changes. It looks good. I'll get somebody else on the team to do a second pass and run the change through our full test pass.

CamileDahdah (approved):

normj:
The PR has been rebased into master and released as part of version 3.7.202.16 of AWSSDK.Core

PR #3060, opened 2023-09-28
Add SSO registration scopes from the shared config file to CredentialProfileOptions.

floralph:
Added CredentialProfileOptions.SsoRegistrationScopes property that is set by SharedCredentialsFile.TryGetProfile.

## Description
Like the other SSO-related fields (sso_region and sso_start_url in an [sso-session] section in the shared config file, this change makes the sso_registration_scopes that may also be defined in the [sso-session] available on CredentialProfileOptions instances.  

This is for reading purposes only as there is no way to save an [sso-session] section from the SDK other than manually creating the section via ProfileIniFile, so no changes have been made for writing this field.

## Motivation and Context
This change is to make sso_registration_scopes from [sso-session] sections in the the shared config file available in code for comparison and filtering.

## Testing
Created a console app to verify that the CredentialFileOptions.SsoRegistrationScopes property matched that of the [sso-session] defined in my shared config file.  Updated SharedConfigurationFile tests to include SsoRegistrationScopes property.

## Types of changes

## Checklist
- [x] My change requires a change to the documentation (should just be API reference generated from XML doc comments)
- [ ] All new and existing tests passed (there appears to be some concurrency issues with unit tests even before I made my changes; the tests I changed passed and continue to pass after the changes)

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/CredentialManagement/CredentialProfileOptions.cs
+++ b/sdk/src/Core/Amazon.Runtime/CredentialManagement/CredentialProfileOptions.cs
@@ -122,6 +122,11 @@ namespace Amazon.Runtime.CredentialManagement
         public string SsoRegion { get; set; }
 
         /// <summary>
+        /// The SSO scopes that are provided for authorization when using AWS SSO.
+        /// </summary>
+        public string SsoRegistrationScopes { get; set; }
+
+        /// <summary>
         /// The corresponding IAM role in the AWS account that temporary AWS credentials will be resolved for using AWS SSO.
         /// </summary>
         public string SsoRoleName { get; set; }
--- a/sdk/src/Core/Amazon.Runtime/CredentialManagement/CredentialProfileOptions.cs
+++ b/sdk/src/Core/Amazon.Runtime/CredentialManagement/CredentialProfileOptions.cs
@@ -194,6 +194,7 @@ namespace Amazon.Runtime.CredentialManagement
 #if !BCL35
                 ", " + "SsoAccountId=" + SsoAccountId +
                 ", " + "SsoRegion=" + SsoRegion +
+                ", " + "SsoRegistrationScopes=" + SsoRegistrationScopes +
                 ", " + "SsoRoleName=" + SsoRoleName +
                 ", " + "SsoStartUrl=" + SsoStartUrl +
                 ", " + "SsoSession=" + SsoSession +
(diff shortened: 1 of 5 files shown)


sdk/src/Core/Amazon.Runtime/CredentialManagement/CredentialProfileOptions.cs
```diff
@@ -121,6 +121,11 @@ public class CredentialProfileOptions
         /// </summary>
         public string SsoRegion { get; set; }
 
+        /// <summary>
+        /// The SSO scopes that are provided for authorization when using AWS SSO.
+        /// </summary>
+        public string SsoRegistrationScopes { get; set; }
```
normj, line 127: Can you update the `ToString`, `Equals` and `HashCode` methods to take include this new property.

normj (changes requested):

sdk/src/Core/Amazon.Runtime/CredentialManagement/CredentialProfileOptions.cs
floralph: Done

normj (approved):

sdk/src/Core/Amazon.Runtime/CredentialManagement/CredentialProfileOptions.cs
ashovlin: And can you update `IsEmpty` as well?

sdk/src/Core/Amazon.Runtime/CredentialManagement/NetSDKCredentialsFile.cs
```diff
@@ -63,6 +63,7 @@ public class NetSDKCredentialsFile : ICredentialProfileStore
 
         private const string SsoAccountId = "sso_account_id";
         private const string SsoRegion = "sso_region";
+        private const string SsoRegistrationScopes = "sso_registration_scopes";
```
ashovlin, line 66: Can you add this to `ReservedPropertyNames` alongside the others for SSO?

sdk/src/Core/Amazon.Runtime/CredentialManagement/CredentialProfileOptions.cs
floralph: Done

sdk/src/Core/Amazon.Runtime/CredentialManagement/NetSDKCredentialsFile.cs
floralph: Done

ashovlin (approved):

ashovlin:
This has been released in https://www.nuget.org/packages/AWSSDK.Core/3.7.203 via https://github.com/aws/aws-sdk-net/commit/68aef4e681fda621f54058d1f2240306cbc72352, https://github.com/aws/aws-sdk-net/commit/09346de46a68dde4bf648d6a5eb2f55b0f5b8498, and https://github.com/aws/aws-sdk-net/commit/0de21ad3e27b23f35a7411582ced81640798a2cd. The commit hashes changed when we rebased during the release, so GitHub didn't auto-close.

Thanks for the contribution!

PR #3070, opened 2023-10-16
Normj/net8 ci changes

normj:
## Description
Changes to make the `feature/net8` build, test and pack in the CI system.

* Add DevConfig file with the ignore list for backwards compact check.
* Add `net8.0` as target framework for test projects.
* Update XUnit and Microsoft.NET.Test.Sdk versions. The .NET 8.0 RC2 install was not working with the old versions.
* Bug fix for unit and integ tests that were failing after running the full test suite for the first time on the `feature/net8.0` branch.

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Internal/InternalConfiguration.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/InternalConfiguration.cs
@@ -182,9 +182,17 @@ namespace Amazon.Runtime.Internal
                 {
                     convertedValue = int.Parse(value);
                 }
+                else if (typeof(T) == typeof(long))
+                {
+                    convertedValue = long.Parse(value);
+                }
                 else if (typeof(T).IsEnum)
                 {
-                    convertedValue = Enum.Parse(typeof(T), value);
+                    convertedValue = Enum.Parse(typeof(T), value, true);
+                }
+                else if (typeof(T) == typeof(string))
+                {
+                    convertedValue = value.ToString();
                 }
                 else
                 {
--- a/sdk/test/NetStandard/IntegrationTests/IntegrationTests/Route53Domains.cs
+++ b/sdk/test/NetStandard/IntegrationTests/IntegrationTests/Route53Domains.cs
@@ -45,11 +45,11 @@ namespace Amazon.DNXCore.IntegrationTests
             var domains = (await Client.ListDomainsAsync()).Domains;
             if (domains.Count > 0)
             {
-                await TestTagging(Client, domains[0].DomainName);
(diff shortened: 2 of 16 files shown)


sdk/src/Core/Amazon.Runtime/Internal/InternalConfiguration.cs
```diff
@@ -182,9 +182,17 @@ private bool TryGetEnvironmentVariable(string environmentVariableName, out strin
                 {
                     convertedValue = int.Parse(value);
                 }
+                else if (typeof(T) == typeof(long))
```
normj, line 185: These changes were bug fixes after I change this code in previous PR to longer use the [TypeDescriptor](https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Core/Amazon.Runtime/Internal/InternalConfiguration.cs#L177C29-L177C43) because that wasn't compatible for trimming. I looked where these private method was being used and confirmed these are the only types used for the generic `T`. Actually the `string` is not used but I added it for completeness.

sdk/test/NetStandard/IntegrationTests/IntegrationTests/Route53Domains.cs
```diff
@@ -45,11 +45,11 @@ public async Task TestTagging()
             var domains = (await Client.ListDomainsAsync()).Domains;
             if (domains.Count > 0)
             {
-                await TestTagging(Client, domains[0].DomainName);
+                await TestTaggingAsync(Client, domains[0].DomainName);
             }
         }
 
-        private async Task TestTagging(AmazonRoute53DomainsClient client, string domain)
+        private async Task TestTaggingAsync(AmazonRoute53DomainsClient client, string domain)
```
normj, line 52: I renamed this with the `Async` because for some reason the new version of XUnit didn't like that there was a test method and a private method with the same name in the same class.

sdk/test/Services/S3/UnitTests/Custom/EncryptionChangeTest.cs
```diff
@@ -23,7 +23,7 @@ namespace AWSSDK.UnitTests
     [TestClass]
     public class EncryptionChangeTest
     {
-        private const string ExpectedHash = "4D0CB6D89DD6320ED4F554B281717800D1554FB438E5768380243BBBD8BAB2FC";
+        private const string ExpectedHash = "7B226418F7D7877680A0E2C10E09AE59CA44E9BC1DBB6A2C10E7E8C0872C4489";
```
normj, line 26: The file this test does a hash check was changed because it has new `#if` for .NET 8.

dscpinheiro (approved):

PR #3076, opened 2023-10-23
For .NET 8 use HttpClient's sync Send method

normj:
The .NET Core request pipeline has always had both a sync and async code path. Since .NET Core did not originally have a sync Send method on the `HttpClient` the pipeline would call the async code path and then block. Now that we are adding a .NET 8 target we can use the new sync `Send` method on the `HttpClient` instead of the blocking async approach.

A later discussion can be head on whether we should make the service methods public for .NET 8.

CI dry run was successful for this branch.

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_netstandard/HttpRequestMessageFactory.cs
+++ b/sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_netstandard/HttpRequestMessageFactory.cs
@@ -507,16 +507,6 @@ namespace Amazon.Runtime
 
                 throw;
             }
-            catch (OperationCanceledException canceledException)
-            {
-                if (canceledException.InnerException != null)
-                {
-                    throw canceledException.InnerException;
-                }
-
-                throw;
-            }
-
         }
 #else
         public IWebResponseData GetResponse()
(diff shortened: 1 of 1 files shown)


sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_netstandard/HttpRequestMessageFactory.cs
```diff
@@ -483,6 +483,42 @@ public HttpContent GetRequestContent()
...
+
+                    if (httpException.InnerException is WebException)
+                        throw httpException.InnerException;
+                }
+
+                throw;
+            }
+            catch (OperationCanceledException canceledException)
+            {
+                if (canceledException.InnerException != null)
```
boblodgett, line 512: Does this need to do a check if it isn't an actual cancellation request before pulling out the inner exception? Example from below: https://github.com/aws/aws-sdk-net/pull/3076/files#diff-2c03c359ab65379b0eb91fd4fb6e2f7833cf3a0a380dacce9a053f8bd2d4b4a0R575

My concern is unwrapping it and throwing can hide IsCancellationRequest value on the rethrown inner exception when it should never have been unwrapped.

normj: There is no `cancellationToken` in the sync code path. In the code you linked before the `GetResponseAsync` takes in a `CancellationToken` but the sync `GetResponse` does not pass in a `CancellationToken`. That is why I removed that code from the catch block. I don't think I have every seen sync APIs take in `CancellationToken`.

dscpinheiro (approved):

boblodgett (approved):

PR #3083, opened 2023-10-27
Petesong/sso token fix

peterrsongg:
## Description
This adds `RegistrationExpiresAt` to the SSOToken that is returned when `GetTokenAsync` is called. This value is then used to generate a new SSO token if the registrationExpiresAt is within 5 minutes of expiration.

## Motivation and Context
This change is for a better user experience within the VSToolkit for AWS

## Testing
Dry Run successful

## Types of changes

## Checklist

## License

### DIFF
--- a/generator/.DevConfigs/petesong-sso-token-fix.json
+++ b/generator/.DevConfigs/petesong-sso-token-fix.json
@@ -1,0 +1,14 @@
+{
+  "core":{
+    "changeLogMessages":["Update SSO Token Manager to generate a new token if the token's client registration has expired."],
+    "type": "patch",
+    "updateMinimum":true
+  },
+  "services": [
+    {
+      "serviceName": "SSOOIDC",
+      "type": "patch",
+      "changeLogMessages": ["Add RegistrationExpiresAt field to GetSsoTokenResponse."]
+    }
+  ]
+}
--- a/sdk/src/Core/Amazon.Runtime/Credentials/Internal/_bcl45+netstandard/SSOTokenManager.cs
+++ b/sdk/src/Core/Amazon.Runtime/Credentials/Internal/_bcl45+netstandard/SSOTokenManager.cs
@@ -144,7 +144,11 @@ namespace Amazon.Runtime.Credentials.Internal
                     {
                         throw new AmazonClientException("SSO Token has expired and can not be refreshed");
                     }
-
+                    //if registration is within 5 minutes of expiration, generate a new token
+                    if (ssoToken.RegisteredClientExpired() && options.SupportsGettingNewToken)
+                    {
+                        return GenerateNewToken(options);
+                    }
                     // did we recently try to refresh the token?
                     if (true == inMemoryToken?.RefreshState.IsInRefreshCoolDown())
                     {
@@ -332,7 +336,11 @@ namespace Amazon.Runtime.Credentials.Internal
                     {
                         throw new AmazonClientException("SSO Token has expired and can not be refreshed");
                     }
-
+                    //if registration is within 5 minutes of expiration, generate a new token
+                    if (ssoToken.RegisteredClientExpired() && options.SupportsGettingNewToken)
+                    {
+                        return await GenerateNewTokenAsync(options, cancellationToken).ConfigureAwait(false);
+                    }
                     // did we recently try to refresh the token?
                     if (true == inMemoryToken?.RefreshState.IsInRefreshCoolDown())
                     {
--- a/sdk/src/Core/Amazon.Runtime/Credentials/Internal/_bcl45+netstandard/SsoTokenUtils.cs
+++ b/sdk/src/Core/Amazon.Runtime/Credentials/Internal/_bcl45+netstandard/SsoTokenUtils.cs
@@ -92,6 +92,21 @@ namespace Amazon.Runtime.Credentials.Internal
             return ToJson(token);
         }
 
+        /// <summary>
+        /// This determines whether the <seealso cref="SsoToken.RegistrationExpiresAt"/> field is 5 minutes within expiration
+        /// </summary>
+        /// <param name="token">The sso token</param>
+        /// <returns>This returns true if <seealso cref="SsoToken.RegistrationExpiresAt"/> is within 5 minutes of expiration. False otherwise</returns>
+        public static bool RegisteredClientExpired(this SsoToken token)
+        {
+            if (null == token)
+                throw new ArgumentNullException(nameof(token));
+            DateTime dateTime = ConvertRFC3339StringToDateTime(token.RegistrationExpiresAt);
+#pragma warning disable CS0618 // Type or member is obsolete
+            return AWSSDKUtils.CorrectedUtcNow >= dateTime.AddMinutes(-5);
+#pragma warning restore CS0618 // Type or member is obsolete               
+        }
+
         #endregion
 
         /// <summary>
@@ -162,5 +177,21 @@ namespace Amazon.Runtime.Credentials.Internal
             
             return token;
         }
+        #region private methods
+        /// <summary>
+        /// This method takes in an RFC3339 string representing a datetime and converts it to a DateTime object 
+        /// </summary>
+        /// <param name="stringFormattedDate">The RFC3339 formatted string</param>
+        /// <returns>Returns the datetime object serialized to UTC</returns>
+        /// <exception cref="ArgumentNullException"></exception>
+        private static DateTime ConvertRFC3339StringToDateTime(string stringFormattedDate)
+        {
+            if (string.IsNullOrEmpty(stringFormattedDate))
+            {
+                throw new ArgumentNullException(nameof(stringFormattedDate));
+            }
+            return XmlConvert.ToDateTime(stringFormattedDate, XmlDateTimeSerializationMode.Utc);
+        }
+        #endregion
     }
 }
--- a/sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
+++ b/sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
@@ -172,10 +172,9 @@ namespace Amazon.Util
         /// The RFC822Date Format string. Used when parsing date objects
         /// </summary>
         public const string RFC822DateFormat = "ddd, dd MMM yyyy HH:mm:ss \\G\\M\\T";
-
 #endregion
 
-#region Internal Methods
+        #region Internal Methods
 
         /// <summary>
         /// Returns an extension of a path.
--- a/sdk/src/Services/SSOOIDC/Custom/Internal/_bcl45+netstandard/CoreAmazonSSOOIDC.cs
+++ b/sdk/src/Services/SSOOIDC/Custom/Internal/_bcl45+netstandard/CoreAmazonSSOOIDC.cs
@@ -17,6 +17,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using System.Xml;
 using Amazon.Runtime;
 using Amazon.Runtime.Internal.Util;
 using Amazon.Runtime.SharedInterfaces;
@@ -101,13 +102,14 @@ namespace Amazon.SSOOIDC.Internal
                 startDeviceAuthorizationResponse.Interval,
                 deviceCodeExpiration,
                 context);
-
+            var clientSecretExpiresAtString = XmlConvert.ToString(AWSSDKUtils.ConvertFromUnixEpochSeconds((int)registerClientResponse.ClientSecretExpiresAt), XmlDateTimeSerializationMode.Utc);
             return new GetSsoTokenResponse
             {
                 AccessToken = ssoToken.AccessToken,
                 Region = client.Config.RegionEndpoint.SystemName,
                 ClientId = registerClientResponse.ClientId,
                 ClientSecret = registerClientResponse.ClientSecret,
+                RegistrationExpiresAt = clientSecretExpiresAtString,
                 RefreshToken = ssoToken.RefreshToken,
                 ExpiresAt = DateTime.UtcNow.AddSeconds(ssoToken.ExpiresIn),
                 StartUrl = request.StartUrl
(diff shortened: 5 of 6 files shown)


awschristou (commented):
Looks good to me, I have a couple of comments that will help the longer term maintenance of this code.

sdk/src/Services/SSOOIDC/Custom/Internal/_bcl45+netstandard/CoreAmazonSSOOIDC.cs
```diff
@@ -176,17 +178,20 @@ public static async Task<GetSsoTokenResponse> GetSsoTokenAsync(IAmazonSSOOIDC cl
...
-
+            string format = "yyyy-MM-ddTHH:mm:ss.fffZ";
+            var clientSecretExpiresAtString = AWSSDKUtils.ConvertFromUnixEpochSeconds((int)registerClientResponse.ClientSecretExpiresAt).ToString(format);
             return new GetSsoTokenResponse
             {
                 AccessToken = ssoToken.AccessToken,
                 Region = client.Config.RegionEndpoint.SystemName,
                 ClientId = registerClientResponse.ClientId,
                 ClientSecret = registerClientResponse.ClientSecret,
+                RegistrationExpiresAt = clientSecretExpiresAtString,
```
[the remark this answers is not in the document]
peterrsongg: I think it's a good idea to do that, and more maintanable:
```
XmlConvert.ToString(AWSSDKUtils.ConvertFromUnixEpochSeconds((int)registerClientResponse.ClientSecretExpiresAt), XmlDateTimeSerializationMode.Utc);
```
is what I had in mind.

sdk/src/Core/Amazon.Runtime/Credentials/Internal/_bcl45+netstandard/SsoTokenUtils.cs
```diff
@@ -162,5 +179,17 @@ public static SsoToken FromJson(string json)
...
+        #region private methods
+        //The string formatted date which comes in RFC3339 format with no milliseconds
+        private static bool ConvertStringToDateTime(string stringFormattedDate, out DateTime dateTime)
+        {
+            dateTime = new DateTime();
+            if (stringFormattedDate == null)
+            {
+                return false;
+            }
+            return DateTime.TryParseExact(stringFormattedDate, AWSSDKUtils.RFC3339NoMsFormat, null, System.Globalization.DateTimeStyles.AssumeUniversal, out dateTime);
```
awschristou, line 191: Should this use `XmlConvert.ToDateTime` to give symmetry to the `XmlConvert.ToString` saving code over in CoreAmazonSSOOIDC.cs?

That would match [Sso Token loading](https://github.com/aws/aws-sdk-net/blob/8c6a77b03b0f967388d57e912bc67238f2c7ab19/sdk/src/Core/Amazon.Runtime/Credentials/Internal/_bcl45%2Bnetstandard/SsoTokenUtils.cs#L141)

ashovlin:
I cannot comment on the line directly since it's not part of the diff, but do we need to guard these locations against attempting to generate a new token with an expired registration as well?

https://github.com/aws/aws-sdk-net/blob/9be0493c0eafb30a7bec5f5ebc179a6eb12b2672/sdk/src/Core/Amazon.Runtime/Credentials/Internal/_bcl45%2Bnetstandard/SSOTokenManager.cs#L161

https://github.com/aws/aws-sdk-net/blob/9be0493c0eafb30a7bec5f5ebc179a6eb12b2672/sdk/src/Core/Amazon.Runtime/Credentials/Internal/_bcl45%2Bnetstandard/SSOTokenManager.cs#L353

sdk/src/Core/Amazon.Runtime/Credentials/Internal/_bcl45+netstandard/SsoTokenUtils.cs
```diff
@@ -92,6 +92,23 @@ public static string AsJson(this SsoToken token)
             return ToJson(token);
         }
 
+        /// <summary>
+        /// This determines whether the SSO token's registrationExpiresAt field is 5 minutes within expiration
+        /// </summary>
+        /// <param name="token"></param>
+        /// <returns></returns>
```
ashovlin, line 99: Fill these out.

sdk/src/Core/Amazon.Runtime/Credentials/Internal/_bcl45+netstandard/SsoTokenUtils.cs
```diff
@@ -92,6 +92,23 @@ public static string AsJson(this SsoToken token)
...
+        public static bool RegisteredClientExpired(this SsoToken token)
```
ashovlin, line 100: nit: It's not _technically_ expired yet if it's within the 5 minute window. Just don't want to mislead somebody attempting to call this in the future if they have a stricter use case. `IsClientRegistrationAboutToExpire` or something like that instead is more accurate.

sdk/src/Core/Amazon.Runtime/Credentials/Internal/_bcl45+netstandard/SsoTokenUtils.cs
```diff
@@ -92,6 +92,23 @@ public static string AsJson(this SsoToken token)
...
+        {
+            DateTime dateTime;
+            if(ConvertStringToDateTime(token.RegistrationExpiresAt,out dateTime))
+            {
+#pragma warning disable CS0618 // Type or member is obsolete
+                return AWSSDKUtils.CorrectedUtcNow >= dateTime.AddMinutes(-5);
```
ashovlin, line 106: Just curious, how'd you arrive at 5 minutes?

ashovlin (changes requested):

awschristou:
> I cannot comment on the line directly since it's not part of the diff, but do we need to guard these locations against attempting to generate a new token with an expired registration as well?
> 
> https://github.com/aws/aws-sdk-net/blob/9be0493c0eafb30a7bec5f5ebc179a6eb12b2672/sdk/src/Core/Amazon.Runtime/Credentials/Internal/_bcl45%2Bnetstandard/SSOTokenManager.cs#L161
> 
> https://github.com/aws/aws-sdk-net/blob/9be0493c0eafb30a7bec5f5ebc179a6eb12b2672/sdk/src/Core/Amazon.Runtime/Credentials/Internal/_bcl45%2Bnetstandard/SSOTokenManager.cs#L353

I don't believe so. In those cases, we're already going to request a new token, which performs a new client registration.

sdk/src/Core/Amazon.Runtime/Credentials/Internal/_bcl45+netstandard/SsoTokenUtils.cs
peterrsongg: I will see if it's possible to use `XmlConvert.ToDateTime`, if it doesn't complain about formatting then I will use that for better maintainability.

generator/.DevConfigs/petesong-sso-token-fix.json
```diff
@@ -0,0 +1,9 @@
+{
+    "services": [
+      {
+        "serviceName": "SSO OIDC",
```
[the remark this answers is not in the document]
peterrsongg: I ran the dry-run immediately after pushing a new revision and it failed on the merge step. I think it might've been because of the service name having the space but I'll need to see.

sdk/src/Core/Amazon.Runtime/Credentials/Internal/_bcl45+netstandard/SSOTokenManager.cs
```diff
@@ -332,7 +336,11 @@ public async Task<SsoToken> GetTokenAsync(SSOTokenManagerGetTokenOptions options
                     {
                         throw new AmazonClientException("SSO Token has expired and can not be refreshed");
                     }
-
+                    //if registration is within 5 minutes of expiration, generate a new token
+                    if (ssoToken.RegisteredClientExpired() && options.SupportsGettingNewToken)
+                    {
+                        return GenerateNewToken(options);
```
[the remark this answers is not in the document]
peterrsongg: Good catch, I did do the dry run immediately after pushing the revision up expecting it to succeed but it failed. I'll get it to succeed first.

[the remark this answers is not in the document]
ashovlin: When working in Core, I'll ensure that the build + tests pass locally for the slimmed down `AWSSDK.CoreAndCustomUnitTests.Net45` and `AWSSDK.CoreAndCustomUnitTests.NetStandard` solutions. Fast feedback cycle than waiting on dry runs :-)

sdk/src/Core/Amazon.Runtime/Credentials/Internal/_bcl45+netstandard/SsoTokenUtils.cs
```diff
@@ -162,5 +179,17 @@ public static SsoToken FromJson(string json)
             
             return token;
         }
+        #region private methods
+        //The string formatted date which comes in RFC3339 format with no milliseconds
```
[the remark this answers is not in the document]
peterrsongg: I used just the double slash here because this was a private method. But I will change it

sdk/src/Core/Amazon.Runtime/Credentials/Internal/_bcl45+netstandard/SsoTokenUtils.cs
peterrsongg: My thought process was similar to how we refresh credentials if `PreEmptExpiryTime` is within 15 minutes of expiration. Also in SsoTokenUtils, it uses 6 minutes to determine if the token should refresh, so I just did something similar.

sdk/src/Core/Amazon.Runtime/Credentials/Internal/_bcl45+netstandard/SsoTokenUtils.cs
peterrsongg: While it's true that it is not technically expired, I think the summary with the intellisense is enough for the user to know it's within 5 minutes.

sdk/src/Core/Amazon.Runtime/Credentials/Internal/_bcl45+netstandard/SsoTokenUtils.cs
peterrsongg: After testing, I decided to leave it as `TryParseExact`. `XmlConvert.ToDateTime` is converting it incorrectly, (for example, when the test says it expires in 1 hour, xmlConvert.ToDateTime return something that expires in 5 hours). I think it is because the original `ssoToken.RegistrationExpiresAt` was converted from a timestamp in Unix Epoch time but not completely certain. But to be safe I want to be exact with the parsing and leave it as is.

awschristou: That sounds like some additional "Kind" handling is needed. You're working in a time-zone 4 hours off UTC, right? Add in the additional hour and that explains your 5 hour discrepancy. 

My concern is that the field doesn't have parity with how the other datetime field is de/serialized with this implementation.

peterrsongg: So the issue was when I made the `XmlConvert.ToDateTime` change in the code , I didn't also use `XmlConvert.ToDateTime` in the test case, causing the mismatch. Once I did that it passed all cases. So in the next revision I will push that out.

sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
```diff
@@ -172,10 +172,13 @@ private static string DetermineValidPathCharacters()
         /// The RFC822Date Format string. Used when parsing date objects
         /// </summary>
         public const string RFC822DateFormat = "ddd, dd MMM yyyy HH:mm:ss \\G\\M\\T";
-
+        /// <summary>
+        /// The RFC3339 format string. Used when parsing date objects
+        /// </summary>
+        public const string RFC3339NoMsFormat = "yyyy-MM-ddTHH:mm:ssZ";
```
awschristou, line 178: This property is no longer used

awschristou (commented):
Looks good to me

sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
peterrsongg: removed.

peterrsongg:
> > I cannot comment on the line directly since it's not part of the diff, but do we need to guard these locations against attempting to generate a new token with an expired registration as well?
> > https://github.com/aws/aws-sdk-net/blob/9be0493c0eafb30a7bec5f5ebc179a6eb12b2672/sdk/src/Core/Amazon.Runtime/Credentials/Internal/_bcl45%2Bnetstandard/SSOTokenManager.cs#L161
> > 
> > https://github.com/aws/aws-sdk-net/blob/9be0493c0eafb30a7bec5f5ebc179a6eb12b2672/sdk/src/Core/Amazon.Runtime/Credentials/Internal/_bcl45%2Bnetstandard/SSOTokenManager.cs#L353
> 
> I don't believe so. In those cases, we're already going to request a new token, which performs a new client registration.

yup in this case the new token would have a new client registration

sdk/src/Core/Amazon.Runtime/Credentials/Internal/_bcl45+netstandard/SsoTokenUtils.cs
ashovlin: Okay now that you added better XML docs. 👍

ashovlin (approved):

normj (approved):

sdk/src/Core/Amazon.Runtime/Credentials/Internal/_bcl45+netstandard/SsoTokenUtils.cs
```diff
@@ -92,6 +92,21 @@ public static string AsJson(this SsoToken token)
...
+        /// <summary>
+        /// This determines whether the <seealso cref="SsoToken.RegistrationExpiresAt"/> field is 5 minutes within expiration
+        /// </summary>
+        /// <param name="token">The sso token</param>
+        /// <returns>This returns true if <seealso cref="SsoToken.RegistrationExpiresAt"/> is within 5 minutes of expiration. False otherwise</returns>
+        public static bool RegisteredClientExpired(this SsoToken token)
+        {
+            if (null == token)
+                throw new ArgumentNullException(nameof(token));
+            DateTime dateTime = ConvertRFC3339StringToDateTime(token.RegistrationExpiresAt);
```
awschristou, line 104: @peterrsongg In order to remain compatible with existing cached tokens, or tokens written out by other systems, tokens will be loaded that contain null values for the RegistrationExpiresAt field. In these cases, can we avoid the exception that is thrown in ConvertRFC3339StringToDateTime by early-exiting before this line?

The RegistrationExpiresAt field is optional, so if RegistrationExpiresAt is null, we have two options:

- we can return false and assume the registered client hasn't expired. An attempt to refresh the token using the SDK changes in this PR would succeed if the client registration hasn't actually expired (only the backend knows at this point). At some point, token refreshes may start to fail due to client registration expiry. I believe this enters into the state that we're trying to fix with this change.
- we can return true and assume the registered client has expired. This would cause the code to request an all-new token (if options.SupportsGettingNewToken is true), and in doing so, will register a new device, and update the registration expires at

I recommend option two, but believe that either handling will be an improvement over the exception.

PR #3084, opened 2023-10-28
Update CopyObject and CopyPart requests not to remove leading slash from object keys

### LINKED ISSUE #1720: AmazonS3Client.CopyObject() is incorrectly reporting missing DestinationKey
## Description
When copying an object using `Amazon.S3.AmazonS3Client.CopyObject`, if the `DestinationKey` is set to `"/"`, then an exception is thrown reporting that the value is not set.

## Use Case

I have an actual object "/" (without the quotes) in my bucket that I want copied.

## Reproduction Steps

```csharp
var client = new Amazon.S3.AmazonS3Client(Amazon.RegionEndpoint.USEast1);

var request = new Amazon.S3.Model.CopyObjectRequest()
{
  SourceKey = "/",
  DestinationKey = "/",

  SourceBucket = "bucket1",
  DestinationBucket = "bucket2",
};

try
{
  var response = client.CopyObject(request);
}
catch (Exception e)
{
  Console.WriteLine("{0}", e.Message);
}

Console.WriteLine("Press Enter to continue...");
Console.ReadLine();
```

The above code throws an exception and outputs the following:

```
DestinationKey is a required property and must be set before making this call.
Parameter name: CopyObjectRequest.DestinationKey
Press Enter to continue...
```

### Environment

* SDK Version: AWSSDK.S3 version 3.5.1.9
* Package Version: AWSSDK.S3 version 3.5.1.9
* OS Info: Windows 10
* Build Environment Visual Studio 2019
* Targeted .NET Platform:  4.7.2

## Additional Notes

1. This appears to only happen when the object is "/". Copying of other objects works correctly.

2. This issue occurs without a valid AWS role or key. So it appears to be an SDK validation check before the `s3:CopyObject` API is actually called.

---

This is a :bug: bug-report

dscpinheiro:
## Description
Fixes #1720 (also internal issue `DOTNET-4410`).

## Motivation and Context
This PR updates the `CopyObject` and `CopyPart` marshallers not to remove leading slashes from the specified object keys.

## Testing
- Dry-run: `DRY_RUN-48cd51a1-d13d-4917-a705-49004234b630`
- Added integration tests
- Also tested the example from the original issue with the built `.nupkg` and confirmed it works (using either a bucket and an access point for `SourceBucket`):

```
var client = new AmazonS3Client(RegionEndpoint.USEast1);

var response = client.CopyObject(new CopyObjectRequest
{
    SourceKey = "/",
    DestinationKey = "/",

    SourceBucket = "bucket1",
    DestinationBucket = "bucket2",
});

// Outputs: OK
Console.WriteLine(response.HttpStatusCode);
```

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/Util/AmazonS3Util.cs
+++ b/sdk/src/Services/S3/Custom/Util/AmazonS3Util.cs
@@ -570,13 +570,6 @@ namespace Amazon.S3.Util
                 key.EndsWith(S3Constants.EncryptionInstructionfileSuffixV2, StringComparison.Ordinal);
         }
 
-        internal static string RemoveLeadingSlash(string key)
-        {
-            return key.StartsWith("/", StringComparison.Ordinal)
-                                    ? key.Substring(1)
-                                    : key;
-        }
-
         /// <summary>
         /// Check if the request resource is an outpost resource
         /// </summary>
--- a/sdk/test/Services/S3/IntegrationTests/CopyObjectPartTests.cs
+++ b/sdk/test/Services/S3/IntegrationTests/CopyObjectPartTests.cs
@@ -1,0 +1,110 @@
+using Amazon.S3;
+using Amazon.S3.Model;
+using Amazon.S3.Util;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System.Collections.Generic;
+using System.IO;
+using System.Net;
+
+namespace AWSSDK_DotNet.IntegrationTests.Tests.S3
+{
+    [TestClass]
+    public class CopyObjectPartTests : TestBase<AmazonS3Client>
+    {
+        private const string testContent = "This is the content body!";
+        private const string testKeyWithSlash = "/testsourcekey";
+        private string bucketName;
+
+        [TestInitialize]
+        public void Initialize()
+        {
+            bucketName = S3TestUtils.CreateBucketWithWait(Client);
+
+            Client.PutObject(new PutObjectRequest
+            {
+                BucketName = bucketName,
+                Key = testKeyWithSlash,
+                ContentBody = testContent
+            });
+        }
+
+        [TestCleanup]
+        public void TestCleanup()
+        {
+            AmazonS3Util.DeleteS3BucketWithObjects(Client, bucketName);
+            BaseClean();
+        }
+
+        [TestMethod]
+        [TestCategory("S3")]
+        public void TestCopyObjectPartWithLeadingSlash()
+        {
+            var destinationKeyWithSlash = "/testdestkey";
+            string uploadId = null;
+
+            try
+            {
+                var multiPartUploadResponse = Client.InitiateMultipartUpload(new InitiateMultipartUploadRequest
+                {
+                    BucketName = bucketName,
+                    Key = destinationKeyWithSlash,
+                });
+
+                uploadId = multiPartUploadResponse.UploadId;
+                var copyPartResponse = Client.CopyPart(new CopyPartRequest
+                {
+                    DestinationBucket = bucketName,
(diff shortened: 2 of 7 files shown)


sdk/src/Services/S3/Custom/Util/AmazonS3Util.cs
```diff
@@ -570,13 +570,6 @@ internal static bool IsInstructionFile(string key)
                 key.EndsWith(S3Constants.EncryptionInstructionfileSuffixV2, StringComparison.Ordinal);
         }
 
-        internal static string RemoveLeadingSlash(string key)
```
dscpinheiro, line 573: This method was only being used in the 2 marshallers.

sdk/test/Services/S3/IntegrationTests/CopyObjectPartTests.cs
```diff
@@ -0,0 +1,110 @@
...
+using System.IO;
+using System.Net;
+
+namespace AWSSDK_DotNet.IntegrationTests.Tests.S3
+{
+    [TestClass]
+    public class CopyObjectPartTests : TestBase<AmazonS3Client>
+    {
+        private const string testContent = "This is the content body!";
+        private const string testKeyWithSlash = "/testsourcekey";
```
dscpinheiro, line 15: There were no existing `CopyPart` integration tests, but it was being used in these two places:
- https://github.com/aws/aws-sdk-net/blob/8c6a77b03b0f967388d57e912bc67238f2c7ab19/sdk/test/Services/S3/IntegrationTests/KMSTests.cs#L536
- https://github.com/aws/aws-sdk-net/blob/8c6a77b03b0f967388d57e912bc67238f2c7ab19/sdk/test/Services/S3/IntegrationTests/ChecksumTests.cs#L328

I created this new file to verify the leading `/` works as expected.

boblodgett (approved):

normj (commented):
Can you do a double check and run this scenario using .NET Core? I want to make sure the difference in HttpClient and changes to Uri class don't produce any different behavior.

dscpinheiro:
> Can you do a double check and run this scenario using .NET Core? I want to make sure the difference in HttpClient and changes to Uri class don't produce any different behavior.

Yes, I can. I tried the [Copy an object from one Amazon S3 bucket to another](https://docs.aws.amazon.com/AmazonS3/latest/userguide/example_s3_CopyObject_section.html) and [Get an object from an Amazon S3 bucket](https://docs.aws.amazon.com/AmazonS3/latest/userguide/example_s3_GetObject_section.html) examples from the docs (using `/testfile.txt` instead of `testfile.txt` as the object keys) and my applications in (`net6.0` and `net48`) produced the same results.

afroz429:
We may have to change PowerShell to do the same.

https://github.com/aws/aws-tools-for-powershell/blob/master/modules/AWSPowerShell/Cmdlets/S3/Advanced/Copy-S3Object-Cmdlet.cs#L582

https://github.com/aws/aws-tools-for-powershell/blob/master/modules/AWSPowerShell/Cmdlets/S3/Advanced/AmazonS3Helper.cs#L39

https://github.com/aws/aws-tools-for-powershell/blob/master/modules/AWSPowerShell/Cmdlets/S3/Advanced/Copy-S3Object-Cmdlet.cs#L675

https://github.com/aws/aws-tools-for-powershell/blob/master/modules/AWSPowerShell/Cmdlets/S3/Advanced/Copy-S3Object-Cmdlet.cs#L691

dscpinheiro:
> We may have to change PowerShell to do the same.

Thanks for pointing that out, I'll create a PR in the PowerShell repository to address those changes.

normj (approved):

PR #3085, opened 2023-10-30
Set assembly versions to 3.7.300 for .NET 8 release

normj:
Update the DevConfig file for the .NET 8 release to set 3.7.300.0 so all SDK assemblies will have the same version number at the point .NET 8 support was added.

### DIFF
--- a/generator/.DevConfigs/feature-net8.json
+++ b/generator/.DevConfigs/feature-net8.json
@@ -1,8 +1,11 @@
 {
   "core": {
-    "changeLogMessages": ["Add trimming support for .NET 8."],
-    "type": "patch",
-    "updateMinimum": true,
+    "changeLogMessages": [
+		"Add .NET 8 framework target for core and service projects.",
+		"For .NET 8 mark all SDK assemblies as trimmable and address all trim warnings.",
+		"For .NET 8 prevent the System.Uri class from canonicalizing the resource path. This broke scenarios with S3 object keys like `foo/../bar.txt`.",
+		"For .NET 8 use the System.Net.Http.HttpClient synchronous `Send` method when then internal synchronous service request methods are invoked."
+	],
     "backwardIncompatibilitiesToIgnore": [
       "Amazon.Util.PaginatedResourceFactory/TypeRemoved",
       "Amazon.Runtime.URIBasedRefreshingCredentialHelper/MethodOverloadAdded",
@@ -10,5 +13,6 @@
       "Amazon.Util.PaginatedResourceInfo/TypeRemoved",
       "Amazon.Runtime.Internal.Settings.SettingsCollection/MethodRemoved"
     ]
-  }
+  },
+  "overrideVersion": "3.7.300.0"
 }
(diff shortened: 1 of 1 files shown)


M4urici0GM (approved):

generator/.DevConfigs/feature-net8.json
```diff
@@ -1,14 +1,18 @@
 {
   "core": {
-    "changeLogMessages": ["Add trimming support for .NET 8."],
-    "type": "patch",
-    "updateMinimum": true,
+    "changeLogMessages": [
+		"Add .NET 8 framework target for core and service projects.",
+		"For .NET 8 mark all SDK assemblies as trimmable and address all trim warnings.",
+		"For .NET 8 prevent the System.Uri class from canonicalizing the resource path. This broke scenarios with S3 object keys like `foo/../bar.txt`.",
+		"For .NET 8 use the System.Net.Http.HttpClient synchronous `Send` method when then internal synchronous service request methods are invoked."
```
dscpinheiro, line 7: I don't follow the last changelog message, did you mean:
```suggestion
		"For .NET 8 use the System.Net.Http.HttpClient synchronous `Send` method when their internal synchronous service request methods are invoked."
```

boblodgett: I am thinking `then` should be deleted and replaced with nothing.

generator/.DevConfigs/feature-net8.json
```diff
@@ -1,14 +1,18 @@
...
     "backwardIncompatibilitiesToIgnore": [
       "Amazon.Util.PaginatedResourceFactory/TypeRemoved",
       "Amazon.Runtime.URIBasedRefreshingCredentialHelper/MethodOverloadAdded",
       "Amazon.Util.Internal.ITypeInfo/MethodRemoved",
       "Amazon.Util.PaginatedResourceInfo/TypeRemoved",
       "Amazon.Runtime.Internal.Settings.SettingsCollection/MethodRemoved"
     ]
-  }
+  },
+  "overrideVersion": "3.7.300.0"
```
boblodgett, line 17: Can you confirm that without the `"updateMinimum": true` value that all services were correctly versioned in the dry-run nupkgs?

normj: Yes, I have confirmed via dry runs with this branch that core, services and all of the dependencies for services get updated to the `overrideVersion`. The unit test I added to the CI system tests also test this. Here is the assert code from the unit test.

```
Version expectedVersion = Version.Parse("3.7.2000.0");
Assert.AreEqual(expectedVersion, sdkVersions.CoreVersion);
Assert.AreEqual(expectedVersion, sdkVersions.NewServiceVersion);

Assert.IsTrue(sdkVersions.ServiceVersions.Count > 0);

foreach(var service in sdkVersions.ServiceVersions)
{
	Assert.AreEqual(expectedVersion, service.Value.Version);
	Assert.IsTrue(service.Value.Dependencies.Count > 0);
	foreach (var dependency in service.Value.Dependencies.Keys)
	{
		Assert.AreEqual(expectedVersion, service.Value.Dependencies[dependency]);
	}
}
```

dscpinheiro (approved):

boblodgett (approved):

Issue #3059, opened 2023-09-28
[S3Client] AddPartETags when completing multi-part copy should include all checksum values


dhhoang:
### Describe the bug

I am trying to copy an S3 object (following the [user guide](https://docs.aws.amazon.com/AmazonS3/latest/userguide/CopyingObjectsMPUapi.html)) and set checksum as SHA256:
```csharp
var initResponse = await s3Client.InitiateMultipartUploadAsync(new InitiateMultipartUploadRequest
{
    BucketName = destinationBucket,
    Key = destinationKey,
    ChecksumAlgorithm = ChecksumAlgorithm.SHA256
});
```
However the error occurs when `CompleteMultipartUploadRequest` is called.

### Expected Behavior

The multi-part copy should succeed after all the parts are copied.

### Current Behavior

`CompleteMultipartUpload` failed with the following exception
```
failed: Amazon.S3.AmazonS3Exception: The upload was created using a sha256 checksum. The complete request must include the checksum for each part. It was missing for part 1 in the request.
 ---> Amazon.Runtime.Internal.HttpErrorResponseException: Exception of type 'Amazon.Runtime.Internal.HttpErrorResponseException' was thrown.
```

### Reproduction Steps

Follow the [user guide](https://docs.aws.amazon.com/AmazonS3/latest/userguide/CopyingObjectsMPUapi.html) and set checksum function as SHA256

### Possible Solution

It turns out that `AddPartEtag` only add the `Etag` part of the `CopyPartResponse` object, and not the rest of the checksum:
https://github.com/aws/aws-sdk-net/blob/094b7c7e7c3237b5597647e2264ef4877ba5cec1/sdk/src/Services/S3/Custom/Model/CompleteMultipartUploadRequest.cs#L423

In my opinion this call should add all the checksum fields to the `PartETag` object.

### AWS .NET SDK and/or Package version used

AWSSDK.S3 3.7.205.2

### Targeted .NET Platform

net6.0

### Operating System and version

Windows, Linux, AWS Lambda

ashishdhingra:
Reproducible using customer's code. Below are the observations:
[...]

Issue #3082, opened 2023-10-27
S3 TransferUtility UploadAsync with non seekable stream is not sending metadata or headers


ymor:
### Describe the bug

When using the UploadAsync method on the S3 TransferUtility with an input stream that is non seekable.  Provided metadata is not been stored against the object

### Expected Behavior

The metadata I set on the TransferUtilityUploadRequest  to be stored with the object

### Current Behavior

The metadata provided on the TransferUtilityUploadRequest  is not stored against the object

### Reproduction Steps

Essentially I'm using this example https://github.com/dotnet/AspNetCore.Docs/tree/main/aspnetcore/mvc/models/file-uploads/samples/5.x/LargeFilesSample/Controllers
However instead of storing it to disk I'm passing it to the TransferUtility to upload.  The underlying stream is the HttpRequestStream which is non seekable

### Possible Solution

https://github.com/aws/aws-sdk-net/blob/0015a303eeba3e74c0c5bdf06003182a36a3c582/sdk/src/Services/S3/Custom/Transfer/Internal/_async/MultipartUploadCommand.async.cs#L198
Construct the InitiateMultipartUploadRequest with the metadata provided in the  TransferUtilityUploadRequest 

### AWS .NET SDK and/or Package version used

AWSSDK.S3 3.7.205.15

### Targeted .NET Platform

.net 7

### Operating System and version

Ubuntu

ashishdhingra:
Reproducible with below code:
```C#
string bucketName = "some-testbucket";

using (AmazonS3Client amazonS3Client = new AmazonS3Client(RegionEndpoint.USEast2))
{
    // Stream returned by HttpClient.GetStreamAsync() is non-seekable
    HttpClient hc = new HttpClient();
    ServicePointManager.SecurityProtocol = SecurityProtocolType.Tls12;
    Stream nonSeekableWebStream = await hc.GetStreamAsync("https://www.google.com");

    TransferUtilityUploadRequest transferUtilityUploadRequest = new TransferUtilityUploadRequest()
    {
        BucketName = bucketName,
        Key = "testfile1.txt",
        InputStream = nonSeekableWebStream
    };

    transferUtilityUploadRequest.Metadata.Add("testmetadata", "testvalue");
    TransferUtility transferUtility = new TransferUtility(amazonS3Client);
    await transferUtility.UploadAsync(transferUtilityUploadRequest); // The uploaded file doesn't has metadata

    // Get seekable stream
[...]

PR #3090, opened 2023-11-02
Fixed an issue where TransferUtility UploadAsync() is not sending metadata or headers with non-seekable stream.

ashishdhingra:
## Description
Fixed an issue where `TransferUtility`'s `UploadAsync()` is not sending metadata or headers with **non-seekable** stream.

## Motivation and Context
Issue: https://github.com/aws/aws-sdk-net/issues/3082

## Changes:
The support for handling non-seekable stream via `TransferUtility` was added in commit https://github.com/aws/aws-sdk-net/commit/305e5c9b70600d89cd54e4876fcef5fa277ea9e2. The logic added does not reuse existing code which already handled adding headers and metadata to `InitiateMultipartUploadRequest`. Instead, for non-seekable steam, it only uses bucket and key [here](https://github.com/aws/aws-sdk-net/blob/0015a303eeba3e74c0c5bdf06003182a36a3c582/sdk/src/Services/S3/Custom/Transfer/Internal/_async/MultipartUploadCommand.async.cs#L198).

The existing [ConstructInitiateMultipartUploadRequest()](https://github.com/aws/aws-sdk-net/blob/488ac37fc1daceb9fad4e7e03a79edbc07312b72/sdk/src/Services/S3/Custom/Transfer/Internal/MultipartUploadCommand.cs#L233) adds `Metadata` and `Headers` in addition to other properties set by user.

This PR includes the following changes:
- Refactors existing code to reuse existing `ConstructInitiateMultipartUploadRequest()` and `ConstructCompleteMultipartUploadRequest()` methods. These are shared between BCL35 (synchronous) and BCL45 (and `async`) versions.
- For non-seekable streams, instead of tracking `PartETag` list for uploaded parts, it would track `UploadPartResponse`. This is done for 2 reasons:
  - For reusability
  - The current logic for non-seekable streams only uses `PartNumber` and `ETag` properties. The [PartETag](https://github.com/aws/aws-sdk-net/blob/488ac37fc1daceb9fad4e7e03a79edbc07312b72/sdk/src/Services/S3/Custom/Model/PartEtag.cs#L31) object also has other **Checksum** related properties, one of which is set if `InitiateMultipartUploadRequest.ChecksumAlgorithm` is set. In such scenario, one of these checksum values must be set in `CompleteMultipartUploadRequest`, else error is returned by S3 service (refer fixed issue https://github.com/aws/aws-sdk-net/issues/3059).
- Adds new helper method `ConstructUploadPartRequestForNonSeekableStream()` in [MultipartUploadCommand](https://github.com/aws/aws-sdk-net/blob/488ac37fc1daceb9fad4e7e03a79edbc07312b72/sdk/src/Services/S3/Custom/Transfer/Internal/MultipartUploadCommand.cs#L40). The logic is shared between `BCL35` and `async` versions.
- Fixes an issue in BCL35 version for non-seekable stream where in [UploadUnseekableStream()](https://github.com/aws/aws-sdk-net/blob/488ac37fc1daceb9fad4e7e03a79edbc07312b72/sdk/src/Services/S3/Custom/Transfer/Internal/_bcl35/MultipartUploadCommand.bcl35.cs#L107), `CompleteMultipartUpload()` is incorrectly placed inside [finally{}](https://github.com/aws/aws-sdk-net/blob/488ac37fc1daceb9fad4e7e03a79edbc07312b72/sdk/src/Services/S3/Custom/Transfer/Internal/_bcl35/MultipartUploadCommand.bcl35.cs#L174C35-L174C58) block. The `finally` block would be invoked in both success and exception conditions. The `CompleteMultipartUpload()` should not be called in exception conditions. Instead, for exception condition, `AbortMultipartUpload()` should be called (which was correctly invoked in outer finally block [here](https://github.com/aws/aws-sdk-net/blob/488ac37fc1daceb9fad4e7e03a79edbc07312b72/sdk/src/Services/S3/Custom/Transfer/Internal/_bcl35/MultipartUploadCommand.bcl35.cs#L189C27-L189C47)).

**Note for reviewer:**
In inner `finally` for both `BCL35` and `async` versions, call to `nextUploadBuffer.Dispose();` is added. I'm unsure if this object should be disposed in `while` loop earlier where this object is re-initialized to new `MemoryStream` object (unsure if I could re-assign new object to a disposed object instance; didn't test it).

## Testing
- Added test case.
  - Tested locally on via Test Explorer for both `BCL35` and `BCL45` targets.
- Dry run completed successfully.

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/Transfer/Internal/MultipartUploadCommand.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/Internal/MultipartUploadCommand.cs
@@ -154,10 +154,18 @@ namespace Amazon.S3.Transfer.Internal
 
         private CompleteMultipartUploadRequest ConstructCompleteMultipartUploadRequest(InitiateMultipartUploadResponse initResponse)
         {
-            if(this._uploadResponses.Count != this._totalNumberOfParts)
+            return ConstructCompleteMultipartUploadRequest(initResponse, false, null);
+        }
+
+        private CompleteMultipartUploadRequest ConstructCompleteMultipartUploadRequest(InitiateMultipartUploadResponse initResponse, bool skipPartValidation, RequestEventHandler requestEventHandler)
+        {
+            if (!skipPartValidation)
             {
-                throw new InvalidOperationException($"Cannot complete multipart upload request. The total number of completed parts ({this._uploadResponses.Count}) " +
-                    $"does not equal the total number of parts created ({this._totalNumberOfParts}).");
+                if (this._uploadResponses.Count != this._totalNumberOfParts)
+                {
+                    throw new InvalidOperationException($"Cannot complete multipart upload request. The total number of completed parts ({this._uploadResponses.Count}) " +
+                        $"does not equal the total number of parts created ({this._totalNumberOfParts}).");
+                }
             }
 
             var compRequest = new CompleteMultipartUploadRequest()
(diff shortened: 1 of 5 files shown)


sdk/src/Services/S3/Custom/Transfer/Internal/MultipartUploadCommand.cs
```diff
@@ -230,7 +247,43 @@ private UploadPartRequest ConstructUploadPartRequest(int partNumber, long filePo
...
+#if (BCL && !BCL45)
+                                    Timeout = ClientConfig.GetTimeoutValue(this._config.DefaultTimeout, this._fileTransporterRequest.Timeout),
+#endif
+                DisableMD5Stream = this._fileTransporterRequest.DisableMD5Stream,
+                DisablePayloadSigning = this._fileTransporterRequest.DisablePayloadSigning,
+                ChecksumAlgorithm = this._fileTransporterRequest.ChecksumAlgorithm,
+                CalculateContentMD5Header = this._fileTransporterRequest.CalculateContentMD5Header
+            };
+
+            // If the InitiateMultipartUploadResponse indicates that this upload is using KMS, force SigV4 for each UploadPart request
```
peterrsongg, line 273: Minor: Will defer to norm here. Do we need to add the progress Handler here like we do for regular streams?

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartUploadCommand.cs
```diff
@@ -230,7 +247,43 @@ private UploadPartRequest ConstructUploadPartRequest(int partNumber, long filePo
             return uploadRequest;
         }
 
+        private UploadPartRequest ConstructUploadPartRequestForNonSeekableStream(Stream inputStream, int partNumber, long partSize, bool isLastPart, InitiateMultipartUploadResponse initiateResponse)
+        {
+            UploadPartRequest uploadPartRequest = new UploadPartRequest()
```
normj, line 252: It would be good to refactor this a bit more so there is a method to setup the generic `UploadPartRequest` that also sets the SigV4 and then have the callers just add the settings that are specific to either streaming or not. We know we keep adding properties to the S3 upload and it seems like these methods will go out of sync.

sdk/src/Services/S3/Custom/Transfer/Internal/_async/MultipartUploadCommand.async.cs
```diff
@@ -240,22 +238,16 @@ private async Task UploadNonSeekableStreamAsync(TransferUtilityUploadRequest req
...
-                                    IsLastPart = isLastPart
-                                }).ConfigureAwait(false);
+                                UploadPartRequest uploadPartRequest = ConstructUploadPartRequestForNonSeekableStream(nextUploadBuffer, partNumber, partSize, isLastPart, initiateResponse);
+
+                                var partResponse = await _s3Client.UploadPartAsync(uploadPartRequest).ConfigureAwait(false);
                                 Logger.DebugFormat($"Uploaded part {partNumber}. (Last part = {isLastPart}, Part size = {partSize}, Upload Id: {initiateResponse.UploadId}");
-                                partETags.Add(new PartETag { PartNumber = partResponse.PartNumber, ETag = partResponse.ETag });
+                                uploadPartResponses.Add(partResponse);
                                 partNumber++;
                                 nextUploadBuffer = new MemoryStream(partBuffer);
```
normj, line 247: We are creating a new `MemoryStream` without disposing the previous `MemoryStream`.

sdk/src/Services/S3/Custom/Transfer/Internal/_async/MultipartUploadCommand.async.cs
```diff
@@ -240,22 +238,16 @@ private async Task UploadNonSeekableStreamAsync(TransferUtilityUploadRequest req
...
-                                    UploadId = initiateResponse.UploadId,
-                                    InputStream = nextUploadBuffer,
-                                    PartSize = partSize,
-                                    PartNumber = partNumber,
-                                    IsLastPart = isLastPart
-                                }).ConfigureAwait(false);
+                                UploadPartRequest uploadPartRequest = ConstructUploadPartRequestForNonSeekableStream(nextUploadBuffer, partNumber, partSize, isLastPart, initiateResponse);
+
+                                var partResponse = await _s3Client.UploadPartAsync(uploadPartRequest).ConfigureAwait(false);
                                 Logger.DebugFormat($"Uploaded part {partNumber}. (Last part = {isLastPart}, Part size = {partSize}, Upload Id: {initiateResponse.UploadId}");
```
normj, line 244: Can you fix the `DebugFormat` calls to not use string interpolation using the `$`. This is mean to pass in a format string with parameters so that we don't make allocations creating a debug string that is never used. It is done a few places in  this function and possible other places updated for Unseekable streams.

```
Logger.DebugFormat($"Uploaded part {0}. (Last part = {1}, Part size = {2}, Upload Id: {3}", partNumber, isLastPart, partSize, initiateResponse.UploadId);
```

sdk/src/Services/S3/Custom/Transfer/Internal/_bcl35/MultipartUploadCommand.bcl35.cs
```diff
@@ -147,41 +145,32 @@ private void UploadUnseekableStream(TransferUtilityUploadRequest request)
...
-                                });
+                                UploadPartRequest uploadPartRequest = ConstructUploadPartRequestForNonSeekableStream(nextUploadBuffer, partNumber, partSize, isLastPart, initiateResponse);
+
+                                var partResponse = _s3Client.UploadPart(uploadPartRequest);
 
                                 Logger.DebugFormat($"Uploaded part {partNumber}. (Last part = {isLastPart}, Part size = {partSize}, Upload Id: {initiateResponse.UploadId}");
-                                partETags.Add(new PartETag { PartNumber = partResponse.PartNumber, ETag = partResponse.ETag });
+                                uploadPartResponses.Add(partResponse);
                                 partNumber++;
                                 nextUploadBuffer = new MemoryStream(partBuffer);
```
normj, line 155: Not disposing previous MemoryStream.

normj (changes requested):

sdk/src/Services/S3/Custom/Transfer/Internal/_async/MultipartUploadCommand.async.cs
ashishdhingra: Fixed. Please review.

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartUploadCommand.cs
ashishdhingra: Fixed. Please review.

sdk/src/Services/S3/Custom/Transfer/Internal/_async/MultipartUploadCommand.async.cs
ashishdhingra: Fixed. Please review.

sdk/src/Services/S3/Custom/Transfer/Internal/_bcl35/MultipartUploadCommand.bcl35.cs
ashishdhingra: Fixed. Please review.

normj (approved):

peterrsongg (approved):

PR #3102, opened 2023-11-15
Fix regression in 3.7.300 where the DataModel stop mapping to non public properties.

normj:
With the recent 3.7.300 release we removed the usage of the `TypeWrapper` abstraction and directly used the `Type` class. There was an unintended change in behavior with DDB DataModel library that when it was doing its reflection it changed from getting all instance properties including non public properties to only getting public. This PR restores the reflection to get members including non-public properties.

I also searched the SDK for other usages of `GetMember` from the `Type` class in case there was other regressions but found no usages.

CI build pending.

### DIFF
--- a/generator/.DevConfigs/fix-datamodel-regression-2023-11-15.json
+++ b/generator/.DevConfigs/fix-datamodel-regression-2023-11-15.json
@@ -1,0 +1,9 @@
+{
+    "services": [
+      {
+        "serviceName": "DynamoDBv2",
+        "type": "patch",
+        "changeLogMessages": [ "Fix regression in 3.7.300 where the DataModel stop mapping to non public properties." ]
+      }
+    ]
+  }
--- a/sdk/src/Services/DynamoDBv2/Custom/DataModel/InternalModel.cs
+++ b/sdk/src/Services/DynamoDBv2/Custom/DataModel/InternalModel.cs
@@ -323,7 +323,7 @@ namespace Amazon.DynamoDBv2.DataModel
             Dictionary<string, MemberInfo> dictionary = new Dictionary<string, MemberInfo>(StringComparer.Ordinal);
 
             var members = type
-                .GetMembers()
+                .GetMembers(BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Public)
                 .Where(IsValidMemberInfo)
                 .ToList();
             foreach (var member in members)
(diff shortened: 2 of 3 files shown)


peterrsongg:
> With the recent 3.7.300 release we removed the usage of the `TypeWrapper` abstraction and directly used the `Type` class. There was an intended change in behavior with DDB DataModel library that when it was doing its reflection it changed from 

Do you mean there was an unintended change in behavior? The description says intended

peterrsongg (approved):

normj:
> > With the recent 3.7.300 release we removed the usage of the `TypeWrapper` abstraction and directly used the `Type` class. There was an intended change in behavior with DDB DataModel library that when it was doing its reflection it changed from
> 
> Do you mean there was an unintended change in behavior? The description says intended

Whoops corrected. Yes I meant **unintended**.

boblodgett (approved):

jknobel:
@peterrsongg Possible another unintended consequence of this that remains is that properties with private setters are no longer able to be saved. 

I have a simple Id property
```
[DynamoDBHashKey]
public String Id { get; private set; }
```
that used to work fine, but since this change, I get an error whenever I try to save any object to the DB. If I make the setter public (remove "private"), then it works fine. This wasn't required before the regression. 

The exception looks like this:
```
Unable to locate property for key attribute Id

   at Amazon.DynamoDBv2.DataModel.ItemStorageConfigCache.Validate(Boolean value, String messageFormat, Object[] args)
   at Amazon.DynamoDBv2.DataModel.ItemStorageConfigCache.GetProperty(ItemStorageConfig config, String attributeName, Boolean optional)
   at Amazon.DynamoDBv2.DataModel.ItemStorageConfigCache.PopulateConfigFromTable(ItemStorageConfig config, Table table)
   at Amazon.DynamoDBv2.DataModel.ItemStorageConfigCache.CreateStorageConfig(Type baseType, String actualTableName, DynamoDBFlatConfig flatConfig)
   at Amazon.DynamoDBv2.DataModel.ItemStorageConfigCache.GetConfig(Type type, DynamoDBFlatConfig flatConfig, Boolean conversionOnly)
   at Amazon.DynamoDBv2.DataModel.DynamoDBContext.SaveHelperAsync(Type valueType, Object value, DynamoDBOperationConfig operationConfig, CancellationToken cancellationToken)
   at Amazon.DynamoDBv2.DataModel.DynamoDBContext.SaveHelperAsync[T](T value, DynamoDBOperationConfig operationConfig, CancellationToken cancellationToken)
```

PR #3128, opened 2023-12-12
fix: Update endpoint resolution logic in all services to allow the combination of the ServiceURL and AuthenticationRegion on the client config to drive the SDK's internal endpoint selection logic.

ashovlin:
## Description
The .NET SDK has long disallowed setting both `ServiceURL` and `RegionEndpoint` on a service client configuration. From https://docs.aws.amazon.com/sdkfornet/v3/apidocs/items/Runtime/TClientConfig.html
> RegionEndpoint and ServiceURL are mutually exclusive properties. Whichever property is set last will cause the other to automatically be reset to null.

However this is problematic after the new endpoint resolution logic introduced in 3.7.100 where AWS services provide rulesets that the SDKs use to resolve the endpoint for a given request. Valid endpoint resolution logic can now use both `ServiceURL` and `RegionEndpoint` as inputs to derive an endpoint.

An example of this is in S3 Control for [Working with S3 buckets on a Snowball Edge device](https://docs.aws.amazon.com/snowball/latest/developer-guide/working-s3-snow-buckets.html). See https://github.com/aws/aws-sdk-net/blob/930216aed32bc39844c8690dd50d459ae86ef582/sdk/src/Services/S3Control/Generated/Internal/AmazonS3ControlEndpointProvider.cs#L63-L76

This change allows `AuthenticationRegion` to be passed to the input to the endpoint resolver for the case where `ServiceURL` is set. It continues to NOT allow both `ServiceURL` and `RegionEndpoint` to be set on a client config.

## Motivation and Context
Internal ticket V1152090454

## Testing
* Added a test case to cover the endpoint resolution for S3 Control and Snowball.
* IN PROGRESS - dry-run on internal CI
* IN PROGRESS - manual testing with PowerShell, since that's where the case originated.

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/test/Services/S3Control/UnitTests/Custom/S3ControlArnTestUtils.cs
+++ b/sdk/test/Services/S3Control/UnitTests/Custom/S3ControlArnTestUtils.cs
@@ -16,14 +16,11 @@ using Amazon.Runtime;
 using Amazon.Runtime.Internal;
 using Amazon.Runtime.Internal.Auth;
 using Amazon.Runtime.Internal.Transform;
-using Amazon.Runtime.Internal.Util;
 using Amazon.S3Control;
 using Amazon.S3Control.Internal;
-using System;
+using AWSSDK.UnitTests.Mocking;
 using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
+using static AWSSDK.UnitTests.Mocking.TestUtils;
 
 namespace AWSSDK.UnitTests
 {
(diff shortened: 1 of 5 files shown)


sdk/test/Services/S3Control/UnitTests/Custom/S3ControlArnTestUtils.cs
```diff
@@ -16,71 +16,28 @@
...
 {
     public class S3ControlArnTestUtils
     {
         public static IRequest RunMockRequest(AmazonWebServiceRequest request, IMarshaller<IRequest, AmazonWebServiceRequest> marshaller, AmazonS3ControlConfig config)
         {
-            var pipeline = new RuntimePipeline(new List<IPipelineHandler>
+            var pipelineHandlers = new List<IPipelineHandler>
             {
                 new NoopPipelineHandler(),
+                new Signer(),
```
ashovlin, line 34: Adding this so we can assert that the signing region in the `Authorization` header is correct in the new test.

sdk/test/Services/S3Control/UnitTests/Custom/S3ControlArnTestUtils.cs
```diff
@@ -16,71 +16,28 @@
...
                 new AmazonS3ControlEndpointResolver(),
                 new Marshaller()
-            });
-
-            var requestContext = new RequestContext(config.LogMetrics, new AWS4Signer())
```
ashovlin, line 41: Just replacing this with the centralized `TestUtils.RunMockRequest` since it's the same logic.

dscpinheiro (approved):

normj (approved):

Issue #3132, opened 2023-12-15
AmazonApiGatewayManagementApiClient.PostToConnectionAsync wrong signature


usix79:
### Describe the bug

AWSSDK.ApiGatewayManagementApi Version="3.7.300.19" 

AmazonApiGatewayManagementApiClient.PostToConnectionAsync it is not capable to properly sign requests to ServiceURL like
```
https://EDITED.execute-api.eu-central-1.amazonaws.com/$default
```

Produced canonical resource path:
```
 /%2524default/%2540connections/P-S_9e-7liACH9A%253D
```
but expected:
```
/%24default/%2540connections/P-S_9e-7liACH9A%253D
```

### Expected Behavior

Calling AmazonApiGatewayManagementApiClient.PostToConnectionAsync should not throw a signature validation exception assuming that the correct permissions are in place.

### Current Behavior

Exception in AmazonApiGatewayManagementApiClient.PostToConnectionAsync:

```
AmazonApiGatewayManagementApiException making request PostToConnectionRequest to https://EDITED.execute-api.eu-central-1.amazonaws.com/$default. Attempt 1. --> Amazon.ApiGatewayManagementApi.AmazonApiGatewayManagementApiException: The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method. Consult the service documentation for details.

The Canonical String for this request should have been
'POST
/%24default/%2540connections/P-S_9e-7liACH9A%253D

content-length:1367
content-type:binary/octet-stream
host:EDITED.execute-api.eu-central-1.amazonaws.com
user-agent:aws-sdk-dotnet-coreclr/3.7.300.19 ua/2.0 os/linux#4.14.255.327 md/ARCH#X64 lang/.NET_Core#6.0.22 exec-env/AWS_Lambda_dotnet6 md/aws-sdk-dotnet-core#3.7.300.19 api/ApiGatewayManagementApi#3.7.300.19 cfg/retry-mode#legacy md/ClientAsync
x-amz-api-version:2018-11-29
x-amz-content-sha256:1402029c938875639856ff94f4ee60faaeb80dfab44ed68fd780f7e91615efde
x-amz-date:20231215T072312Z
x-amz-security-token:EDITED

content-length;content-type;host;user-agent;x-amz-api-version;x-amz-content-sha256;x-amz-date;x-amz-security-token
1402029c938875639856ff94f4ee60faaeb80dfab44ed68fd780f7e91615efde'

[...]

PR #3139, opened 2024-01-04
Fix: Single encode resource path components from the endpoint and dou…

peterrsongg:
…ble encode resource path components from the request

## Description

This fix ensures that when we canonicalize the resource path we single encode the portion that comes from the configured endpoint (via serviceURL or the default endpoint) and double encode the resource path that comes from the request. ApiGatewayManagementClient allows special characters to be included in the configured endpoint as part of its resource path, which was being incorrectly double encoded because `CanonicalizeResourcePathV2` treated the resource path from the request and the resource path from the endpoint as the same. This change ensures that we separate these two resource paths and encode them accordingly, first encoding the resource path from the request, then concatenating that to the resource path from the endpoint and encoding that altogether.

This change also adds unit tests for these scenarios and also updates the existing unit tests to use the V2 version of `CanonicalizeResourcePath`.
## Motivation and Context

Fixes this issue: https://github.com/aws/aws-sdk-net/issues/3132
## Testing

Full dry run on prod passed and added unit test coverage for all cases, as well as updated existing unit test with newer method.
## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
+++ b/sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
@@ -427,25 +427,32 @@ namespace Amazon.Util
         /// <returns>Canonicalized resource path for the endpoint.</returns>
         public static string CanonicalizeResourcePathV2(Uri endpoint, string resourcePath, bool encode, IDictionary<string, string> pathResources)
         {
+            // Single encode the resourcepath from the endpoint i.e. treat it like a regular uri.
+            // Double encode the resource path from the request
+            string resourcePathFromRequest = resourcePath != null ? resourcePath : string.Empty;
+            string resourcePathFromEndpoint = endpoint != null? endpoint.AbsolutePath : string.Empty;
+
             if (endpoint != null)
             {
-                var path = endpoint.AbsolutePath;
-                if (string.IsNullOrEmpty(path) || string.Equals(path, Slash, StringComparison.Ordinal))
-                    path = string.Empty;
+                // If the resource path from the endpoint is just a slash, it is the equivalent of being empty, so just set it to empty.
+                if (string.IsNullOrEmpty(resourcePathFromEndpoint) || string.Equals(resourcePathFromEndpoint, Slash, StringComparison.Ordinal))
+                    resourcePathFromEndpoint = string.Empty;
 
-                if (!string.IsNullOrEmpty(resourcePath) && resourcePath.StartsWith(Slash, StringComparison.Ordinal))
-                    resourcePath = resourcePath.Substring(1);
+                if (!string.IsNullOrEmpty(resourcePathFromRequest) && resourcePathFromRequest.StartsWith(Slash, StringComparison.Ordinal))
+                    resourcePathFromRequest = resourcePathFromRequest.Substring(1);
 
-                if (!string.IsNullOrEmpty(resourcePath))
-                    path = path + Slash + resourcePath;
+                // If the resource path from the endpoint is empty, add a slash in front of the resource path from the request,
+                // so that when we split and join the resource path segments, we don't encode the slash. Removing this will cause odd unwanted behavior.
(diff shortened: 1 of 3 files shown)


sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
```diff
@@ -427,25 +427,32 @@ public static string CanonicalizeResourcePath(Uri endpoint, string resourcePath,
         /// <returns>Canonicalized resource path for the endpoint.</returns>
         public static string CanonicalizeResourcePathV2(Uri endpoint, string resourcePath, bool encode, IDictionary<string, string> pathResources)
         {
+            // Single encode the resourcepath from the endpoint i.e. treat it like a regular uri.
+            // Double encode the resource path from the request
+            string resourcePathFromRequest = resourcePath != null ? resourcePath : string.Empty;
```
normj, line 432: After this variable is set here can we only use `resourcePathFromRequest` instead of `resourcePath` anywhere else in the function? I like that we have 2 clearly named variables and still using the `resourcePath` leaves some confusion.

sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
```diff
@@ -469,6 +475,30 @@ public static string CanonicalizeResourcePathV2(Uri endpoint, string resourcePat
             return canonicalizedResourcePath;
         }
 
+        /// <summary>
+        /// This method returns the full canonicalized resource path which includes the resource path from the request and the resource path from the endpoint.
+        /// This is just a helper method to be used internally.
```
normj, line 480: That fact that this is a `private static` method is enough to say this is "just a helper method".

sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
```diff
@@ -469,6 +475,30 @@ public static string CanonicalizeResourcePathV2(Uri endpoint, string resourcePat
...
+        /// </summary>
+        /// <param name="encodedSegments">If double encoded, encoded segments contains the encoded segments from the resource path from the request</param>
+        /// <param name="resourcePathFromEndpoint">This is the resource path that comes from the endpoint itself and not the request</param>
+        /// <returns>The full canonicalized resource path that contains the single encoded canonicalized resource path from the endpoint and the double encoded
+        /// resource path from the request</returns>
+        private static string GetFullCanonicalizedResourcePath(IEnumerable<string> encodedSegments, string resourcePathFromEndpoint)
```
normj, line 486: nit: I would put `resourcePathFromEndpoint` as the first parameter since it is the first part of the path.

sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
```diff
@@ -469,6 +475,30 @@ public static string CanonicalizeResourcePathV2(Uri endpoint, string resourcePat
...
+        {
+            // Most of the time resourcePathFromEndpoint will just be "/", meaning it is essentially empty, so check to see if the length is > 1.
+            if (resourcePathFromEndpoint.Length > 1)
+            {
+                // Since we are resolving the configured endpoint here, we don't want to subsitute any key/value pairs from path resources.
+                IEnumerable<string> absolutePathSegment = AWSSDKUtils.SplitResourcePathIntoSegments(resourcePathFromEndpoint, null);
+                // If encodedSegments' first value is empty, that means the resourcePath was empty, so no need to concatenate it since that will produce an extra "/" character at the end.
+                IEnumerable<string> fullPath = string.IsNullOrEmpty(encodedSegments.First()) ? absolutePathSegment : absolutePathSegment.Concat(encodedSegments);
```
normj, line 494: Can we use `FirstOrDefault` instead of `First` to protect ourselves in case `encodedSegments` is ever an empty collection.

normj (changes requested):

normj (approved):

muhammad-othman (approved):

Issue #3172, opened 2024-02-05
Performance downgrade with AmazonCloudFrontUrlSigner in Linux Environment (Ubuntu 22.04.3 LTS) 


SonalCalcey:
### Describe the bug

We've identified a performance degradation in **AmazonCloudFrontUrlSigner** when generating a signed URL using the **GetCannedSignedURL** method in **Ubuntu 22.04.3 LTS**. After some further investigation, we have identified that the issue only appears on  **Ubuntu 22.04.3 LTS**.

Furthermore, we have done a Benchmark test for the **Ubuntu** versions such as

- Ubuntu 22.04.3 LTS
- Ubuntu 20.04.3 LTS
- Ubuntu 18.04.6 LTS

To help you better understand the issue, we have prepared a sample description highlighting the key observations and differences. Kindly be aware that the tests were executed on **Ubuntu** versions within the **WSL 2** environment.

(Benchmark results were taken using [BenchmarkDotNet ](https://github.com/dotnet/BenchmarkDotNet)v0.13.11)

**Summary for Ubuntu 22.04.3 LTS**

![Signed Url-Ubuntu 22 04 LTS 1](https://github.com/aws/aws-sdk-net/assets/139344044/34a54633-9560-4921-b2b4-cb5c9ecac707)

![Signed Url-Ubuntu 22 04 LTS 2](https://github.com/aws/aws-sdk-net/assets/139344044/2ea63fe1-1825-4292-9661-1b6a6a95490c)

**Benchmark log 1:**
[benchmark_tester.SignedUrlRunner-20240205-104837-Ubuntu 22.04 LTS-Run1.log](https://github.com/aws/aws-sdk-net/files/14165507/benchmark_tester.SignedUrlRunner-20240205-104837-Ubuntu.22.04.LTS-Run1.log)

**Benchmark log 2:**
[benchmark_tester.SignedUrlRunner-20240205-105435-Ubuntu 22.04 LTS-Run2.log](https://github.com/aws/aws-sdk-net/files/14165508/benchmark_tester.SignedUrlRunner-20240205-105435-Ubuntu.22.04.LTS-Run2.log)

**Summary for Ubuntu 18.04.6 LTS**

![Signed Url-Ubuntu 18 04 LTS 1](https://github.com/aws/aws-sdk-net/assets/139344044/7c73533c-0855-4f4a-9f74-4b5865f6af5c)

[...]

Issue #3165, opened 2024-01-30
There are performance issues with openssl3.0 and AmazonCloudFrontUrlSigner should reuse RSA object


adomass:
### Describe the bug

Most information about the cause can be found [here](https://github.com/dotnet/runtime/issues/97727).
Since openssl3 operation for ImportParameters is very expensive it should not be called twice when possible to avoid this.

### Expected Behavior

It should be possible to reuse RSA as it became very expensive operation.

### Current Behavior

https://github.com/aws/aws-sdk-net/blob/d5ec588535f39d2dad37a91eac9ee465672e7335/sdk/src/Services/CloudFront/Custom/AmazonCloudFrontUrlSigner.cs#L529-L531

```
var rsa = RSA.Create();
rsa.ImportParameters(rsaParams);
return rsaParams;
```

Importing the parameters into an RSA object then throwing it away. But then just import it again here:

https://github.com/aws/aws-sdk-net/blob/d5ec588535f39d2dad37a91eac9ee465672e7335/sdk/src/Services/CloudFront/Custom/AmazonCloudFrontUrlSigner.cs#L510

Essentially calling `ImportParameters `twice for every key.

### Reproduction Steps

![image](https://github.com/aws/aws-sdk-net/assets/4628301/c4eb3885-69d3-4598-b842-a10edd166b30)

### Possible Solution

       public static string SignUrl(string resourceUrlOrPath, string keyPairId, TextReader privateKey, string policy)
        {
            RSA rsaProvider = ConvertPEMToRSA(privateKey);
            byte[] signatureBytes = SignWithSha1RSA(UTF8Encoding.UTF8.GetBytes(policy), rsaProvider);

            string urlSafePolicy = MakeStringUrlSafe(policy);
            string urlSafeSignature = MakeBytesUrlSafe(signatureBytes);

            string signedUrl = resourceUrlOrPath + (resourceUrlOrPath.IndexOf('?') >= 0 ? "&" : "?") + "Policy="
                    + urlSafePolicy + "&Signature=" + urlSafeSignature + "&Key-Pair-Id=" + keyPairId;
            return signedUrl;
        }

 	internal static RSA ConvertPEMToRSAParameters(TextReader privateKeyReader)
        {
            RSAParameters rsaParams;
            try
            {
                rsaParams = new PemReader(privateKeyReader).ReadPrivatekey();
            }
            catch (Exception e)
            {
                throw new AmazonClientException("Invalid RSA Private Key", e);
            }

[...]

Issue #3130, opened 2023-12-14
AWSSDK.CloudFront is slower than  AWSSDK.S3 - Signed URL Generation


SonalCalcey:
### Describe the bug

 [Previous issue](https://github.com/aws/aws-sdk-net/issues/3122) 
Because of the above issue, we are just wondering if there is any issue on the CloudFront as well. Because compared to S3 the CloudFront Response time is x6 or x5 times slower than S3.

These two methods **(PlayerSummaryFull and UpdatePlayerProfilePicture)** use S3 or CloudFront to Generate Signed URLs. 
The Following two benchmark snapshots are by using S3 and CloudFront respectively as media content providers.

------
**AWSSDK.S3 - 3.7.304.7**

![s3 using latest only](https://github.com/aws/aws-sdk-net/assets/139344044/3e8671ca-0627-41ef-be5d-0e57aba88b7a)
**Benchmark log:** [Benchmark-using-S3.log](https://github.com/aws/aws-sdk-net/files/13670007/Benchmark-using-S3.log)

------
**AWSSDK.CloudFront - 3.7.301.11**

![cloud front using latest only](https://github.com/aws/aws-sdk-net/assets/139344044/10fbf04b-a54d-4418-88f8-7fcaf88d410d)
**Benchmark log:** [Benchmark-using-CloudFront.log](https://github.com/aws/aws-sdk-net/files/13670364/Benchmark-using-CloudFront.log)

------

Sorry for not having Multiple version comparisons or more details. The reason for providing these details is FYI to identify if there is any performance degradation on the CloudFront as well.

### Expected Behavior

No response 

### Current Behavior

No response

### Reproduction Steps

No response

### AWS .NET SDK and/or Package version used

AWSSDK.CORE - 3.7.300.18
AWSSDK.S3 - 3.7.304.7
AWSSDK.CloudFront - 3.7.301.11

### Targeted .NET Platform

.NET Framework 7

### Operating System and version

Windows 10 Enterprise

ashishdhingra:
@SonalCalcey Good morning. The fix merged in other issue https://github.com/aws/aws-sdk-net/issues/3122 was made in `AWSSDK.Core` and hence should have propagated to CloudFront package as well. Do you have performance benchmarks between different versions of `AWSSDK.CloudFront` rather than comparing it with `AWSSDK.S3`?

Thanks,
Ashish

SonalCalcey:
[...]

PR #3176, opened 2024-02-07
Remove duplicate RSA call from AmazonCloudFrontUrlSigner

dscpinheiro:
## Description
Our `AmazonCloudFrontUrlSigner` helper class invokes the `RSACryptoServiceProvider.ImportParameters` method twice, but it ignores the results of the first call. Customers using OpenSSL 3 have reported a performance regression as that operation is slower than previous versions of OpenSSL (see #3165 and #3172).

## Testing
- Dry-run: `DRY_RUN-4b9e956b-2815-4fc2-9d55-656d5938c53d`
- With the dry-run artifacts, I ran benchmark tests on an Ubuntu 2022 EC2 instance (benchmarks are based on the example from #3172):

```
$ uname -a
Linux ip-172-31-37-170 6.2.0-1018-aws #18~22.04.1-Ubuntu SMP Wed Jan 10 22:54:16 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux

$ openssl version
OpenSSL 3.0.2 15 Mar 2022 (Library: OpenSSL 3.0.2 15 Mar 2022)

$ curl -H "X-aws-ec2-metadata-token: $TOKEN" http://169.254.169.254/latest/meta-data/instance-type
m7i.large
```

Before:
```
BenchmarkDotNet v0.13.12, Ubuntu 22.04.3 LTS (Jammy Jellyfish)
Intel Xeon Platinum 8488C, 1 CPU, 2 logical cores and 1 physical core
.NET SDK 8.0.101
  [Host]     : .NET 8.0.1 (8.0.123.58001), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI
  DefaultJob : .NET 8.0.1 (8.0.123.58001), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI
```
| Method                      | Mean     | Error    | StdDev   | Allocated |
|---------------------------- |---------:|---------:|---------:|----------:|
| CloudFrontGenerateSignedUrl | 63.64 ms | 0.102 ms | 0.086 ms |  37.29 KB |
| WildCardCloudFrontSignedUrl | 63.62 ms | 0.093 ms | 0.077 ms |  38.84 KB |

After:
```
BenchmarkDotNet v0.13.12, Ubuntu 22.04.3 LTS (Jammy Jellyfish)
Intel Xeon Platinum 8488C, 1 CPU, 2 logical cores and 1 physical core
.NET SDK 8.0.101
  [Host]     : .NET 8.0.1 (8.0.123.58001), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI
  DefaultJob : .NET 8.0.1 (8.0.123.58001), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI
```
| Method                      | Mean     | Error    | StdDev   | Allocated |
|---------------------------- |---------:|---------:|---------:|----------:|
| CloudFrontGenerateSignedUrl | 32.10 ms | 0.092 ms | 0.077 ms |  30.53 KB |
| WildCardCloudFrontSignedUrl | 32.11 ms | 0.081 ms | 0.072 ms |  32.02 KB |

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/CloudFront/Custom/AmazonCloudFrontUrlSigner.cs
+++ b/sdk/src/Services/CloudFront/Custom/AmazonCloudFrontUrlSigner.cs
@@ -525,9 +525,6 @@ namespace Amazon.CloudFront
             {
                 throw new AmazonClientException("Invalid RSA Private Key", e);
             }
-
-            var rsa = RSA.Create();
-            rsa.ImportParameters(rsaParams);
             return rsaParams;
         }
 
--- a/sdk/test/Services/CloudFront/UnitTests/Custom/CookieSignerTest.cs
+++ b/sdk/test/Services/CloudFront/UnitTests/Custom/CookieSignerTest.cs
@@ -1,0 +1,78 @@
+﻿using System;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System.IO;
+using Amazon.CloudFront;
+
+namespace AWSSDK_DotNet35.UnitTests
+{
+    [TestClass]
+    public class CookieSignerTest
+    {
+        private static StreamReader privateRSAKeyStreamReader = null;
+
+        [TestInitialize]
+        public void Initialize()
+        {
+            privateRSAKeyStreamReader = new StreamReader(Utils.GetResourceStream("sample.rsa.private.key.txt"));
+        }
+
+        [TestMethod]
+        [TestCategory("CloudFront")]
+        public void CannedPolicyValidation()
+        {
+            string expectedSignature = 
+                "IGIL8ItK6uqLzFAR1338Mf5RqqU9VIhG9G0nUmgl4VPdozBDS5fyRMZH39cFqWfK7gXSCOGBGt9X0yYRwqO-yikx5f9eQUQ3pyNukf-iQ~cp4upSVG6ETsR8NQQ1eeiP3DruH4hzwXqmh-ozD~bSzqbvbFW24LclBFG8f6QH~QY_";
+            string expectedKeyPair = "CannedKeyPairId";
+
+            var cookies = AmazonCloudFrontCookieSigner.GetCookiesForCannedPolicy(
+                "http://awesome.dot.com/amazing/uri/",
+                expectedKeyPair,
+                privateRSAKeyStreamReader,
+                new DateTime(2024, 1, 1, 12, 0, 0, DateTimeKind.Utc)
+            );
+            Assert.IsNotNull(cookies);
+
+            Assert.AreEqual("CloudFront-Expires", cookies.Expires.Key);
+            Assert.AreEqual("1704110400", cookies.Expires.Value);
+
+            Assert.AreEqual("CloudFront-Signature", cookies.Signature.Key);
+            Assert.AreEqual(expectedSignature, cookies.Signature.Value);
+
+            Assert.AreEqual("CloudFront-Key-Pair-Id", cookies.KeyPairId.Key);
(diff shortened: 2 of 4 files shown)


sdk/test/Services/CloudFront/UnitTests/Custom/CookieSignerTest.cs
```diff
@@ -0,0 +1,78 @@
+﻿using System;
```
dscpinheiro, line 1: There were no existing tests for `AmazonCloudFrontCookieSigner`. 

I added these to make sure I didn't break anything when removing the duplicated call (they passed before and after making the change).

sdk/src/Services/CloudFront/Custom/AmazonCloudFrontUrlSigner.cs
```diff
@@ -525,9 +525,6 @@ internal static RSAParameters ConvertPEMToRSAParameters(TextReader privateKeyRea
             {
                 throw new AmazonClientException("Invalid RSA Private Key", e);
             }
-
-            var rsa = RSA.Create();
```
dscpinheiro, line 529: We call `ImportParameters` again here: https://github.com/aws/aws-sdk-net/blob/871aed7af1271bf39fa1992b4e29d51a1d13f809/sdk/src/Services/CloudFront/Custom/AmazonCloudFrontUrlSigner.cs#L509-L510

That method is invoked right after `ConvertPEMToRSAParameters`.

SonalCalcey:
@dscpinheiro As for the above benchmarks the performance went from 60ms to 30ms that's good. but these [#3172](https://github.com/aws/aws-sdk-net/issues/3172) benchmark shows it is only taking 5ms in Ubuntu 20 version right?

dscpinheiro:
> @dscpinheiro As for the above benchmarks the performance went from 60ms to 30ms that's good. but these [#3172](https://github.com/aws/aws-sdk-net/issues/3172) benchmark shows it is only taking 5ms in Ubuntu 20 version right?

I saw similar numbers on Ubuntu 20 (5ms before and 3.5ms after removing this duplicate call), that operation is just unfortunately slower in OpenSSL 3.

I know it doesn’t help your .NET 7 use case, but apparently there’ll be improvements included in .NET 8.0.3: https://github.com/dotnet/runtime/issues/97727#issuecomment-1930809385

sdk/src/Services/CloudFront/Custom/AmazonCloudFrontUrlSigner.cs
dscpinheiro: You're right, it shouldn't but I decided not to change the interface of the signer in this PR. There's a separate issue (#3130) about improving performance of the CloudFront signers, that's probably a better place to discuss / design changes.

Another thing is that `ConcurrentDictionary` is not available in `net35`, which the SDK still supports: https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Services/CloudFront/AWSSDK.CloudFront.Net35.csproj#L4

muhammad-othman (approved):

dscpinheiro:
This PR was included in version `3.7.301.41` of the `AWSSDK.CloudFront` package. I'm afraid the operation will still be slower when using OpenSSL 3 (compared to OpenSSL 1), but now the AWS SDK won't make the expensive call twice (and hopefully we'll see improvements in .NET 8.0.3 as well).

And as I mentioned above, we do have a separate issue for improving the overall performance of the CloudFront signers: https://github.com/aws/aws-sdk-net/issues/3130

PR #3199, opened 2024-02-26
Update manifest file to use net462 instead of net45

dscpinheiro:
## Description
Updates the generator to create `net462` projects instead of `net45`. I also modified a few other projects so that the service solutions could compile successfully (e.g. CRT extensions, Core, etc...)

Example of generated changes for one service: https://github.com/aws/aws-sdk-net/commit/249bf2add1344d9c179891e75af8ba8153ccbc93

## Testing
- Ran generator locally: `msbuild .\buildtools\build.proj /t:run-generator`
- Built extensions package and confirmed it contained `net462` assets: `dotnet msbuild .\buildtools\build.proj /t:package-crt-extensions`
- Built service solutions: `dotnet build -c Release .\sdk\src\Services\AccessAnalyzer\AccessAnalyzer.sln` and `dotnet build -c Release .\sdk\src\Services\S3\S3.sln`

## Screenshots (if appropriate)
Solution after renaming and changing the target framework for the `net45` projects:

> Note: This solution is missing the `AWSSDK.IntegrationTests.Service.NetFramework` project. Those files are not generated automatically, I'll handle them in a separate PR as this one was already getting somewhat large (probably at the same time when modifying the hand-written test projects such as `sdk\test\CSMTest`).

net462-solution

## License
- [X] I confirm that this pull request can be released under the Apache 2 license

PR #3202, opened 2024-02-27
Update integration test projects to target net462

dscpinheiro:
## Description
Follow-up of https://github.com/aws/aws-sdk-net/pull/3199, updates the integration test projects (which are not generated) to target `net462`. Apologies for the number of files, but I had to make the same changes to around 60 services.

I also intentionally didn't modify the `sdk\test\CSMTest` solution, will handle it in a separate PR.

## Testing
- Ran generator locally: `msbuild .\buildtools\build.proj /t:run-generator`
- Built solutions for services that contain integration tests: `dotnet build -c Release .\sdk\src\Services\APIGateway\APIGateway.sln` and `dotnet build -c Release .\sdk\src\Services\S3\S3.sln`

## License


sdk/test/Services/AutoScaling/IntegrationTests/Config/462/App.config
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="utf-8"?>
```
dscpinheiro, line 1: I don't know why, but in some places this diff seems to think a new file was added (when what happened instead was I renamed them from `45/App.config` to `462/App.config`, like this: https://github.com/aws/aws-sdk-net/pull/3202/files?diff=split&w=1#diff-8ddb50d4f9f3924b59e5f4a683dfc3c2c7b4bea75991dbaa3169a3cdcb88ee04).

sdk/test/Services/AutoScaling/IntegrationTests/Config/462/App.config
```diff
@@ -0,0 +1,51 @@
...
+          <!--<add name="console" type="System.Diagnostics.ConsoleTraceListener" traceOutputOptions="DateTime"/>-->
+          <add name="console" type="AWSSDK_DotNet.IntegrationTests.Utils.ConsoleTraceListener, AWSSDK.IntegrationTestUtilities.NetFramework"/>
+        </listeners>
+      </source>
+    </sources>
+  </system.diagnostics>
+  <aws correctForClockSkew="true">
+    <s3 useSignatureVersion4="true"/>
+    <dynamoDB>
+      <dynamoDBContext>
```
dscpinheiro, line 28: But speaking of configs, all of them include this `dynamoDBContext`... Should this only be included in the DynamoDB projects?

sdk/test/IntegrationTests/Tests/462/Test462.cs
```diff
@@ -17,7 +17,7 @@
 namespace AWSSDK_DotNet.IntegrationTests.Tests
 {
     [TestClass]
-    public class Tests45
+    public class Tests462
```
normj, line 20: Should we just name this to TestsNetFramework?

sdk/test/Services/AutoScaling/IntegrationTests/Config/462/App.config
normj: I assume this is a leftover artifact of when the integ tests were all in a single project and we used tooling to split up the tests. Would make sense to only be in DynamoDB test projects but doesn't hurt.

normj (approved):

sdk/test/Services/AutoScaling/IntegrationTests/Config/462/App.config
dscpinheiro: Got it. I ended up removing the `dynamoDB` section for all other services (https://github.com/aws/aws-sdk-net/pull/3202/commits/7ecbfa0fc6a7d97fa430a6c633e096888c9cf9c7) to avoid confusion later.

normj (approved):

afroz429 (approved):

PR #3203, opened 2024-02-28
Validate security credentials obtained from Instance Metadata Service

opottone:
## Description

Getting security credentials from Instance Metadata Service fails when an EC2 instance does not have permissions to assume a role. When that happens, we get a very unhelpful error message `Value cannot be null. (Parameter 'awsAccessKeyId')` Do response validation so that we get better exception message.

See also https://github.com/aws/aws-cli/issues/2060

## Types of changes

## Checklist

## License


ashishdhingra:
The change needs to be tested in all compute environments (like EC2, ECS, EKS, etc.).

ashishdhingra:
Reviewed the PR with the team. The unit tests are broken and dry run is failing. This needs to be reworked internally.

dscpinheiro:
I'm going to close this PR as Ashish mentioned there are some tests that are broken with this change. 

We're also currently working on other IMDS changes, and we'll re-evaluate whether the fix can be included there.

PR #3208, opened 2024-03-05
Update SDK to use BCL constant instead of BCL45

dscpinheiro:
## Description
Updates the generator to use `BCL` as a constant instead of `BCL45`. I also had to update some hand-written code in `Core` and `S3`.

Note: I also left out the integration tests (e.g. `CSMTest` and `AWSSDK.IntegrationTests.APIGateway.NetFramework.csproj`); the changes will be similar but I didn't want to add >100 files to this already large PR.

## Testing
- Ran generator locally: `msbuild .\buildtools\build.proj /t:run-generator`
- Built solution for service that included hand-written code: `dotnet build -c Release .\sdk\src\Services\S3\S3.sln`

## License
- [X] I confirm that this pull request can be released under the Apache 2 license

PR #3210, opened 2024-03-06
Update remaining BCL45 constants

dscpinheiro:
## Description
Follow-up of #3208, updates the remaining places to use `BCL` instead of `BCL45`. Ended up with more modified files than expected (I had to move / delete some hand-written bcl folders), but I was able to compile the `AWSSDK.NetFramework.sln` solution (the one with all 300+ services):

build

## License


generator/ServiceClientGeneratorLib/GeneratorDriver.cs
```diff
@@ -1470,6 +1470,9 @@ public static void RemoveLegacyFiles(string sdkRootFolder)
             var legacyFolderNames = new HashSet<string>
             {
                 "_bcl35",
+                Utils.PathCombineAlt("Generated", "_bcl45"),
```
dscpinheiro, line 1473: Added these so that we also delete folders for the "test services" (e.g. https://github.com/aws/aws-sdk-net/tree/main/sdk/test/Services/PaginatorsTest/Generated/Model/_bcl45%2Bnetstandard)

sdk/test/UnitTests/AWSSDK.UnitTests.NetFramework.csproj
```diff
@@ -23,7 +23,7 @@
   <Choose>
     <When Condition=" '$(AWSKeyFile)' == '' ">
       <PropertyGroup>
-        <AssemblyOriginatorKeyFile>../../../awssdk.dll.snk</AssemblyOriginatorKeyFile>
```
dscpinheiro, line 26: This file keeps changing (the correct is the one on the right), but I found out why: In the generator, we have a hard-coded check for `net35` and `net45`: https://github.com/aws/aws-sdk-net/blob/fad56d05b452da7623071a1517378223ff38db27/generator/ServiceClientGeneratorLib/Generators/ProjectFiles/VS2017ProjectFile.tt#L133

I'll update that in a separate PR (we also have `namespace AWSSDK_DotNet35` and `namespace AWSSDK.UnitTests.Net45` hardcoded in a few places...)

sdk/AWSSDK.CoreAndCustomUnitTests.NetFramework.sln
```diff
@@ -48,6 +48,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "AWSSDK.CloudWatch.NetFramew
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "AWSSDK.SSO.NetFramework", "src\Services\SSO\AWSSDK.SSO.NetFramework.csproj", "{685155AC-D4AE-41ED-A76E-388D0F66D6F8}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "AWSSDK.KeyManagementService.NetFramework", "src\Services\KeyManagementService\AWSSDK.KeyManagementService.NetFramework.csproj", "{F0397317-53D7-4B42-8D15-9E02083BF5D3}"
```
dscpinheiro, line 51: This is the hand-written solution file. I forgot to include KMS at first.

normj (approved):

PR #3212, opened 2024-03-06
Update test namespaces not to use 35 or 45

dscpinheiro:
## Description
Update the namespaces we use in unit tests not to reference `35` or `45` (I intentionally didn't use `NetFramework` as a suffix as we may run the same tests targeting .NET Core in the future).

After running the generator, I verified a few solutions built successfully (`AWSSDK.CoreAndCustomUnitTests.NetFramework.sln`, `DynamoDBv2.sln`, `S3.sln`, and `AWSSDK.NetFramework.sln`). 

## License


sdk/test/Services/BedrockRuntime/IntegrationTests/BedrockRuntimeEventStreamTests.cs
```diff
@@ -25,48 +25,7 @@ namespace AWSSDK_DotNet.IntegrationTests.Tests
     [TestClass]
     public class BedrockRuntimeEventStreamTests : TestBase<AmazonBedrockRuntimeClient>
     {
-#if BCL35
```
dscpinheiro, line 28: Not sure why these didn't show up before (my guess is there's a lot of generated files and my local search is struggling a bit).

normj (approved):

afroz429 (approved):

PR #3222, opened 2024-03-14
[V4] Change value types on model types to be nullable.

normj:
## Description
Update the value types for the generated model types and manually written S3 model types to use nullables for value types.

When updating code that was affected by the change I would use the `Value` property if there was a check like `IsSet<name>` before accessing the property. In the SDK code itself I used `GetValueOrDefault` if there was no check. In the tests I generally used `Value` because the tests were written with the expectation there was a value and if there wasn't a value that would be a bug that shouldn't be hidden by the `GetValueOrDefault`.

I removed some of the customizations for nullablity but more exists that are tied to collections. I'll remove the rest that make sense to remove as part of that story.

## Motivation and Context
This is part of V4 effort to better model behavior and allow users to understand if a value is not set in a service or not.

## Testing
Confirmed both the .NET Framework and .NET Solution uber solution files compile. Running all of the tests is delayed till we setup the new CI system for V4.

The PR is kind of large because I wanted the PR to compile which involved pretty much making all of the changes. I broke up the changes into component commits that might make it easier to review.

I did not change collections to use nullables. So we would still use types like `List<int>` instead of `List<int?>`. I think the experience if using nullables for all collections would be a pretty bad user experience. I'm going to create a separate story to figure out how to special case the theoretical use case of having collections that could possibly have nullable value types.

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Internal/Util/StringUtils.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Util/StringUtils.cs
@@ -68,9 +68,22 @@ namespace Amazon.Runtime.Internal.Util
             return value.ToString(CultureInfo.InvariantCulture);
         }
 
+        public static string FromLong(long? value)
+        {
+            if (value != null)
+                return value.Value.ToString(CultureInfo.InvariantCulture);
+
+            return null;
+        }
+
         public static string FromFloat(float value)
         {
             return value.ToString(CultureInfo.InvariantCulture);
+        }
+
+        public static string FromBool(bool? value)
+        {
+            return FromBool(value.GetValueOrDefault());
         }
 
         public static string FromBool(bool value)
@@ -91,6 +104,20 @@ namespace Amazon.Runtime.Internal.Util
         {
             return value.ToUniversalTime().ToString(AWSSDKUtils.ISO8601DateFormat, CultureInfo.InvariantCulture);
         }
+
+        /// <summary>
+        /// Converts a DateTime to ISO8601 formatted string.
+        /// </summary>
+        public static string FromDateTimeToISO8601(DateTime? value)
+        {
+            if (!value.HasValue)
+            {
+                return null;
+            }
+
+            return value.Value.ToUniversalTime().ToString(AWSSDKUtils.ISO8601DateFormat, CultureInfo.InvariantCulture);
+        }
+
         /// <summary>
         /// Converts a DateTime to ISO8601 formatted string.
         /// </summary>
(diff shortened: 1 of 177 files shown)


sdk/src/Core/Amazon.Runtime/Internal/Util/StringUtils.cs
```diff
@@ -120,11 +147,29 @@ public static string FromDouble(double value)
             return value.ToString(CultureInfo.InvariantCulture);
         }
 
+        public static string FromDouble(double? value)
```
dscpinheiro, line 150: Not super important, but any reason you used `value.HasValue` here but `value != null` below on line 167?

normj: I went back through the file and made it consistent to use the `value.HasValue` property.

peterrsongg:
For S3, I found a few areas where a non-nullable value type is used. Just going to point them out, I believe some are probably intentional.
1. `ByteRange` I believe this is something we created internally, since ScanRange is the closest thing I could find in the model:  `Start` and `End` still use the `long` non-nullable type. Was that intentional?
1. `Expiration` -> The POCO uses `DateTime` for `expiryDate` and `expiryDateUtc`
1. `UploadPartRequest` -> `private bool lastPart`
1. `TransferUtilityUploadDirectoryRequest` - > `private bool disablePayloadSigning`.

normj:
> For S3, I found a few areas where a non-nullable value type is used. Just going to point them out, I believe some are probably intentional.
> 
> 1. `ByteRange` I believe this is something we created internally, since ScanRange is the closest thing I could find in the model:  `Start` and `End` still use the `long` non-nullable type. Was that intentional?
> 2. `Expiration` -> The POCO uses `DateTime` for `expiryDate` and `expiryDateUtc`
> 3. `UploadPartRequest` -> `private bool lastPart`
> 4. `TransferUtilityUploadDirectoryRequest` - > `private bool disablePayloadSigning`.

Yes those were intentional because they are not modeled fields from the service they are part of our SDK's logic. I didn't want to change the nullability of fields that are used for SDK logic, just change fields used for serialization.

1. `ByteRange` - This was a helper class create the string formatted header value.
2. `Expiration` - Similar to `ByteRange` except used on responses. The constructors make sure there is always a value for the date fields.
3. `lastPart` - This is not a modeled fields. The SDK uses it to communicate for subclasses like the encryption client to always communicate when it is the last part.
4. `disablePayloadSigning` - Expect, we want the value to default to false to know not to skip disabling payload signing.

peterrsongg (approved):

PR #3231, opened 2024-03-20
Add ability to initialize collections to null instead of empty

normj:
## Description
For the .NET types used for service API request and responses the collection properties have always been initialized to an empty collection. This was done for ease of use. It does cause problems though with the SDK and consumers of the SDK to not understand if a property that currently set to an empty string is that way because it is the default value or the user wants to clear out the values of a collection. On a response users don't know if an collection is empty on the resource or just the default value for the collection.

This behavior also cause memory allocation issues because a lot of collections are created that are not even used. This is most pronounced with DynamoDB. In that case every attribute for an item is represented by an `AttributeValue`. This type creates 5 collections when at most 1 collection would actually have a value. A DynamoDB query that returns back 100 items with 10 attributes would end up creating 5000 collection when at most 1000 could be used. The actual value would mostly likely be less considering the frequency of scaler attributes.

This PR adds a new `AWSConfigs.InitializeCollection` property which defaults to `true`. With the value being `true` the behavior of the SDK stays as it has been. SDK decides to send a property as part of the request payload if the collection is not null and has a count greater then 0. This is where the problem the SDK has about clearing out collections. With the value `false` the collections for property on request and response types will be initialized to null. The SDK will send the collection property as part of the request payload whenever the collection is not null. That includes sending the property when the count is 0.

For V3 `AWSConfigs.InitializeCollection` this property will stay default to `true` to not break any behavior. We hope to change the default of this property to `false` in the next major version to address the modeling and performance issues with auto initializing collections.

The generated code changes are not included in this PR. That will be a massive amount of file changes. 

My process for updating S3 since it is hand coded.
* Add the check `AWSConfigs.InitializeCollection` when declaring and initializing the collections.
* Updated the `IsSet` operations to match our new desired behavior.
* To make things consistent I replaced code that was lazily creating the collection in the getter
* Did a VS Find Reference and made sure the calling code in the SDK was null safe. This usually involved updating the unmarshallers.
* A few `IsSet` methods I deleted because they were response only objects that were never called. I seemed better then updating them knowing there was nothing that was going to test them.

I also tried to address some compiler warnings as I was going though the code.

## Testing
Ran successful dry runs with `AWSConfigs.InitializeCollection` set to both `true` and `false`.

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_bcl/HttpWebRequestFactory.cs
+++ b/sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_bcl/HttpWebRequestFactory.cs
@@ -479,11 +479,18 @@ namespace Amazon.Runtime.Internal
             {
                 if (_request.RequestUri.Scheme == Uri.UriSchemeHttp)
                 {
-                    _request.Proxy = requestContext.ClientConfig.GetHttpProxy();
+                    proxy = requestContext.ClientConfig.GetHttpProxy();
                 }
                 else if (_request.RequestUri.Scheme == Uri.UriSchemeHttps)
                 {
-                    _request.Proxy = requestContext.ClientConfig.GetHttpsProxy();
+                    proxy = requestContext.ClientConfig.GetHttpsProxy();
+                }
+
+                // Only set the HttpWebRequest Proxy property if we have a value so
+                // that we don't override any OS level proxy settings.
+                if (proxy != null)
+                {
+                    _request.Proxy = proxy;
                 }
             }
 
--- a/sdk/src/Services/DynamoDBv2/Custom/DocumentModel/Search.cs
+++ b/sdk/src/Services/DynamoDBv2/Custom/DocumentModel/Search.cs
@@ -343,10 +343,14 @@ namespace Amazon.DynamoDBv2.DocumentModel
                             Limit = Limit,
                             TableName = TableName,
                             AttributesToGet = AttributesToGet,
-                            ScanFilter = Filter.ToConditions(SourceTable),
                             Select = EnumMapper.Convert(Select),
                             ConsistentRead = IsConsistentRead
                         };
+
+                        var scanFilter = Filter.ToConditions(SourceTable);
+                        if (scanFilter?.Count > 0)
+                            scanReq.ScanFilter = scanFilter;
+
                         if (!string.IsNullOrEmpty(this.IndexName))
                             scanReq.IndexName = this.IndexName;
                         if (this.FilterExpression != null && this.FilterExpression.IsSet)
@@ -396,8 +400,8 @@ namespace Amazon.DynamoDBv2.DocumentModel
 
                         Dictionary<string, Condition> keyConditions, filterConditions;
                         SplitQueryFilter(Filter, SourceTable, queryReq.IndexName, out keyConditions, out filterConditions);
-                        queryReq.KeyConditions = keyConditions;
-                        queryReq.QueryFilter = filterConditions;
+                        queryReq.KeyConditions = keyConditions.Count > 0 ? keyConditions : null;
+                        queryReq.QueryFilter = filterConditions.Count > 0 ? filterConditions : null;
                         Common.ConvertAttributesToGetToProjectionExpression(queryReq);
 
                         if (queryReq.QueryFilter != null && queryReq.QueryFilter.Count > 1)
--- a/sdk/src/Services/DynamoDBv2/Custom/DocumentModel/Search.cs
+++ b/sdk/src/Services/DynamoDBv2/Custom/DocumentModel/Search.cs
@@ -223,16 +223,26 @@ namespace Amazon.DynamoDBv2.DocumentModel
                             Limit = Limit,
                             TableName = TableName,
                             AttributesToGet = AttributesToGet,
-                            ScanFilter = Filter.ToConditions(SourceTable),
                             Select = EnumMapper.Convert(Select),
                             ConsistentRead = IsConsistentRead
                         };
+
                         if (!string.IsNullOrEmpty(this.IndexName))
                             scanReq.IndexName = this.IndexName;
                         if (this.FilterExpression != null && this.FilterExpression.IsSet)
                             this.FilterExpression.ApplyExpression(scanReq, SourceTable);
-                        if (scanReq.ScanFilter != null && scanReq.ScanFilter.Count > 1)
-                            scanReq.ConditionalOperator = EnumMapper.Convert(ConditionalOperator);
+
+
+                        var scanFilter = Filter.ToConditions(SourceTable);
+                        if (scanFilter != null && scanFilter.Count > 0)
+                        {
+                            scanReq.ScanFilter = scanFilter;
+
+                            if (scanFilter.Count > 1)
+                            {
+                                scanReq.ConditionalOperator = EnumMapper.Convert(ConditionalOperator);
+                            }
+                        }
                         Common.ConvertAttributesToGetToProjectionExpression(scanReq);
 
                         if (this.TotalSegments != 0)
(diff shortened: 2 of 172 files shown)


sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_bcl/HttpWebRequestFactory.cs
```diff
@@ -479,11 +479,18 @@ public virtual void ConfigureRequest(IRequestContext requestContext)
...
                 {
-                    _request.Proxy = requestContext.ClientConfig.GetHttpProxy();
+                    proxy = requestContext.ClientConfig.GetHttpProxy();
                 }
                 else if (_request.RequestUri.Scheme == Uri.UriSchemeHttps)
                 {
-                    _request.Proxy = requestContext.ClientConfig.GetHttpsProxy();
+                    proxy = requestContext.ClientConfig.GetHttpsProxy();
+                }
+
```
normj, line 488: This change was done because I found we broke debugging with fiddler because we are overriding the OS proxy by explicitly setting the `Proxy` property to null.

sdk/src/Services/DynamoDBv2/Custom/DocumentModel/Search.cs
```diff
@@ -378,8 +400,8 @@ internal async Task<List<Document>> GetNextSetHelperAsync(CancellationToken canc
 
                         Dictionary<string, Condition> keyConditions, filterConditions;
                         SplitQueryFilter(Filter, SourceTable, queryReq.IndexName, out keyConditions, out filterConditions);
-                        queryReq.KeyConditions = keyConditions;
-                        queryReq.QueryFilter = filterConditions;
+                        queryReq.KeyConditions = keyConditions.Count > 0 ? keyConditions : null;
```
dscpinheiro, line 403: Can `keyConditions` or `filterConditions` be null here? Earlier in this file you used `keyConditions?.Count` for example.

sdk/test/NetStandard/IntegrationTests/IntegrationTests/EC2/BasicDescribes.cs
```diff
@@ -48,7 +52,7 @@ public async System.Threading.Tasks.Task TestDescribeAmazonImagesCancellationTes
             });
 
             //Assert.Equal(token, exception.CancellationToken);
-            Assert.Equal(true, exception.CancellationToken.IsCancellationRequested);
+            Assert.True(exception.CancellationToken.IsCancellationRequested);
```
dscpinheiro, line 55: Nice, there are so many new (at least since .NET 8) xUnit warnings in the .NET Standard tests...

sdk/src/Services/S3/Custom/Model/_bcl45+netstandard/ListVersionsPaginator.cs
```diff
@@ -36,12 +36,12 @@ internal sealed partial class ListVersionsPaginator : IPaginator<ListVersionsRes
...
         public IPaginatedEnumerable<ListVersionsResponse> Responses => new PaginatedResponse<ListVersionsResponse>(this);
-        
+
         /// <summary>
         /// Enumerable containing all of the Versions
         /// </summary>
-        public IPaginatedEnumerable<S3ObjectVersion> Versions => 
-            new PaginatedResultKeyResponse<ListVersionsResponse, S3ObjectVersion>(this, (i) => i.Versions);
+        public IPaginatedEnumerable<S3ObjectVersion> Versions =>
+            new PaginatedResultKeyResponse<ListVersionsResponse, S3ObjectVersion>(this, (i) => i.Versions ?? new List<S3ObjectVersion>());
```
dscpinheiro, line 44: I don't see any changes to the paginator generators (e.g. https://github.com/aws/aws-sdk-net/blob/main/generator/ServiceClientGeneratorLib/Generators/SourceFiles/BasePaginator.tt); are they not impacted?

sdk/src/Services/EC2/Custom/Model/IpPermission.cs
```diff
@@ -74,7 +74,15 @@ internal List<IpRange> PreIpv4Ranges
         /// <returns></returns>
         internal IpRangeValue CanModify()
         {
-            List<string> ipv4Ranges = this._ipv4Ranges.Select(p => p.CidrIp).ToList();
+#pragma warning disable CS0618
+            if (this._ipv4Ranges != null && this.IpRanges == null)
```
dscpinheiro, line 78: I don't follow this change... Are we returning a different value now? I tried reading the `summary` but it seems out-of-date (for example, it mentions `ArgumentException` which this method doesn't throw at all).

sdk/src/Services/S3/Custom/Model/_bcl45+netstandard/ListVersionsPaginator.cs
normj: Good catch, fixed.

peterrsongg:
Still looking through the PR, but I believe there is another place that requires us to check the AWSConfigs variable which is in the [RestXMLrequestMarshaller.tt file](https://github.com/aws/aws-sdk-net/blob/3f0ce55d93e0ddada07a905fdddfde08aa29dfdd/generator/ServiceClientGeneratorLib/Generators/Marshallers/RestXmlRequestMarshaller.tt#L232) I ran the generator and checked Route53 since it is one of the few rest-xml services, and saw the marshaller was still checking to see if the count > 0.

normj:
> ChangeCidrCollectionRequestMarshaller

There was a null check already in the `RestXMLRequestMarshaller` generator.
```
                var publicRequestChanges = publicRequest.Changes;
                if (publicRequestChanges != null && publicRequestChanges.Count > 0) 
                {                        
                    xmlWriter.WriteStartElement("Changes", "https://route53.amazonaws.com/doc/2013-04-01/");
                    foreach (var publicRequestChangesValue in publicRequestChanges) 
                    {
```

sdk/src/Services/EC2/Custom/Model/IpPermission.cs
normj: I updated the doc comment to reflect the return is actually `IpRangeValue.Invalid` in the case of an invalid situation. 

The changes I made were to add the `pragma` to get rid of the warnings for calling obsolete methods and checks if one or both collection is null at the start. If only one has a value and the other is null then we know to use the one with a value. If both are null default `Ipv4Ranges` which is the same behavior as if both collections are empty.

This is some confusing code dealing with the obsolete `IpRanges` property and might be worth removing in V4.

peterrsongg:
> > I believe there is another place that requires us to check the AWSConfigs variable which is in the [RestXMLrequestMarshaller.tt file](https://github.com/aws/aws-sdk-net/blob/3f0ce55d93e0ddada07a905fdddfde08aa29dfdd/generator/ServiceClientGeneratorLib/Generators/Marshallers/RestXmlRequestMarshaller.tt#L232) I ran the generator and checked Route53 since it is one of the few rest-xml services, and saw the marshaller was still checking to see if the count > 0. ChangeCidrCollectionRequestMarshaller in Route53 is an example of where this occurs.
> 
> There was a null check already in the `RestXMLRequestMarshaller` generator.
> 
> ```
>                 var publicRequestChanges = publicRequest.Changes;
>                 if (publicRequestChanges != null && publicRequestChanges.Count > 0) 
>                 {                        
>                     xmlWriter.WriteStartElement("Changes", "https://route53.amazonaws.com/doc/2013-04-01/");
>                     foreach (var publicRequestChangesValue in publicRequestChanges) 
>                     {
> ```

Maybe I'm misunderstanding, but if the collection is not null but empty and AWSConfigs.InitializeCollections is set to false, wouldn't that if block be set to false and not send the empty collection?

dscpinheiro (approved):

normj:
> Maybe I'm misunderstanding, but if the collection is not null but empty and AWSConfigs.InitializeCollections is set to false, wouldn't that if block be set to false and not send the empty collection?

You were right. I was looking at avoiding the null issue but you are right it needed to be updated to send the empty collection if `AWSConfigs.InitializeCollections` is false. I just pushed that change up.

peterrsongg (approved):

PR #3235, opened 2024-03-26
Remove Legacy Enums

dscpinheiro:
## Description
Removes the `RequestRetryMode.Legacy` and `DefaultConfigurationMode.Legacy` enum values; in V4, the .NET SDK will default to `Standard`.

There are 2 commits in this PR:
- [Must be reviewed] Manual changes to update the enums: https://github.com/aws/aws-sdk-net/pull/3235/commits/0d23038b269f6ce3de28bce9904e058b4edb79e9
- Generated changes to the `DefaultConfiguration` class for all services: https://github.com/aws/aws-sdk-net/pull/3235/commits/4f56a074aa409d387676c499abf862e9e1541711

## Testing
- Built the `AWSSDK.CoreAndCustomUnitTests.NetFramework.sln` and `AWSSDK.NetFramework.sln` solutions
- Ran the unit tests for the custom solution; there are a few failures but those happen in the `v4` branch as well:
  - `RetryForHttpStatus200WithErrorResponse`
  - `LongerTokenCaseAsync`
  - `MinimalBearerAuthCaseAsync`
  - `SingerShouldOverrideExistingHeaderAsync`

## License

### DIFF
--- a/generator/ServiceClientGeneratorLib/DefaultConfiguration/DefaultConfigurationController.cs
+++ b/generator/ServiceClientGeneratorLib/DefaultConfiguration/DefaultConfigurationController.cs
@@ -9,8 +9,7 @@ namespace ServiceClientGenerator.DefaultConfiguration
     {
         /// <summary>
         /// Fully loads and populates <see cref="DefaultConfigurationModel"/> by parsing the
-        /// sdk-default-configurations.json file from disk.  Also adds the 'Legacy'
-        /// <see cref="DefaultConfigurationMode"/> with the correct defaults.
+        /// sdk-default-configurations.json file from disk.
         /// </summary>
         /// <param name="repositoryRootDirectoryPath">
         /// Full path to the root directory, ie contains the 'sdk' directory.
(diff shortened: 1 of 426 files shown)


sdk/test/UnitTests/Custom/Runtime/EndpointResolverTests.cs
```diff
@@ -19,15 +19,16 @@ public class EndpointResolverTests
...
         private const string DefaultStsEndpoint = @"https://sts.amazonaws.com/";
 
-        [TestMethod][TestCategory("UnitTest")]
+        [TestMethod]
+        [TestCategory("UnitTest")]
         [TestCategory("Runtime")]
         public void TestSuccessfulCall()
         {
             var endpointResolver = new EndpointResolver();            
             var executionContext = CreateExecutionContext(SetupConfig());
```
dscpinheiro, line 28: The `config` returned by the setup method now uses the `Standard` mode, so the determined endpoints have changed.

sdk/test/UnitTests/Custom/Runtime/DefaultConfigurationAutoModeResolverTests.cs
```diff
@@ -16,14 +16,14 @@
 using Amazon.Runtime;
 using Amazon;
 using Moq;
-using Xunit;
```
dscpinheiro, line 19: This class was using xUnit, but the project is configured for MSTest so I think the tests weren't running... I updated it to match the other custom tests.

sdk/src/Core/Amazon.Runtime/ClientConfig.cs
```diff
@@ -720,33 +711,6 @@ protected ClientConfig(IDefaultConfigurationProvider defaultConfigurationProvide
             Initialize();
         }
 
-        public ClientConfig() : this(new LegacyOnlyDefaultConfigurationProvider())
```
dscpinheiro, line 723: This constructor was not used by the service clients; I did have to modify one test to pass in a mocked provider.

normj (approved):

generator/ServiceClientGeneratorLib/DefaultConfiguration/DefaultConfigurationController.cs
```diff
@@ -52,33 +51,11 @@ public DefaultConfigurationModel LoadDefaultConfiguration(string repositoryRootD
             }
 
             var parsedModel = _defaultConfigurationParser.Parse(json);
-
-            EnrichLegacyMode(parsedModel);
-
+            
+            var validModes = parsedModel.Modes.Where(m => m.Name.ToPascalCase() != "Legacy").ToList();
```
muhammad-othman, line 55: Not sure if we should add this part, maybe we will need to add a new mode named `Legacy` in the future for any reason.

muhammad-othman (approved):

generator/ServiceClientGeneratorLib/DefaultConfiguration/DefaultConfigurationController.cs
dscpinheiro: I think that's unlikely, `Legacy` is already one of the modes (https://docs.aws.amazon.com/sdkref/latest/guide/feature-smart-config-defaults.html), if a new one is added in the future we'd have to pick a new name.

Either way, it'd require us to update `generator/ServiceClientGeneratorLib/DefaultConfiguration/DefaultConfigurationModel.cs` as well.

PR #3236, opened 2024-03-26
Update STS to always use regional endpoints

dscpinheiro:
## Description
Updates STS to default to `regional` endpoints (https://docs.aws.amazon.com/sdkref/latest/guide/feature-sts-regionalized-endpoints.html).

## Testing
- Built the `AWSSDK.CoreAndCustomUnitTests.NetFramework.sln` and `AWSSDK.NetFramework.sln` solutions
- Ran the unit tests that were modified in this PR
- Ran the unit tests in the `SecurityToken` solution
- Ran all generated endpoint tests (`AWSSDK_DotNet.UnitTests.Endpoints`)

## License


generator/ServiceClientGeneratorLib/Generators/Endpoints/EndpointResolver.partial.cs
```diff
@@ -24,7 +24,7 @@ private string GetValueSource(Parameter parameter)
                 case "AWS::S3Control::UseArnRegion": return "config.UseArnRegion";
                 case "AWS::S3::DisableMultiRegionAccessPoints": return "config.DisableMultiregionAccessPoints";
                 case "AWS::S3::UseGlobalEndpoint": return "config.USEast1RegionalEndpointValue == S3UsEast1RegionalEndpointValue.Legacy";
-                case "AWS::STS::UseGlobalEndpoint": return "config.StsRegionalEndpoints == StsRegionalEndpointsValue.Legacy";
+                case "AWS::STS::UseGlobalEndpoint": return "false";
```
dscpinheiro, line 27: I had to keep this built-in otherwise the generator would fail. It's only used in the STS endpoint rules: https://github.com/aws/aws-sdk-net/blob/cb7de169cc3666096b9f38b2cb13d30404d258cc/generator/ServiceModels/sts/sts-2011-06-15.endpoint-rule-set.json#L30-L36

I'll double check if there's any other impact, there are endpoint rules which set `"AWS::STS::UseGlobalEndpoint": true` but all of our generated tests are still passing.

sdk/test/Services/SecurityToken/UnitTests/Custom/StsRegionalFlagTests.cs
```diff
@@ -13,77 +13,23 @@
...
         private const string AwsStsRegionalEndpointsEnvironmentVariable = "AWS_STS_REGIONAL_ENDPOINTS";
 
-        [Ignore]
-        [DataTestMethod]
-        [DataRow("default", StsRegionalEndpointsValue.Legacy)]
-        [DataRow("other", StsRegionalEndpointsValue.Regional)]
-        [TestCategory("SecurityToken")]
-        /// Ensure that the STS Regional Flag is set according to the value
-        /// in the Shared Credentials File
-        public void TestCredentialsFileStsFlag(string profile, StsRegionalEndpointsValue expectedValue)
```
dscpinheiro, line 54: Moved this test to `sdk/test/UnitTests/Custom/Runtime/Credentials/SharedCredentialsFileTest.cs` (which has access to the `SharedCredentialsFileTestFixture`).

generator/ServiceClientGeneratorLib/DefaultConfiguration/DefaultConfigurationModel.cs
```diff
@@ -91,18 +91,11 @@ public enum RequestRetryMode
     }
 
     /// <summary>
-    /// Sts Regional Endpoints Value determines whether or not
-    /// to send the sts request to the regional endpoint or to
-    /// the global sts endpoint
+    /// StsRegionalEndpointsValue determines the service endpoint used to talk to STS.
     /// </summary>
     /// <remarks>This is a copy of Amazon.Runtime.StsRegionalEndpointsValue</remarks>
     public enum StsRegionalEndpointsValue
```
normj, line 97: Do we even need this enum anymore or can we just delete it. I think we can just delete it because the only reason we had it before was to support the legacy use case. Without the legacy endpoint resolution STS works like any other service and shouldn't need this enum.

sdk/src/Core/Amazon.Runtime/Enumerations.cs
```diff
@@ -48,17 +48,10 @@ public enum ErrorType
     }
 
     /// <summary>
-    /// Sts Regional Endpoints Value determines whether or not
-    /// to send the sts request to the regional endpoint or to
-    /// the global sts endpoint
+    /// StsRegionalEndpointsValue determines the service endpoint used to talk to STS.
     /// </summary>
     public enum StsRegionalEndpointsValue
```
normj, line 53: Do we even need this enum anymore or can we just delete it. I think we can just delete it because the only reason we had it before was to support the legacy use case. Without the legacy endpoint resolution STS works like any other service and shouldn't need this enum.

normj (changes requested):

generator/ServiceClientGeneratorLib/DefaultConfiguration/DefaultConfigurationModel.cs
dscpinheiro: I can delete the enums; I don't expect any new values to be added, but then we'd ignore the `AWS_STS_REGIONAL_ENDPOINTS` / `sts_regional_endpoints` options and the value from the default configuration file: https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Core/sdk-default-configuration.json#L5

Do you have any concerns with that?

dscpinheiro: I found the internal design for `STS Regionalized Endpoints`, seems safe to remove it:
> SDK major version bumps should ignore the STS Regional Endpoints option, and only resolve STS endpoints to the partition's STS regional endpoints. A major version bump SDK may choose to keep the STS Regional Endpoints option, but the option's default value must be regional.

Removed both enums (from Core and the generator) in: https://github.com/aws/aws-sdk-net/pull/3236/commits/d17f17508101fa70e4ebcabecdc5286e446205aa

This also changed the generated files, included them in a separate commit: https://github.com/aws/aws-sdk-net/pull/3236/commits/cfdf9f98777e79f443fde4823e7b77e009ed2471

normj (approved):

muhammad-othman (approved):

PR #3252, opened 2024-04-03
Protocol tests issue 24

peterrsongg:
## Description

This PR fixes DOTNET-7357 issue 24. `Issue 24: Trying to add a key to the query parameters that already exists`

For some context, this is to account for the [httpQueryParams smithy  trait](https://smithy.io/2.0/spec/http-bindings.html#httpqueryparams-trait). Since there is no such trait in C2J, we instead look at the target of the member, if it is a map we safely assume the httpQueryParams trait was applied (confirmed with smithy team this is ok). 

If you look at the smithy docs you'll see the following, which will help when reviewing the PR:

> If a member with the httpQueryParams trait and a member with the [httpQuery trait](https://smithy.io/2.0/spec/http-bindings.html#httpquery-trait) conflict, clients MUST use the value set by the member with the [httpQuery trait](https://smithy.io/2.0/spec/http-bindings.html#httpquery-trait) and disregard the value set by httpQueryParams. For example, given the following model:

* (Review manual changes) commit a9c3a40c87794a67d58ef952cdbb3f6c9ad72950 - Fixes for issue 24.
* (Reference generated changes) commit 980e91dfad8e1341fbb6e76d3ca69a1eb0a557cd - Generated changes for issue 24.

Builds on prior PR: https://github.com/aws/aws-sdk-net/pull/3251
Merge order: 104 (I'm going to start mine from 100 and increment to keep separate from Bo's)
Ticket: DOTNET-7357

## Testing

## Types of changes

## Checklist

## License

### DIFF
--- a/generator/ServiceClientGeneratorLib/Generators/Marshallers/BaseMarshaller.tt
+++ b/generator/ServiceClientGeneratorLib/Generators/Marshallers/BaseMarshaller.tt
@@ -187,11 +187,41 @@ using Amazon.Runtime.Internal.Util;
 			}
 		}
 	}
-
+#>
+<#+
 	// Only applicable for rest-json and rest-xml
     // Generates code to add any operation query members to the query string for the request
     // If the marshaller is customized, it marshalls the value before changing it to a string
 	protected void ProcessQueryStringMembers(string variableName, Operation operation)
+	{
+		if(operation.RequestHasQueryStringMembers)		
+		{
+            bool containsQueryParamMap = operation.RequestQueryStringMembers.Any(item => item.IsMap);
+            if(!containsQueryParamMap)
+            {
+                // Most services don't contain a query param map, and since the smithy spec specifies that if a query param map
+                // exists it must be used over a normal query param, we will separate these two cases out to different functions
+                // since handling queryparam maps + overwriting previous normal query param entries requires a lot more generated code.
+                ProcessQueryStringMembersNoMap(variableName,operation);
+            }
+            else
+            {
+
+			    foreach(var member in operation.RequestQueryStringMembers)
+			    {
+                    ProcessQueryStringMemberWithMap(3, member,variableName);
+                }
+#>
+<#+        
+			}
+		}
+	}
+#>
+<#+
+	// Only applicable for rest-json and rest-xml
+    // Generates code to add any operation query members to the query string for the request
+    // If the marshaller is customized, it marshalls the value before changing it to a string
+	protected void ProcessQueryStringMembersNoMap(string variableName, Operation operation)
 	{
 		if(operation.RequestHasQueryStringMembers)		
 		{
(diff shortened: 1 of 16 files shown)


generator/ServiceClientGeneratorLib/Generators/Marshallers/BaseMarshaller.tt
```diff
@@ -187,11 +187,41 @@ using Amazon.Runtime.Internal.Util;
 			}
 		}
 	}
-
+#>
+<#+
```
boblodgett, line 191: nit: shouldn't be needed.

sdk/src/Services/APIGateway/Generated/Model/Internal/MarshallTransformations/ImportRestApiRequestMarshaller.cs
```diff
@@ -60,14 +60,17 @@ public IRequest Marshall(ImportRestApiRequest publicRequest)
             request.HttpMethod = "POST";
 
             request.AddSubResource("mode", "import");
-            
             if (publicRequest.IsSetFailOnWarnings())
-                request.Parameters.Add("failonwarnings", StringUtils.FromBool(publicRequest.FailOnWarnings));
-            
+            {
+                if(!request.Parameters.ContainsKey("failonwarnings"))
```
boblodgett, line 65: I am confused how it can ever already exist and if so we don't add the value on the request. Can you clarify?

sdk/src/Services/APIGateway/Generated/Model/Internal/MarshallTransformations/GetSdkRequestMarshaller.cs
```diff
@@ -67,11 +67,12 @@ public IRequest Marshall(GetSdkRequest publicRequest)
             if (!publicRequest.IsSetStageName())
                 throw new AmazonAPIGatewayException("Request object does not have required field StageName set");
             request.AddPathResource("{stage_name}", StringUtils.FromString(publicRequest.StageName));
-            
             if (publicRequest.IsSetParameters())
             {
                 foreach(var kvp in publicRequest.Parameters)
                 {
+                    if(request.Parameters.ContainsKey(kvp.Key))
```
boblodgett, line 74: I am confused how the key would exist here in this case because Parameters is already a Dictionary. Is this the case where it could have already been a set property on the request that was already added? If that situation doesn't exist though it seems like we shouldn't be trying to remove keys.

peterrsongg: If you look at `PutRestApiRequestMarshaller`, if Mode is set the key will be added, but later down the line when going through the map of query params, it will try to add a key that is already there if that remove statement is not there. Having said that though, I just realized from reading the smithy doc that I implemented this backwards :( . I should be disregarding the value set by the query param of maps and taking the value of the httpQuery trait. I will need to double check the protocol test b/c it did pass for some reason

sdk/src/Services/APIGateway/Generated/Model/Internal/MarshallTransformations/ImportRestApiRequestMarshaller.cs
peterrsongg: Since the order of the properties being generated is not gauranteed, it is possible that the query param of maps inserted the key first, in which case we would have to check if the key already existed. This may not matter however, if you read my comment above

boblodgett (approved):

normj (approved):

PR #3253, opened 2024-04-03
Protocol tests issue 25

peterrsongg:
## Description

This PR fixes DOTNET-7357 issue 25. `Issue 25 Unable to unmarshall exceptions that send a namespace or uri in the __type key of the json payload. All JSON protocols. We are able to parse the __type key, but we don't cut off everything after the :

* (Review manual changes) commit 02e2cf40521- Fixes for issue 25.
* (Reference generated changes) N/A

Builds on prior PR: https://github.com/aws/aws-sdk-net/pull/3251
Merge order: 105 (I'm going to start mine from 100 and increment to keep separate from Bo's)
Ticket: DOTNET-7357

## Testing

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Internal/Transform/JsonErrorResponseUnmarshaller.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Transform/JsonErrorResponseUnmarshaller.cs
@@ -65,12 +65,7 @@ namespace Amazon.Runtime.Internal.Transform
                     {
                         // The error type can contain additional information, with ":" as a delimiter
                         // We are only interested in the initial part which is the error type
-                        var index = errorType.IndexOf(":", StringComparison.Ordinal);
-                        if(index != -1)
-                        {
-                            errorType = errorType.Substring(0, index);
-                        }
-                        type = errorType;
+                        type = ParseType(errorType);
                     }
                 }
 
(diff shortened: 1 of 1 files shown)


sdk/src/Core/Amazon.Runtime/Internal/Transform/JsonErrorResponseUnmarshaller.cs
```diff
@@ -92,8 +87,7 @@ public ErrorResponse Unmarshall(JsonUnmarshallerContext context)
                 }
 
                 // strip extra data from type, leaving only the exception type name
-                type = type == null ? null : type.Substring(type.LastIndexOf("#", StringComparison.Ordinal) + 1);
-
+                type = type == null ? null : ParseType(type.Substring(type.LastIndexOf("#", StringComparison.Ordinal) + 1));
```
boblodgett, line 90: If type != null and if code is null or empty it is going to call ParseType on type twice. That seems it would produce the wrong type value in that case. Or is it only passing because type wouldn't have another `:` in it?

peterrsongg: Hmm, if there was another `:` before the actual error, then it would return the wrong error type. But, in that case the first ParseType would return the wrong error anyways. 

I added the second ParseType here because if both the `__type` and header don't specify the error, type is populated in the if block (83-86), in which case it is the full type (i.e. `aws.protocoltests.restjson#FooError:http://internal.amazon.com/coral/com.amazon.coral.validate/`). In this case we need to call ParseType to get the error. 

The only case I can think of this returning the wrong error type is if there is a `:` in the the namespace, but the docs specify: 

> If a : character is present, then take only the contents before the first : character in the value.

so either way we just need to blindly take the index of the first `:`

boblodgett (approved):

muhammad-othman (approved):

normj (approved):

PR #3265, opened 2024-04-05
Protocol tests issue 31

peterrsongg:
## Description

This PR fixes DOTNET-7357 issue 31. `Issue 31 Rest-XML Requests, not setting the xmlNamespace on individual member values`
**note: I commented where the real changes are for an easier review**
* (Review manual changes) commit ac7f65da382 - Fixes for issue 31.
* (Reference generated changes) commit 0f424130564 - Generated changes for issue 31.

**MERGE into `protocol-tests` feature branch only**

Builds on prior PR: https://github.com/aws/aws-sdk-net/pull/3261
Merge order: 110 (I'm going to start mine from 100 and increment to keep separate from Bo's)
Ticket: DOTNET-7357

## Testing

N/A - the test cases are the protocol tests. A final dry-run will be done on the `protocol-tests` feature branch once everything is merged into the feature branch.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project
- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [ ] I have read the **README** document
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed

## License

- [x] I confirm that this pull request can be released under the Apache 2 license

[issues]: https://github.com/aws/aws-sdk-net/issues
[license]: http://aws.amazon.com/apache2.0/
[cla]: http://en.wikipedia.org/wiki/Contributor_License_Agreement

PR #3266, opened 2024-04-05
Protocol tests issue 30

peterrsongg:
## Description

This PR fixes DOTNET-7357 issue 30. `Issue 30 (Related to issue 29) Incorrectly unmarshalling responses which used xmlName on the request`

A little more context: This specifically fixes the case where xml name is used on a key or value of a map. This makes it so that we don't search for a hardcoded "key" or "value" and instead pass that down from the model.

* (Review manual changes) commit 86f10b32041 - Fixes for issue 30.
* (Reference generated changes) commit 675e5049eec - Generated changes for issue 30.

Builds on prior PR: https://github.com/aws/aws-sdk-net/pull/3265
Merge order: 111 (I'm going to start mine from 100 and increment to keep separate from Bo's)
Ticket: DOTNET-7357

## Testing

## Types of changes

## Checklist

## License


sdk/src/Core/Amazon.Runtime/Internal/Transform/UnmarshallerContext.cs
```diff
@@ -361,6 +363,39 @@ public Stream Stream
             }
         }
 
+        /// <summary>
+        /// For map shapes the KeyXmlName represents the "locationName" on the map shape's key member. This can be changed
+        /// via the @xmlName trait. This ensures that we don't check for a hardcoded "key" value, if the key's name was changed.
+        /// </summary>
+        public string KeyXmlName
```
boblodgett, line 370: nit: these could be `public string KeyXmlName { get; set; }` without needing member variables.

sdk/src/Core/Amazon.Runtime/Internal/Transform/UnmarshallerContext.cs
```diff
@@ -352,6 +352,8 @@ public class XmlUnmarshallerContext : UnmarshallerContext
         private bool disposed = false;
         private bool currentlyProcessingEmptyElement;
         private bool processEmptyElements = false; //Flip to true in v4
+        private string valueXmlName = "";
+        private string keyXmlName = "";
```
boblodgett, line 356: Should the defaults be `key` and `value`?

dscpinheiro (approved):
I'd follow Bo's suggestion to use auto-properties, but other than that PR looks good.

sdk/src/Core/Amazon.Runtime/Internal/Transform/UnmarshallerContext.cs
peterrsongg: added defaults

sdk/src/Core/Amazon.Runtime/Internal/Transform/UnmarshallerContext.cs
peterrsongg: used auto properties

generator/ServiceClientGeneratorLib/Generators/Marshallers/AWSQueryEC2ResponseUnmarshaller.tt
```diff
@@ -72,6 +72,8 @@ namespace <#=this.Config.Namespace #>.Model.Internal.MarshallTransformations
 					if (member.IsMap || member.IsList)
 					{
 #>
+                        context.KeyXmlName = "<#=member.Shape.KeyMarshallName#>";
```
normj, line 75: This mechanism of storing the current key and value name for the shape being marshalled to the context seems like it would be broken if we had values that were maps as well that had different `KeyMarshallName`. When we made the recursive call to to the child map being marshalled the context would be changed and then when we process the next element in the parent map we would have different `KeyXmlName` or `ValueXmlName`.

normj (changes requested):

generator/ServiceClientGeneratorLib/Generators/Marshallers/AWSQueryEC2ResponseUnmarshaller.tt
peterrsongg: Going to have to update this, to handle nested xmlNames on key/values. Will push an update with that fix later

peterrsongg:
Nested maps with xml names need more investigation. closing

PR #3277, opened 2024-04-09
fix(DynamoDBv2): check for properties marked `required` when loading objects via DynamoDBContext

### LINKED ISSUE #3276: DynamoDB Object persistence Model does not respect `required` keyword
### Describe the bug

C# 11 introduced the [required](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/required) keyword to signal to the compiler that class members must be initialized using an object initializer.

This is also used by `System.Text.Json.JsonSerializer` to perform data validation at runtime.

```csharp
using System;
using System.Text.Json;

var json = """{"Email": "onefish@example.com", "DisplayName": "Fish One"}""";

// This will fail if Email or DisplayName are missing above
var user = JsonSerializer.Deserialize<User>(json);

if (user is null) {
    return;
}

// No complaints about dereferencing a possible null-reference
Console.WriteLine(user.DisplayName.ToUpper());

// Class has two properties, Email and DisplayName, that can not be null.
class User {
    public required string Email {get; set;}
    public required string DisplayName {get; set;}
}
```

When retrieving a POCO from DynamoDB via `DynamoDBContext` this modifier is ignored, making it possible to create invalid objects accidentally.

Take the following DynamoDB table storing users with their email and a display name.

| Email (pk, string)  | DisplayName (string) |
| ------------------- | -------------------- |
| onefish@example.com | Fish One             |
| twofish@example.com | _null_               |
| redfish@example.com | _empty_              |

```csharp
var client = new AmazonDynamoDBClient();
var context = new DynamoDBContext(client);

var users = await context.ScanAsync<User>([]).GetRemainingAsync();

// It *should* be safe to assume that DisplayName is not null and .ToUpper() will not fail with a NullReferenceException
Console.WriteLine(string.Join('\n', users.Select(user => user.DisplayName.ToUpper()));)

// Same as in previous example
class User {
    [DynamoDBHashKey]
    public required string Email {get; set;}
    public required string DisplayName {get; set;}
}
```

Creating a `User` instance via `DynamoDBContext.ScanAsync<User>([])` can create an invalid object if a property marked `required` does not exist on the corresponding item in DynamoDB, or is null, thus breaking expected guarantees.

### Expected Behavior

In the code snippet above I would expect the DynamoDB SDK to fail fast when encountering invalid values, instead of passing them along, which can cause the kind of unexpected behavior that `required` is supposed to guard against.

### Current Behavior

The SDK creates invalid objects at runtime.

### Reproduction Steps

1. Create the following DynamoDB Table:

| Email (string, pk)  | DisplayName (string) |
| ------------------- | -------------------- |
| onefish@example.com | Fish One             |
| twofish@example.com | _null_               |
| redfish@example.com | _no value_           |

2. Apply the following patch to the main branch using `git apply`:

```diff
diff --git a/testapp/Program.cs b/testapp/Program.cs
new file mode 100644
index 00000000000..c12fd3b4a30
--- /dev/null
+++ b/testapp/Program.cs
@@ -0,0 +1,16 @@
+using Amazon.DynamoDBv2;
+using Amazon.DynamoDBv2.DataModel;
+
+var client = new AmazonDynamoDBClient();
+var context = new DynamoDBContext(client);
+
+var users = await context.ScanAsync<User>([], new() {OverrideTableName = Environment.GetEnvironmentVariable("DDB_TABLE")}).GetRemainingAsync();
+
+Console.WriteLine(string.Join('\n', users.Select(user => user.DisplayName.ToUpper())));
+
+// Class has two properties, Email and DisplayName, that should never be null.
+class User {
+    [DynamoDBHashKey]
+    public required string Email {get; set;}
+    public required string DisplayName {get; set;}
+}
\ No newline at end of file
diff --git a/testapp/testapp.csproj b/testapp/testapp.csproj
new file mode 100644
index 00000000000..db7d421377c
--- /dev/null
+++ b/testapp/testapp.csproj
@@ -0,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <ItemGroup>
+    <ProjectReference Include="..\sdk\src\Services\S3\AWSSDK.S3.NetStandard.csproj" />
+    <ProjectReference Include="..\sdk\src\Services\DynamoDBv2\AWSSDK.DynamoDBv2.NetStandard.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="AWSSDK.IdentityManagement" Version="3.7.300.58" />
+    <PackageReference Include="AWSSDK.SecurityToken" Version="3.7.300.59" />
+  </ItemGroup>
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+</Project>
```

3. Run the example

```bash
cd ./testapp
DDB_TABLE="<name of table>" dotnet run
```

It will fail when trying to invoke `ToUpper()` on a null value.

### Possible Solution

_No response_

I consider this a bug, since `System.Text.Json.JsonSerializer` upholds these guarantees at runtime, whereas `using Amazon.DynamoDBv2.DataModel.DynamoDBContext` does not.

AWSSDK.DynamoDBv2
Git commit 23b74540be451eda8bb789980b4863bd8cc4241f (HEAD at time of writing)

### Targeted .NET Platform

.NET 8

Windows 11 22H2

mstern02:
## Description
Add conditionally compiled checks for `RequiredMemberAttribute`, which is generated by the compiler for class members marked with the `required` keyword.

## Motivation and Context
C# 11 introduced the [required](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/required) keyword to signal to the compiler that class members must be initialized using an object initializer.

This is also used by `System.Text.Json.JsonSerializer` to perform data validation at runtime.

When retrieving a POCO from DynamoDB via `DynamoDBContext` this modifier is ignored, making it possible to create invalid objects accidentally.

Fixes #3276 

## Testing

1. Create the following DynamoDB Table:

| Email (string, pk)  | DisplayName (string) |
| ------------------- | -------------------- |
| onefish@example.com | Fish One             |
| twofish@example.com | _null_               |
| redfish@example.com | _no value_           |

2. Apply the following patch to the main branch using `git apply`:

```diff
diff --git a/testapp/Program.cs b/testapp/Program.cs
new file mode 100644
index 00000000000..c12fd3b4a30
--- /dev/null
+++ b/testapp/Program.cs
@@ -0,0 +1,16 @@
+using Amazon.DynamoDBv2;
+using Amazon.DynamoDBv2.DataModel;
+
+var client = new AmazonDynamoDBClient();
+var context = new DynamoDBContext(client);
+
+var users = await context.ScanAsync<User>([], new() {OverrideTableName = Environment.GetEnvironmentVariable("DDB_TABLE")}).GetRemainingAsync();
+
+Console.WriteLine(string.Join('\n', users.Select(user => user.DisplayName.ToUpper())));
+
+// Class has two properties, Email and DisplayName, that should never be null.
+class User {
+    [DynamoDBHashKey]
+    public required string Email {get; set;}
+    public required string DisplayName {get; set;}
+}
\ No newline at end of file
diff --git a/testapp/testapp.csproj b/testapp/testapp.csproj
new file mode 100644
index 00000000000..db7d421377c
--- /dev/null
+++ b/testapp/testapp.csproj
@@ -0,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <ItemGroup>
+    <ProjectReference Include="..\sdk\src\Services\S3\AWSSDK.S3.NetStandard.csproj" />
+    <ProjectReference Include="..\sdk\src\Services\DynamoDBv2\AWSSDK.DynamoDBv2.NetStandard.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="AWSSDK.IdentityManagement" Version="3.7.300.58" />
+    <PackageReference Include="AWSSDK.SecurityToken" Version="3.7.300.59" />
+  </ItemGroup>
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+</Project>
```

3. Run the example

```bash
cd ./testapp
DDB_TABLE="<name of table>" dotnet run
```

4. After these changes the program should fail with an `InvalidOperationException` thrown in the library, whereas in the current version an unexpected `NullReferenceException`  is thrown in the application code.

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextInternal.cs
+++ b/sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextInternal.cs
@@ -19,6 +19,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.IO;
 using System.Reflection;
+using System.Runtime.CompilerServices;
 
 using Amazon.DynamoDBv2.DocumentModel;
 using Amazon.DynamoDBv2.Model;
@@ -388,6 +389,11 @@ namespace Amazon.DynamoDBv2.DataModel
                         if (propertyStorage.IsVersion)
                             storage.CurrentVersion = entry as Primitive;
                     }
+                    #if NET7_0_OR_GREATER
+                    else if (Attribute.IsDefined(propertyStorage.Member, typeof(RequiredMemberAttribute))) {
+                        throw new InvalidOperationException("Unable to retrieve value from " + attributeName);
+                    }
+                    #endif
                 }
             }
         }
@@ -785,6 +791,12 @@ namespace Amazon.DynamoDBv2.DataModel
         // Get/Set object properties
         private static bool TrySetValue(object instance, MemberInfo member, object value)
         {
+            #if NET7_0_OR_GREATER
+            if (Attribute.IsDefined(member, typeof(RequiredMemberAttribute)) && value == null) {
+                return false;
+            }
+            #endif
+
             FieldInfo fieldInfo = member as FieldInfo;
             PropertyInfo propertyInfo = member as PropertyInfo;
 


normj:
My understanding of the change is this would prevent saving POCOs with the `required` keyword and that makes sense. But it looks like it is also blocking loading POCO that would violate the required keyword. That seems bad to give users no way to access their data. Maybe if there was a config option on the context config that would allow turning off the required check to give people a back door just in case something was going weird for them. I might be misunderstanding the situation for loading options today and it fails already because I don't think we are setting the required properties as part of the object initializer.

normj:
Closing for lack of activity.

PR #3279, opened 2024-04-09
Protocol tests issue 36

peterrsongg:
## Description

This PR fixes DOTNET-7357 issue 36. `Issue 36 Incorrectly handling @xmlAttribute trait.` The only service/shape where this is used is S3 Grantee. The generator change for the test expression was based on the S3Grantee unmarshaller.

* (Review manual changes) commit cfd5b3363e0 - Fixes for issue 36.
* (Reference generated changes) commit 01a22a68135 - Generated changes for issue 36

**MERGE into `protocol-tests` feature branch only**

Builds on prior PR: https://github.com/aws/aws-sdk-net/pull/3275 ( diff got messed up so recreated)
Merge order: 115 (I'm going to start mine from 100 and increment to keep separate from Bo's)
Ticket: DOTNET-7357

## Testing

N/A - the test cases are the protocol tests. A final dry-run will be done on the `protocol-tests` feature branch once everything is merged into the feature branch.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project
- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [ ] I have read the **README** document
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed

## License

- [x] I confirm that this pull request can be released under the Apache 2 license

[issues]: https://github.com/aws/aws-sdk-net/issues
[license]: http://aws.amazon.com/apache2.0/
[cla]: http://en.wikipedia.org/wiki/Contributor_License_Agreement

PR #3281, opened 2024-04-11
Protocol tests issue 37

peterrsongg:
## Description

This PR fixes DOTNET-7357 issue 37. ` Resolved hosts with prefix/labels not working for EC2Query / AWSQuery protocols` No query based service does this, only the protocol test services. We have this working in the Json/Xml based services, and already have a method that resolves the hostprefix with labels. All I did was add that method to query requests as well.

* (Review manual changes) commit a625ca2e8f3 - Fixes for issue 37.
* (Reference generated changes) commit 30b50855946 - Generated changes for issue 37

Builds on prior PR: https://github.com/aws/aws-sdk-net/pull/3279
Merge order: 116 (I'm going to start mine from 100 and increment to keep separate from Bo's)
Ticket: DOTNET-7357

## Testing

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/test/Services/EC2Protocol/Generated/Model/Internal/MarshallTransformations/EndpointOperationRequestMarshaller.cs
+++ b/sdk/test/Services/EC2Protocol/Generated/Model/Internal/MarshallTransformations/EndpointOperationRequestMarshaller.cs
@@ -60,7 +60,7 @@ namespace Amazon.EC2Protocol.Model.Internal.MarshallTransformations
             if(publicRequest != null)
             {
             
-            request.HostPrefix = $"foo.";
+                request.HostPrefix = $"foo.";
             }
             return request;
         }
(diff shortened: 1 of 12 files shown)


sdk/test/Services/EC2Protocol/Generated/Model/Internal/MarshallTransformations/EndpointOperationRequestMarshaller.cs
```diff
@@ -59,6 +59,8 @@ public IRequest Marshall(EndpointOperationRequest publicRequest)
 
             if(publicRequest != null)
             {
+            
+            request.HostPrefix = $"foo.";
```
boblodgett, line 63: This is using `request` but it is in a check for `publicRequest != null`. This doesn't look right and the indentation should be increased by 1 even in the case below where it doesn't have the if check.

peterrsongg: The indentation was wrong so I fixed it. But, it is correct to change the `request` object because that is what we are returning in the function. If you look a couple lines above as well, we alter the `Parameters` dictionary of the `request` object. We wouldn't want to alter the `publicRequest` object because that isn't returned to the user.

boblodgett (approved):

muhammad-othman (approved):

PR #3287, opened 2024-04-13
Protocol tests fix dryrun

boblodgett:
## Description

This PR fixes DOTNET-7357 dryrun build failures for all PRs merged in to the protocol-tests feature branch so far. This contains the following fixes:

* Tuple was used but didn't build on net35. Changed to a class to build on all targets.
* Fixes a crashing bug where context.Stream.Length was being checked but Length through an exception. IsEmptyResponse checking was done instead. This change belongs to @peterrsongg.
* JSON and REST XML marshalling tests were using hardcoded strings instead of HeaderKeys.* values used by the SDK. WHen a HeaderKeys.* value was changed the tests were not changed at the same time resulting in failure. Moved some hardcoded values to HeaderKeys.*. This is quality of life improvement at this point because we ended up rolling back issue 19 in the next point.
* Rolled back issue 19 `X-Amzn-Errortype` casing change as it is a breaking change based on services using the previous casing. Although headers are case insensitive they are not treated case insensitive by the SDK in all situations. This requires a larger fix.
* Misc build failure fixes in tests for timestamp changes from issue 4. Missing `</Param>` tag causing build failure.

**Commits**

* (Review manual changes) commit be6511a18068c5865a3a9d5f944d6ee02399de15 - Protocol tests fix dry-run build errors.
* (Reference generated changes) commit 5f68c08dfb0aa28bdc0d7559321c1877a2de4b8d - Generated changes for dryrun fixes.

Note: some changes in this PR having to do with IsEmptyResponse belong to @peterrsongg. They were included so that specific changes didn't need to be broken out into another branch after getting the dry run to pass.

Ticket: DOTNET-7357

## Testing

Dry run succeeded for the commits up to this point in this PR. There will be later dry runs as more issues are merged into the feature branch.

## Types of changes

## Checklist

## License


sdk/src/Core/Amazon.Runtime/Internal/Util/MultiValueHeaderParser.cs
```diff
@@ -51,12 +51,12 @@ public static List<string> ToStringList(string header)
 
             int startAtIndex = 0;
             var bytes = System.Text.Encoding.UTF8.GetBytes(header);
-            Tuple<string, int> readResponse;
```
dscpinheiro, line 54: There's another place where we need to use a `Tuple` in 3.5: https://github.com/aws/aws-sdk-net/blob/fab713ecd779581b23a46f4b6c500934e1744b9d/sdk/src/Core/Amazon.Runtime/Internal/Util/IniFile.cs#L50-L61

Would it make sense to follow the same pattern here? Then when `BCL35` is removed in `v4` we'll use the `Tuple` type from the BCL.

sdk/test/ProtocolTests/AWSSDK.ProtocolTests.Net45.sln
```diff
@@ -0,0 +1,103 @@
+﻿
```
dscpinheiro, line 1: Did you mean to include solution files for other target frameworks? (I'm assuming these are for running the protocol tests in our build system).

dscpinheiro (approved):

sdk/test/ProtocolTests/AWSSDK.ProtocolTests.Net45.sln
boblodgett: We haven't created the multi targeted netstandard solution file yet because some of the utility projects are targeted specifically for net framework. It requires upgrades there which are in progress.

sdk/src/Core/Amazon.Runtime/Internal/Util/MultiValueHeaderParser.cs
boblodgett: I will make this change as it does make since to automatically drop an extra class when we drop BCL35 even if it is all private now.

peterrsongg: I also think having the Tuple class like above would be good. That way the resulting code can stay the same when we drop BCL35.

peterrsongg (approved):

peterrsongg (approved):

dscpinheiro (approved):

PR #3294, opened 2024-04-23
DynamoDB: add support for transactional writes that include update operations with custom update and condition expressions

RenesansJG:
## Description
DynamoDB transactional writes that are created using the document model or the object model can now include update operations with custom update and condition expressions:
- Document model:
  - `AddKeyToUpdate` methods have been added to the `DocumentTransactWrite` class. The methods enable adding an update operation to the transaction with an arbitrary update expression instead of providing a `Document`.
  - Since the `TransactWriteItemOperationConfig` can also include a condition expression, the `AddKeyToUpdate` call can include both an update expression and a condition expression. This enables conflicts to occur between the two expressions if they use the same parameter names for attribute names and/or values. If both expressions contain the same parameter, but with different values, an exception is thrown.
- Object model:
  - `AddSaveKey` methods have been added to the `TransactWrite` class. The methods enable adding a save operation with an arbitrary update expression instead of providing an object. Optionally, a condition expression can be provided because due to the lack of object, no version check can occur. This also enables more complex condition expressions than the simple version check.
  - The `AddDeleteKey` methods now include an optional parameter for the condition expression. This is comparable to the same parameter of the `AddSaveKey` methods and is added for symmetry.

## Motivation and Context
The new operations are useful in scenarios when an item needs to be updated without retrieving it first. This can help avoid conflicts in some cases.

Example scenario:
- A table includes items with a partition key and a range key.
- We want to limit the number of items per partition.
- Another table contains items which include a counter. The partition key is the same as for the other table and there is no range key. Thus, the counters count the number of items per partition in the other table.
- Adding an item to the first table with checking the limit can be implemented the following way using the existing object model and document model operations:
  - Retrieve the counter.
  - Check whether the counter will reach its limit when incremented.
  - Add the new item and update the counter in a transaction.
- However, conflicts can occur if many such operations happen concurrently. The counter needs to include a version property to ensure each added item increments the counter by one. The version check may fail if the counter has been updated by another concurrent operation and the operation needs to be retried with retrieving the counter again. Multiple retries may be necessary if there are lots of concurrent operations.

DynamoDB update expressions can be used to increment a value by a given amount (e.g `SET counter = counter + 1`). Using an update expression without needing to use low-level operations is possible with the changes in this PR.

The example above can be implemented by using a transaction that includes the an operation that adds the new item and an update operation with an update expression that increments the counter and a condition expression that checks that the counter cannot surpass the limit.

## Testing
Existing DynamoDB integration tests have been extended to test the new operations for both models. All tests passed.

## Types of changes

## Checklist

## License


RenesansJG:
Hi! Do you have any update on reviewing this?

ashovlin:
> Hi! Do you have any update on reviewing this?

I'm working on our initial pass, aiming to wrap up tomorrow morning eastern time. Then I'll line up a second reviewer from the maintainers.

sdk/src/Services/DynamoDBv2/Custom/DocumentModel/DocumentTransactWrite.cs
```diff
@@ -325,6 +362,38 @@ internal void AddDocumentToUpdateHelper(Document document, Key key, TransactWrit
...
+            {
+                // Filter attribute names and values from the update expression that are also found in the condition expression
+                // If the value is different for the same key between the two expressions, an exception is thrown
+                // This ensures no exception is thrown when the attribute names and values of the update expression are merged into the request object
+                if (!TryFilterDuplicates(updateExpression.ExpressionAttributeValues,
+                        operationConfig.ConditionalExpression.ExpressionAttributeValues, out var attributeValues) ||
+                    !TryFilterDuplicates(updateExpression.ExpressionAttributeNames,
+                        operationConfig.ConditionalExpression.ExpressionAttributeNames, out var attributeNames))
+                {
+                    throw new ArgumentException("Conflicting values for key", "updateExpression");
```
ashovlin, line 377: 1. Could you expand this exception message to explain how to correct this, and which key triggered it if not too much more work?
2. Could you also cover the duplicate filtering via tests - assert that we're seeing this `ArgumentException` when expected?

sdk/src/Services/DynamoDBv2/Custom/DataModel/TransactWrite.cs
```diff
@@ -143,6 +143,39 @@ public void AddSaveItem(T item)
...
         }
 
+        /// <summary>
+        /// Add a single item to be saved in the current transaction operation.
+        /// Item is identified by its hash primary key and will be updated using the update expression provided.
+        /// </summary>
+        /// <param name="hashKey">Hash key of the item to delete.</param>
+        /// <param name="updateExpression">Update expression to use.</param>
+        /// <param name="conditionExpression">Condition to check before the operation.</param>
+        public void AddSaveKey(object hashKey, Expression updateExpression, Expression conditionExpression = null)
```
ashovlin, line 153: Thinking about naming here, `Key` feels implied since all items will have one. I wonder if something like `AddSaveExpression` or `AddSaveViaExpression` is more meaningful relative to the existing `AddSaveItem`.

(not asking for a change yet, curious what you or other reviewers think)

ashovlin (changes requested):

RenesansJG:
> I'm working on our initial pass, aiming to wrap up tomorrow morning eastern time. Then I'll line up a second reviewer from the maintainers.

Hi @ashovlin, thanks for the review, I'll address your comments soon.

sdk/src/Services/DynamoDBv2/Custom/DataModel/TransactWrite.cs
RenesansJG: Hmm... I think we have two alternate paths to take here instead of `AddSaveKey`. The first is to include the term "expression" in the name, like `AddSave(Item)(Via|By|With)(Update)Expression`. The second is to call this a new operation entirely, e.g. `AddUpdateItem`. I'm happy with any of these, whatever you decide.

sdk/src/Services/DynamoDBv2/Custom/DataModel/TransactWrite.cs
```diff
@@ -197,9 +231,18 @@ public void AddDeleteKey(object hashKey)
         /// </summary>
         /// <param name="hashKey">Hash key of the item to delete.</param>
         /// <param name="rangeKey">Range key of the item to delete.</param>
-        public void AddDeleteKey(object hashKey, object rangeKey)
+        /// <param name="conditionExpression">Condition to check before the operation.</param>
+        public void AddDeleteKey(object hashKey, object rangeKey, Expression conditionExpression = null)
```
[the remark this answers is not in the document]
RenesansJG: Done

sdk/src/Services/DynamoDBv2/Custom/DocumentModel/DocumentTransactWrite.cs
RenesansJG: Done, please check if it's okay this way

sdk/src/Services/DynamoDBv2/Custom/DocumentModel/DocumentTransactWrite.cs
```diff
@@ -325,6 +362,52 @@ internal void AddDocumentToUpdateHelper(Document document, Key key, TransactWrit
...
+                // If the value is different for the same key between the two expressions, an exception is thrown
+                // This ensures no exception is thrown when the attribute names and values of the update expression are merged into the request object
+                if (!TryFilterDuplicates(updateExpression.ExpressionAttributeValues,
+                        operationConfig.ConditionalExpression.ExpressionAttributeValues, out var attributeValues,
+                        out var conflictingValueKeys))
+                {
+                    throw new ArgumentException(
+                        "Conflicting attribute values found between the update expression and the condition expression. " +
+                        "Use a different key instead of each affected attribute value key in one of the expressions. " +
+                        $"Affected attribute value keys: {string.Join(", ", conflictingValueKeys)}.",
```
ashovlin, line 379: For the two new `string.Join` calls here, can you add `ToArray()

We still target .NET Framework 3.5, which doesn't have the `Join` overload for a list of strings.
```
Severity	Code	Description	Project	File	Line	Suppression State	Details
Error	CS1503	Argument 2: cannot convert from 'System.Collections.Generic.List<string>' to 'string[]'	AWSSDK.DynamoDBv2.Net35	C:\dev\aws-sdk-net-staging\sdk\src\Services\DynamoDBv2\Custom\DocumentModel\DocumentTransactWrite.cs	379	Active	
``

Comment (RenesansJG) [reply sdk/src/Services/DynamoDBv2/Custom/DocumentModel/DocumentTransactWrite.cs]: Done

**Comment by RenesansJG** [comment]
Hi @ashovlin, @96malhar, is there any update on when this PR will be merged?

File: sdk/src/Services/DynamoDBv2/Custom/DataModel/TransactWrite.cs (hunk 2 above)
Comment (ashovlin) [reply sdk/src/Services/DynamoDBv2/Custom/DataModel/TransactWrite.cs:153]: Looking at it again, I think I like just overloading `AddSaveItem`. That looks like what Java does here https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/enhanced/dynamodb/model/WriteBatch.Builder.html

But not requesting changes yet, maybe a second opinion from @normj or @boblodgett since the reviewed this API originally?

**Review by ashovlin — APPROVED** [review]

**Comment by ashovlin** [comment]
I just approved, verified it passes the build+tests in our internal release system, and prepared the versioning and changelog metadata via #3336. 

We still need a second approval, but having the first should help. I'll bring up at our internal stand-up. 

I also tagged a few others for the open naming thread https://github.com/aws/aws-sdk-net/pull/3294#discussion_r1601794330, but I'd be okay with what's in the current diff if no one else objects.

File: sdk/src/Services/DynamoDBv2/Custom/DataModel/TransactWrite.cs (hunk 2 above)
Comment (ashovlin) [reply sdk/src/Services/DynamoDBv2/Custom/DataModel/TransactWrite.cs:153]: Discussed with the team internally today - could you adjust it to keep overloading `AddSaveItem`?

We feel that functionally we're just adding another action to the transaction, so lets keep the same operation name regardless of the input variations.

File: sdk/src/Services/DynamoDBv2/Custom/DocumentModel/DocumentTransactWrite.cs [hunk 5]
```diff
@@ -158,6 +161,40 @@ public void AddDocumentToUpdate(Document document, TransactWriteItemOperationCon
             AddDocumentToUpdateHelper(document, TargetTable.MakeKey(document), operationConfig);
         }
 
+        /// <summary>
+        /// Add a single item to update, identified by its hash primary key, using the specified update expression and config.
+        /// </summary>
+        /// <param name="hashKey">Hash key of the document.</param>
+        /// <param name="updateExpression">Update expression to use.</param>
+        /// <param name="operationConfig">Configuration to use.</param>
+        public void AddKeyToUpdate(Primitive hashKey, Expression updateExpression, TransactWriteItemOperationConfig operationConfig = null)
```
ashovlin, line 170: Similar to https://github.com/aws/aws-sdk-net/pull/3294/files#r1648167618 above, could we keep calling these three `AddDocumentToUpdate`?

sdk/src/Services/DynamoDBv2/Custom/DataModel/TransactWrite.cs
RenesansJG: Done

sdk/src/Services/DynamoDBv2/Custom/DocumentModel/DocumentTransactWrite.cs
RenesansJG: Done

ashovlin (approved):
Reapproved, and verified that 833177b8a712fce16363a11ff6dc59eb18d57a45 builds and passes all tests on our internal CI system.

I'll mention at our internal stand-up again today that it's ready for another approver.

normj (approved):
Another great PR @RenesansJG. Thanks for your help improving the DynamoDB high level libraries.

RenesansJG:
> Another great PR @RenesansJG. Thanks for your help improving the DynamoDB high level libraries.

My pleasure, thank you for the review!

ashovlin:
This shipped yesterday in [AWSSDK.DynamoDBv2 3.7.304](https://www.nuget.org/packages/AWSSDK.DynamoDBv2/3.7.304). Thanks again for your contribution!

Issue #3218, opened 2024-03-08
AmazonCloudWatchLogsClient.GetLogEvents does not return any event


gerdb42:
### Describe the bug

I use the following code to retrieve LogEvents from CloudWatch (shortened for brevity):

```
var request = new GetLogEventsRequest
{
	LogGroupName = <valid log group name>,
	LogStreamName = <valid log stream name>
};

var response = await amazonCloudWatchLogsClient.GetLogEventsAsync(request);

response.Events.ForEach(t =>
{
    Console.Write ($"{t.Timestamp}: {t.Message}");
});
```

However, `response.Events` always contains 0 items. And despite `response.NextForwardToken` being non-empty, the request never returns any events and may be repeated forever. And I know the connection works and the log stream name is valid, because the name comes from an enumeration of log streams done just before. And the log steam contains events since I can see them in CloudWatch console.

### Expected Behavior

The returned response contains a reasonable number of events together with a non-`null` `NextForwardToken` if more events are available or `NextForwardToken` being `null` if there are no more events to be retrieved.

### Current Behavior

In the returned response, the Events collection contains 0 elements and the `NextForwardToken` is set. Repeating the request with `NextToken` set returns 0 events again.

### Reproduction Steps

See code sample above

### AWS .NET SDK and/or Package version used

AWSSDK.CloudWatchLogs 3.7.304.8

### Targeted .NET Platform

.NET 8

### Operating System and version

Windows 10

gerdb42:
Yes, setting `StartFromHead` to `true` does the Trick! So this means we can only get events in ascending order (oldest first). But I still get several results with 0 events before `NextForwardToken` equals the previous token.

It would be nice if `StartFromHead` would work as described and `NextForwardToken` was set to `null` upon delivery of the last chunk of events (to avoid unnecessary roundtrips).

peterrsongg:
@gerdb42 If tailing a log is what you want, CloudWatchLogs added a StartLiveTail API recently, and there is an example of how to configure it here. I helped write the example so I know that the code in there works. It might feel more natural than the current API you're using.

[Example in docs](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/example_cloudwatch-logs_StartLiveTail_section.html)

dscpinheiro:
[...]

PR #3296, opened 2024-04-24
Add a StopPaginationOnSameToken customization to CloudWatchLogs.GetLogEvents operation

muhammad-othman:
## Description
This PR fixes DOTNET-7441 by adding StopPaginationOnSameToken flag to Operation customizations in the generator and adds this flag to CloudWatchLogs.GetLogEvents operation customization.

## Motivation and Context

CloudWatchLogs.GetLogEvents operation response includes two pagination tokens (`NextBackwardToken` and `NextForwardToken`), but they won't be null once all events are retrieved (instead, CloudWatch will return the same value sent in the request)
This means the .[NET paginator](https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Services/CloudWatchLogs/Generated/Model/_bcl45%2Bnetstandard/GetLogEventsPaginator.cs#L63-L72) (which is generated from the model to keep invoking the API until the response token is empty) enters an infinite loop as the `NextForwardToken` will never be null.
Additionally this issue was opened #3218 and we couldn't advise to use the paginator since it will enter an infinite loop.

This PR contains the following commits to fix this issue:
* [Added StopPaginationOnSameToken flag to Operation customization in the generator](https://github.com/aws/aws-sdk-net/pull/3296/commits/f5b289aa6a8e94ca34c8e95b4c1bfe6189d00fd4) which updates the paginator to stop when the same token is sent on the response instead of expecting it to be null.
* [Added the newly created StopPaginationOnSameToken customization to CloudWatchLogs.GetLogEvents operation](https://github.com/aws/aws-sdk-net/pull/3296/commits/414e8853a2a219076ee33de8f0466acf184bbec3) 

## Testing

* Created a console app to test `GetLogEvents` pagination with a long and a short `LogStream` and validated that it the pagination stops when it reaches the end of the stream.
* Updated `GetLogEventsTest_TwoResponses` test to reflect the service response.
* Did a full dry run `DRY_RUN-36b366d9-c447-45f7-8764-df58b7cb716a`.

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/CloudWatchLogs/Generated/Model/_bcl45+netstandard/GetLogEventsPaginator.cs
+++ b/sdk/src/Services/CloudWatchLogs/Generated/Model/_bcl45+netstandard/GetLogEventsPaginator.cs
@@ -70,7 +70,7 @@ namespace Amazon.CloudWatchLogs.Model
                 nextToken = response.NextForwardToken;
                 yield return response;
             }
-            while (!string.IsNullOrEmpty(nextToken));
+            while (nextToken != _request.NextToken);
         }
 #endif
 #if AWS_ASYNC_ENUMERABLES_API
@@ -91,7 +91,7 @@ namespace Amazon.CloudWatchLogs.Model
                 cancellationToken.ThrowIfCancellationRequested();
                 yield return response;
             }
-            while (!string.IsNullOrEmpty(nextToken));
+            while (nextToken != _request.NextToken);
         }
 #endif
     }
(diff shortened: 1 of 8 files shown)


afroz429 (approved):

sdk/src/Services/CloudWatchLogs/Generated/Model/_bcl45+netstandard/GetLogEventsPaginator.cs
```diff
@@ -70,7 +70,7 @@ IEnumerable<GetLogEventsResponse> IPaginator<GetLogEventsResponse>.Paginate()
                 nextToken = response.NextForwardToken;
                 yield return response;
             }
-            while (!string.IsNullOrEmpty(nextToken));
```
dscpinheiro, line 73: The changes look good, but can you double check if the other SDKs (Java / JS) do something different for `NextBackwardToken`? (asking because this API allows customers to paginate in both directions)

dscpinheiro (approved):

PR #3302, opened 2024-04-25
Update InternalSDKUTils.DetermineFramework to return standard string …

peterrsongg:
…and additional metadata if version parsing fails

## Description
This updates the `DetermineFramework` method, which previously could return junk like "1-10-2023" or whatever was returned from `GetValidSubstringOrUnknown`. This is because people can build their own custom version of `System.Runtime.InteropServices.RuntimeInformation.dll`. 

In these cases, where the version is not in the standard format i.e. `n.n.n`, we now return a consistent string.
In order to mock the `DetermineFramework` method, I had to mock the static `RuntimeInformation` class. This class doesn't implement and interface so I had to create an interface and a wrapper to allow me to inject the mocked version of `RuntimeInformation`.

## Motivation and Context

Internal Ticket 7368
## Testing

Added Unit Tests
Dry run passes
## Types of changes

## Checklist

## License


sdk/src/Core/Amazon.Util/Internal/_netstandard/InternalSDKUtils.netstandard.cs
```diff
@@ -25,14 +25,30 @@
 using System.Text;
 
 namespace Amazon.Util.Internal
-{
+{        
+    /// <summary>
+    /// An interface for <see cref="RuntimeInformationWrapper"/> which allows for mocking the <see cref="RuntimeInformation"/> class.
+    /// A wrapper is necessary because the class is static.
+    /// </summary>
+    internal interface IRuntimeInformationWrapper
```
dscpinheiro, line 33: Does this interface need to marked `internal`? It's already in the `Internal` namespace, and in my opinion made this PR much more complicated than necessary.

peterrsongg: Norm asked me to make these internal and utilize `InternalsVisibleTo`😅.

sdk/src/Core/Amazon.Util/Internal/_netstandard/InternalSDKUtils.netstandard.cs
```diff
@@ -57,8 +73,13 @@ public static string DetermineFramework()
         {
             try
             {
-                var desc = RuntimeInformation.FrameworkDescription.Trim();
-                return string.Format(CultureInfo.InvariantCulture, ".NET_Core#{0}", GetValidSubstringOrUnknown(desc, desc.LastIndexOf(' ') + 1, desc.Length));
+                var desc = _runtimeInformationWrapper.FrameworkDescription.Trim();
+                var version = GetValidSubstringOrUnknown(desc, desc.LastIndexOf(' ') + 1, desc.Length);
+                if (!Version.TryParse(version, out var _))
+                {
+                    return "Unknown" + $" md/framework-raw-version#{desc}";
```
normj, line 80: For the `desc` being added to there can be spaces in it which could mess up parsing logic assuming the value after the `#` has no spaces. Can you replaces spaces in `desc` with underscores? `desc.Replace(' ', '_');`

sdk/src/Core/Amazon.Util/Internal/_netstandard/InternalSDKUtils.netstandard.cs
normj: I'm fine with either approach. My intention was I wanted the new interface to be treated as internal. Probably not a bad thing to get our unit tests project setup to access internals for future testing.

normj (changes requested):

sdk/src/Core/Amazon.Util/Internal/_netstandard/InternalSDKUtils.netstandard.cs
peterrsongg: I agree it's not a bad thing to have it set up so that we can use it in the future.

normj (approved):

sdk/test/NetStandard/UnitTests/AWSSDK.UnitTests.Custom.NetStandard.csproj
```diff
@@ -44,12 +56,10 @@ This project file should not be used as part of a release pipeline.
 
     <ItemGroup>
         <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.7.2" />
-        <PackageReference Include="xunit.runner.visualstudio" Version="2.4.5">
```
dscpinheiro, line 47: Was removing `PrivateAssets` intentional?

peterrsongg: Ah I forgot to put that back after some experimentation. I just updated it and added it back!

PR #3307, opened 2024-04-28
Optimize UrlEncode in Utils

danielmarbach:
## Description

Optimizes UrlEncode for better memory efficiency

## Motivation and Context

```

BenchmarkDotNet v0.13.12, macOS Sonoma 14.4.1 (23E224) [Darwin 23.4.0]
Apple M3 Max, 1 CPU, 14 logical and 14 physical cores
.NET SDK 8.0.202
  [Host]     : .NET 8.0.3 (8.0.324.11423), Arm64 RyuJIT AdvSIMD
  DefaultJob : .NET 8.0.3 (8.0.324.11423), Arm64 RyuJIT AdvSIMD


```

image

https://github.com/danielmarbach/aws-sdk-net-benchmarks/blob/main/UrlEncode.cs

## Testing

Added tests to verify the code still does the same as before. 

## Types of changes

## Checklist

## License


sdk/src/Core/AWSSDK.Core.NetStandard.csproj
```diff
@@ -23,6 +23,8 @@
     <NoWarn>$(NoWarn);CS1591;CA1822</NoWarn>
 
     <SignAssembly>True</SignAssembly>
+
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
danielmarbach, line 27: Like discussed before this is required for some of the shim APIs.

sdk/src/Core/Amazon.Runtime/Internal/Util/_netstandard/EncodingExtensions.cs
```diff
@@ -0,0 +1,81 @@
...
+using System;
+using System.Text;
+
+namespace Amazon.Runtime.Internal.Util
+{
+    /// <summary>
+    ///     Encoding extensions used to provide a compatibility shim for encoding methods that are missing in .NET Standard
+    ///     2.0.
+    /// </summary>
+    internal static class EncodingExtensions
```
danielmarbach, line 26: Let me know if you prefer another place to put it

sdk/src/Core/Amazon.Runtime/Internal/_bcl+netstandard/SkipLocalsInitAttribute.cs
```diff
@@ -0,0 +1,10 @@
+#if NETSTANDARD || NETCOREAPP
+namespace System.Runtime.CompilerServices
+{
+    /// <summary>Indicates to the compiler that the .locals init flag should not be set in nested method headers when emitting to metadata.</summary>
+    [AttributeUsage(AttributeTargets.Module | AttributeTargets.Class | AttributeTargets.Struct | AttributeTargets.Constructor | AttributeTargets.Method | AttributeTargets.Property | AttributeTargets.Event | AttributeTargets.Interface, Inherited = false)]
+    internal sealed class SkipLocalsInitAttribute : Attribute
```
danielmarbach, line 6: Let me know if you prefer another place to put it

sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
```diff
@@ -75,12 +77,6 @@ public static partial class AWSSDKUtils
         // Default value of progress update interval for streaming is 100KB.
         public const long DefaultProgressUpdateInterval = 102400;
 
-        internal static Dictionary<int, string> RFCEncodingSchemes = new Dictionary<int, string>
```
danielmarbach, line 78: I did check the entire SDK and couldn't find another reference so I got rid of it. Is it possible to entirely get rid of the different RFC implementations? If that is possible we could potentially even get rid of the custom code and replace it with the System.Net implementation? 

https://learn.microsoft.com/en-us/dotnet/api/system.net.webutility.urlencode?view=net-8.0

sdk/test/NetStandard/UnitTests/AWSSDK.UnitTests.Custom.NetStandard.csproj
```diff
@@ -6,7 +6,7 @@ This project file should not be used as part of a release pipeline.
 -->
 <Project Sdk="Microsoft.NET.Sdk">
     <PropertyGroup>
-        <TargetFrameworks>netstandard2.0;netcoreapp3.1;net8.0</TargetFrameworks>
+        <TargetFrameworks>netcoreapp3.1;net8.0</TargetFrameworks>
```
danielmarbach, line 9: Netstandard is not a valid target for unit tests

sdk/test/NetStandard/UnitTests/Core/AWSSDKUtilsTests.cs
```diff
@@ -0,0 +1,29 @@
...
+        public void UrlEncodeWithoutPath(string input, string expected)
+        {
+            var encoded = AWSSDKUtils.UrlEncode(input, path: false);
+
+            Assert.Equal(expected, encoded);
+        }
+
+        [Theory]
+        [InlineData("value, with special chars!", "value%2C%20with%20special%20chars!")]
+        [InlineData("value, with special chars and path {/+:}", "value%2C%20with%20special%20chars%20and%20path%20%7B/%2B:%7D")]
```
danielmarbach, line 21: Should I add more data from the valid path chars?

sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
```diff
@@ -1107,35 +1103,75 @@ public static string UrlEncode(string data, bool path)
...
             {
-                if (unreservedChars.IndexOf(symbol) != -1)
-                {
-                    encoded.Append(symbol);
-                }
-                else
+                if (!TryGetRFCEncodingSchemes(rfcNumber, out var validUrlCharacters))
+                    validUrlCharacters = ValidUrlCharacters;
+
+                var unreservedChars = string.Concat(validUrlCharacters, path ? ValidPathCharacters : "");
```
danielmarbach, line 1116: Given the encoding schemes are only used here the string concat could also go by already have the concatenated string available as costs and returning them by TryGetRFCEncodingSchemes

sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
```diff
@@ -1107,35 +1103,75 @@ public static string UrlEncode(string data, bool path)
...
+
+                var dataAsSpan = data.AsSpan();
+                var encoding = Encoding.UTF8;
+
+                var dataByteLength = encoding.GetMaxByteCount(dataAsSpan.Length);
+                var encodedByteLength = 2 * dataByteLength;
```
danielmarbach, line 1122: I left the 2 * dataLength in here since the previous method made similar assumptions.I assume this was chosen because it is unrealistic to only have data that will have to be encoded all the way like a string with only spaces or for example a string like "!!!!"

danielmarbach: image

danielmarbach: Never mind we should be good with 2 * data length

danielmarbach:
@normj @dscpinheiro I know this is still draft but would you mind having a look at it at some point? I have added a bunch of inline comments that might influence this PR.

danielmarbach:
Ping @normj

sdk/src/Core/AWSSDK.Core.NetStandard.csproj
danielmarbach: I don't understand this argument
The net runtime allows unsafe and runs pretty much on any platform. Azure SDK allows unsafe and runs on a variety of platforms too. Rabbitmq client which I contribute too is used in all sorts of environments and allows unsafe too. With the choice of your TFMs you will have to backport some amount of helpers that sometimes require "unsafe" modifier in order to balance getting perf and also reducing the complexity of the code base by reducing the number of redundant code paths that have to be implemented across the board.

Unsafe is unsafe in the sense that you need to know what you are doing and yes ultimately there should be as little as possible unsafe code but that should not mean there can't be any because of a fear of making a mistake.

Fyi the code here was ported from NET originally and brought into the Azure SDK and has been battle tested against a plethora of targets over years

sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
```diff
@@ -1107,35 +1103,75 @@ public static string UrlEncode(string data, bool path)
...
+                var dataBuffer = encodedByteLength <= MaxStackLimit
+                    ? stackalloc byte[MaxStackLimit]
+                    : sharedDataBuffer = ArrayPool<byte>.Shared.Rent(dataByteLength);
+                // instead of stack allocating or renting two buffers we over rent and add the data
```
normj, line 1126: The wording of this comment is confusing. Maybe something like

```
// Instead of stack allocating or renting two buffers we use one buffer with at least twice the capacity of the
// max encoding length. Then store the character data as bytes in the second half reserving the first half of the buffer 
// for the encoded representation.
```

sdk/test/NetStandard/UnitTests/Core/AWSSDKUtilsTests.cs
```diff
@@ -0,0 +1,29 @@
+using Amazon.Util;
+using Xunit;
+
+namespace UnitTests.NetStandard.Core
+{
+    [Trait("Category", "Core")]
+    public class AWSSDKUtilsTests
+    {
+        [Theory]
+        [InlineData("value, with special chars!", "value%2C%20with%20special%20chars%21")]
```
normj, line 10: Can we add some tests with multi byte characters?

sdk/test/NetStandard/UnitTests/Core/AWSSDKUtilsTests.cs
normj: You have proven that the characters from the valid path are being taken into account. I don't think this test needs to test each value that is in the collection.

sdk/src/Core/Amazon.Runtime/Internal/Util/_netstandard/EncodingExtensions.cs
normj: You can add these extensions method to this .NET Standard extensions file. https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Core/Amazon.Util/_netstandard/Extensions.cs

normj (changes requested):

sdk/src/Core/Amazon.Runtime/Internal/Util/_netstandard/EncodingExtensions.cs
danielmarbach: Done

sdk/src/Core/Amazon.Runtime/Internal/_bcl+netstandard/SkipLocalsInitAttribute.cs
danielmarbach: Done

sdk/test/NetStandard/UnitTests/Core/AWSSDKUtilsTests.cs
danielmarbach: I wanted to play extra safe TBH

sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
danielmarbach: Haha. My bad English shines through 👯

danielmarbach: Done

sdk/test/NetStandard/UnitTests/Core/AWSSDKUtilsTests.cs
danielmarbach: Done

danielmarbach:
I raised https://github.com/aws/aws-sdk-net/issues/3326 because the constants are a bit messy and need an overhaul. I did my best to get things into compilable state for now.

danielmarbach:
Ok I'll do a rebase

danielmarbach:
@normj rebased

normj (approved):
Looks good. I'll get somebody else on the team to do a second pass.

danielmarbach:
Rebased to solve the conflict

danielmarbach:
Would be nice if this could be merged since I have a bunch of other PRs that introduce more tests to the test file that probably require a rebase then ;)

boblodgett:
> Would be nice if this could be merged since I have a bunch of other PRs that introduce more tests to the test file that probably require a rebase then ;)

Taking a look

boblodgett:
I don't visually see issues in the PR. However, this didn't build in our build system because a change is required to support the /unsafe switch. I will make the change so we can ensure this PR will build and test correctly before I approve the PR.

sdk/src/Core/AWSSDK.Core.NetStandard.csproj
```diff
@@ -23,6 +23,8 @@
     <NoWarn>$(NoWarn);CS1591;CA1822</NoWarn>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <SignAssembly>True</SignAssembly>
+
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
normj, line 27: @danielmarbach can you add this same line to the `AWSSDK.Core.NetFramework.csproj` file? The internal builds are failing for the .NET Framework project because the PR doesn't allow unsafe code in .NET Framework.

normj (changes requested):

sdk/src/Core/AWSSDK.Core.NetStandard.csproj
danielmarbach: Sure. FYI all my PRs have push right

sdk/test/NetStandard/UnitTests/AWSSDK.UnitTests.Custom.NetStandard.csproj
boblodgett: Ran the build system again and got the following error:

```
C:\Program Files\dotnet\sdk\8.0.300\Sdks\Microsoft.NET.Sdk\targets\Microsoft.PackageDependencyResolution.targets(266,5): error NETSDK1005: Assets file 'C:\codebuild\tmp\output\src3842370628\src\aws-sdk-net\sdk\test\NetStandard\UnitTests\obj\AWSSDK.UnitTests.Custom.NetStandard\project.assets.json' doesn't have a target for 'netstandard2.0'. Ensure that restore has run and that you have included 'netstandard2.0' in the TargetFrameworks for your project.
```

I believe it is likely because of this change which needs to be reverted. Although you are correct that it really isn't a correct target to run for unit tests it is part of the solution. So when the solution is built with:

```
dotnet build -c Release --no-incremental $buildTarget -f netstandard2.0 /p:AWSKeyFile=$CustomSnkLocation /p:SignAssembly=true
```

The `-f netstandard2.0` switch which we used for parallel target building forces that it finds every project with that target in the solution. Since we have a multitargeted NetStandard.csproj and a single sln for Netstandard that doesn't have netstandard2.0 broken out it is required here.

boblodgett (changes requested):
https://github.com/aws/aws-sdk-net/pull/3307#discussion_r1664600695

sdk/test/NetStandard/UnitTests/AWSSDK.UnitTests.Custom.NetStandard.csproj
danielmarbach: Ok I'm going to revert this for now to not further derail things but I have questions 😂

danielmarbach, line 9:
```suggestion
        <TargetFrameworks>netstandard2.0;netcoreapp3.1;net8.0</TargetFrameworks>
```

danielmarbach: Done

normj (approved):
Approve with assumption @boblodgett gets a successful build through the internal system.

dscpinheiro:
@danielmarbach I approved and merged #3363 and #3365, but we saw integration test failures when running this PR on our build system (they were all related to this CloudFormation section: https://github.com/aws/aws-sdk-net/blob/3.7.844.0/sdk/test/NetStandard/IntegrationTests/IntegrationTests/CloudFormation.cs#L20).

Essentially that value is passed to `UrlEncode` (via `GetParametersAsString`), but `AWSSDKUtils.UrlEncode(TEMPLATE_TEXT, false)` in your branch is not returning the same value as `v4-development`.

danielmarbach:
@dscpinheiro Ok I'll check

danielmarbach:
@dscpinheiro I'm approaching my leave period so it might take a bit longer for me to respond. Apologies in advance

danielmarbach:
@dscpinheiro I fixed the problem. It was between the chair and the keyboard 👯

sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
```diff
@@ -1028,35 +1023,76 @@ public static string UrlEncode(string data, bool path)
...
-                    // Break apart the byte into two four-bit components and
-                    // then convert each into their hexadecimal equivalent.
-                    byte b = (byte)symbol;
-                    int hiNibble = b >> 4;
-                    int loNibble = b & 0xF;
-                    encoded.Append(ToUpperHex(hiNibble));
-                    encoded.Append(ToUpperHex(loNibble));
+                    if (unreservedChars.IndexOf((char)symbol) != -1)
+                    {
+                        dataBuffer[index++] = symbol;
```
danielmarbach, line 1057: Technically we could potentially remove bound checks here by doing something like 

```
Unsafe.Add(ref MemoryMarshal.GetReference(dataBuffer), index++) = symbol;
```

but I decided to not go down that path since it increases the complexity of the code

danielmarbach:
Give it another run through the benchmark

image

Still very good allocation reduction and the throughput hit is still in very acceptable range

boblodgett:
All the new changes are being run through the backend build system now.

boblodgett:
One test failure for netcoreapp3.1 target running tests.

```
error : Test failure in XUnitTestRunner: UnitTests.NetStandard.Core.AWSSDKUtilsTests.UrlEncodeWithoutPath(input: "{\r\n\t\"AWSTemplateFormatVersion\" : \"2010-09-09"..., expected: "%7B%0A%09%22AWSTemplateFormatVersion%22%20%3A%20%2"...)
```

danielmarbach:
I ran against net8 and as far as I recall net4.x. I'm not installing a runtime that is out of support for almost two years and hasn't received security updates since them. Sorry. 

So if you can provide me the entire test output with some information that makes it possible for me to guess what's going on I'm happy to give it another look.

Otherwise this PR has also edit rights. Feel free to push updates.

boblodgett:
> I ran against net8 and as far as I recall net4.x. I'm not installing a runtime that is out of support for almost two years and hasn't received security updates since them. Sorry.
> 
> So if you can provide me the entire test output with some information that makes it possible for me to guess what's going on I'm happy to give it another look.
> 
> Otherwise this PR has also edit rights. Feel free to push updates.


Thank you for working with us on this PR while we shake out the test issues. I ran the test in Visual Studio locally and it also fails in .NET 8.

The project containing the failed test is: `.\aws-sdk-net\sdk\test\NetStandard\UnitTests\AWSSDK.UnitTests.Custom.NetStandard.csproj`

The file containing the failed test within that project is: `.\aws-sdk-net\sdk\test\NetStandard\UnitTests\Core\AWSSDKUtilsTests.cs`

The test is: `public void UrlEncodeWithoutPath(string input, string expected)` which has three [InlineData] test cases. The case that is failing specifically is the third one which uses the `TemplateText`.

Here is a link to the test case: https://github.com/danielmarbach/aws-sdk-net/blob/f5fe55e95b98d07bb7cd2c2a8e8f4451f0b1b819/sdk/test/NetStandard/UnitTests/Core/AWSSDKUtilsTests.cs#L70

TemplateText can be found here: https://github.com/danielmarbach/aws-sdk-net/blob/f5fe55e95b98d07bb7cd2c2a8e8f4451f0b1b819/sdk/test/NetStandard/UnitTests/Core/AWSSDKUtilsTests.cs#L78

All other tests in this project pass.

boblodgett:
From the look of it `\r` (0x0D) characters are being injected and encoded when the test case hits a `\n` (0x0A):

image

sdk/test/NetStandard/UnitTests/Core/AWSSDKUtilsTests.cs
```diff
@@ -63,5 +63,76 @@ public void DetermineService(string url, string expectedService)
...
...roperties%22%20%3A%20%7B%0A%09%09%09%09%22QueueName%22%20%3A%20%7B%20%22Ref%22%20%3A%20%22QueueName%22%20%7D%0A%09%09%20%20%20%20%7D%0A%09%09%7D%2C%0A%0A%20%20%20%20%20%20%20%20%22TheTopic%22%20%3A%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20%22Type%22%20%3A%20%22AWS%3A%3ASNS%3A%3ATopic%22%2C%0A%20%20%20%20%20%20%20%20%20%20%20%20%22Properties%22%20%3A%20%7B%0A%09%09%09%09%22TopicName%22%20%3A%20%7B%20%22Ref%22%20%3A%20%22TopicName%22%20%7D%2C%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%22Subscription%22%20%3A%20%5B%0A%09%09%09%09%09%7B%22Protocol%22%20%3A%20%22sqs%22%2C%20%22Endpoint%22%20%3A%20%7B%22Fn%3A%3AGetAtt%22%20%3A%20%5B%20%22TheQueue%22%2C%20%22Arn%22%5D%7D%7D%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%5D%0A%20%20%20%20%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%20%20%7D%0A%09%7D%2C%0A%0A%09%22Outputs%22%20%3A%20%7B%0A%09%09%22TopicARN%22%20%3A%20%7B%0A%09%09%20%20%20%20%22Value%22%20%3A%20%7B%20%22Ref%22%20%3A%20%22TheTopic%22%20%7D%0A%09%09%7D%2C%0A%20%20%20%20%20%20%20%20%22QueueURL%22%20%3A%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20%22Value%22%20%3A%20%7B%20%22Ref%22%20%3A%20%22TheQueue%22%20%7D%0A%20%20%20%20%20%20%20%20%7D%0A%09%7D%0A%7D%0A")]
```
boblodgett, line 70: This is the test case that is failing which I didn't remember was a new test case until reviewing the changes again. This likely passes on a Linux based machine but will fail on Windows because of how some of the methods automatically translate `\n` to `\r\n` in windows.

boblodgett (changes requested):

sdk/test/NetStandard/UnitTests/Core/AWSSDKUtilsTests.cs
danielmarbach: Ah yeah the input needs to be normalized. I can only do that after my vacation. Feel free to push it if you have time.

danielmarbach: Does it happen on checkout on the build agent?

danielmarbach:
Honestly looking at this again on my phone I'd argue the only thing that would have been needed to cover the bug in https://github.com/aws/aws-sdk-net/pull/3307/commits/618ef3e80e8748c4f40162fe61bc95ba446ed521 would have been a long enough string. Fighting line endings and going back and forth on this PR for trying to fit the template text into here seems a bit pointless and unnecessarily holding this up. How about I remove this test case for now?

sdk/test/NetStandard/UnitTests/Core/AWSSDKUtilsTests.cs
danielmarbach, line 70:
```suggestion
```

boblodgett: Yes, there are going to be problems with the line endings depending on where the test is run. I would have to dig into seeing if the build system or where it is even happening on my Windows system but it will easily get switched to `\r\n` when a developer touches this file on Windows. Thus this test case is too fragile and should be removed.

boblodgett: I will strip that test case and run it again to make sure nothing else it blocking in the meantime.

danielmarbach:
@boblodgett based on the above comment I removed the template test for now. Having some tests in place for the method where there previously have been none is already a good start and the template was only a temporary thing anyway which helped me discover the buffer length problem. Ideally, there would be another test with a longer input but I cannot do this right now since I'm not around my laptop or workstation for a while

danielmarbach:
And thanks for your support here and sorry for me being less helpful this time

boblodgett (approved):

PR #3311, opened 2024-05-02
Remove DetermineServiceURL and refactor its usages to use config.DetermineServiceOperationEndpoint

muhammad-othman:
## Description
Remove DetermineServiceURL and refactor its usages to use config.DetermineServiceOperationEndpoint
There are 2 commits in this PR:
 * [Must be reviewed] Manual changes to Remove DetermineServiceURL and refactor its usages: https://github.com/aws/aws-sdk-net/pull/3311/commits/24d595ea70ea79de48fb516cd88d70cbbbe561ef
* Generated changes to move`DetermineServiceOperationEndpoint` from Client to ClientConfig: https://github.com/aws/aws-sdk-net/pull/3311/commits/ffc4c40500f7bd23d6324bc2179f1dda65026f84

## Motivation and Context
In order to remove `endpoints.json`, this PR start by  `DetermineServiceURL` as it is one of the most used methods that depends on  `endpoints.json`, and did the following to fix its usages:
* Moved `DetermineServiceOperationEndpoint` from Client to ClientConfig to make it easier to use in the pipelineHandlers since it doesn't depend on anything at the Client anyway.
* Updated unittests utilities that used `DetermineServiceURL` to `SetClockCorrectionForEndpoint` to instead set `GlobalClockCorrection`.
* Obsoleted `ClientConfig.ClockOffset` and suggested using `CorrectClockSkew.GetClockCorrectionForEndpoint(string endpoint)` instead.
* Updated the ClockOffsets that used to depend on `DetermineServiceURL`  to use `DetermineServiceOperationEndpoint` if possible or just depend on ManualClockCorrections.
   * This should be safe since `CorrectClockSkew.SetClockCorrectionForEndpoint` is set based on `executionContext.RequestContext.Request.Endpoint` (the actual endpoint that the request was sent to) and the `DetermineServiceURL` wouldn't return the exact endpoint that was used for the request in most cases and `GetClockCorrectionForEndpoint` will return the default value.

## Testing
Built `AWSSDK.NetFramework.sln` and `AWSSDK.NetStandard.sln` and made sure no compilation errors there via
```
dotnet build --force .\AWSSDK.NetFramework.sln && dotnet build --force .\AWSSDK.NetStandard.sln
```
Ran all the unit tests and integration tests of S3 and S3Control solutions since these were the most affected services.
The following test fails here and at v4-development branch as well:
* `S3ObjectLambdaBucketFieldInputTests`
    * This one depends on data related to `endpoints.json` so it will either be removed or changed significantly. Will double check it after finishing `endpoints.json` removal.

## License


generator/ServiceClientGeneratorLib/Generators/SourceFiles/ServiceConfig.tt
```diff
@@ -115,5 +116,26 @@ namespace <#=this.Config.Namespace#>
         public <#=parameter.nativeType#> <#=parameter.name#> { get; set; }
 
 <#}#>
+<# if (this.Config.EndpointsRuleSet != null) { #>
+        /// <summary>
+        /// Returns the endpoint that will be used for a particular request.
+        /// </summary>
+        /// <param name="request">Request for the desired service operation.</param>
+        /// <returns>The resolved endpoint for the given request.</returns>
+        public override Amazon.Runtime.Endpoints.Endpoint DetermineServiceOperationEndpoint(AmazonWebServiceRequest request)
```
normj, line 125: There is some possible work coming up that might involve account based endpoints which would be tied with the `AWSCredentials`. Putting this logic in the service config would make that work hard/impossible because service config has no concept of `AWSCredentials`. 

Maybe instead of taking in `AmazonWebServiceRequest` as a parameter take in a new container class like `EndpointParameters` which has a property for the `AmazonWebServiceRequest` and we can add other properties like account id in the future.

generator/ServiceClientGeneratorLib/Generators/SourceFiles/ServiceConfig.tt
```diff
@@ -115,5 +116,26 @@ namespace <#=this.Config.Namespace#>
...
+        {
+            var requestContext = new RequestContext(false, new <#=GeneratorHelpers.DetermineSigner(this.Config.ServiceModel.SignatureVersion, this.Config.ClassName)#>())
+            {
+                ClientConfig = this,
+                OriginalRequest = request,
+                Request = new DefaultRequest(request, ServiceId)
+            };
+
+            var executionContext = new Amazon.Runtime.Internal.ExecutionContext(requestContext, null);
+            var resolver = new Amazon<#=this.Config.ClassName#>EndpointResolver();
```
normj, line 135: Should we store the resolver as a member variable so we don't have to recreate for every request and possibly implement some caching in the resolver in the future.

normj (changes requested):

sdk/src/Core/Amazon.Runtime/ClientConfig.cs
```diff
@@ -1115,6 +1082,17 @@ private static void ValidateTcpKeepAliveTimeSpan(TimeSpan? value, string paramNa
                 : (clientTimeout.HasValue ? clientTimeout : null);
         }
 
+        /// <summary>
+        /// Returns the endpoint that will be used for a particular request.
+        /// </summary>
+        /// <param name="parameters">A Container class for parameters used for endpoint resolution.</param>
+        /// <returns>The resolved endpoint for the given request.</returns>
+        public virtual Endpoint DetermineServiceOperationEndpoint(ServiceOperationEndpointParameters parameters)
```
normj, line 1090: Seems like if this is ever called the null will cause a NullReferenceException. Should this method actually be `abstract`? Knowing this method will cause a failure in the request pipeline if called if there is a reason for it to not be `abstract` then maybe the base method should throw a meaningful error message.

muhammad-othman: I don't think the request pipeline would call it considering how the generator is written. I just wanted to follow the same structure that we have in `BaseEndpointResolver.cs` https://github.com/aws/aws-sdk-net/blob/043879987f2d23429b15ceec73d8979c258e731a/sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseEndpointResolver.cs#L261
But I guess making it `abstract` would be reasonable.

normj (approved):

dscpinheiro (approved):

PR #3312, opened 2024-05-07
Fix failing collection related tests for v4 dev

peterrsongg:
## Description

With the initializeCollections flag set to false by default, this caused some protocol tests to fail. This fixes a couple issues.
1. When comparing objects, if the initializeCollection flag is set to false then we will treat an empty collection and a null collection as the same (look at `Comparer.CompareObjects` for more detail)
2. If initializeCollection is set to false, then when unmarshalling a dictionary or a list which is sent back by a service, create the collection as to not get NullReferenceExceptions
3. If we are unmarshalling a collection and it is empty, then return a new instance of the collection if InitializeCollections is set to true, otherwise return null.

manual fixes - 6c259a95346
generated fixes - 3e68e7306d0
## Testing

Fixes Failed Tests:
```
    QueryXmlEmptyListsResponse
    Ec2XmlEmptyListsResponse
    FlatNestedXmlMapResponseResponse
    QueryXmlEmptyMapsResponse
    FlattenedXmlMapResponse
    QueryXmlEmptySelfClosedMapsResponse
    HttpPrefixHeadersAreNotPresentResponse
    XmlEmptySelfClosedMapsResponse
    AwsJson10DeserializeAllowNullsResponse
```
## Types of changes

## Checklist

## License

### DIFF
--- a/generator/ServiceClientGeneratorLib/Generators/Marshallers/RestXmlResponseUnmarshaller.tt
+++ b/generator/ServiceClientGeneratorLib/Generators/Marshallers/RestXmlResponseUnmarshaller.tt
@@ -187,10 +187,6 @@ namespace <#=this.Config.Namespace #>.Model.Internal.MarshallTransformations
 #>
 					if (context.TestExpression("<#=DetermineXmlMarshallName(member)#>", targetDepth))
 					{
-                        if (response.<#=member.PropertyName#> == null)
-                        {
-                            response.<#=member.PropertyName#> = new <#=member.DetermineType()#>();
-                        }
 						var unmarshaller = <#= member.DetermineTypeUnmarshallerInstantiate() #>;
 						response.<#=member.PropertyName#> = unmarshaller.Unmarshall(context);
 						continue;
--- a/generator/ServiceClientGeneratorLib/Generators/Marshallers/RestXmlResponseUnmarshaller.tt
+++ b/generator/ServiceClientGeneratorLib/Generators/Marshallers/RestXmlResponseUnmarshaller.tt
@@ -172,6 +172,10 @@ namespace <#=this.Config.Namespace #>.Model.Internal.MarshallTransformations
 #>
 					if (context.TestExpression("<#=DetermineXmlMarshallName(member)#>", targetDepth))
 					{
+                        if (response.<#=member.PropertyName#> == null)
+                        {
+                            response.<#=member.PropertyName#> = new <#=member.DetermineType()#>();
+                        }
 						var unmarshaller = <#= member.DetermineTypeUnmarshallerInstantiate() #>;
 						response.<#=member.PropertyName#>.Add(unmarshaller.Unmarshall(context));
 						continue;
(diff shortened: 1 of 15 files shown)


sdk/src/Core/Amazon.Runtime/Internal/Transform/SimpleTypeUnmarshaller.cs
```diff
@@ -921,7 +921,10 @@ public List<I> Unmarshall(JsonUnmarshallerContext context)
         {
             context.Read(); // Read [ or null
             if (context.CurrentTokenType == JsonToken.Null)
```
normj, line 923: Since this is no longer just a simple one line return use brackets for this `if` block.

generator/ServiceClientGeneratorLib/Generators/Marshallers/RestXmlResponseUnmarshaller.tt
```diff
@@ -172,6 +172,10 @@ namespace <#=this.Config.Namespace #>.Model.Internal.MarshallTransformations
 #>
 					if (context.TestExpression("<#=DetermineXmlMarshallName(member)#>", targetDepth))
 					{
+                        if (response.<#=member.PropertyName#> == null)
```
normj, line 175: I don't understand why are we initializing a value for these collections. The goal in V4 is to not initialize collections to avoid unused allocations and communicate to users what actually had a value from the service.

generator/ServiceClientGeneratorLib/Generators/Marshallers/RestXmlResponseUnmarshaller.tt
```diff
@@ -183,6 +187,10 @@ namespace <#=this.Config.Namespace #>.Model.Internal.MarshallTransformations
...
```
normj, line 190: Same comment above about initializing collections.

sdk/src/Core/Amazon.Runtime/Internal/Transform/SimpleTypeUnmarshaller.cs
```diff
@@ -975,7 +978,10 @@ public DictionaryUnmarshaller(TKeyUnmarshaller kUnmarshaller, TValueUnmarshaller
         {
             context.Read(); // Read { or null
             if (context.CurrentTokenType == JsonToken.Null)
```
normj, line 980: Add brackets for this now more complex `if` block.

sdk/test/UnitTests/Custom/TestTools/Comparer.cs
```diff
@@ -51,6 +53,14 @@ private static void Compare(object x, object y, Type type)
             if (x == null && y == null)
                 return;
 
+            if (x != null && x.GetType().GetInterfaces().Any(i => i.IsGenericType && i.GetGenericTypeDefinition() == typeof(ICollection<>)) && !AWSConfigs.InitializeCollections)
```
normj, line 56: The comparer should handle both possible boolean values for the `InitializeCollections` property.

normj (changes requested):

sdk/test/UnitTests/Custom/TestTools/Comparer.cs
peterrsongg: The other if statements handle when `InitializeCollections` is true, but I did simplify the if statement and the statement below.

normj (approved):

boblodgett (approved):

PR #3315, opened 2024-05-08
Improve Randomness in JitteredDelay Class

isaacschepp:
## Description
This pull request replaces the usage of System.Random with System.Security.Cryptography.RandomNumberGenerator in the JitteredDelay class. This enhances the randomness and security of random number generation. We generate a random long value, normalize it to a double in the range [0, 1), and use this double value to add a random amount of jitter to the retry delay.

## Motivation and Context
The change is required to enhance the security and randomness of the JitteredDelay class, particularly for security-sensitive operations. It improves the overall robustness of the system's retry mechanism.

## Testing
The changes have been tested to ensure the JitteredDelay class still functions as expected, and the new randomness generation does not introduce any errors or unexpected behavior. Tests were run in a local development environment.

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
+++ b/sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
@@ -1802,7 +1802,7 @@ namespace Amazon.Util
         private TimeSpan _maxDelay;
         private TimeSpan _variance;
         private TimeSpan _baseIncrement;
-        private Random _rand = null;
+        private RandomNumberGenerator _rand = null;
         private int _count = 0;
 
         public JitteredDelay(TimeSpan baseIncrement, TimeSpan variance)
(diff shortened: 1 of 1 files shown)


normj:
I ran the change through `Benchmark.NET` comparing using `System.Random` and `RandomNumberGenerator`. The `RandomNumberGenerator` version was quite a bit slower then using `System.Random`. Given the use of random in this case is not for any crypto graphic purposes, just adding a bit of randomness in retries, I don't think switching to `RandomNumberGenerator` is worth the performance hit.

```
| Method           | Mean      | Error     | StdDev    |
|----------------- |----------:|----------:|----------:|
| RandomTest       |  2.101 ns | 0.0661 ns | 0.1104 ns |
| RandomCryptoTest | 52.501 ns | 1.0247 ns | 1.0064 ns |
```

PR #3324, opened 2024-05-25
Optimize AWS4Signer

danielmarbach:
This PR heavily optimizes the HTTP header path on the AWS4Signer to reduce the number of allocations.

## Description

See screenshots

## Motivation and Context

Making sure the request path is less allocation heavy will drastically improve the throughput of the AWS SDK.

### CanonicalHeaderNames

CanonicalizeHeaderNames

https://github.com/danielmarbach/aws-sdk-net-benchmarks/blob/main/CanonicalHeaderNames.cs

### CanonicalHeaders

CanonicalHeaders

https://github.com/danielmarbach/aws-sdk-net-benchmarks/blob/main/CanonicalHeaderNames.cs

### CompressSpace

CompressSpace

https://github.com/danielmarbach/aws-sdk-net-benchmarks/blob/main/CompressSpace.cs

### CanonicalHeaders with CompressSpace

CanonicalHeadersCompressSpace

https://github.com/danielmarbach/aws-sdk-net-benchmarks/blob/main/CanonicalHeaderNames.cs

Attention this was just a short run because things take ages otherwise

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4Signer.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4Signer.cs
@@ -870,16 +870,21 @@ namespace Amazon.Runtime.Internal.Auth
         /// <returns>Canonicalized string of headers, with the header names in lower case.</returns>
         protected internal static string CanonicalizeHeaders(IEnumerable<KeyValuePair<string, string>> sortedHeaders)
         {
-            if (sortedHeaders == null || sortedHeaders.Count() == 0)
+            if (sortedHeaders == null)
                 return string.Empty;
 
-            var builder = new StringBuilder();
+            // Majority of the cases we will always have a IDictionary<string, string> for headers which implements ICollection<KeyValuePair<string, string>>.
+            var materializedSortedHeaders = sortedHeaders as ICollection<KeyValuePair<string, string>> ?? sortedHeaders.ToList();
+            if (materializedSortedHeaders.Count == 0)
+                return string.Empty;
+
+            using var builder = new ValueStringBuilder(512);
             
-            foreach (var entry in sortedHeaders)
+            foreach (var entry in materializedSortedHeaders)
             {
                 // Refer https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html. (Step #4: "To create the canonical headers list, convert all header names to lowercase and remove leading spaces and trailing spaces. Convert sequential spaces in the header value to a single space.").
                 builder.Append(entry.Key.ToLowerInvariant());
-                builder.Append(":");
+                builder.Append(':');
                 builder.Append(AWSSDKUtils.CompressSpaces(entry.Value)?.Trim());
                 builder.Append("\n");
             }
--- a/sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4Signer.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4Signer.cs
@@ -22,7 +22,7 @@ using Amazon.Internal;
 using Amazon.Util;
 using Amazon.Runtime.Internal.Util;
 using Amazon.Runtime.Endpoints;
-using AWSSDK.Core.NetStandard.Amazon.Runtime.Internal.Util;
+using ThirdParty.RuntimeBackports;
 
 namespace Amazon.Runtime.Internal.Auth
 {
--- a/sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4Signer.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4Signer.cs
@@ -819,7 +819,7 @@ namespace Amazon.Runtime.Internal.Auth
                                                     IDictionary<string, string> pathResources,
                                                     bool doubleEncode)
         {
-            using var canonicalRequest = new ValueStringBuilder(512);
+            var canonicalRequest = new ValueStringBuilder(512);
             canonicalRequest.Append($"{httpMethod}\n");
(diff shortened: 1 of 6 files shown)


danielmarbach:
@normj I would appreciate your input ;)

sdk/src/Core/Amazon.Runtime/Internal/Util/ValueStringBuilder.cs
```diff
@@ -0,0 +1,309 @@
...
+
+using System;
+using System.Buffers;
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+#nullable enable
+
+namespace AWSSDK.Core.NetStandard.Amazon.Runtime.Internal.Util
```
normj, line 13: Since this is source coming from the `dotnet/runtime` repo lets move this to a folder called `RuntimeBackports` under the [ThirdParty](https://github.com/aws/aws-sdk-net/tree/main/sdk/src/Core/ThirdParty) namespace of core. The file's namespace should be `ThirdParty.RuntimeBackports`. I'll figure out the attribution.

sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4Signer.cs
```diff
@@ -886,6 +887,30 @@ protected internal static string CanonicalizeHeaders(IEnumerable<KeyValuePair<st
...
         }
 
+        /// <summary>
+        /// Computes the canonical headers with values for the request. Only headers included in the signature
+        /// are included in the canonicalization process.
+        /// </summary>
+        /// <param name="sortedHeaders">All request headers, sorted into canonical order</param>
+        /// <param name="builder">The value string builder to append to.</param>
+        /// <returns>Canonicalized string of headers, with the header names in lower case.</returns>
+        protected internal static void AppendCanonicalizedHeaders(ICollection<KeyValuePair<string, string>> sortedHeaders, ref ValueStringBuilder builder)
```
normj, line 897: I'll need to do some research on why this was made protected but we don't want to expose `ValueStringBuilder` as part of the external API.

normj (changes requested):

sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4Signer.cs
```diff
@@ -818,26 +819,26 @@ public static string DetermineService(IClientConfig clientConfig, IRequest reque
...
         {
-            var canonicalRequest = new StringBuilder();
-            canonicalRequest.AppendFormat("{0}\n", httpMethod);
-            canonicalRequest.AppendFormat("{0}\n", AWSSDKUtils.CanonicalizeResourcePathV2(endpoint, resourcePath, doubleEncode, pathResources));
-            canonicalRequest.AppendFormat("{0}\n", canonicalQueryString);
-
-            canonicalRequest.AppendFormat("{0}\n", CanonicalizeHeaders(sortedHeaders));
-            canonicalRequest.AppendFormat("{0}\n", CanonicalizeHeaderNames(sortedHeaders));
+            var canonicalRequestBuilder = new ValueStringBuilder(512);
+            canonicalRequestBuilder.Append($"{httpMethod}\n");
```
Dreamescaper, line 823: Does it make sense to replace such cases with
```csharp
            canonicalRequestBuilder.Append(httpMethod);
            canonicalRequestBuilder.Append('\n');
```

To avoid allocating new strings with '\n' in the end?

danielmarbach: I would argue given the limited number of http methods the entire span could probably be optimized further
As far as I recall I could pull off a compiler trick here. I haven't gone as far as optimizing it to that level yet because there are still some outstanding inputs from the team about why some of these methods are even protected and whether the need to continue to be so

sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4Signer.cs
danielmarbach: Since I haven't received more feedback I'm going with a lighter variant that will use ValueStringBuilder internally within the methods only. The impact should already be quite high and this will leave the API surface untouched for now.

sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4Signer.cs
danielmarbach: Addressed

danielmarbach:
@normj I have to update the description of the PR but I'm running out of my fun coding time allotted budget for now. I have added some benchmark results that should speak though themselves. Of course ideally it would be better if we could bring the ValueStringBuilder ref struct into the helper methods but since I did not get feedback on the usage of those protected static methods I went ahead and used an approach that doesn't require signature changes which I think is in your favor anyway for now.

@Dreamescaper I also addressed one of your feedbacks

sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4Signer.cs
```diff
@@ -893,16 +898,7 @@ protected internal static string CanonicalizeHeaders(IEnumerable<KeyValuePair<st
...
-            
-            foreach (var header in sortedHeaders)
-            {
-                if (builder.Length > 0)
-                    builder.Append(";");
-                builder.Append(header.Key.ToLowerInvariant());
-            }
-            
-            return builder.ToString();
+            return string.Join(";", sortedHeaders.Select(x => x.Key.ToLowerInvariant()));
```
danielmarbach, line 901: Ran another benchmark on Fedora and decided to switch to ValueStringBuilder here too

image

peterrsongg (approved):
I ran the core and custom unit tests, mainly was concerned about the existing Signer tests which pass. Change looks good to me.

normj (approved):
Thanks @danielmarbach!

PR #3330, opened 2024-06-04
Add abstraction project for identity and auth

dscpinheiro:
## Description
This is the initial PR for updating the .NET SDK to be compliant with the Smithy Reference Architecture (SRA) Identity and Auth (internal issue: `DOTNET-7217`).

## Testing
- Built `AWSSDK.NetFramework.sln` and `AWSSDK.NetStandard.sln` solutions
- Ran unit tests locally
- Updated integration test to validate new property

## License


generator/ServiceClientGeneratorLib/GeneratorDriver.cs
```diff
@@ -812,7 +812,6 @@ public static void UpdateUnitTestProjects(GenerationManifest generationManifest,
         /// </summary>
         public static List<string> UpdateUnitTestProjects(GenerationManifest generationManifest, GeneratorOptions options, string serviceTestFilesRoot, ServiceConfiguration serviceConfiguration)
         {
-            Console.WriteLine("Updating unit test project files.");
```
dscpinheiro, line 815: This method calls an overload of `UpdateUnitTestProjects` which _also_ writes the same message, so I removed this to avoid duplication.

generator/ServiceClientGeneratorLib/Generators/SourceFiles/StructureGenerator.tt
```diff
@@ -234,8 +234,20 @@ namespace <#=this.Config.Namespace#>.Model
 
                     // Because some services have model properties named "Equals" which conflicts with .NET's Equals operation. Add the "new" keyword to override the behavior.
                     // The Retryable is there because the CloudHsmServiceException exception from CloudHSM models Retryable with a different return then what comes from the base class.
-#>
-        <#=member.AccessModifier#> <#=member.PropertyName.Equals("Equals") || (this.Structure is ExceptionShape && member.PropertyName.Equals("Retryable")) ? "new " : ""#><#=member.DetermineType()#> <#=member.PropertyName#>
+                    var shouldAddNewKeyword = 
+                        member.PropertyName.Equals("Equals") || 
+                        (this.Structure is ExceptionShape && member.PropertyName.Equals("Retryable"));
+
+                    // Cognito and SecurityToken also have a modeled property named Expiration, which conflicts with the interface used in SRA.
```
dscpinheiro, line 241: Both Cognito [1] and SecurityToken [2] have hand-written code that extend `AWSCredentials`. If I didn't add the `new` keyword, the solution would not compile.

[1] https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Services/CognitoIdentity/Custom/Model/Credentials.Extensions.cs
[2] https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Services/SecurityToken/Custom/Model/Credentials.Extensions.cs

sdk/test/IntegrationTests/Tests/General.cs
```diff
@@ -659,6 +661,7 @@ private static void TestCredentials(ProxyRefreshingAWSCredentials creds, bool ex
                 {
                     client.ListBuckets();
                     Assert.IsFalse(expectFailure);
+                    Assert.IsNotNull(creds.Expiration);
```
dscpinheiro, line 664: `ProxyRefreshingAWSCredentials` (the type for this `creds` object) extends `RefreshingAWSCredentials` so I'm verifying the property was populated correctly.

sdk/src/Core/Amazon.Runtime/Credentials/RefreshingAWSCredentials.cs
```diff
@@ -28,6 +28,20 @@ public abstract class RefreshingAWSCredentials : AWSCredentials, IDisposable
     {
         private Logger _logger = Logger.GetLogger(typeof(RefreshingAWSCredentials));
 
+        public override DateTimeOffset? Expiration
```
dscpinheiro, line 31: A few classes inherit from `ImmutableCredentials` instead, and they have an `Expiration` / `Expires` property as well (e.g. [AssumeRoleImmutableCredentials](https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Core/Amazon.Runtime/Credentials/AssumeRoleImmutableCredentials.cs) and [SSOImmutableCredentials](https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Core/Amazon.Runtime/Credentials/_bcl45%2Bnetstandard/SSOImmutableCredentials.cs)).

Please correct my understanding if it's wrong, but I assumed I didn't have to change them since they're used to create new `CredentialsRefreshState` objects (which end up in the `currentState` field): https://github.com/aws/aws-sdk-net/blob/dabf1d1241924fa3bc8ebc5f55eaac99881b9c4d/sdk/src/Core/Amazon.Runtime/Credentials/AssumeRoleWithWebIdentityCredentials.cs#L211-L217

extensions/src/Smithy.Identity.Abstractions/IIdentity.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+
+namespace Smithy.Identity.Abstractions
+{
+    /// <summary>
+    /// An identity is an entity within the SDK representing who the customer is.
+    /// For example, the customer's identity could be anonymouus, a token, a public / private key pair, etc...
+    /// </summary>
+    public abstract class IIdentity
```
normj, line 9: Since we are not actually an interface we should not start with the extra letter `I`. Since we are using abstract classes I would suggest for all of these abstract class uses the prefix `Base` or `Abstract`. My preference is `Base`.

extensions/src/Smithy.Identity.Abstractions/Smithy.Identity.Abstractions.csproj
```diff
@@ -0,0 +1,30 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <AssemblyName>Smithy.Identity.Abstractions</AssemblyName>
+    <GenerateAssemblyTitleAttribute>false</GenerateAssemblyTitleAttribute>
```
normj, line 6: I don't see an `AssemblyInfo.cs` for this project where you would specify all of these things that you have turned off generation. Either add one or remove these and add the required information. I'm not sure how you will handle the assembly versioning strategy we use in the SDK with `AssemblyVersion` and `AssemblyFileVersion` purely with the csproj file so you might need to add an `AssemblyInfo.cs`.

generator/ServiceClientGeneratorLib/Generators/SourceFiles/StructureGenerator.tt
normj: Since this is V4 I think we should just remove the `Expiration in the Cognito and SecurityToken credentials and let them just use the base class implementation.

I definitely don't want to use the `new` as that could cause problems of the wrong `Expiration` property being used depending on the cast of the credentials variable.

sdk/src/Core/Amazon.Runtime/Credentials/RefreshingAWSCredentials.cs
normj: But isn't the Expiration on these ImmutableCredentials sub class then redundant and possibly conflicting? I'm assuming these expiration properties were added on immutablecredentials because we didn't have the concept at AWSCredentials.

normj (changes requested):

generator/ServiceClientGeneratorLib/Generators/SourceFiles/StructureGenerator.tt
normj: I miss understood these were generated classes that we added a `partial` to make the be `AWSCredentials` objects.

sdk/src/Core/Amazon.Runtime/Credentials/RefreshingAWSCredentials.cs
normj: I miss understood these were generated classes that we added a `partial` to make the be `AWSCredentials` objects.

extensions/src/Smithy.Identity.Abstractions/IIdentity.cs
```diff
@@ -0,0 +1,17 @@
...
+    {
+        /// <summary>
+        /// The time after which this identity will no longer be valid. If this is null, an expiration time
+        /// is not known (for example, when using long-term credentials).
+        /// </summary>
+        public virtual DateTimeOffset? Expiration { get; }
```
normj, line 15: Since we have generated types like Cognito `Credentials` that are using `DateTime` we need use `DateTime` here. We can't change the return type in the subclasses.

Once we align on `DateTime` make the credential classes in Cognito and SecurityToken use `override` instead of `new`. We can't use `new` in this case because everything in Core will treat the object as an `IIdentity` so it will use the implementation on `IIdentity` instead of the subclass.

The alternative is we change everywhere in the SDK to use `DateTimeOffset`. There are some valid arguments for changing so if you want to have a wider discussion about changing globally to `DateTimeOffset` we can. @boblodgett you had some thoughts about switching to `DateTimeOffset` for  V4. I suspect the change is to big of a breaking change but given we did change already to nullable types maybe it is worth doing.

normj (changes requested):

extensions/src/Smithy.Identity.Abstractions/IIdentity.cs
dscpinheiro: Updated the type to `DateTime` in https://github.com/aws/aws-sdk-net/commit/285e421c304ad6d5c21deb54e6f85ec374226275

extensions/src/Smithy.Identity.Abstractions/Smithy.Identity.Abstractions.csproj
dscpinheiro: Added the assembly info file in https://github.com/aws/aws-sdk-net/pull/3330/commits/285e421c304ad6d5c21deb54e6f85ec374226275#diff-37d2151f0ebc200ab67497b246f32ca3d73c3a9d1a1ac999098f5229eb7ebdcf

extensions/src/Smithy.Identity.Abstractions/IIdentity.cs
dscpinheiro: Renamed the class to `BaseIdentity`: https://github.com/aws/aws-sdk-net/pull/3330/commits/285e421c304ad6d5c21deb54e6f85ec374226275#diff-37d2151f0ebc200ab67497b246f32ca3d73c3a9d1a1ac999098f5229eb7ebdcf

extensions/src/Smithy.Identity.Abstractions/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Reflection;
+
+[assembly: AssemblyTitle("Smithy.Identity.Abstractions")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyProduct("Smithy identity and authentication abstractions for the .NET SDK")]
+[assembly: AssemblyCompany("Amazon.com, Inc")]
+[assembly: AssemblyCopyright("Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.")]
+[assembly: AssemblyTrademark("")]
+
+[assembly: AssemblyVersion("0.1")]
```
normj, line 10: Can you do the same as @boblodgett did on the extensions AssemblyInfo and add the `#if` for .NET Framework with how we handle `AssemblyVersion`? https://github.com/aws/aws-sdk-net/blob/v4-development/extensions/src/AWSSDK.Extensions.CrtIntegration/Properties/AssemblyInfo.cs#L19

This also means in the csproj you need to add the .NET Framework target 4.7.2.

extensions/src/Smithy.Identity.Abstractions/Smithy.Identity.Abstractions.csproj
```diff
@@ -0,0 +1,30 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
normj, line 4: Add the .NET Framework targets as I mentioned above so we can do conditional compilations for .NET Framework.

extensions/src/Smithy.Identity.Abstractions/Smithy.Identity.Abstractions.csproj
```diff
@@ -0,0 +1,30 @@
...
+    <GenerateAssemblyCopyrightAttribute>false</GenerateAssemblyCopyrightAttribute>
+    <GenerateAssemblyVersionAttribute>false</GenerateAssemblyVersionAttribute>
+    <GenerateAssemblyFileVersionAttribute>false</GenerateAssemblyFileVersionAttribute>
+    <SignAssembly>True</SignAssembly>
+  </PropertyGroup>
+
+  <Choose>
+    <When Condition=" '$(AWSKeyFile)' == '' ">
+      <PropertyGroup>
+        <AssemblyOriginatorKeyFile>..\..\..\sdk\awssdk.dll.snk</AssemblyOriginatorKeyFile>
```
normj, line 20: Because we might extract this into a separate repo and product sometime in the future I wonder if we should have a separate `snk` file for Smithy extensions. If we think X years in future and .NET Smithy being own standalone product it would be weird to have it share the AWS SDK `snk` file.

normj (approved):
A few minor comments but approving trusting you will take care of the comments.

extensions/src/Smithy.Identity.Abstractions/Properties/AssemblyInfo.cs
dscpinheiro: Yes, I added the `#if` statements in https://github.com/aws/aws-sdk-net/pull/3330/commits/8d996a29a697a419323c524209c5af250b4d5280 (I ended up using `0.1` for both versions for now, let me know if I should use a different value instead).

extensions/src/Smithy.Identity.Abstractions/Smithy.Identity.Abstractions.csproj
dscpinheiro: Changed the target frameworks in https://github.com/aws/aws-sdk-net/pull/3330/commits/8d996a29a697a419323c524209c5af250b4d5280 (using `net462` as this is the current value in the `v4` branch, will update it as soon we make the switch to `net472`).

extensions/src/Smithy.Identity.Abstractions/Smithy.Identity.Abstractions.csproj
dscpinheiro: That makes sense, I created a new `snk` file in https://github.com/aws/aws-sdk-net/pull/3330/commits/8d996a29a697a419323c524209c5af250b4d5280 (using the `sn` command described in https://learn.microsoft.com/en-us/dotnet/standard/assembly/create-public-private-key-pair).

extensions/src/Smithy.Identity.Abstractions/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System.Reflection;
+
+[assembly: AssemblyTitle("Smithy.Identity.Abstractions")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyProduct("Smithy identity and authentication abstractions for the .NET SDK")]
```
boblodgett, line 5: `AWS SDK for .NET` or we have used `AWS .NET SDK` in the assembly info is more accurate  for customer facing descriptions.

sdk/src/Core/Amazon.Runtime/Credentials/RefreshingAWSCredentials.cs
```diff
@@ -28,6 +28,19 @@ public abstract class RefreshingAWSCredentials : AWSCredentials, IDisposable
     {
         private Logger _logger = Logger.GetLogger(typeof(RefreshingAWSCredentials));
 
+        public override DateTime? Expiration
```
boblodgett, line 31: I would like to see docs on public properties as we add new ones at the minimum.

dscpinheiro: I added an `<inheritdoc>` tag so the documentation from the base class is propagated.

extensions/src/Smithy.Identity.Abstractions/Properties/AssemblyInfo.cs
dscpinheiro: Changed the product suffix to `for the AWS SDK for .NET`

boblodgett (approved):

PR #3337, opened 2024-06-07
Fix SuppressNextMessage nullabilty for DialogAction

Gutse:
Fix nullability of property SuppressNextMessage  for class Amazon.LexModelsV2.Model.DialogAction
## Description

change property from bool to bool?
## Motivation and Context

Unable to build bot after updating slot with next action=FulfillIntent. In this case this property should be null.

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/LexModelsV2/Generated/Model/DialogAction.cs
+++ b/sdk/src/Services/LexModelsV2/Generated/Model/DialogAction.cs
@@ -64,9 +64,9 @@ namespace Amazon.LexModelsV2.Model
         /// When true the next message for the intent is not used.
         /// </para>
         /// </summary>
-        public bool SuppressNextMessage
+        public bool? SuppressNextMessage
         {
-            get { return this._suppressNextMessage.GetValueOrDefault(); }
+            get { return this._suppressNextMessage; }
             set { this._suppressNextMessage = value; }
         }
 


bhoradc:
Hi @Gutse,

Can you please share the code snippet for the issue you are facing? So, we can take a deeper look at the issue. 

Regards, 
Chaitanya

normj:
@Gutse The code you are changing is generated code so it won't work. There mechanism in our code generator to customize the property to force nullablity but that would be a breaking change.

We are working on V4 of the SDK that we hope to start shipping preview packages "soon". In V4 we have changed all the value types for properties in the generated model namespaces to use nullability to avoid problems like this. Here you can see the V4 version of the class you changed. https://github.com/aws/aws-sdk-net/blob/v4-development/sdk/src/Services/LexModelsV2/Generated/Model/DialogAction.cs#L67

Can you use your reflection trick for now till we get V4 out or are you blocked?

Gutse:
> @Gutse The code you are changing is generated code so it won't work. 

yeah, I was thinking about this, but had to give it a try anyway =)

> We are working on V4 of the SDK that we hope to start shipping preview packages "soon". In V4 we have changed all the value types for properties in the generated model namespaces to use nullability to avoid problems like this. Here you can see the V4 version of the class you changed. https://github.com/aws/aws-sdk-net/blob/v4-development/sdk/src/Services/LexModelsV2/Generated/Model/DialogAction.cs#L67

Great, thanks!

> Can you use your reflection trick for now till we get V4 out or are you blocked? To make the change in V3 we would have to do some impact analysis to see if we can justify the breaking change.

sure

PR #3335, opened 2024-06-06
Change version strategy in V4 to set AssemblyVersion

normj:
V3 we always set the `AssemblyVersion` because the version number is part of the assembly identity. If you change the version number any assemblies that depend on have to recompile for the new version. This strategy dated back to when the SDK only targeted .NET Framework. 

In .NET Core the version number is not part of the assembly identity so users are not required to recompile. This PR changes the strategy for .NET Core set the `AssemblyVersion` to the same value of `AssemblyFileVersion`. For the .NET Framework targets we continue to use the same strategy.

Example of a generated `AssemblyInfo.cs
```csharp
using System;
using System.Reflection;
using System.Runtime.InteropServices;
using System.Runtime.CompilerServices;

// General Information about an assembly is controlled through the following 
// set of attributes. Change these attribute values to modify the information
// associated with an assembly.
[assembly: AssemblyTitle("AWSSDK.AmplifyBackend")]
#if BCL
[assembly: AssemblyDescription("The Amazon Web Services SDK for .NET (4.6.2) - AmplifyBackend. (New Service) The Amplify Admin UI offers an accessible way to develop app backends and manage app content. We recommend that you use the Amplify Admin UI to manage the backend of your Amplify app.")]
#elif NETSTANDARD20
[assembly: AssemblyDescription("The Amazon Web Services SDK for .NET (NetStandard 2.0) - AmplifyBackend. (New Service) The Amplify Admin UI offers an accessible way to develop app backends and manage app content. We recommend that you use the Amplify Admin UI to manage the backend of your Amplify app.")]
#elif NETCOREAPP3_1
[assembly: AssemblyDescription("The Amazon Web Services SDK for .NET (.NET Core 3.1) - AmplifyBackend. (New Service) The Amplify Admin UI offers an accessible way to develop app backends and manage app content. We recommend that you use the Amplify Admin UI to manage the backend of your Amplify app.")]
#elif NET8_0
[assembly: AssemblyDescription("The Amazon Web Services SDK for .NET (.NET 8.0) - AmplifyBackend. (New Service) The Amplify Admin UI offers an accessible way to develop app backends and manage app content. We recommend that you use the Amplify Admin UI to manage the backend of your Amplify app.")]
#else
#error Unknown platform constant - unable to set correct AssemblyDescription
#endif

[assembly: AssemblyConfiguration("")]
[assembly: AssemblyProduct("Amazon Web Services SDK for .NET")]
[assembly: AssemblyCompany("Amazon.com, Inc")]
[...]

PR #3341, opened 2024-06-10
Change extension projects version strategy in V4 to set AssemblyVersion.

boblodgett:
## Description

This PR builds on the generator change PR of the same type here: https://github.com/aws/aws-sdk-net/pull/3335

The extensions based projects do not go through the generator and must be manually set. There is no BCL typedef. Instead NETFRAMEWORK is used as is specified here: https://learn.microsoft.com/en-us/dotnet/core/tutorials/libraries#preprocessor-symbols

## Testing

Built AWSSDK.Extensions.sln and ran all tests within the solution. All pass.

## Types of changes

## Checklist

## License

### DIFF
--- a/extensions/src/AWSSDK.Extensions.NETCore.Setup/Properties/AssemblyInfo.cs
+++ b/extensions/src/AWSSDK.Extensions.NETCore.Setup/Properties/AssemblyInfo.cs
@@ -16,5 +16,5 @@ using System.Runtime.InteropServices;
 [assembly: ComVisible(false)]
 
 
-[assembly: AssemblyVersion("4.0")]
+[assembly: AssemblyVersion("4.0.0.0")]
 [assembly: AssemblyFileVersion("4.0.0.0")]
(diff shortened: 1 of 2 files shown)


extensions/src/AWSSDK.Extensions.NETCore.Setup/Properties/AssemblyInfo.cs
```diff
@@ -16,5 +16,9 @@
 [assembly: ComVisible(false)]
 
 
+#if NETFRAMEWORK
```
normj, line 19: AWSSDK.Extenstions.NETCore.Setup doesn't have any .NET Framework targets so you can get rid of the `#if` and just always set the `AssemblyVersion` to the same as `AssemblyFileVersion`.

normj (changes requested):

normj (approved):

dscpinheiro (approved):

PR #3344, opened 2024-06-14
Add Observability APIs to the SDK and add telemetry provider reference to AWSConfigs and ClientConfig

muhammad-othman:
## Description

Added Observability APIs to the SDK along with NoOp implementation that will be used by default when the user doesn't provide a custom implementation.
Added `TelemetryProvider` to `AWSConfigs` and `ClientConfig`, to allow the users to set it globally or per client.

## Motivation and Context
* https://sim.amazon.com/issues/DOTNET-7517
* https://sim.amazon.com/issues/DOTNET-7540

## Testing
Compared the added Observability APIs with the SRA and OpenTelemetry APIs.

## Types of changes

## Checklist

## License


sdk/src/Core/Amazon.Runtime/Telemetry/Tracing/TracerProvider.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// </summary>
+    public abstract class TracerProvider
+    {
+        /// <summary>
+        /// Creates a new <see cref="Tracer"/> instance for a given scope.
+        /// Tracers are used as an entry point for creating spans.
+        /// </summary>
+        /// <param name="scope">The name of the instrumentation scope that uniquely identifies the tracer.</param>
+        /// <returns>A <see cref="Tracer"/> instance for the specified scope.</returns>
+        public abstract Tracer GetTracer(string scope);
```
muhammad-othman, line 30: The SRA contained an optional attributes for `getTracer`, however `OTel.GetTracer` doesn't accept attributes as a parameter.
I checked c++ implementation and they take this parameter and ignore it when for the OTel wrapper 
image
And Kotlin only allows `scope` for this API 
image

I chose to only allow `scope` rather than allowing an additional attributes parameter that we will be ignored by OTel implementation.

sdk/src/Core/Amazon.Runtime/Telemetry/DefaultTelemetryProvider.cs
```diff
@@ -0,0 +1,71 @@
...
+    public class DefaultTelemetryProvider : TelemetryProvider
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DefaultTelemetryProvider"/> class with no-op providers.
+        /// This setup prevents any telemetry actions from being performed unless explicitly registered using the
+        /// registration methods.
+        /// </summary>
+        public DefaultTelemetryProvider()
+        {
+            TracerProvider = new NoOpTracerProvider();
```
muhammad-othman, line 33: After we add `System.Diagnostics`, Will add a condition here to use `NoOpTracerProvider` only for .NET framework 3.5 and 4.5, and use the `System.Diagnostics` implementation for .net standard and above

sdk/src/Core/Amazon.Runtime/Telemetry/Attributes.cs
```diff
@@ -0,0 +1,84 @@
...
+            return null;
+        }
+
+        /// <summary>
+        /// Retrieves all the key-value pairs in the order they were added.
+        /// </summary>
+        /// <returns>An IEnumerable of key-value pairs.</returns>
+        public IEnumerable<KeyValuePair<string, object>> GetAll()
+        {
+            return _attributes.AsReadOnly();
```
boblodgett, line 81: Why is `AsReadOnly` needed here? IEnumerable already prevents you from modifying the underlying list.

sdk/src/Core/Amazon.Runtime/Telemetry/Attributes.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// </summary>
+        /// <param name="key">The key of the attribute to retrieve.</param>
+        /// <returns>The value associated with the key, or null if the key does not exist.</returns>
+        public object Get(string key)
+        {
+            // Check if the key already exists
+            var index = _attributes.FindIndex(kvp => kvp.Key == key);
+            if (index != -1)
+            {
+                // Replace the existing key-value pair
```
boblodgett, line 68: This isn't replacing the KVP. Comment is incorrect.

sdk/src/Core/Amazon.Runtime/Telemetry/Attributes.cs
```diff
@@ -0,0 +1,84 @@
...
+    {
+        private readonly List<KeyValuePair<string, object>> _attributes;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Attributes"/> class.
+        /// </summary>
+        /// <param name="attributes">An initial collection of key-value pairs to populate the attributes.</param>
+        public Attributes(IEnumerable<KeyValuePair<string, object>> attributes)
+        {
+            _attributes = new List<KeyValuePair<string, object>>(attributes);
```
boblodgett, line 33: This allows for duplicate keys:

```
var attributes = new Attributes(new List<KeyValuePair<string, object>>
{
    new KeyValuePair<string, object>("name", "John"),
    new KeyValuePair<string, object>("name", "Fred"),
    new KeyValuePair<string, object>("age", 30),
    new KeyValuePair<string, object>("isStudent", true)
});
```

Which in turn will cause a problem with Set/Get only using the first one found.

sdk/src/Core/Amazon.Runtime/Telemetry/Attributes.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// <param name="key">The key of the attribute to retrieve.</param>
+        /// <returns>The value associated with the key, or null if the key does not exist.</returns>
+        public object Get(string key)
+        {
+            // Check if the key already exists
+            var index = _attributes.FindIndex(kvp => kvp.Key == key);
+            if (index != -1)
+            {
+                // Replace the existing key-value pair
+                return _attributes[index];
```
boblodgett, line 69: This should be `return _attributes[index].Value` per the method comments. There is also a problem with this method in general because the Value could be `null` which would indicate that the key wasn't found. Should `null` be allowed as a value?

sdk/src/Core/Amazon.Runtime/Telemetry/Attributes.cs
```diff
@@ -0,0 +1,84 @@
...
+using System.Collections.Generic;
+
+namespace Amazon.Runtime.Telemetry
+{
+    /// <summary>
+    /// Represents a collection of key-value pairs for attributes.
+    /// </summary>
+    public class Attributes
+    {
+        private readonly List<KeyValuePair<string, object>> _attributes;
```
boblodgett, line 25: Why not just use a Dictionary or OrderedDictionary? It would solve some of the problems in this class and give you o(1) access.

sdk/src/Core/Amazon.Runtime/Telemetry/Attributes.cs
```diff
@@ -0,0 +1,84 @@
...
+        {
+            _attributes = new List<KeyValuePair<string, object>>(attributes);
+        }
+
+        /// <summary>
+        /// Sets a key-value pair in the attributes collection.
+        /// </summary>
+        /// <param name="key">The key of the attribute.</param>
+        /// <param name="value">The value of the attribute.</param>
+        public void Set(string key, object value)
```
boblodgett, line 41: Should this allow setting null?

sdk/src/Core/Amazon.Runtime/Telemetry/Metrics/AsyncMeasurement.cs
```diff
@@ -0,0 +1,54 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+namespace Amazon.Runtime.Telemetry.Metrics
+{
+    /// <summary>
+    /// Represents an async measurement.
+    /// An async measurement records values, such as periodically or based on events.
+    /// The generic type parameter T specifies the type of value being recorded, such as int, long, etc.
```
boblodgett, line 21: Did you intend these lines to show up with newlines in intelliSense? They won't without <para>. This same comment applies to other /// comments in this review

boblodgett (changes requested):

sdk/src/Core/Amazon.Runtime/Telemetry/Tracing/SpanKind.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// </summary>
+        INTERNAL,
+
+        /// <summary>
+        /// Represents a request to a remote service.
+        /// </summary>
+        CLIENT,
+
+        /// <summary>
+        /// Represents a request received from a remote client.
```
boblodgett, line 34: When it is `CLIENT` it comes from the service but when it is `SERVER` it comes from the remote client? Are the comments correct?

sdk/src/Core/Amazon.Runtime/Telemetry/Attributes.cs
muhammad-othman: Used regular Dictionary since the `OrderedDictionary` doesn't match the regular dictionary performance for insertion and deletion, also the order restriction wasn't mentioned in the OTel specifications.

sdk/src/Core/Amazon.Runtime/Telemetry/Attributes.cs
muhammad-othman: After updating the internal attributes to be a dictionary, the duplicates will replace the previous version, similar to the OTel tags behavior

sdk/src/Core/Amazon.Runtime/Telemetry/Attributes.cs
muhammad-othman: Update it to remove the existing item if the value is null.

sdk/src/Core/Amazon.Runtime/Telemetry/Attributes.cs
muhammad-othman: Fixed

sdk/src/Core/Amazon.Runtime/Telemetry/Attributes.cs
muhammad-othman: Fixed after using dictionary.

sdk/src/Core/Amazon.Runtime/Telemetry/Attributes.cs
muhammad-othman: Removed `AsReadOnly` and used Getter instead.

sdk/src/Core/Amazon.Runtime/Telemetry/Metrics/AsyncMeasurement.cs
boblodgett: I think it looks fine/better without the `<para>` tags. I was only mentioning if you were trying to format your comment which it looks like you were that it wouldn't show up that way in intelliSense.

sdk/src/Core/Amazon.Runtime/Telemetry/Attributes.cs
```diff
@@ -0,0 +1,86 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+using System.Collections.Generic;
+
+namespace Amazon.Runtime.Telemetry
+{
+    /// <summary>
+    /// Represents a collection of key-value pairs for attributes.
```
normj, line 21: Explain what attributes are and what they are used for.

sdk/src/Core/Amazon.Runtime/Telemetry/DefaultTelemetryProvider.cs
normj: What do you mean add `System.Diagnostics`? That is part of the base class library in .NET Standard/Core and nothing needs to be added.

sdk/src/Core/Amazon.Runtime/Telemetry/Metrics/AsyncMeasurementHandle.cs
```diff
@@ -0,0 +1,48 @@
...
+    /// Represents a handle for async measurements.
+    /// The handle is used to manage and stop the async recording of measurements.
+    /// </summary>
+    /// <typeparam name="T">The type of value being measured.</typeparam>
+    public abstract class AsyncMeasurementHandle<T> where T : struct
+    {
+        /// <summary>
+        /// Stops the async measurement.
+        /// </summary>
+        public abstract void Stop();
```
normj, line 28: Instead of having to call `Stop` methods should we use the pattern of making the type implement `IDisposable` and then we can rely on the dispose pattern?

sdk/src/Core/Amazon.Runtime/Telemetry/Tracing/TraceSpan.cs
```diff
@@ -0,0 +1,122 @@
...
+        {
+        }
+    }
+
+    /// <summary>
+    /// No-op implementation of the <see cref="TraceSpan"/> class.
+    /// This implementation does nothing when methods are called, and is used as a placeholder or default
+    /// implementation when tracing is not enabled or required.
+    /// </summary>
+    public class NoOpTraceSpan : TraceSpan
```
normj, line 79: Can we put all of the NoOp types in their own files and in their own separate namespace `Amazon.Runtime.Telemetry.Tracing.NoOp`. Also should they be internal?

normj (changes requested):
I would suggest looking into adding the System.Diagnostics implementation to prove out the abstract API will work for our needs. If you want to do that in a different PR that is okay accepting this might need changes when we start trying to do things for real.

sdk/src/Core/Amazon.Runtime/Telemetry/Attributes.cs
muhammad-othman: The spec wasn't very detailed and didn't contain any directions on the Attributes API other than set and get in addition to ` may contain additional methods`
I was trying to match how it is implemented in the ActivityTagsCollection https://github.com/dotnet/runtime/blob/5535e31a712343a63f5d7d796cd874e563e5ac14/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/ActivityTagsCollection.cs#L57, but we don't have to stick to what they have if we think it is a code smell, will update it.

sdk/src/Core/Amazon.Runtime/Telemetry/Attributes.cs
```diff
@@ -0,0 +1,86 @@
...
+        /// </summary>
+        public Attributes()
+        {
+            _attributes = new Dictionary<string, object>();
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Attributes"/> class.
+        /// </summary>
+        /// <param name="attributes">An initial collection of key-value pairs to populate the attributes.</param>
```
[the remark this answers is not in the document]
muhammad-othman: Nope, no details from the specs. Will add some details to the comments.

sdk/src/Core/Amazon.Runtime/Telemetry/Attributes.cs
```diff
@@ -0,0 +1,86 @@
...
+        /// <param name="value">The value of the attribute. If null, the attribute is removed.</param>
+        public void Set(string key, object value)
+        {
+            if (value == null)
+            {
+                _attributes.Remove(key);
+            }
+            else
+            {
+                _attributes[key] = value;
```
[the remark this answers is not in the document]
muhammad-othman: Yep the keys shouldn't be null https://opentelemetry.io/docs/specs/otel/common/#:~:text=The%20attribute%20key%20MUST%20be%20a%20non%2Dnull%20and%20non%2Dempty%20string
So either we throw an exception or leave it to the dictionary to throw the exception, I think it is fine to leave it to the dictionary unless you see any downside to that.

sdk/src/Core/Amazon.Runtime/Telemetry/Tracing/TraceSpan.cs
muhammad-othman: Will move them and update their access modifier.

muhammad-othman:
> I would suggest looking into adding the System.Diagnostics implementation to prove out the abstract API will work for our needs. If you want to do that in a different PR that is okay accepting this might need changes when we start trying to do things for real.

@normj When I worked on the design I created a PoC that used `System.Diagnostics` and I did test tracing and one metric before working on this implementation and both matched the APIs that we have here.
I may find discrepancies between the APIs that we added and `System.Diagnostics` but it should be minor since the design was based on OpenTelemetry specs, and in this case I would prefer to make changes later since this PR is already kinda long, and we can safely do so as we are merging into a feature branch and we can make any needed changes.

sdk/src/Core/Amazon.Runtime/Telemetry/DefaultTelemetryProvider.cs
muhammad-othman: I mean the adapter that we will create to wrap `System.Diagnostics` instead of the NoOp, it should be something like this 

```
#if BCL
    TracerProvider = new NoOpTracerProvider();
    MeterProvider = new NoOpMeterProvider();
#else
   
    TracerProvider = new DiagnosticsTracerAdapter();
    MeterProvider = new DiagnosticsMeterAdapter();
#endif
```

sdk/src/Core/Amazon.Runtime/Telemetry/Metrics/AsyncMeasurementHandle.cs
muhammad-othman: Actually I should remove this stop method, this is one of the discrepancies between OpenTelemtery specs and the .net implementation.
From OTel specs: https://opentelemetry.io/docs/specs/otel/metrics/api/#asynchronous-instrument-api:~:text=Where%20the%20API%20supports%20registration%20of%20callback%20functions%20after%20asynchronous%20instrumentation%20creation%2C%20the%20user%20MUST%20be%20able%20to%20undo%20registration%20of%20the%20specific%20callback%20after%20its%20registration%20by%20some%20means.

And I couldn't find a way to do the unregistration, we should map `AsyncMeasurementHandle` to `ObservableInstrument` and there is no way to stop it and we have to dispose the whole meter we used (which now implements IDisposable) based on this https://github.com/open-telemetry/opentelemetry-dotnet/issues/3460#issuecomment-1190364356

Will add a reference for the `Meter` instead of the stop method.

boblodgett (approved):

normj (approved):

PR #3348, opened 2024-06-19
Switch AWSSDK.Extensions.NETCore.Setup to use project reference

normj:
## Description

Changes for getting `AWSSDK.Extensions.NETCore.Setup` ready for V4.

* Switch `AWSSDK.Extensions.NETCore.Setup` to use a project reference to `AWSSDK.Core`. The package reference was a legacy issue when the repo used to swap out different `snk` files.
* Update version of AWSSDK.Core in nuspec file to `4.0.0.0-preview`
* Add individual dependency groups in nuspec file to remove pack warnings
* Address issues in the `DefaultConfig` now that it taking a dependency on V4
* Enable treat warnings as errors. 

I tested this by building the NuGet package locally and confirming no warnings and the package contained the correct layout.


normj:
Closing. Found issues when running the unit tests. It needs more work.

PR #3352, opened 2024-06-24
Create tracing spans to cover the SDK operations

muhammad-othman:
## Description

Created new tracing spans in the following places:
* Service Operations.
* Credentials retrieval.
* Compression.
* Http Requests.

## Motivation and Context
https://sim.amazon.com/issues/DOTNET-7519

## Testing
Created a hacked OTel provider, ran same requests and checked the output in the console and visualized these traces in observability tools, here are some examples:

* Full tracing timeline
image

* PutObject attributes: 
image

* GetObject attributes: 
image

* DynamoDB.UpdateItem
image

* DynamoDB.DescribeTableRequest with exception attributes:
image
image

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/MetricsHandler.cs
+++ b/sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/MetricsHandler.cs
@@ -17,6 +17,8 @@ using Amazon.Util;
 using Amazon.Runtime.Internal.Util;
 using System.Globalization;
 using System;
+using Amazon.Runtime.Telemetry.Tracing;
+using Amazon.Runtime.Telemetry;
 
 namespace Amazon.Runtime.Internal
 {
@@ -34,15 +36,28 @@ namespace Amazon.Runtime.Internal
         public override void InvokeSync(IExecutionContext executionContext)
         {
             executionContext.RequestContext.Metrics.AddProperty(Metric.AsyncCall, false);
+
+            var operationName = AWSSDKUtils.ExtractOperationName(executionContext.RequestContext.RequestName);
+            var spanName = $"{executionContext.RequestContext.ServiceMetaData.ServiceId}.{operationName}";
+            var span = TracingUtilities.CreateSpan(executionContext.RequestContext, spanName);
+
             try
             {
                 executionContext.RequestContext.Metrics.StartEvent(Metric.ClientExecuteTime);
                 base.InvokeSync(executionContext);
+
+                span.SetAttribute(TelemetryConstants.RequestIdAttributeKey, executionContext.ResponseContext.Response.ResponseMetadata.RequestId);
+            }
+            catch(Exception ex)
+            {
+                span.SetExceptionAttributes(ex);
+                throw;
             }
             finally
             {
                 executionContext.RequestContext.Metrics.StopEvent(Metric.ClientExecuteTime);
                 this.LogMetrics(executionContext);
+                span.Dispose();
             }
         }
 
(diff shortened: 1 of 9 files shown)


sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/MetricsHandler.cs
```diff
@@ -59,16 +74,29 @@ public override void InvokeSync(IExecutionContext executionContext)
...
+                span.SetExceptionAttributes(ex);
+                throw;
             }
             finally
             {
                 executionContext.RequestContext.Metrics.StopEvent(Metric.ClientExecuteTime);
                 this.LogMetrics(executionContext);
+                span.Dispose();
             }            
         }
```
muhammad-othman, line 102: I intentionally left out adding traces for `AWS_APM_API (.NET Framework 3.5)` because we plan to remove it in v4 and including it would introduce unnecessary complexity to store the spans and retrieve them in the InvokeCallback, which will not be needed in the newer versions.
Additionally, none of our customers will benefit from it as OTel is not supported for .NET Framework versions older than 4.6.2.

sdk/src/Core/Amazon.Runtime/Telemetry/Tracing/TracingUtilities.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// <summary>
+        /// Creates a new span from the requestContext using the span details.
+        /// </summary>
+        /// <param name="requestContext">The request context containing the tracer provider and service details.</param>
+        /// <param name="spanName">The name of the span.</param>
+        /// <param name="initialAttributes">Optional initial set of attributes for the span.</param>
+        /// <param name="spanKind">Optional type of span to create.</param>
+        /// <param name="parentContext">Optional parent context for the span.</param>
+        /// <returns>A <see cref="TraceSpan"/> instance representing the created span.</returns>
+        public static TraceSpan CreateSpanHelper(
```
normj, line 35: Adding the `Helper` suffix is redundant because you are in a class of utilities. It also makes me think it returns back an object that has helper methods for creating spans. I would remove the `Helper` suffix.

sdk/src/Core/Amazon.Runtime/Telemetry/Tracing/TracingUtilities.cs
```diff
@@ -0,0 +1,83 @@
...
+            IRequestContext requestContext,
+            string spanName,
+            Attributes initialAttributes = null,
```
normj, line 38: Since this is a public method that will be used across the service clients. I would recommend putting all of the optional parameters into an options type. That way we can add new optional parameters in the future without changing the public method's signature.

sdk/src/Core/Amazon.Runtime/Telemetry/Tracing/TracingUtilities.cs
```diff
@@ -0,0 +1,83 @@
...
+                initialAttributes = new Attributes();
+
+            // Add common attributes
+            var operationName = AWSSDKUtils.ExtractOperationName(requestContext.RequestName);
+            initialAttributes.Set(TelemetryConstants.MethodAttributeKey, operationName);
+
+            initialAttributes.Set(TelemetryConstants.SystemAttributeKey, TelemetryConstants.SystemAttributeValue);
+            initialAttributes.Set(TelemetryConstants.ServiceAttributeKey, serviceId);
+
+            var tracerProvider = requestContext.ClientConfig.TelemetryProvider.TracerProvider;
```
normj, line 54: Do we block users from setting the TraceProvider to null? If not should we?

normj (changes requested):

sdk/src/Core/Amazon.Runtime/Telemetry/Tracing/TracingUtilities.cs
muhammad-othman: Actually I marked the `TracingUtilities` as internal to allow us to change it based on our usage, since the span creation can be done using the tracer and this is method is just of convenience and will be using it only in the pipeline, so no need to share it with the users.
Do I have to also mark this method as internal to make it clear? I've found both approaches in our SDK (internal class with internal members, internal class with public members) and the members will be internal in practice anyway.

sdk/src/Core/Amazon.Runtime/Telemetry/Tracing/TracingUtilities.cs
muhammad-othman: Nope we don't, it doesn't hurt to check for nullability in the registration methods.

normj (approved):

afroz429 (approved):

PR #3354, opened 2024-06-25
v4: Delete obsolete S3 Encryption client

normj:
In V4 we are removing the S3 encryption client that is embedded in the AWSSDK.S3 package. This client has been marked as obsolete and a replacement has been created in the `Amazon.Extensions.S3.Encryption` package. Migration to this package will does not require code changes other then updating using statements and adding theNuGet package reference.

Confirmed both the .NET Framework and .NET Standard solutions compile after deleting the client.


dscpinheiro:
I found a reference to `Amazon.S3.Encryption` here: https://github.com/aws/aws-sdk-net/blob/54138724bcdfc1a9758cca720c60f423fd1d614d/sdk/src/Core/RuntimeDependencies/GlobalRuntimeDependencyRegistry.cs#L75-L77

Should that also be removed? From the comment it looks like it was only used for the encryption client.

normj:
> I found a reference to `Amazon.S3.Encryption` here:
> 
> https://github.com/aws/aws-sdk-net/blob/54138724bcdfc1a9758cca720c60f423fd1d614d/sdk/src/Core/RuntimeDependencies/GlobalRuntimeDependencyRegistry.cs#L75-L77
> 
> Should that also be removed? From the comment it looks like it was only used in the legacy encryption client.

Good catch. I deleted the ICoreAmazonKMS and references since it was only used for the S3 encryption client. I confirmed after being able recompile both Framework and Standard solutions.

dscpinheiro (approved):

peterrsongg:
Is there any need to keep `IAmazonS3Encryption` as well? I see that interface is still there. I see a lot of logic that follows this sort of pattern
```
if (_s3Client is Amazon.S3.Internal.IAmazonS3Encryption)
```
is this why this was left there?

normj:
> Is there any need to keep `IAmazonS3Encryption` as well? I see that interface is still there. I see a lot of logic that follows this sort of pattern
> 
> ```
> if (_s3Client is Amazon.S3.Internal.IAmazonS3Encryption)
> ```
> 
> is this why this was left there?

Good catch but the S3 encryption clients in the separate `Amazon.Extensions.S3.Encryption` package also implement the `Amazon.S3.Internal.IAmazonS3Encryption` interface. It is how features the TransferUtility know the uploads are being encrypted and change some of its behavior.

peterrsongg (approved):

PR #3357, opened 2024-06-26
Remove the obsolete ActionIdentifiers in v4

normj:
The ActionIdentifiers have been marked as obsolete for several years. They are obsolete because there is no mechanism to keep them up to date and they are terribly out of date. They also add unnecessary bloat to Core library for a lot of constants rarely used and out of date. As part of V4 the ActionIdentifiers constants are being removed.

To make it easier to migrate to use strings the `ActionIdentifier` class has a new implicit string assignment operator so if there are any users actually using one of these obsolete constants they can just swap in the string value.


sdk/src/Services/SimpleNotificationService/Custom/AmazonSimpleNotificationServiceClient.Extensions.cs
```diff
@@ -39,7 +32,7 @@ private static void AddSQSPermission(Policy policy, string topicArn, string sqsQ
         {
             Statement statement = new Statement(Statement.StatementEffect.Allow);
 #pragma warning disable CS0612,CS0618
```
dscpinheiro, line 34: The `pragma warning disable CS0612,CS0618` can be removed here too, right?

---

Not related to this PR, but there are also a lot of places with these `pragma` for `AWSSDKUtils.CorrectedUtcNow` (which we unmarked as deprecated); it'd be nice to remove them for `v4` as well.

dscpinheiro (approved):

sdk/test/UnitTests/Custom/RegionEndpointTests.cs
```diff
@@ -6,7 +6,6 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
 using Amazon.Auth.AccessControlPolicy;
```
muhammad-othman, line 8: nit: seems we can also remove this statement `using Amazon.Auth.AccessControlPolicy;`

muhammad-othman (approved):

normj:
@dscpinheiro @muhammad-othman I took care of both of your nits and will merge.

Issue #3316, opened 2024-05-09
V4 Development: Sensible LangVersion


danielmarbach:
### Describe the feature

Came up as a discussion point in https://github.com/aws/aws-sdk-net/pull/3293#discussion_r1593689240

### Use Case

At least make it possible to define static lambdas to avoid closure capturing.

### Proposed Solution

Target LangVersion or higher

Azure SDK targets now LangVersion 11 https://github.com/Azure/azure-sdk-for-net/pull/33890 even though they support net standard 2.0

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS .NET SDK and/or Package version used

v4

### Targeted .NET Platform

All

### Operating System and version

All

dscpinheiro:
@danielmarbach I'm closing this since we're now using C# 11 for .NET 8 and C# 9 for the other TFMs (`net472`, `netstandard2.0`, and `netcoreapp3.1`).

I think that should be enough to use newer features, but please let us know if you have any other concerns.

danielmarbach:
Is it worth the complexity of two lang versions? Why not have 11 across the board? You can shim/poly fill a lot (I already brought this up elsewhere) and for the very few things that requires runtime support you could add an analyzer or some other form of early capturing mechanism. Probably even a PR review checklist would do it.

dscpinheiro:
I can bring this up with Norm when he gets back from vacation. Like you said, I believe the concern was getting C# 11 to work with the .NET Framework (I don't think he had a chance to look at / make a decision on whether we're going to use Polyfill / PolySharp).

PR #3359, opened 2024-06-27
Optimize GetExtension execution time on NET8

stevenaw:
## Description
Newer versions of .NET have improved the performance of the IndexOf() methods such that it's often quicker for many cases to do two `LastIndexOf[Any]()` calls rather than one manual loop. This PR uses this to provide a 4x speedup (on my hardware) over the "worst-case" time of the old algorithm where no delimeters can be found, such as when processing a GUID in string format. The "best case" for the original implementation where a value ends in a short 3-letter extension remains comparable, +/- 0.5ns.

Despite the APIs being available as far back as netcore2.1, I have chosen to target this hotpath against v4 and NET8+ as the BCL speed-ups to make this worthwhile are only available on later runtimes.

## Motivation and Context

This function appears used by S3 client. Optimizing handling of non-file names such as GUIDS as keys will help when used against S3 buckets using GUIDs or similar keys.

## Testing
The changes are essentially .NET8-only though I added equivalent unit tests for both the NetStandard and NetFramework projects to demonstrate matching functionality.

image

## Benchmarks

<details>
  
  ```csharp
[MemoryDiagnoser]
public class GetExtension
{
    [Params(".htaccess", "amazon-s3.pdf", "relative/path-no-file-sep", "086219F2-E4AC-4ECA-8962-8C2879F9B6B2")]
    public string Scenario { get; set; }

    [Benchmark(Baseline = true)]
    public string Original() => OriginalImpl.GetExtension(Scenario);

    [Benchmark]
    public string Optimized() => OptimizedImpl.GetExtension(Scenario);

    [Benchmark]
    public string BuiltIn() => Path.GetExtension(Scenario);

    public static class OptimizedImpl
    {
        public static string GetExtension(string path)
        {
            if (path is null)
                return null;

            int extensionIndex = path.AsSpan().LastIndexOf('.');
            if (extensionIndex == -1)
            {
                return string.Empty;
            }

            int directoryIndex = path.AsSpan().LastIndexOfAny('/', '\\', ':');

            // extension separator is found and exists before path separator or path separator doesn't exist
            // AND it's not the last one in the string
            if (directoryIndex < extensionIndex && extensionIndex < path.Length - 1)
            {
                return path.Substring(extensionIndex);
            }

            return string.Empty;
        }
    }

    private static class OriginalImpl
    {
        public static string GetExtension(string path)
        {
            if (path == null)
                return null;
            int length = path.Length;
            int index = length;

            while (--index >= 0)
            {
                char ch = path[index];
                if (ch == '.')
                {
                    if (index != length - 1)
                        return path.Substring(index, length - index);
                    else
                        return string.Empty;
                }
                else if (IsPathSeparator(ch))
                    break;
            }
            return string.Empty;
        }

        // Checks if the character is one \ / :
        private static bool IsPathSeparator(char ch)
        {
            return (ch == '\\' ||
                    ch == '/' ||
                    ch == ':');
        }
    }
}
  ```
</details>

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
+++ b/sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
@@ -195,33 +195,29 @@ namespace Amazon.Util
         /// <returns></returns>
         public static string GetExtension(string path)
         {
-            if (path == null)
+            if (path is null)
                 return null;
-            int length = path.Length;
-            int index = length;
 
-            while (--index >= 0)
+#if NETFRAMEWORK
+            int sepIdx = path.LastIndexOf('.');
+            int dirIdx = path.LastIndexOfAny(PathSeps);
+#else
+            int sepIdx = MemoryExtensions.LastIndexOf(path.AsSpan(), '.');
+            int dirIdx = MemoryExtensions.LastIndexOfAny(path.AsSpan(), '/', '\\', ':');
+#endif
+
+            // extension separator is found and exists before path separator or path separator doesn't exist
+            // AND it's not the last one in the string
+            if (dirIdx < sepIdx && sepIdx < path.Length - 1)
             {
-                char ch = path[index];
-                if (ch == '.')
-                {
-                    if (index != length - 1)
-                        return path.Substring(index, length - index);
-                    else
-                        return string.Empty;
-                }
-                else if (IsPathSeparator(ch))
-                    break;
+                return path.Substring(sepIdx);
             }
-            return string.Empty;
-        }
 
-        // Checks if the character is one \ / :
-        private static bool IsPathSeparator(char ch)
-        {
-            return (ch == '\\' ||
-                    ch == '/' ||
-                    ch == ':');
+            // If we're here we either:
+            // 1. Can't find the extension separator
+            // 2. Found but it's at the end of the string
+            // 3. Found but there is a directory separator after
+            return string.Empty;
         }
 
         /*
--- a/sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
+++ b/sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
@@ -198,26 +198,45 @@ namespace Amazon.Util
             if (path is null)
                 return null;
 
-#if NETFRAMEWORK
-            int sepIdx = path.LastIndexOf('.');
-            int dirIdx = path.LastIndexOfAny(PathSeps);
-#else
-            int sepIdx = MemoryExtensions.LastIndexOf(path.AsSpan(), '.');
-            int dirIdx = MemoryExtensions.LastIndexOfAny(path.AsSpan(), '/', '\\', ':');
-#endif
+#if NET8_0_OR_GREATER
+            int extensionIndex = path.AsSpan().LastIndexOf('.');
+            int directoryIndex = path.AsSpan().LastIndexOfAny('/', '\\', ':');
 
             // extension separator is found and exists before path separator or path separator doesn't exist
             // AND it's not the last one in the string
-            if (dirIdx < sepIdx && sepIdx < path.Length - 1)
+            if (directoryIndex < extensionIndex && extensionIndex < path.Length - 1)
             {
-                return path.Substring(sepIdx);
+                return path.Substring(extensionIndex);
             }
 
-            // If we're here we either:
-            // 1. Can't find the extension separator
-            // 2. Found but it's at the end of the string
-            // 3. Found but there is a directory separator after
             return string.Empty;
+#else
+            int length = path.Length;
+            int index = length;
+
+            while (--index >= 0)
+            {
+                char ch = path[index];
+                if (ch == '.')
+                {
+                    if (index != length - 1)
+                        return path.Substring(index, length - index);
+                    else
+                        return string.Empty;
+                }
+                else if (IsPathSeparator(ch))
+                    break;
+            }
+
+            return string.Empty;
+
+            bool IsPathSeparator(char ch)
+            {
+                return (ch == '\\' ||
+                        ch == '/' ||
+                        ch == ':');
+            }
+#endif
         }
 
         /*
--- a/sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
+++ b/sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
@@ -200,21 +200,22 @@ namespace Amazon.Util
 
 #if NET8_0_OR_GREATER
             // LastIndexOf and LastIndexOfAny is vectorized on .NET8+ and is
-            // signifigantly faster for cases where 'path' does not end with a short file
+            // significantly faster for cases where 'path' does not end with a short file
             // extension, such as GUIDs
-            int extensionIndex = path.AsSpan().LastIndexOf('.');
+            ReadOnlySpan<char> pathSpan = path.AsSpan();
+            int extensionIndex = pathSpan.LastIndexOf('.');
             if (extensionIndex == -1)
             {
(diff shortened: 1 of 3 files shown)


sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
```diff
@@ -199,8 +199,22 @@ private static string DetermineValidPathCharacters()
         /// <returns></returns>
         public static string GetExtension(string path)
         {
-            if (path == null)
+            if (path is null)
```
danielmarbach, line 202: Why not simply replace this with https://learn.microsoft.com/en-us/dotnet/api/system.io.path.getextension?view=net-8.0?

It seems it covers all the edge cases already and also has the directory and alt directory separator char. That version is already heavily optimized, and then the complexity would go away of having to manage TFM specific code.

stevenaw: @danielmarbach  I actually gave that a try :)

Benchmarking yielded numbers similar to the existing AWSSDK code, which as it turns out use [very similar algorithms](https://github.com/dotnet/runtime/blob/5535e31a712343a63f5d7d796cd874e563e5ac14/src/libraries/System.Private.CoreLib/src/System/IO/Path.cs#L197C13-L212C45). It has indeed been heavily optimized for the filesystem "happy path" where a short file extension exists.

There is one notable difference though in that the current dotnet runtime is _more permissive_ than the current AWSSDK implementation. The AWSSDK will currently return an empty string if any of `/`, `\` or `:` are present, regardless of underlying runtime. The current built-in Windows dotnet runtime, however, will [no longer detect volume separator](https://github.com/dotnet/runtime/blob/main/src/libraries/Common/src/System/IO/PathInternal.Windows.cs#L280-L283) while the Unix runtime will [not detect a volume separator or a backslash](https://github.com/dotnet/runtime/blob/main/src/libraries/Common/src/System/IO/PathInternal.Unix.cs#L27-L33) - only forward slash. I believe the the opposite may've been true for the full framework side, where the built-in one was _too restrictive_ in that it may throw for various invalid filesystem characters like `*`, `|` etc.

@normj Reaching out as I see you've been requested for review here. This does introduce added complexity and I recognize that optimizing for GUID (or similar) keys may be a less-common use case than I may imagine. I'm happy to go in any direction you like here. Note that if optimizing for GUIDs were of interest then this could be taken a bit further by adding this after the first `LastIndexOf()` call effectively bring the GUID case down to around 2ns on my machine at the expense of another 0.5ns for the cases where a file extension _does_ exist.

```diff
int extensionIndex = path.AsSpan().LastIndexOf('.');
+if (extensionIndex == -1) {
+  return string.Empty;
+}

int directoryIndex = path.AsSpan().LastIndexOfAny('/', '\\', ':');
```

normj: Are you saying using the extension method that @danielmarbach shows no significant performance boost over what is shipped in the SDK today?

The code change you have for .NET 8+ is pretty simple and straight forward so I'm good with taking the change assuming we are saying this approach is still faster then the extension method coming from the framework. I would ask that you add a comment in the `#if` explaining how newer versions of .NET have an optimized version of `LastIndexOf`. Basically give a hint for feature reads of the code wondering why we have 2 separate implementations.

stevenaw: Apologies, my last comment was a little verbose as I'd been rushing to send it before the bottom of the hour here.

> Are you saying using the extension method that @danielmarbach shows no significant performance boost over what is shipped in the SDK today?

@normj 
This is correct. I've re-benchmarked to include the built-in .NET version as well as including the "early return" in this version. I've updated the original PR description to include the benchmarking code and latest results. I'll push the changes shortly, after resolving some conflicts.

sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
```diff
@@ -200,8 +200,30 @@ private static string DetermineValidPathCharacters()
...
                 return null;
+
+#if NET8_0_OR_GREATER
+            // LastIndexOf and LastIndexOfAny is vectorized on .NET8+ and is
+            // signifigantly faster for cases where 'path' does not end with a short file
+            // extension, such as GUIDs
+            int extensionIndex = path.AsSpan().LastIndexOf('.');
```
danielmarbach, line 210: Would it be cleaner to assign the span to a local and then use it through the method?

sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
```diff
@@ -200,8 +200,30 @@ private static string DetermineValidPathCharacters()
...
+                return string.Empty;
+            }
+
+            int directoryIndex = path.AsSpan().LastIndexOfAny('/', '\\', ':');
+
+            // extension separator is found and exists before path separator or path separator doesn't exist
+            // AND it's not the last one in the string
+            if (directoryIndex < extensionIndex && extensionIndex < path.Length - 1)
+            {
+                return path.Substring(extensionIndex);
```
danielmarbach, line 222: If we had the span assigned to a local like suggested above we could slice the span here directly

sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
```diff
@@ -218,15 +240,16 @@ public static string GetExtension(string path)
...
             return string.Empty;
-        }
 
-        // Checks if the character is one \ / :
-        private static bool IsPathSeparator(char ch)
-        {
-            return (ch == '\\' ||
-                    ch == '/' ||
-                    ch == ':');
+            bool IsPathSeparator(char ch)
```
danielmarbach, line 246: Should this still be static?

danielmarbach:
@stevenaw I haven't looked in details at the PR but I wonder why the conditional compile is required. I was able to make similar optimizations in https://github.com/aws/aws-sdk-net/pull/3365 without conditionals

stevenaw:
Thanks for the feedback @normj 
Thanks as well for the advice @danielmarbach re: keeping the code simple when contending with multiple TFMs. I'd forgotten that suggestion of yours from my that other PR 🙂

I've left recently on vacation but can take a look at all these when I return middle next week.

sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
stevenaw: Good catch. Unfortunately, at the time I had looked at it this was limited by the `<LangVersion>` on .NET Framework (https://github.com/aws/aws-sdk-net/issues/3316). Static local functions required version 9.

stevenaw:
Thanks for your patience on this one. I think I've incorporated or replied to all the feedback here and it is ready for another review

stevenaw:
@danielmarbach 

> I haven't looked in details at the PR but I wonder why the conditional compile is required. MemoryExtensions etc. should be available. You might not get the same speed as on NET8 but that doesn't really matter especially because you could then also have just one code path and one test case for both TFMs

This is a good point. The main drawback would be how long it takes when processing a file path with an extension (ex: `amazon.pdf`). Testing on my machine shows the `LastIndexOf()` approach will be ~3 times slower than the current implementation on .NET Framework, going from 8ns -> 25ns. However the change would mean simpler code.

 @normj I'm happy to push another commit with this change if you'd prefer it

normj (approved):
I'm fine leaving the old targets using the same code path not giving users any surprises and focus on the performance improvements being in .NET 8+.

muhammad-othman (approved):

Issue #3360, opened 2024-06-27
There is no way to get bedrock streaming api responses using async io.


lucasmeijer:
### Describe the bug

```c# 
       var response = await _client.InvokeModelWithResponseStreamAsync(request);
        
        //there are two ways to get the streaming events.  one is enumeration.
        foreach (var e in response.Body)
        {
        }
        
        //the other is setting up events + StartProcessing():
        response.Body.ChunkReceived += (sender, args) => { /* handle event*/};
        response.Body.StartProcessing();
        //(in this 2nd way there does not seem to be a natural way to know when it's finished?
        
        //both techniques end up in this blocking NetworkStream.Read() call:
        //https://github.com/aws/aws-sdk-net/blob/976e5b161fc1d41b1b8e57e47597c834ba3c40a4/sdk/src/Core/Amazon.Runtime/EventStreams/Internal/EventStream.cs#L310
```

This is unfortunate.  The threads that are supposed to handle new incoming requests for my server, are now blocked on very long running blocking network calls. (AI model responses are easily 30s, (spread out over many events)).

### Expected Behavior

I would expect a way that allows me to use async enumeration:

```c#
await foreach(var event in response.EnumerateEventsAsync())
{
    //handle event.
}
```

EnumerateEventsAsync would under the hood not use blocking io, leaving my servers
threads available to do server work,  and come back to this network request only when
data is there.

### Current Behavior

server threads being blocked.

### Reproduction Steps

see code snippet in description

### AWS .NET SDK and/or Package version used

<PackageReference Include="AWSSDK.BedrockRuntime" Version="3.7.306" />

### Targeted .NET Platform

net8

### Operating System and version

all

ashishdhingra:
Needs review with the team.

normj:
The normal request / response APIs are using async don't the http client but I agree after revisiting the for response streaming that the `StartProcessing` method does eventually go down to a loop calling `Stream.Read` on the stream return back for the response and that will block till there is data coming back from Bedrock.

At a minimum we should rework the [ProcessLoop](https://github.com/aws/aws-sdk-net/blob/db5608cb0eb56e0cba08dea180ea53577e71636e/sdk/src/Core/Amazon.Runtime/EventStreams/Internal/EventStream.cs#L278) to use async / await.

normj:
I created a PR to add a new `StartProcessingAsync` method that will use async methods down to the network stream read. https://github.com/aws/aws-sdk-net/pull/3364

[...]

PR #3364, opened 2024-06-29
Add async version of the StartProcessingAsync for event streams

normj:
## Description
https://github.com/aws/aws-sdk-net/issues/3360

When users use response streaming for APIs like Bedrock the `StartProcessing` method eventually goes down to a blocking read on the network stream. This uses up a thread for the process which can cause scaling concerns.

This PR creates an async version of the `StartProcessing` method all the way down to the network read. This new method has the benefit that callers can await the `Task` returned from `StartProcessingAsync` which pausing execution in the thread that called `StartProcessingAsync` till all of the events are processed.

```csharp
var invokeResponse = await bedrockClient.InvokeModelWithResponseStreamAsync(invokeRequest);

invokeResponse.Body.ChunkReceived += Body_ChunkReceived;
await invokeResponse.Body.StartProcessingAsync();

Console.WriteLine("Done");

async void Body_ChunkReceived(object? sender, Amazon.Runtime.EventStreams.EventStreamEventReceivedArgs<PayloadPart> e)
{
    Console.WriteLine(UTF8Encoding.UTF8.GetString(e.EventStreamEvent.Bytes.ToArray()));
}
```

## Testing
Added new integ tests and wrote test app to confirm behavior.

Internal dry run was successful.


danielmarbach:
@normj something is off with this entire API. StartProcessing like the name suggests starts back processing which the when data is available invokes the corresponding event handlers. So the method is designed to be fire and forgot so that the caller can continue and effectively it is the responsibility of the caller to have synchronization primitives in place to know when the event handlers are called. This PR seems to deviate from the design to solve the thread blocking problem but introduces weird assymetry on the API surface. Technical rhe existing start method can use the async path already which would remove the thread offloading and unlock the threads during the IO bound operation without having to introduce a new API

If you look at what the user requests that would also address the threading part. Then the next problem is that the higher level event APIs are not async friendly which an async enumerable would solve

normj:
> @normj something is off with this entire API. StartProcessing like the name suggests starts back processing which the when data is available invokes the corresponding event handlers. So the method is designed to be fire and forgot so that the caller can continue and effectively it is the responsibility of the caller to have synchronization primitives in place to know when the event handlers are called. This PR seems to deviate from the design to solve the thread blocking problem but introduces weird assymetry on the API surface. Technical rhe existing start method can use the async path already which would remove the thread offloading and unlock the threads during the IO bound operation without having to introduce a new API
> 
> If you look at what the user requests that would also address the threading part. Then the next problem is that the higher level event APIs are not async friendly which an async enumerable would solve

I also updated the original `StartProcess` method to use the new async code path down to the network stream getting rid of the blocking stream. Except for .NET Framework 3.5. So user's request would be satisfied by updating to the SDK.

```csharp
        public virtual void StartProcessing()
        {
            if (IsProcessing) return;

            IsProcessing = true;
            Process();
        }

        protected void Process()
        {
#if AWS_ASYNC_API
            // Task only exists in framework 4.5 and up, and Standard.
            _ = ProcessLoopAsync();
#else
            // ThreadPool only exists in 3.5 and below. These implementations do not have the Task library.
            ThreadPool.QueueUserWorkItem(ProcessLoop);
#endif
        }
```

I thought it would be useful to add the `StartProcessingAsync` returning the `Task` because like you said "responsibility of the caller to have synchronization primitives in place to know when the event handlers are called." but the SDK doesn't give any primitives back to the caller to use. And there is no telling how many times the callbacks will be called. So I thought if I gave back the `Task` the user could choose when to end for example the HTTP response from a server while using the callbacks to send data through the response while data is coming back from Bedrock. If they don't care they can always ignore the `Task` and use it for fire and forget.

In the PR I created an async version of the existing S3 select integ test. You can see in that test when using the feature it was written waiting till it got the known number of callbacks or a timeout. That works if you know the number of callbacks you should get but likely you don't know especially in Bedrocks case.

```csharp
[TestMethod]
[TestCategory("S3")]
public void TestCallEvents()
{
    var eventStream = GetSelectObjectContentEventStream();

    var testContent = "";
    var recordsEvents = new List<RecordsEvent>();

    eventStream.RecordsEventReceived += (sender, args) => recordsEvents.Add(args.EventStreamEvent);
    eventStream.StartProcessing();
    SpinWait.SpinUntil(() => recordsEvents.Count > 0, TimeSpan.FromSeconds(5));

    using (var streamReader = new StreamReader(recordsEvents[0].Payload))
    {
        testContent = streamReader.ReadToEnd();
    }
    eventStream.Dispose();

    AssertRecordsEqualsExpected(testContent);
}
```

The event stream API is not my favorite part of the SDK's user experience. The original request asked for the use of `IAsyncEnumerable` user experience which I agree would be better for most use cases. But that is a larger effort that I have bandwidth right now and this seemed like a reasonable way to unblock the original request. The event stream API predates `IAsyncEnumerable` existing which is why it wasn't done that way from the start.

sdk/src/Core/Amazon.Runtime/EventStreams/Internal/EventStream.cs
```diff
@@ -253,22 +262,39 @@ protected void Process()
         {
 #if AWS_ASYNC_API
             // Task only exists in framework 4.5 and up, and Standard.
-            Task.Run(() => ProcessLoop());
+            _ = ProcessLoopAsync();
```
danielmarbach, line 265: You still want to wrap this async version with a Task.Run. Otherwise, the caller is still blocked until the first actual IO bound operation happens, which is deep down in the network stream read.

danielmarbach:
> The event stream API is not my favorite part of the SDK's user experience. The original request asked for the use of `IAsyncEnumerable` user experience which I agree would be better for most use cases. But that is a larger effort that I have bandwidth right now and this seemed like a reasonable way to unblock the original request.

I think my comment came primarily from the fact that both methods are called `StartProcessing`, but they behave differently. One starts the processing by starting to call the event handler on a background thread and leaves it up to the caller to do the necessary synchronization while the other starts the processing but then holds up the calling thread until the processing is done. This is confusing to me from an API surface perspective.

I would argue either you have `StartProcessing` starting to process and raising the event and then a `StopProcessing` and `StopProcessingAsync` that allows the caller to either synchronously wait or asynchronously "wait" until the processing is done, or you would have to rename the new async method to be something like `StartProcessingAndWaitForCompletionAsync`.

sdk/src/Core/Amazon.Runtime/EventStreams/Internal/EventStream.cs
```diff
@@ -253,22 +262,39 @@ protected void Process()
...
-        /// The background thread main loop. It will constantly read from the network stream until IsProcessing is false, or an error occurs.
-        /// <para></para>
-        /// This stub exists due to FXCop.
-        /// </summary>
-        private void ProcessLoop()
+#if AWS_ASYNC_API
+        private async Task ProcessLoopAsync()
         {
-            ProcessLoop(null);
+            var buffer = new byte[BufferSize];
```
peterrsongg, line 275: How do you feel about optimizing here for memory? Something like
```
#If !NETFRAMEWORK
            var arrayPool = ArrayPool<byte>.Shared;
            var buffer = arrayPool.Rent(BufferSize);
#else
             var buffer = new byte[BufferSize];
#endif
```
Then conditionally adding the using statement based off `!NETFRAMEWORK`

danielmarbach: FYI I would be very careful with pooling here. The problem is that the byte array is handed out to an event handler that then theoretically can do whatever with the byte array. While this problem already exists today with the current approach at least the byte array is local to the EventStream implementation and therefore "buffered" locally over subsequent reads until no data is available anymore. With the pool you would acquire a buffer from the pool and keep it for a long period and then return the buffer to the pool. If the event handler does something weird that buffer reference would be handed back to the pool and then eventually handed out to someone else

If you want to do pooling it might be better to pool for every read and return after the handler returned or does the decoder do some memory copying under the covers?

peterrsongg: that's a good point, the decoder does do some memory copying under the covers, but the added complexity of pooling probably isn't worth the benefit here.

peterrsongg (approved):

danielmarbach:
> > The event stream API is not my favorite part of the SDK's user experience. The original request asked for the use of `IAsyncEnumerable` user experience which I agree would be better for most use cases. But that is a larger effort that I have bandwidth right now and this seemed like a reasonable way to unblock the original request.
> 
> I think my comment came primarily from the fact that both methods are called `StartProcessing`, but they behave differently. One starts the processing by starting to call the event handler on a background thread and leaves it up to the caller to do the necessary synchronization while the other starts the processing but then holds up the caller until the processing is done. This is confusing to me from an API surface perspective.
> 
> I would argue either you have `StartProcessing` starting to process and raising the event and then a `StopProcessing` and `StopProcessingAsync` that allows the caller to either synchronously wait or asynchronously "wait" until the processing is done, or you would have to rename the new async method to be something like `StartProcessingAndWaitForCompletionAsync`.

@peterrsongg what do you think about the comment above?

sdk/src/Core/Amazon.Runtime/EventStreams/Internal/EventStream.cs
normj: Good call on adding the `Task.Run`. Added that back in.

normj:
> starts the processing but then holds up the caller until the processing is done.

I don't view that we are forcing the caller to be held up because the caller can choose to await the `Task` or not. It gives the caller the option to use the `Task` as a synchronization context. I don't think a `StopProcessingAsync` makes sense because a caller doesn't have any knowledge how far they are in the data being sent through the stream and wouldn't have any context on when to stop. I can kind of see the argument for a renaming to `StartProcessingAndWaitAsync` but the reality the method isn't the one waiting. It is up to the caller to choose to wait by awaiting the `Task` returned.

danielmarbach:
I brought up the term Stop because that is what message / event pumps usually use to indicate as stop processing and wait until currently processed items are successfully processed but I do understand your concern there that Stop is not the right terminology. 

I'm mostly concerned about the API behavioral differences. When you look how other message or event processors work is the essentially represent a pump. When you call `Start` you start the pump. Any task returned there would represent the IO-bound operations that may occur until the pump is started.

The synchronous start method exactly works like that. The async version would behave differently. So from an API perspective you can argue, even in the synchronous case when you have a way to start the pump you also want to have a way to indicate that you would like to wait until the pump is done. So by splitting the starting from "waiting for done or all signals arrived" into dedicated API methods you would get that, and it would be symmetrical across the synchronous and asynchronous code path, which, I think, is ultimately the more consistent behavior from the SDK.

peterrsongg:
> would be symmetrical ac

So is your concern that `StartProcessingAsync()` contains both the logic for starting the pump and  waiting until all events are done processing and that you'd like them separated out? Also just so I understand correctly, are you saying that "waiting for done **or** all signals arrived" should be separate APIs? All signals arriving would signal that we are done, so this part has me a bit confused.

The original purpose of this PR was to get rid of the blocking IO network read, and adding this functionality would cause the scope of this PR to get bigger than intended, but perhaps this is something we could address in a separate PR.

peterrsongg:
> I'm mostly concerned about the API behavioral differences. When you look how other message or event processors work is the essentially represent a pump. When you call `Start` you start the pump. Any task returned there would represent the IO-bound operations that may occur until the pump is started.
> 
> The synchronous start method exactly works like that. The async version would behave differently. So from an API perspective you can argue, even in the synchronous case when you have a way to start the pump you also want to have a way to indicate that you would like to wait until the pump is done. So by splitting the starting from "waiting for done or all signals arrived" into dedicated API methods you would get that, and it would be symmetrical across the synchronous and asynchronous code path, which, I think, is ultimately the more consistent behavior from the SDK.

So is your concern that `StartProcessingAsync()` contains both the logic for starting the pump and  waiting until all events are done processing and that you'd like them separated out? Also just so I understand correctly, are you saying that "waiting for done **or** all signals arrived" should be separate APIs? All signals arriving would signal that we are done, so this part has me a bit confused.

The original purpose of this PR was to get rid of the blocking IO network read, and adding this functionality would cause the scope of this PR to get bigger than intended, but perhaps this is something we could address in a separate PR.

danielmarbach:
> The original purpose of this PR was to get rid of the blocking IO network read, and adding this functionality would cause the scope of this PR to get bigger than intended, but perhaps this is something we could address in a separate PR.

My input early on was that this could have been addressed by not adding the StartProcessingAsync method and simply within the StartProcessing method have two implementation paths. One that offloads to the worker thread pool and uses synchronous IO-bound operation to read until completed and one that does a Task.Run that calls into a path that only uses the asynchronous stream APIs like mentioned here

> This PR seems to deviate from the design to solve the thread blocking problem but introduces weird asymmetry on the API surface. Technical the existing start method can use the async path already which would remove the thread offloading and unlock the threads during the IO bound operation without having to introduce a new API

By adding the new StartProcessingAsync the scope of this PR was in my view already expanded and that's why I brought my thoughts up here. 

I realize though nobody has asked me about my opinion and from this point on I might come across like a broken record and I definitely don't want to be that guy.

normj:
@danielmarbach your opinions are welcome, don't feel you need to be asked.

What I'm wrestling with is the `StartProcessing` is pretty challenging to use correctly for API's like Bedrock's event stream. I'm thinking in v4 we should deprecate `StartProcessing` in fact and could be convinced to do it in v3. The event stream mechanism was added when S3 added their [SelectObjectContent](https://docs.aws.amazon.com/AmazonS3/latest/API/API_SelectObjectContent.html#API_SelectObjectContent_ResponseElements) operation. In that API it returns multiple events including an `EndEvent`. So callers could use the event stream API and run the operation till they got the `EndEvent`.

Later as Lambda and now Bedrock have added response streaming APIs they were modeled with no ending event. So the assumption used for `S3` that users could listen to events till there was an end event was invalidated. So I'm not sure how you would ever use `StartProcessing` without putting basically a sleep hoping it finishes with in that time. That approach has serious problems.

The reason I don't deprecate `StartProcessing` in v3 is it still has the .NET Framework support with no `Task` support. I agree that `StartProcessing` and `StartProcessingAsync` have different semantics but I would rather consider `StartProcessing` as obsolete and not be considered the driver for the behavior. Ultimately I expect most users will just need to await the `StartProcessingAsync` and asking them to know to call 2 different APIs every time seems like unnecessary work.

boblodgett:
> @danielmarbach your opinions are welcome, don't feel you need to be asked.
> 
> What I'm wrestling with is the `StartProcessing` is pretty challenging to use correctly for API's like Bedrock's event stream. I'm thinking in v4 we should deprecate `StartProcessing` in fact and could be convinced to do it in v3. The event stream mechanism was added when S3 added their [SelectObjectContent](https://docs.aws.amazon.com/AmazonS3/latest/API/API_SelectObjectContent.html#API_SelectObjectContent_ResponseElements) operation. In that API it returns multiple events including an `EndEvent`. So callers could use the event stream API and run the operation till they got the `EndEvent`.
> 
> Later as Lambda and now Bedrock have added response streaming APIs they were modeled with no ending event. So the assumption used for `S3` that users could listen to events till there was an end event was invalidated. So I'm not sure how you would ever use `StartProcessing` without putting basically a sleep hoping it finishes with in that time. That approach has serious problems.
> 
> The reason I don't deprecate `StartProcessing` in v3 is it still has the .NET Framework support with no `Task` support. I agree that `StartProcessing` and `StartProcessingAsync` have different semantics but I would rather consider `StartProcessing` as obsolete and not be considered the driver for the behavior. Ultimately I expect most users will just need to await the `StartProcessingAsync` and asking them to know to call 2 different APIs every time seems like unnecessary work.

I think this is reasonable and agree we need to make sure we obsolete `StartProcessing` as soon as possible without breaking customers. This sounds like a v4 change we need to record.

generator/.DevConfigs/E75E0238-4533-4EDC-89B5-08C4038EBE03.json
```diff
@@ -0,0 +1,10 @@
+{
+    "core": {
+        "changeLogMessages": ["Update event stream apis use non blocking async io."],
```
boblodgett, line 3: Non blocking. Change if making other changes: `apis use` => `APIs to use`

boblodgett (approved):

lucasmeijer:
Good news: it works.

Bad news: it makes the thing people are most likely wanting to do very difficult.  I have to write this code myself in order to use bedrock.  This makes bedrock the odd one out, as my openAI, anthropic, azure, and groq implementations all do not ask me to write code like this.  Quite tricky code to get right, I'm not 100% sure I got it right.


```
static async IAsyncEnumerable<PayloadPart> PayloadPartsFrom(InvokeModelWithResponseStreamResponse r)
        {
            ConcurrentQueue<PayloadPart> queue = new();

            using SemaphoreSlim semaphore = new(0);
            
            void OnBodyOnChunkReceived(object? sender, EventStreamEventReceivedArgs<PayloadPart> args)
            {
                queue.Enqueue(args.EventStreamEvent);
                semaphore.Release();
            }
            
            r.Body.ChunkReceived += OnBodyOnChunkReceived; 
            var finishTask = r.Body.StartProcessingAsync();

            while (true)
            {
                var t = await Task.WhenAny(semaphore.WaitAsync(), finishTask);
                if (t == finishTask)
                    yield break;
                while (queue.TryDequeue(out var p))
                    yield return p;
            }
        }
```
 
That said, it does work, and allows me to move forward, so quite happy with that.

PR #3370, opened 2024-07-02
Optimize logger

danielmarbach:
## Description

This PR removes the global lock on the logger and introduces currently internal properties to check on the hot path whether the logger is enabled to avoid resource intensive operations for log levels that are not enabled.

## Motivation and Context

When requests are canonicalized (is that even a word :D) the resource path runs for the canonicalization process which then acquires a logger. This acquires a global lock to then access a dictionary that quite likely already contains the logger. Furthermore it always allocates strings for debug logging that might not be required when that log level is not even requested.

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Internal/Util/Logger.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Util/Logger.cs
@@ -46,8 +46,7 @@ namespace Amazon.Runtime.Internal.Util
             {
                 loggers = new List<InternalLogger>(3);
 #pragma warning disable
-                InternalLog4netLogger log4netLogger = new InternalLog4netLogger(type);
-                loggers.Add(log4netLogger);
+                loggers.Add(new InternalLog4netLogger(type));
 #pragma warning restore
             }
             else
@@ -56,8 +55,7 @@ namespace Amazon.Runtime.Internal.Util
             }
 
             loggers.Add(new InternalConsoleLogger(type));
-            InternalSystemDiagnosticsLogger sdLogger = new InternalSystemDiagnosticsLogger(type);
-            loggers.Add(sdLogger);
+            loggers.Add(new InternalSystemDiagnosticsLogger(type));
             ConfigureLoggers();
             AWSConfigs.PropertyChanged += ConfigsChanged;
         }
--- a/sdk/src/Core/Amazon.Runtime/Internal/Util/Logger.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Util/Logger.cs
@@ -13,6 +13,7 @@
  * permissions and limitations under the License.
  */
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.Diagnostics;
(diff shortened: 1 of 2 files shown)


danielmarbach:
Some of these changes might even benefit from being cherry picked to the current released versions

sdk/src/Core/Amazon.Runtime/Internal/Util/Logger.cs
```diff
@@ -80,39 +77,76 @@ private void ConfigureLoggers()
...
+            } while (Interlocked.CompareExchange(ref cachedLoggers, newLoggerCache, oldLoggerCache) != oldLoggerCache);
+
+            // Unregister all the loggers in the old cache
+            foreach (var item in oldLoggerCache ?? Enumerable.Empty<KeyValuePair<Type, Lazy<Logger>>>())
+            {
+                var lazyLogger = item.Value;
+                if (lazyLogger.IsValueCreated)
+                {
+                    lazyLogger.Value.Unregister();
+                }
```
danielmarbach, line 129: I have realized the config change handler is never unregistered and effectively the old logger instances can still be rooted.

sdk/src/Core/Amazon.Runtime/Internal/Util/Logger.cs
```diff
@@ -80,39 +77,76 @@ private void ConfigureLoggers()
                     il.IsEnabled = (logging & LoggingOptions.Console) == LoggingOptions.Console;
                 if (il is InternalSystemDiagnosticsLogger)
                     il.IsEnabled = (logging & LoggingOptions.SystemDiagnostics) == LoggingOptions.SystemDiagnostics;
+
+                if (!IsEnabled)
```
normj, line 81: For all of these properties the last `InternalLogger` will overwrite the value on the `Logger`. If I understand the purpose the properties on the `Logger` should be true if any of the `InternalLogger` are true.

normj (changes requested):

sdk/src/Core/Amazon.Runtime/Internal/Util/Logger.cs
danielmarbach: I do suck at boolean logic and would probably fail any job interview 👯‍♂️ 

Isn't that exactly what it is doing? If the property is false it will check the property of the internal logger and set it to the corresponding value of the internal logger. If the internal logger is true it sets the property to true and subsequent checks don't need to look at the internal logger value anymore. If the property gets set again to false the next iteration needs to check the value of the next internal logger. So if all are false it remains false. If one is true it will turn to true and subsequent checks omit toggling it. I wanted to avoid 'Any()'

danielmarbach:
@dscpinheiro I think this is ready for review. See also my answer to @normj

danielmarbach:
Ping @normj @dscpinheiro

danielmarbach:
Is there anything else required here to be able to move this forward?

danielmarbach:
Ping @normj @dscpinheiro

danielmarbach:
Closing since there is no interest in this

normj:
Apologize @danielmarbach for never getting to this. I have been wanting to larger changes to logging to support other logging libraries like the Microsoft ILogger. Here is my incomplete branch if you want to have a look at the direction I'm going. https://github.com/aws/aws-sdk-net/tree/normj/v4-logging

PR #3365, opened 2024-06-30
Optimize DetermineService

danielmarbach:
## Description

Heavily optimizes DetermineService method which is also used on the Signing request path

## Motivation and Context

image

https://github.com/danielmarbach/aws-sdk-net-benchmarks/blob/main/DetermineService.cs

Proof of embedding (also works on .NET)

https://sharplab.io/#gist:c381a4555b6ad204c609be729ee4e864

## Testing

I added a theory test with some inline data to make sure the existing functionality is preserved

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project
- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [ ] I have read the **README** document
- [x] I have added tests to cover my changes
- [x] All new and existing tests passed

## License

- [x] I confirm that this pull request can be released under the Apache 2 license

[issues]: https://github.com/aws/aws-sdk-net/issues
[license]: http://aws.amazon.com/apache2.0/
[cla]: http://en.wikipedia.org/wiki/Contributor_License_Agreement

PR #3373, opened 2024-07-03
Update Resource Path canonicalization to follow RFC 3986 in accordance with protocol test.

peterrsongg:
## Description

This PR has 3 commits. 

The first commit - 33d587969a85a5792134abaa40081baf8f014694 is  the update Core and resulting tests to follow RFC 3986, on which characters to encode. We no longer need customizations for 4.5 URI class differences so that logic is removed. According to the protocol test + RFC we should encode labels no matter what, and only encode literals if it is not within the allowed character set (segment-nz-nc). Previously we were encoding regardless of label vs literal, but the protocol test clarifies this. This means that we look at the path resources dictionary to determine if something is a label or a literal and change our encoding behavior based on this.

The second commit- c6ce745471761a6e39ce425fcc2d8d77e04f3c5c is the update to the handwritten S3 marshallers to use resource path dictionaries instead of hardcoding the value of the key in the resource path. This needs to be done since we rely on literal vs label now.

The last commit - 6d9b853bada54cc0980b857fc044bf51c5a8d07c is just housekeeping. I updated the smithy-build.json to match what is currently in main, which moves all the protocol tests to a different folder, which is done automatically on a `./gradlew :smithy-dotnet-protocol-test:build`

## Types of changes

## Checklist

## License


sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
```diff
@@ -478,6 +501,25 @@ public static string JoinResourcePathSegmentsV2(IEnumerable<string> pathSegments
...
+        {
+            List<string> encodedSegments = pathSegments.Select(segment =>
+            {
+                if (segment.SegmentType == SegmentType.Label)
+                    return UrlEncode(segment.Value, false);
+                else
+                    return UrlEncode(segment.Value, true);
+            }).ToList();
+            // join the encoded segments with /
+            return string.Join(Slash, encodedSegments.ToArray());
```
danielmarbach, line 519: Join has enumerable support. Why do you need to eager materialize twice?

sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
```diff
@@ -1761,4 +1803,15 @@ public void Reset()
             _count = 0;
         }
     }
+    public class UriComponent
```
danielmarbach, line 1806: Could this be a readonly struct or a record struct?

sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
```diff
@@ -444,6 +422,51 @@ public static IEnumerable<string> SplitResourcePathIntoSegments(string resourceP
...
+                {
+                    uriComponentSegments.Add(new UriComponent { SegmentType = SegmentType.Literal, Value = segment});
+                }
+                return uriComponentSegments;
+            }
+
+            //Otherwise there are key/values that need to be resolved
+            foreach (var segment in pathSegments)
+            {
+                if (!pathResources.ContainsKey(segment))
```
danielmarbach, line 450: If you do a TryGet here you already have it available down below

sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
```diff
@@ -444,6 +422,51 @@ public static IEnumerable<string> SplitResourcePathIntoSegments(string resourceP
...
+        /// segments for composing a URL we must discern that which is a literal which we do not encode and that which is a label which
+        /// we do encode.
+        /// </summary>
+        /// <param name="resourcePath">The resource path, which includes literals and labels.</param>
+        /// <param name="pathResources">The dictionary for substituting label values.</param>
+        /// <returns></returns>
+        public static IEnumerable<UriComponent> SplitResourcePathIntoSegmentsV2(string resourcePath, IDictionary<string, string> pathResources)
+        {
+            var splitChars = new char[] { SlashChar };
+            var pathSegments = resourcePath.Split(splitChars, StringSplitOptions.None);
```
danielmarbach, line 436: With IndexOf and slicing the resource path as you go you can avoid the additional string allocations which makes this code way more efficient. This would also avoid the additional char array allocation. See https://github.com/aws/aws-sdk-net/pull/3365/files for an example

sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
```diff
@@ -444,6 +422,51 @@ public static IEnumerable<string> SplitResourcePathIntoSegments(string resourceP
...
+                if (!pathResources.ContainsKey(segment))
+                {
+                    uriComponentSegments.Add(new UriComponent { SegmentType = SegmentType.Literal, Value = segment});
+                    continue;
+                }
+
+                //Determine if the path is greedy. If greedy the segment will be split at each / into multiple segments.
+                if (segment.EndsWith("+}", StringComparison.Ordinal))
+                {
+                    uriComponentSegments.AddRange(pathResources[segment].Split(splitChars, StringSplitOptions.None).Select(x => new UriComponent { Value = x, SegmentType = SegmentType.Label}));
```
danielmarbach, line 459: Same comment as above with the splitting and allocations

sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
peterrsongg: Ah, you're right I can get rid of the second `.ToArray()`

sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
peterrsongg: I believe i could make it a readonly struct but is there any reason to?

sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
peterrsongg: I did some digging and found that using the built in Split method requires far less complexity especially for the corner cases (see below for more details). For that reason I decided to stick with the `Split` method. I did a quick benchmark with the following code (first part is a snippet)
```
            while (true)
            {
                var slashIndex = resourcePathSpan.IndexOf(SlashChar);
                if (slashIndex == -1)
                {
                    break;
                }

                ProcessSegment(resourcePathSpan.Slice(0, slashIndex).ToString(), uriComponentSegments);
                resourcePathSpan = resourcePathSpan.Slice(slashIndex + 1);
            }

            // Handle the remaining part of the resourcePathSpan
            if (!resourcePathSpan.IsEmpty)
            {
                ProcessSegment(resourcePathSpan.ToString(), uriComponentSegments);
            }

        private void ProcessSegment(string segment, List<UriComponent> uriComponentSegments)
        {
            if (!pathResources.TryGetValue(segment, out var pathSegment))
            {
                uriComponentSegments.Add(new UriComponent { SegmentType = SegmentType.Literal, Value = segment });
            }
            else if (segment.EndsWith("+}", StringComparison.Ordinal))
            {
                var resolvedSegment = pathSegment.AsSpan();
                while (true)
                {
                    var innerSlashIndex = resolvedSegment.IndexOf(SlashChar);
                    if (innerSlashIndex == -1)
                    {
                        break;
                    }
                    string temp = resolvedSegment.Slice(0, innerSlashIndex).ToString();
                    uriComponentSegments.Add(new UriComponent { Value = temp, SegmentType = SegmentType.Label });
                    resolvedSegment = resolvedSegment.Slice(innerSlashIndex + 1);
                }
            }
            else
            {
                uriComponentSegments.Add(new UriComponent { SegmentType = SegmentType.Label, Value = pathSegment });
            }
        }
```

For example, in the case above if there is a greedy path and it is the last portion of the string, then simply checking if the `slashIndex != -1` won't cut it because it's possible that the resource path doesn't end in a slash. This then requires us to either create a new version of the resourcePath which ends in a `/` which allocates more memory since we have to make a copy of the string or to check again after the loop exits to see if the string is empty. But even that may not be enough because within the greedy path logic we also check for a `/` which may not exist at the end.


The happy path, however, did result in a decrease in allocations. 

```
| Method           | Mean     | Error   | StdDev  | Median   | Gen0   | Allocated |
|----------------- |---------:|--------:|--------:|---------:|-------:|----------:|
| TestRegularSplit | 197.9 ns | 3.90 ns | 9.93 ns | 194.5 ns | 0.0503 |     632 B |
| TestSpanAndSlice | 172.8 ns | 3.42 ns | 5.42 ns | 170.4 ns | 0.0393 |     496 B |
```

boblodgett (approved):

muhammad-othman (approved):

PR #3374, opened 2024-07-04
Fixing bug: IsClockskew incorrectly returing false in multithreading …

Somanko:
### Bug
Our application uses  AWS SDK .NET KinensisClient which throws Signature Expired exception Like: 
```
Signature expired: 20240624T173605Z is now earlier than 20240624T175905Z (20240624T180405Z - 5 min.)
```

As mentioned in [this post](https://aws.amazon.com/blogs/developer/clock-skew-correction/):
```
For the most part, this process is transparent: the SDK will make a request, and if the server responds with a clock skew error, the SDK will calculate a clock offset (how much client time is different from server time) and will then retry the original request with the correct time. 
```
But the problem is - AWS SDK is not doing that.

### Details
When 2 threads with skewed clock call same kinesis endpoint together - the first one succeeds and second one fails:
1. 2 calls are being made at the same time with same region as destination
2. First call fails and Exception caught [here](https://github.com/aws/aws-sdk-net/blob/976e5b161fc1d41b1b8e57e47597c834ba3c40a4/sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryHandler.cs#L137)
3. `RetryPolicy.RetryAsync` is called from [here](https://github.com/aws/aws-sdk-net/blob/976e5b161fc1d41b1b8e57e47597c834ba3c40a4/sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryHandler.cs#L142C46-L142C68)
4. That calls `IsClockskew` [here](https://github.com/aws/aws-sdk-net/blob/976e5b161fc1d41b1b8e57e47597c834ba3c40a4/sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/_async/RetryPolicy.cs#L45)
5. That calls `CorrectClockSkew.SetClockCorrectionForEndpoint` [here](https://github.com/aws/aws-sdk-net/blob/976e5b161fc1d41b1b8e57e47597c834ba3c40a4/sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs#L457)
6. That sets the `clockCorrectionDictionary[endpoint] = correction;` [here](https://github.com/aws/aws-sdk-net/blob/976e5b161fc1d41b1b8e57e47597c834ba3c40a4/sdk/src/Core/Amazon.Runtime/CorrectClockSkew.cs#L131)
7. In the mean time the second thread also failed and the execution came to `IsClockskew`, but this time the `clockCorrectionDictionary` was already set by the thread1 and `CorrectClockSkew.GetCorrectedUtcNowForEndpoint` [here](https://github.com/aws/aws-sdk-net/blob/976e5b161fc1d41b1b8e57e47597c834ba3c40a4/sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs#L433C36-L433C82) would return the corrected time
8. So for the second thread:
a. it would not go inside [this if block](https://github.com/aws/aws-sdk-net/blob/976e5b161fc1d41b1b8e57e47597c834ba3c40a4/sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs#L450) and return `IsClockskew` = false
b. That would cause the execution to not go inside [this if block](https://github.com/aws/aws-sdk-net/blob/976e5b161fc1d41b1b8e57e47597c834ba3c40a4/sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/_async/RetryPolicy.cs#L46) and return false
c. So it calls `LogForError` [here](https://github.com/aws/aws-sdk-net/blob/976e5b161fc1d41b1b8e57e47597c834ba3c40a4/sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryHandler.cs#L145), that prints this **Error** log: `AmazonKinesisClient 64|2024-06-25T17:11:20.839Z|ERROR|AmazonKinesisException making request PutRecordsRequest to https://kinesis.us-east-1.amazonaws.com/. Attempt 1.`
Note: Even if it says `Attempt 1`, its actually attempt zero(for thread2), as it does a `+1` while logging [here](https://github.com/aws/aws-sdk-net/blob/976e5b161fc1d41b1b8e57e47597c834ba3c40a4/sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryHandler.cs#L284).
d. And it throws the exception [here](https://github.com/aws/aws-sdk-net/blob/976e5b161fc1d41b1b8e57e47597c834ba3c40a4/sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryHandler.cs#L146), that is caught by the client calling SDK
9. Whereas for the first thread, 
a. `IsClockskew`was true and 
b. `LogForRetry` was called from [here](https://github.com/aws/aws-sdk-net/blob/976e5b161fc1d41b1b8e57e47597c834ba3c40a4/sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryHandler.cs#L152), that prints this **Info** log: `AmazonKinesisClient 62|2024-06-25T17:11:20.600Z|INFO|AmazonKinesisException making request PutRecordsRequest to https://kinesis.us-east-1.amazonaws.com/. Attempting retry 1 of 4.`
Note: For the other(failure) theread `LogForError` was called and Error log was made
c. And then it retried and succeeded

Note:
1. The `clockCorrectionDictionary` is set per region - thats why it repros only if the 2 calls are made simultaneously for the same region.
2. It also explains why individual threads were succeeding
### Fix
The decision whether to make a retry or not is taken in [this if block](https://github.com/aws/aws-sdk-net/blob/976e5b161fc1d41b1b8e57e47597c834ba3c40a4/sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs#L450), that uses `correctedNow`  which was set by the first thread
Each thread should independently decide whether to retry and set the clock correction - one thread should not depend on another thread for taking this decision. So, I am using the client time(with skewed clock) that was used to call the SDK and caused the exception. That way each thread will set the clock skew and will return true for retry

## Motivation and Context
This exception is causing data loss for our customer

## Testing
Tested this change with the same app that was catching the exception thrown from SDK:
1. Configure the app in a way that it will call 2 kinesis endpoints in the same region simultaneously
2. Run the app - both calls to Kinesis is successful without any clock skew
3. Then change system time - set to 10 min in past
4. Both calls to Kinesis is successful and AWS SDK log shows that both threads are setting the clock correction independently
5. Set the system clock back to normal(synced)
6. Both calls to Kinesis is successful and AWS SDK log shows that both threads are setting the clock correction independently
7. Repeat steps 3 - 6 couple of times

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs
+++ b/sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs
@@ -412,6 +412,8 @@ namespace Amazon.Runtime
         private const string clockSkewMessageParen = "(";
         private const string clockSkewMessagePlusSeparator = " + ";
         private const string clockSkewMessageMinusSeparator = " - ";
+        private const string clockSkewMessageColonSeparator = ":";
+        private const string clockSkewMessageSpaceSeparator = " ";
         private static TimeSpan clockSkewMaxThreshold = TimeSpan.FromMinutes(5);
 
         private bool IsClockskew(IExecutionContext executionContext, Exception exception)
@@ -439,13 +441,18 @@ namespace Amazon.Runtime
 
                 // If that fails, try to parse it from the exception message
                 if (!serverTimeDetermined)
-                    serverTimeDetermined = TryParseExceptionMessage(ase, out serverTime);
+                    serverTimeDetermined = TryParseExceptionMessageToGetServerTime(ase, out serverTime);
 
                 if (serverTimeDetermined)
                 {
                     // using accurate server time, calculate correction if local time is off
                     serverTime = serverTime.ToUniversalTime();
-                    var diff = correctedNow - serverTime;
+
+                    DateTime clientTime;
+                    bool clientTimeDetermined = TryParseExceptionMessageToGetClientTime(ase, out clientTime);
+                    clientTime = clientTime.ToUniversalTime();
+
+                    var diff = clientTimeDetermined ? clientTime - serverTime : correctedNow - serverTime;
                     var absDiff = diff.Ticks < 0 ? -diff : diff;
                     if (absDiff > clockSkewMaxThreshold)
                     {
@@ -497,7 +504,7 @@ namespace Amazon.Runtime
             serverTime = DateTime.MinValue;
             return false;
         }
-        private static bool TryParseExceptionMessage(AmazonServiceException ase, out DateTime serverTime)
+        private static bool TryParseExceptionMessageToGetServerTime(AmazonServiceException ase, out DateTime serverTime)
         {
             if (ase != null && !string.IsNullOrEmpty(ase.Message))
             {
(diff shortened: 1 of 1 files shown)


peterrsongg:
Thank you for opening a PR. As discussed internally we are going to use a different approach. Pr is here:
https://github.com/aws/aws-sdk-net/pull/3378

PR #3376, opened 2024-07-08
Create separate operation-specific configs for the object-mapper Batch and Transact operations

ashovlin:
## Description
Creates new operation-specific config objects for the DynamoDB object-mapper operations.

I'm splitting this into 3 PRs:
1. Batch and Transact - This PR
2. Delete/Load/Save - #3387
3. Query/Scan - #3383 

I have a longer analysis in an internal doc, but I'm planning to apply these properties to all operations because they deal with the table itself or the DynamoDB item <-> .NET object transformation:
1. `OverrideTableName`
2. `TableNamePrefix`
3. `MetadataCachingMode`
4. `IsEmptyStringValueEnabled`
5. `Conversion`
6. `DisableFetchingTableMetadata`

Then these are on a case-by-case basis:
1. `BackwardQuery` - only when querying
2. `IndexName` - only when querying or scanning
3. `ConditionalOperator` - only when querying or scanning
4. `QueryFilter` - only when querying or scanning
5. `ConsistentRead` - when getting items (including querying or scanning)
6. `SkipVersionCheck` - when writing or deleting items
7. `IgnoreNullValues` - when writing items
8. `RetrieveDateTimeInUtc` - when getting items (including querying or scannign)

## Motivation and Context
For V4 of the SDK, we're planning to split the shared `DynamoDBOperationConfig` object into operation-specific configs, since the shared object has grown overtime to include properties that are not relevant for all operations.

DOTNET-7513

## Testing
Ran DynamoDBv2 sln tests locally

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/DynamoDBv2/Custom/DataModel/BatchGetConfig.cs
+++ b/sdk/src/Services/DynamoDBv2/Custom/DataModel/BatchGetConfig.cs
@@ -1,0 +1,115 @@
+﻿/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ * 
+ *  http://aws.amazon.com/apache2.0
+ * 
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+using System;
+
+namespace Amazon.DynamoDBv2.DataModel
+{
+    /// <summary>
+    /// Input for the BatchGet operation in the object-persistence programming model
+    /// </summary>
+#if NET8_0_OR_GREATER
+    [System.Diagnostics.CodeAnalysis.RequiresUnreferencedCode(Custom.Internal.InternalConstants.RequiresUnreferencedCodeMessage)]
+#endif
+    public class BatchGetConfig
+    {
+        /// <summary>
+        /// Property that indicates which DynamoDB table to use, 
+        /// overriding the DynamoDBTable attribute declared for the object type.
+        /// </summary>
+        public string OverrideTableName { get; set; }
+
+        /// <summary>
+        /// Property that directs DynamoDBContext to prefix all table names
+        /// with a specific string.
+        /// If property is null or empty, no prefix is used and default
+        /// table names are used.
+        /// </summary>
+        public string TableNamePrefix { get; set; }
+
+        /// <summary>
+        /// The object persistence model API relies on an internal cache of the DynamoDB table's metadata to construct and validate 
+        /// requests. This controls how the cache key is derived, which influences when the SDK will call 
+        /// IAmazonDynamoDB.DescribeTable(string) internally to populate the cache.
+        /// </summary>
+        /// <remarks>
+        /// For <see cref="MetadataCachingMode.Default"/> the cache key will be a combination of the table name, credentials, region and service URL. 
+        /// For <see cref="MetadataCachingMode.TableNameOnly"/> the cache key will only consist of the table name. This reduces cache misses in contexts
+        /// where you are accessing tables with identical structure but using different credentials or endpoints (such as a multi-tenant application).
+        /// </remarks>
+        public MetadataCachingMode? MetadataCachingMode { get; set; }
+
+        /// <summary>
+        /// If true disables fetching table metadata automatically from DynamoDB. Table metadata must be 
+        /// defined by <see cref="DynamoDBAttribute"/> attributes and/or in <see cref = "AWSConfigsDynamoDB"/>.
+        /// </summary>
+        /// <remarks>
+        /// Setting this to true can avoid latency and thread starvation due to blocking asynchronous 
+        /// IAmazonDynamoDB.DescribeTable(string) calls that are used to populate the SDK's cache of 
+        /// table metadata. It requires that the table's index schema be accurately described via the above methods, 
+        /// otherwise exceptions may be thrown and/or the results of certain DynamoDB operations may change.
+        /// </remarks>
+        public bool? DisableFetchingTableMetadata { get; set; }
+
+        /// <summary>
+        /// Conversion specification which controls how the conversion between
+        /// .NET and DynamoDB types happens.
+        /// </summary>
+        public DynamoDBEntryConversion Conversion { get; set; }
+
+        /// <summary>
+        /// Property that directs DynamoDBContext to enable empty string values
+        /// on attributes during a Save operation.
+        /// If the property is false (or not set), empty string values will be
+        /// interpreted as null values.
+        /// </summary>
+        public bool? IsEmptyStringValueEnabled { get; set; }
+
+        /// <summary>
+        /// Property that directs DynamoDBContext to use consistent reads.
+        /// If property is not set, behavior defaults to non-consistent reads.
+        /// </summary>
+        public bool? ConsistentRead { get; set; }
+
+        /// <summary>
+        /// If true, all <see cref="DateTime"/> properties are retrieved in UTC timezone while reading data from DynamoDB. Else, the local timezone is used.
+        /// </summary>
+        /// <remarks>This setting is only applicable to the high-level library. Service calls made via <see cref="AmazonDynamoDBClient"/> will always return <see cref="DateTime"/> attributes in UTC.</remarks>
+        public bool? RetrieveDateTimeInUtc { get; set; }
+
+        /// <summary>
+        /// Converts this to the shared <see cref="DynamoDBOperationConfig"/>
+        /// </summary>
+        /// <remarks>
+        /// Users should interact with the new, operation-specific configs, but we
+        /// convert to the internal shared config for the internal code paths.
+        /// </remarks>
+        /// <returns>New <see cref="DynamoDBOperationConfig"/></returns>
+        internal DynamoDBOperationConfig ToDynamoDBOperationConfig()
+        {
+            return new DynamoDBOperationConfig()
+            {
+                OverrideTableName = OverrideTableName,
+                TableNamePrefix = TableNamePrefix,
+                MetadataCachingMode = MetadataCachingMode,
+                DisableFetchingTableMetadata = DisableFetchingTableMetadata,
+                Conversion = Conversion,
+                IsEmptyStringValueEnabled = IsEmptyStringValueEnabled,
+                ConsistentRead = ConsistentRead,
+                RetrieveDateTimeInUtc = RetrieveDateTimeInUtc
+            };
+        }
+    }
+}
--- a/sdk/src/Services/DynamoDBv2/Custom/DataModel/BatchGetConfig.cs
+++ b/sdk/src/Services/DynamoDBv2/Custom/DataModel/BatchGetConfig.cs
@@ -21,64 +21,13 @@ namespace Amazon.DynamoDBv2.DataModel
     /// Input for the BatchGet operation in the object-persistence programming model
     /// </summary>
 #if NET8_0_OR_GREATER
+    // The DataModel namespace doesn't support trimming yet, so annotate public classes/methods as incompatible
     [System.Diagnostics.CodeAnalysis.RequiresUnreferencedCode(Custom.Internal.InternalConstants.RequiresUnreferencedCodeMessage)]
 #endif
-    public class BatchGetConfig
+    public class BatchGetConfig : BaseOperationConfig
     {
         /// <summary>
-        /// Property that indicates which DynamoDB table to use, 
-        /// overriding the DynamoDBTable attribute declared for the object type.
-        /// </summary>
-        public string OverrideTableName { get; set; }
-
-        /// <summary>
-        /// Property that directs DynamoDBContext to prefix all table names
-        /// with a specific string.
-        /// If property is null or empty, no prefix is used and default
-        /// table names are used.
-        /// </summary>
-        public string TableNamePrefix { get; set; }
-
-        /// <summary>
-        /// The object persistence model API relies on an internal cache of the DynamoDB table's metadata to construct and validate 
-        /// requests. This controls how the cache key is derived, which influences when the SDK will call 
-        /// IAmazonDynamoDB.DescribeTable(string) internally to populate the cache.
-        /// </summary>
-        /// <remarks>
-        /// For <see cref="MetadataCachingMode.Default"/> the cache key will be a combination of the table name, credentials, region and service URL. 
-        /// For <see cref="MetadataCachingMode.TableNameOnly"/> the cache key will only consist of the table name. This reduces cache misses in contexts
-        /// where you are accessing tables with identical structure but using different credentials or endpoints (such as a multi-tenant application).
-        /// </remarks>
-        public MetadataCachingMode? MetadataCachingMode { get; set; }
-
-        /// <summary>
-        /// If true disables fetching table metadata automatically from DynamoDB. Table metadata must be 
-        /// defined by <see cref="DynamoDBAttribute"/> attributes and/or in <see cref = "AWSConfigsDynamoDB"/>.
-        /// </summary>
-        /// <remarks>
-        /// Setting this to true can avoid latency and thread starvation due to blocking asynchronous 
-        /// IAmazonDynamoDB.DescribeTable(string) calls that are used to populate the SDK's cache of 
-        /// table metadata. It requires that the table's index schema be accurately described via the above methods, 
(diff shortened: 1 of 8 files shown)


normj:
If we are putting the same 6 properties on all operation configs should those be in a base class? Or does that cause confusion when you get to other operations.

ashovlin:
> If we are putting the same 6 properties on all operation configs should those be in a base class? Or does that cause confusion when you get to other operations.

I was on the fence about this. Given that we are unwinding an object that became over-shared, I opted to treat it like generated code and keep all the definitions separate. But I am planning to use those 6 for all of our current operations as they're used either at the table level or anytime we convert POCO <-> items. 

I suppose that if we did use a base class and added a new operation that didn't use one of the 6, then we can not use the base class. 

Any thoughts @96malhar or @philasmar ?

sdk/src/Services/DynamoDBv2/Custom/DataModel/BatchGetConfig.cs
```diff
@@ -0,0 +1,115 @@
...
+        /// with a specific string.
+        /// If property is null or empty, no prefix is used and default
+        /// table names are used.
+        /// </summary>
+        public string TableNamePrefix { get; set; }
+
+        /// <summary>
+        /// The object persistence model API relies on an internal cache of the DynamoDB table's metadata to construct and validate 
+        /// requests. This controls how the cache key is derived, which influences when the SDK will call 
+        /// IAmazonDynamoDB.DescribeTable(string) internally to populate the cache.
```
philasmar, line 45: nit: add doc reference to `IAmazonDynamoDB.DescribeTable(string)`

sdk/src/Services/DynamoDBv2/Custom/DataModel/BatchGetConfig.cs
```diff
@@ -0,0 +1,115 @@
...
+        /// </remarks>
+        public MetadataCachingMode? MetadataCachingMode { get; set; }
+
+        /// <summary>
+        /// If true disables fetching table metadata automatically from DynamoDB. Table metadata must be 
+        /// defined by <see cref="DynamoDBAttribute"/> attributes and/or in <see cref = "AWSConfigsDynamoDB"/>.
+        /// </summary>
+        /// <remarks>
+        /// Setting this to true can avoid latency and thread starvation due to blocking asynchronous 
+        /// IAmazonDynamoDB.DescribeTable(string) calls that are used to populate the SDK's cache of
```
philasmar, line 60: nit: add doc reference to `IAmazonDynamoDB.DescribeTable(string)`

sdk/src/Services/DynamoDBv2/Custom/DataModel/BatchWriteConfig.cs
```diff
@@ -0,0 +1,116 @@
...
+        /// with a specific string.
+        /// If property is null or empty, no prefix is used and default
+        /// table names are used.
+        /// </summary>
+        public string TableNamePrefix { get; set; }
+
+        /// <summary>
+        /// The object persistence model API relies on an internal cache of the DynamoDB table's metadata to construct and validate 
+        /// requests. This controls how the cache key is derived, which influences when the SDK will call 
+        /// IAmazonDynamoDB.DescribeTable(string) internally to populate the cache.
```
philasmar, line 43: nit: add doc reference to `IAmazonDynamoDB.DescribeTable(string)`

sdk/src/Services/DynamoDBv2/Custom/DataModel/BatchWriteConfig.cs
```diff
@@ -0,0 +1,116 @@
...
+        /// </remarks>
+        public MetadataCachingMode? MetadataCachingMode { get; set; }
+
+        /// <summary>
+        /// If true disables fetching table metadata automatically from DynamoDB. Table metadata must be 
+        /// defined by <see cref="DynamoDBAttribute"/> attributes and/or in <see cref = "AWSConfigsDynamoDB"/>.
+        /// </summary>
+        /// <remarks>
+        /// Setting this to true can avoid latency and thread starvation due to blocking asynchronous 
+        /// IAmazonDynamoDB.DescribeTable(string) calls that are used to populate the SDK's cache of
```
philasmar, line 58: nit: add doc reference to `IAmazonDynamoDB.DescribeTable(string)`

sdk/src/Services/DynamoDBv2/Custom/DataModel/IDynamoDBContext.cs
```diff
@@ -113,14 +113,33 @@ public partial interface IDynamoDBContext : IDisposable
...
+
         /// <summary>
         /// Creates a strongly-typed BatchGet object, allowing
         /// a batch-get operation against DynamoDB.
         /// </summary>
         /// <typeparam name="T">Type of objects to get</typeparam>
         /// <param name="operationConfig">Config object which can be used to override that table used.</param>
         /// <returns>Empty strongly-typed BatchGet object</returns>
-        BatchGet<T> CreateBatchGet<T>(DynamoDBOperationConfig operationConfig = null);
+        [Obsolete("Use the CreateBatchGet overload that takes BatchGetConfig instead, since DynamoDBOperationConfig contains properties that are not applicable to BatchGet.")]
```
philasmar, line 132: why did you change the signature of the method?

sdk/src/Services/DynamoDBv2/Custom/DataModel/IDynamoDBContext.cs
```diff
@@ -134,14 +153,38 @@ public partial interface IDynamoDBContext : IDisposable
...
         /// Creates a strongly-typed BatchWrite object, allowing
         /// a batch-write operation against DynamoDB.
         /// </summary>
         /// <typeparam name="T">Type of objects to write</typeparam>
         /// <param name="operationConfig">Config object which can be used to override that table used.</param>
         /// <returns>Empty strongly-typed BatchWrite object</returns>
-        BatchWrite<T> CreateBatchWrite<T>(DynamoDBOperationConfig operationConfig = null);
+        [Obsolete("Use the CreateBatchWrite overload that takes BatchWriteConfig instead, since DynamoDBOperationConfig contains properties that are not applicable to BatchWrite.")]
+
+        BatchWrite<T> CreateBatchWrite<T>(DynamoDBOperationConfig operationConfig);
```
philasmar, line 173: why did you change the signature of the method?

sdk/src/Services/DynamoDBv2/Custom/DataModel/IDynamoDBContext.cs
```diff
@@ -156,7 +199,32 @@ public partial interface IDynamoDBContext : IDisposable
         /// <param name="valuesType">The type of data which will be persisted in this batch.</param>
         /// <param name="operationConfig">Config object which can be used to override that table used.</param>
         /// <returns>Empty strongly-typed BatchWrite object</returns>
-        BatchWrite<object> CreateBatchWrite(Type valuesType, DynamoDBOperationConfig operationConfig = null);
+        [Obsolete("Use the CreateBatchWrite overload that takes BatchWriteConfig instead, since DynamoDBOperationConfig contains properties that are not applicable to BatchWrite.")]
+        BatchWrite<object> CreateBatchWrite(Type valuesType, DynamoDBOperationConfig operationConfig);
```
philasmar, line 203: why did you change the signature of the method?

sdk/src/Services/DynamoDBv2/Custom/DataModel/IDynamoDBContext.cs
```diff
@@ -170,14 +238,32 @@ public partial interface IDynamoDBContext : IDisposable
...
+
         /// <summary>
         /// Creates a strongly-typed TransactGet object, allowing
         /// a transactional get operation against DynamoDB.
         /// </summary>
         /// <typeparam name="T">Type of objects to get.</typeparam>
         /// <param name="operationConfig">Config object which can be used to override that table used.</param>
         /// <returns>Empty strongly-typed TransactGet object.</returns>
-        TransactGet<T> CreateTransactGet<T>(DynamoDBOperationConfig operationConfig = null);
+        [Obsolete("Use the CreateTransactGet overload that takes TransactGetConfig instead, since DynamoDBOperationConfig contains properties that are not applicable to BatchGet.")]
```
philasmar, line 256: why did you change the signature of the method?

sdk/src/Services/DynamoDBv2/Custom/DataModel/IDynamoDBContext.cs
```diff
@@ -191,14 +277,32 @@ public partial interface IDynamoDBContext : IDisposable
...
         /// <summary>
         /// Creates a strongly-typed TransactWrite object, allowing
         /// a transactional write operation against DynamoDB.
         /// </summary>
         /// <typeparam name="T">Type of objects to write.</typeparam>
         /// <param name="operationConfig">Config object which can be used to override that table used.</param>
         /// <returns>Empty strongly-typed TransactWrite object.</returns>
-        TransactWrite<T> CreateTransactWrite<T>(DynamoDBOperationConfig operationConfig = null);
+        [Obsolete("Use the CreateTransactWrite overload that takes TransactWriteConfig instead, since DynamoDBOperationConfig contains properties that are not applicable to CreateTransactWrite.")]
+        TransactWrite<T> CreateTransactWrite<T>(DynamoDBOperationConfig operationConfig);
```
philasmar, line 296: why did you change the signature of the method?

philasmar:
> > If we are putting the same 6 properties on all operation configs should those be in a base class? Or does that cause confusion when you get to other operations.
> 
> I was on the fence about this. Given that we are unwinding an object that became over-shared, I opted to treat it like generated code and keep all the definitions separate. But I am planning to use those 6 for all of our current operations as they're used either at the table level or anytime we convert POCO <-> items.
> 
> I suppose that if we did use a base class and added a new operation that didn't use one of the 6, then we can not use the base class.
> 
> Any thoughts @96malhar or @philasmar ?

I prefer to use a BaseClass. This is going to become a maintainability issue. We should not add any property to this base class that does not apply to all operations. Any property that is operation-specific should go on the operation-specific config object.

ashovlin:
> > > If we are putting the same 6 properties on all operation configs should those be in a base class? Or does that cause confusion when you get to other operations.
> > 
> > 
> > I was on the fence about this. Given that we are unwinding an object that became over-shared, I opted to treat it like generated code and keep all the definitions separate. But I am planning to use those 6 for all of our current operations as they're used either at the table level or anytime we convert POCO <-> items.
> > I suppose that if we did use a base class and added a new operation that didn't use one of the 6, then we can not use the base class.
> > Any thoughts @96malhar or @philasmar ?
> 
> I prefer to use a BaseClass. This is going to become a maintainability issue. We should not add any property to this base class that does not apply to all operations. Any property that is operation-specific should go on the operation-specific config object.

Created https://github.com/aws/aws-sdk-net/pull/3376 to introduce the base class. Once that is merged I'll rebase this on top of it.

sdk/src/Services/DynamoDBv2/Custom/DataModel/BatchGetConfig.cs
ashovlin: `DescribeTable` isn't available in all targets, it's `DescribeTableAync` in .NET/Core/Standard.

I didn't think it was worth `#if` in the doc string, and the API reference generator is going to pick one anyway.

sdk/src/Services/DynamoDBv2/Custom/DataModel/IDynamoDBContext.cs
ashovlin: If we added something similar to what was there, with the new config as optional:
```
BatchGet<T> CreateBatchGet<T>(DynamoDBOperationConfig operationConfig = null);
BatchGet<T> CreateBatchGet<T>(BatchGetConfig batchGetConfig = null);
```

Then these two calls are both ambiguous:
```
context.CreateBatchGet<Foo>();
context.CreateBatchGet<Foo>(null);
```

If we didn't make the new config optional:
```
BatchGet<T> CreateBatchGet<T>(DynamoDBOperationConfig operationConfig = null);
BatchGet<T> CreateBatchGet<T>(BatchGetConfig batchGetConfig);
```

Then
```
context.CreateBatchGet<Foo>();     // Calling the now obsolete overload, which we don't want
context.CreateBatchGet<Foo>(null); // Still ambiguous
```

So for that `context.CreateBatchGet<Foo>();` case, I added a non-obsolete parameterless `CreateBatchGet<T>()`
```
BatchGet<T> CreateBatchGet<T>();
BatchGet<T> CreateBatchGet<T>(DynamoDBOperationConfig operationConfig = null);
BatchGet<T> CreateBatchGet<T>(BatchGetConfig batchGetConfig);
```

With that I was worried `context.CreateBatchGet<Foo>()` would still be ambiguous between the first two, but reviewing https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#12643-better-function-member again the parameterless one will win. So I'll add back the middle call. 

`context.CreateBatchGet<Foo>(null);` is still ambiguous, but I don't think we can do anything about that (nor should we, since that call doesn't add anything over `context.CreateBatchGet<Foo>()` that I can see

sdk/src/Services/DynamoDBv2/Custom/DataModel/BatchGetConfig.cs
ashovlin: Resolved over in #3380

sdk/src/Services/DynamoDBv2/Custom/DataModel/BatchGetConfig.cs
ashovlin: Resolved over in #3380

sdk/src/Services/DynamoDBv2/Custom/DataModel/BatchWriteConfig.cs
ashovlin: Resolved over in #3380

sdk/src/Services/DynamoDBv2/Custom/DataModel/BatchWriteConfig.cs
ashovlin: Resolved over in #3380

sdk/src/Services/DynamoDBv2/Custom/DataModel/IDynamoDBContext.cs
ashovlin: See https://github.com/aws/aws-sdk-net/pull/3376#discussion_r1672274431

sdk/src/Services/DynamoDBv2/Custom/DataModel/IDynamoDBContext.cs
ashovlin: See https://github.com/aws/aws-sdk-net/pull/3376#discussion_r1672274431

sdk/src/Services/DynamoDBv2/Custom/DataModel/IDynamoDBContext.cs
ashovlin: See https://github.com/aws/aws-sdk-net/pull/3376#discussion_r1672274431

sdk/src/Services/DynamoDBv2/Custom/DataModel/IDynamoDBContext.cs
ashovlin: See https://github.com/aws/aws-sdk-net/pull/3376#discussion_r1672274431

sdk/test/Services/DynamoDBv2/UnitTests/Custom/DataModelOperationSpecificConfigTests.cs
```diff
@@ -18,5 +18,37 @@ public void BaseOperationConfig()
             // `ToDynamoDBOperationConfig` before updating this unit test
             Assert.AreEqual(6, typeof(BaseOperationConfig).GetProperties().Length);
         }
+
+        [TestMethod]
+        public void BatchGetConfig()
```
peterrsongg, line 23: non-blocking. but maybe worth adding unit test coverage for `ToDynamoDBOperationConfig`. Don't have to see this added, but just commenting since this is a new method, and these tests mainly seem to be a reminder to add new properties to `ToDynamoDbOperationConfig`

peterrsongg (approved):
Approved with a comment

philasmar (approved):

sdk/test/Services/DynamoDBv2/UnitTests/Custom/DataModelOperationSpecificConfigTests.cs
ashovlin: Added in https://github.com/aws/aws-sdk-net/pull/3376/commits/68959295c0ea195936627845e99afdb54bc3fd91.

It's testing that one of the properties on the new config object is correctly overriding the context-level config by inspecting the low-level request.

PR #3378, opened 2024-07-09
Fix clockskew race condition in multi-threading scenarios

peterrsongg:
## Description

This fixes a race condition that may happen when two threads accessing the same region with the same client have a clockskew. Both threads enter the `IsClockSkew` function and one thread will update the clockskew. The first thread will have a diff > the max allowed threshold and will return true and retry. The second thread will then use the updated time (client time + clockskew) and since this is less than the max threshold, the function will return false and not retry. When done serially this doesn't occur because the request will use the updated time and therefore not return a clockskew error code.

The fix is to always retry if we get a clockskew error code and always update the dictionary which holds the clockskew per endpoint. This way even if two threads were to enter the function, they would both retry and succeed upon the second attempt (1st retry attempt). For HEAD requests, we cannot rely on the error code, sometimes we will get a generic 403, so in this case we rely on the diff between the current corrected clock and the server time. 

## Motivation and Context

Internal issue 7600
## Testing

Added Unit Test
Dry Run Passes
## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs
+++ b/sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs
@@ -392,7 +392,7 @@ namespace Amazon.Runtime
             var serviceException = exception as AmazonServiceException;
             return TimeoutErrorCodesToRetryOn.Contains(serviceException?.ErrorCode);
         }
-        
+
 
         #region Clock skew correction
 
@@ -400,11 +400,10 @@ namespace Amazon.Runtime
         {
             "RequestTimeTooSkewed",
             "RequestExpired",
+            "RequestInTheFuture",
             "InvalidSignatureException",
             "SignatureDoesNotMatch",
-            "AuthFailure",
-            "RequestExpired",
-            "RequestInTheFuture",
+            "AuthFailure"
         };
 
         private const string clockSkewMessageFormat = "Identified clock skew: local time = {0}, local time with correction = {1}, current clock skew correction = {2}, server time = {3}, service endpoint = {4}.";
@@ -430,7 +429,6 @@ namespace Amazon.Runtime
             {
                 var endpoint = executionContext.RequestContext.Request.Endpoint.ToString();
                 var realNow = AWSConfigs.utcNowSource();
-                var correctedNow = CorrectClockSkew.GetCorrectedUtcNowForEndpoint(endpoint);
 
                 DateTime serverTime;
 
@@ -445,28 +443,30 @@ namespace Amazon.Runtime
                 {
                     // using accurate server time, calculate correction if local time is off
                     serverTime = serverTime.ToUniversalTime();
-                    var diff = correctedNow - serverTime;
+                    var correctedNow = CorrectClockSkew.GetCorrectedUtcNowForEndpoint(endpoint);
+                    var diff =  correctedNow - serverTime;
+                    var newCorrection = serverTime - realNow;
                     var absDiff = diff.Ticks < 0 ? -diff : diff;
-                    if (absDiff > clockSkewMaxThreshold)
+
+                    Logger.InfoFormat(clockSkewMessageFormat,
+                        realNow, correctedNow, clientConfig.ClockOffset, serverTime, endpoint);
+
+                    // Always set the correction, for informational purposes
+                    CorrectClockSkew.SetClockCorrectionForEndpoint(endpoint, newCorrection);
+                    var shouldRetry = AWSConfigs.CorrectForClockSkew && !AWSConfigs.ManualClockCorrection.HasValue;
+                    // Some services such as S3 will return a generic 403 error for clock skew errors on HEAD requests
+                    // with no error code. In this case we will always retry the request if the clock skew is greater than
+                    // the threshold.
+                    if (shouldRetry)
                     {
-                        var newCorrection = serverTime - realNow;
-                        Logger.InfoFormat(clockSkewMessageFormat,
-                            realNow, correctedNow, clientConfig.ClockOffset, serverTime, endpoint);
-
-                        // Always set the correction, for informational purposes
-                        CorrectClockSkew.SetClockCorrectionForEndpoint(endpoint, newCorrection);
-
-                        var shouldRetry = AWSConfigs.CorrectForClockSkew && !AWSConfigs.ManualClockCorrection.HasValue;
-
-                        // Only retry if clock skew correction is not disabled
-                        if (shouldRetry)
+                        if ((isHead && absDiff > clockSkewMaxThreshold) || isClockskewErrorCode)
                         {
-                            // Set clock skew correction
                             Logger.InfoFormat(clockSkewUpdatedFormat, newCorrection, endpoint);
                             executionContext.RequestContext.IsSigned = false;
                             return true;
                         }
                     }
+
                 }
             }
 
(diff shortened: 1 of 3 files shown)


sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs
```diff
@@ -446,26 +445,32 @@ private bool IsClockskew(IExecutionContext executionContext, Exception exception
...
-                            // Set clock skew correction
-                            Logger.InfoFormat(clockSkewUpdatedFormat, newCorrection, endpoint);
-                            executionContext.RequestContext.IsSigned = false;
-                            return true;
-                        }
+                    // Some services such as S3 will return a generic 403 error for clock skew errors on HEAD requests
+                    // with no error code. In this case we will always retry the request if the clock skew is greater than
+                    // the threshold.
+                    if (isHead && absDiff > clockSkewMaxThreshold)
+                    {
```
dscpinheiro, line 461: Question: In the existing code path that returns `true`, we also have:
```
Logger.InfoFormat(clockSkewUpdatedFormat, newCorrection, endpoint);
executionContext.RequestContext.IsSigned = false;
```

Are those not required here? (logging would be helpful, I'm more curious about the `IsSigned = false` assignment)

peterrsongg: Ah that's a good catch, I need to add that in this code path as well.

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs
```diff
@@ -446,26 +445,32 @@ private bool IsClockskew(IExecutionContext executionContext, Exception exception
...
-                        {
-                            // Set clock skew correction
-                            Logger.InfoFormat(clockSkewUpdatedFormat, newCorrection, endpoint);
-                            executionContext.RequestContext.IsSigned = false;
-                            return true;
-                        }
+                    // Some services such as S3 will return a generic 403 error for clock skew errors on HEAD requests
+                    // with no error code. In this case we will always retry the request if the clock skew is greater than
+                    // the threshold.
+                    if (isHead && absDiff > clockSkewMaxThreshold)
```
boblodgett, line 460: Shouldn't this if also take into account `AWSConfigs.CorrectForClockSkew && !AWSConfigs.ManualClockCorrection.HasValue` ?

If you move the shouldRetry calculation up here you can just do if(!shouldRetry) return false; and short circuit everything below.

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs
```diff
@@ -446,26 +445,32 @@ private bool IsClockskew(IExecutionContext executionContext, Exception exception
...
+                    {
+                        return true;
+                    }
+
+                    var shouldRetry = AWSConfigs.CorrectForClockSkew && !AWSConfigs.ManualClockCorrection.HasValue && isClockskewErrorCode;
```
boblodgett, line 465: I think this should be moved up before the prior if so it can be included there, removing the isClockskewErrorCode then change the if(shouldRetry) to an else

peterrsongg: I ended up merging the if statements to make it cleaner and moving the Shouldretry logic up.

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs
```diff
@@ -446,27 +445,28 @@ private bool IsClockskew(IExecutionContext executionContext, Exception exception
...
-                        var shouldRetry = AWSConfigs.CorrectForClockSkew && !AWSConfigs.ManualClockCorrection.HasValue;
+                    Logger.InfoFormat(clockSkewMessageFormat,
+                        realNow, correctedNow, clientConfig.ClockOffset, serverTime, endpoint);
 
-                        // Only retry if clock skew correction is not disabled
-                        if (shouldRetry)
+                    // Always set the correction, for informational purposes
+                    CorrectClockSkew.SetClockCorrectionForEndpoint(endpoint, newCorrection);
+                    var shouldRetry = AWSConfigs.CorrectForClockSkew && !AWSConfigs.ManualClockCorrection.HasValue;
+                    // Some services such as S3 will return a generic 403 error for clock skew errors on HEAD requests
```
Somanko, line 457: Is this from JAVA SDK?

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs
```diff
@@ -446,27 +445,28 @@ private bool IsClockskew(IExecutionContext executionContext, Exception exception
                     // using accurate server time, calculate correction if local time is off
                     serverTime = serverTime.ToUniversalTime();
                     var diff = correctedNow - serverTime;
+                    var newCorrection = serverTime - realNow;
                     var absDiff = diff.Ticks < 0 ? -diff : diff;
-                    if (absDiff > clockSkewMaxThreshold)
-                    {
-                        var newCorrection = serverTime - realNow;
-                        Logger.InfoFormat(clockSkewMessageFormat,
```
girishrdatamzn, line 453: nit: What's the rationale for setting this log at INFO level? Just concerned about ROI and log real-estate in some of the consumes such as KTap using SDK calls in multi threaded environment. 

Can it be either a WARN to seek attention from consumer or a DEBUG to indicate the SDK is taking care of such issues.

Somanko: This log helped us debugging the error.

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs
```diff
@@ -446,27 +445,28 @@ private bool IsClockskew(IExecutionContext executionContext, Exception exception
...
-                        if (shouldRetry)
+                    // Always set the correction, for informational purposes
+                    CorrectClockSkew.SetClockCorrectionForEndpoint(endpoint, newCorrection);
+                    var shouldRetry = AWSConfigs.CorrectForClockSkew && !AWSConfigs.ManualClockCorrection.HasValue;
+                    // Some services such as S3 will return a generic 403 error for clock skew errors on HEAD requests
+                    // with no error code. In this case we will always retry the request if the clock skew is greater than
+                    // the threshold.
+                    if (shouldRetry)
+                    {
+                        if ((isHead && absDiff > clockSkewMaxThreshold) || isClockskewErrorCode)
```
Somanko, line 462: As per this code, the bug is not fixed for S3. In the repro scenario when 2 threads are coming together, the second one will not go inside the if block and will return false causing it to not retry and the throw the exception back to caller of SDK

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs
peterrsongg: If you're referring to the HEAD request comment, that was already there and is because of a COE from a while back where a HEAD request was not returning an error code since there is no body

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs
peterrsongg: ack, let's unblock you guys first and come back to the head request as a follow up.

boblodgett (approved):
Approving knowing we will likely need to make an adjustment for S3 and Head which we can do in another PR. The HEAD behavior is the same as today.

dscpinheiro (approved):

peterrsongg:
The force push is me just squasing the commits :)

PR #3380, opened 2024-07-10
Add base class for upcoming object persistence operation-specific config objects

ashovlin:
## Description
This is adding a base class to use for the new set of operation-specific config objects for the object persistence APIs.

## Motivation and Context
See #3376 for the full context for this change, this PR came from https://github.com/aws/aws-sdk-net/pull/3376#issuecomment-2220525789. I'd rather merge this into the feature branch quickly, then rebase the PRs that are introducing the derived classes on top of it.

## Testing
Existing DynamoDBv2 sln tests pass.

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/DynamoDBv2/Custom/DataModel/BaseOperationConfig.cs
+++ b/sdk/src/Services/DynamoDBv2/Custom/DataModel/BaseOperationConfig.cs
@@ -23,6 +23,7 @@ namespace Amazon.DynamoDBv2.DataModel
     /// anything operation-specific should be added to derived classes.
     /// </remarks>
 #if NET8_0_OR_GREATER
+    // The DataModel namespace doesn't support trimming yet, so annotate public classes/methods as incompatible
     [System.Diagnostics.CodeAnalysis.RequiresUnreferencedCode(Custom.Internal.InternalConstants.RequiresUnreferencedCodeMessage)]
 #endif
     public abstract class BaseOperationConfig
@@ -41,9 +42,9 @@ namespace Amazon.DynamoDBv2.DataModel
         public string TableNamePrefix { get; set; }
 
         /// <summary>
-        /// The object persistence model API relies on an internal cache of the DynamoDB table's metadata to construct and validate 
-        /// requests. This controls how the cache key is derived, which influences when the SDK will call
-        /// IAmazonDynamoDB.DescribeTable(string) internally to populate the cache.
+        /// The object persistence model API relies on an internal cache of the DynamoDB table's metadata to 
+        /// construct and validate  requests. This controls how the cache key is derived, which influences 
+        /// when the SDK will call DescribeTable internally to populate the cache.
         /// </summary>
         /// <remarks>
         /// For <see cref="MetadataCachingMode.Default"/> the cache key will be a combination of the table name, credentials, region and service URL. 
@@ -58,8 +59,8 @@ namespace Amazon.DynamoDBv2.DataModel
         /// </summary>
         /// <remarks>
         /// Setting this to true can avoid latency and thread starvation due to blocking asynchronous 
-        /// IAmazonDynamoDB.DescribeTable(string) calls that are used to populate the SDK's cache of 
-        /// table metadata. It requires that the table's index schema be accurately described via the above methods, 
+        /// DescribeTable calls that are used to populate the SDK's cache of table metadata. 
+        /// It requires that the table's index schema be accurately described via the above methods, 
         /// otherwise exceptions may be thrown and/or the results of certain DynamoDB operations may change.
         /// </remarks>
         public bool? DisableFetchingTableMetadata { get; set; }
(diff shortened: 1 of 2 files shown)


philasmar (approved):

sdk/src/Services/DynamoDBv2/Custom/DataModel/BaseOperationConfig.cs
```diff
@@ -0,0 +1,100 @@
...
+{
+    /// <summary>
+    /// Base class for operation-specific configurations for DynamoDB object persistence operations.
+    /// </summary>
+    /// <remarks>
+    /// This should only contain members that are relevant to all object persistence operations, 
+    /// anything operation-specific should be added to derived classes.
+    /// </remarks>
+#if NET8_0_OR_GREATER
+    [System.Diagnostics.CodeAnalysis.RequiresUnreferencedCode(Custom.Internal.InternalConstants.RequiresUnreferencedCodeMessage)]
```
96malhar, line 26: Can you add a code comment for posterity as to why is the `RequiresUnreferencedCodeMessage` attribute needed?

sdk/src/Services/DynamoDBv2/Custom/DataModel/BaseOperationConfig.cs
```diff
@@ -0,0 +1,100 @@
...
+        /// </remarks>
+        public MetadataCachingMode? MetadataCachingMode { get; set; }
+
+        /// <summary>
+        /// If true disables fetching table metadata automatically from DynamoDB. Table metadata must be 
+        /// defined by <see cref="DynamoDBAttribute"/> attributes and/or in <see cref = "AWSConfigsDynamoDB"/>.
+        /// </summary>
+        /// <remarks>
+        /// Setting this to true can avoid latency and thread starvation due to blocking asynchronous 
+        /// IAmazonDynamoDB.DescribeTable(string) calls that are used to populate the SDK's cache of
```
96malhar, line 61: Can you also wrap `IAmazonDynamoDB.DescribeTable(string)` in a `cref`

sdk/src/Services/DynamoDBv2/Custom/DataModel/BaseOperationConfig.cs
```diff
@@ -0,0 +1,100 @@
...
+        /// Directs <see cref="DynamoDBContext"/> to prefix all table names
+        /// with a specific string. If this is null or empty, no prefix is used 
+        /// and default table names are used.
+        /// </summary>
+        public string TableNamePrefix { get; set; }
+
+        /// <summary>
+        /// The object persistence model API relies on an internal cache of the DynamoDB table's metadata to construct and validate 
+        /// requests. This controls how the cache key is derived, which influences when the SDK will call
+        /// IAmazonDynamoDB.DescribeTable(string) internally to populate the cache.
```
96malhar, line 46: Can you also wrap `IAmazonDynamoDB.DescribeTable(string)` in a `cref`

sdk/src/Services/DynamoDBv2/Custom/DataModel/BaseOperationConfig.cs
ashovlin: It's not available in all targets since `DescribeTable` vs. `DescribeTableAsync` vary. 
The API reference generator doesn't generate target-specific pages, so I don't think it's worth `#if` inside the comment block. And technically we're always calling an _internal_ `DescribeTable` even from async targets, which is the root cause of the sync-over-async issues.

sdk/src/Services/DynamoDBv2/Custom/DataModel/BaseOperationConfig.cs
ashovlin: See https://github.com/aws/aws-sdk-net/pull/3380#discussion_r1672791987

sdk/src/Services/DynamoDBv2/Custom/DataModel/BaseOperationConfig.cs
ashovlin: Added in https://github.com/aws/aws-sdk-net/pull/3380/commits/3e03c1ee9c64a1b54987c0e37acbb7e4309b5314

sdk/src/Services/DynamoDBv2/Custom/DataModel/BaseOperationConfig.cs
ashovlin: I switched to just a plain `DescribeTable` in https://github.com/aws/aws-sdk-net/pull/3380/commits/3e03c1ee9c64a1b54987c0e37acbb7e4309b5314 since this keeps coming up. 

From the user's perspective, the important part is that we're calling that API implicitly. The exact invocation is less important.

96malhar (approved):

PR #3382, opened 2024-07-10
Add interfaces for auth scheme resolver

dscpinheiro:
## Description
This PR adds interfaces for the `Authentication Scheme` and `Authentication Scheme Resolver` sections of the SRA identity and auth.

The interfaces will be expanded later (e.g. `AuthScheme` returns an identity resolver and signer to use), but I'm trying to get the initial happy path (AWS service that only uses SigV4) to work first.

## License


extensions/src/Smithy.Identity.Abstractions/Auth/IAuthSchemeParameters.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace Smithy.Identity.Abstractions
+{
+    /// <summary>
+    /// <para>
+    /// The input to the authentication scheme resolver. A code-generated interface for passing
```
normj, line 5: nit: We will code-generate it but since this is a Smithy interface that could someday be used by other users they could write it by hand. I suggest removing "code-generated" from the doc.

extensions/src/Smithy.Identity.Abstractions/Auth/IAuthSchemeResolver.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// An authentication scheme resolver is an entity within the SDK that determines which authentication scheme to use 
+    /// for a service.
+    /// </para>
+    /// 
+    /// <para>
+    /// It accepts data from the request (e.g. request members) and client (e.g. AWS region) and returns a list of
+    /// authentication scheme options that the SDK should use when authenticating that request.
+    /// </para>
+    /// </summary>
+    public interface IAuthSchemeResolver
```
normj, line 16: Do we need to have this be a generic interface where the generic is the type of `IAuthSchemeParameters`? and then use the `T` in the `ResolveAuthScheme`? Otherwise I'm not sure how in the auth resolver which will need the service specific auth parameters without doing an awkward casting.

```
public interface IAuthSchemeResolver<T> : where T : IAuthSchemeParameters
{
    List<IAuthSchemeOption> ResolveAuthScheme(T authParameters);
}
```

normj (commented):
I just have the one significant comment about `IAuthSchemeResolver` but overall I'm good with this as the initial PR to get the auth interfaces in place to build off of.

extensions/src/Smithy.Identity.Abstractions/Auth/IAuthSchemeResolver.cs
dscpinheiro: Refactored this interface in https://github.com/aws/aws-sdk-net/pull/3382/commits/e6ed5012726d5b8cbd6c7d1c0c8838c3bfc55302

sdk/src/Core/Amazon.Runtime/Credentials/Internal/AwsV4aAuthScheme.cs
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+using Smithy.Identity.Abstractions;
+
+namespace Amazon.Runtime.Credentials.Internal
```
peterrsongg, line 18: Do we also have to add HttpBearer? `smithy.api#HTTPBearerAuth`? I see in the description that you're trying to get Sigv4 services to work first, so does that mean this will be added in an upcoming PR?

extensions/src/Smithy.Identity.Abstractions/Auth/IAuthSchemeOption.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace Smithy.Identity.Abstractions
+{
+    /// <summary>
+    /// The output from the authentication scheme resolver. The resolver returns a list of these, in the
```
peterrsongg, line 4: nit: it's a little bit unclear what "these" is referring to when you say "returns a list of these". Can you specify what you're referring to?

sdk/src/Core/Amazon.Runtime/Credentials/Internal/AwsV4aAuthScheme.cs
dscpinheiro: Yes, we do need it but I'll add it separately (bearer token uses a separate implementation - it resolves credentials at request time but that needs to be adapted to the SRA).

extensions/src/Smithy.Identity.Abstractions/Auth/IAuthSchemeOption.cs
dscpinheiro: Do you have a suggestion? These are coming from the SRA book.

peterrsongg: Maybe instead of "these" just say The resolver returns a list of AuthSchemeOptions?

dscpinheiro: Updated in https://github.com/aws/aws-sdk-net/pull/3382/commits/01b7a7cd84e858b5bfb3d6300a2bfa3b35e4aea1

peterrsongg (approved):

extensions/src/Smithy.Identity.Abstractions/Auth/IAuthSchemeResolver.cs
dscpinheiro: That was a good suggestion, the service resolver will be easier to implement (no need to cast anything):
```csharp
public class AutoScalingAuthSchemeParameters : IAuthSchemeParameters
{
    public string Operation { get; set; }
    public string Region { get; set; }
}

public class AutoScalingAuthSchemeResolver : IAuthSchemeResolver<AutoScalingAuthSchemeParameters>
{
    public List<IAuthSchemeOption> ResolveAuthScheme(AutoScalingAuthSchemeParameters authParameters)
    {
        // Can read region here.
        Console.WriteLine(authParameters.Region);
    }
}
```

Issue #1589, opened 2020-04-23
DynamoDB `Table` misrepresents dependency on `IAmazonDynamoDB`


rlyczynski:
I think the `Table` class misrepresents it's dependency on `IAmazonDynamoDB` when it really depends on an `AmazonDynamoDBClient`. The constructor takes an `IAmazonDynamoDB` but then attempts to cast it to a `AmazonDynamoDBClient` just a few lines later. 

https://github.com/aws/aws-sdk-net/blob/41ca5761addd6f1d3dbfcc2752d8569cf1ef93f0/sdk/src/Services/DynamoDBv2/Custom/DocumentModel/Table.cs#L369-L383

This hinders the ability to unit test code that relies on the `Table` class by passing in a mocked version of `IAmazonDynamoDB` and also makes it very difficult to do something like use a decorator instead of an `AmazonDynamoDBClient` specifically. 

At very least I think the constructor should be clear that it depends on a `AmazonDynamoDBClient`, but ideally the `Table` class would really only depended on an `IAmazonDynamoDB`. 

# Steps to Reproduce
Use the `Table` class with an object that implements `IAmazonDynamoDB` but is not a `AmazonDynamoDBClient`. 

```
Table.LoadTable((IAmazonDynamoDB)notAnAmazonDynamoDBClient, "tablename")
```

This will cause the cast to an `AmazonDynamoDBClient` to fail and set `DDBClient` to null and eventually result in a `NullReferenceException`.

ashishdhingra:
Hi @rlyczynski,

Good morning.

If you refer the code at https://github.com/aws/aws-sdk-net/blob/41ca5761addd6f1d3dbfcc2752d8569cf1ef93f0/sdk/src/Services/DynamoDBv2/Custom/DocumentModel/Table.cs#L56, for platforms PCL || UNITY || NETSTANDARD, the following internal property is declared:

```C#
internal AmazonDynamoDBClient DDBClient { get; private set; }
```

The code at line https://github.com/aws/aws-sdk-net/blob/41ca5761addd6f1d3dbfcc2752d8569cf1ef93f0/sdk/src/Services/DynamoDBv2/Custom/DocumentModel/Table.cs#L378 in `Table` constructor simple casts to IAmazonDynamoDb to AmazonDynamoDBClient. I'm not sure the design consideration around this, but cast appears to be correct given that there is an internal property of that type.

Hope this provides some guidance.

Thanks,
Ashish

[...]

PR #3388, opened 2024-07-16
fix: Delay the cast from IAmazonDynamoDB to AmazonDynamoDBClient in the DocumentModel 

ashovlin:
## Description

**Before:** When initializing the document model `Table` object with an `IAmazonDynamoDB`, .NET/Standard/Core, we cast it right away to an `AmazonDynamoDBClient`. 
* Because we didn't have an synchronous HTTP client, we removed the public sync apis from the low-level service client.
* However we never removed the synchronous public `Table` APIs, which still rely on the internal sync low-level APIs internally.
* So to work around this, we kept the sync low-level APIs as `internal` on `AmazonDynamoDBClient`. This is why we need the cast, since the sync low-level APIs are no longer part of the public `IAmazonDynamoDB`.

**Now:** We delay the cast until right before we need to access the internal, low-level sync APIs.
* For users still calling the sync APIs in .NET/Standard/Core, you can avoid this by moving to the async equivalents where possible.
* If you're still relying on the sync APIs but aren't providing an actual `AmazonDynamoDBClient`, you'll now see a clearer `InvalidOperationException` instead of the `NullReferenceException` from the failed cast.

**Future:**
1. We can deprecate the `Table` sync-over-async APIs entirely. This would remove the need for this cast.
2. Since .NET 5 added back a sync HTTP client, we've discussed internally re-exposing sync low-level APIs where possible.

## Motivation and Context
DOTNET-6755 and #1589 (most of the way, until we do on of the _Future_ ideas from above)

This allows `Table` consumers to mock the `IAmazonDynamoDB` from .NET/Standard/Core as long as you stick to async public APIs.

## Testing
```
 dotnet test .\sdk\src\Services\DynamoDBv2\DynamoDBv2.sln
dotnet test .\sdk\AWSSDK.CoreAndCustomUnitTests.NetStandard.sln
```
PENDING - .NET Standard integ tests for DDB

## Types of changes

## Checklist

## License


sdk/src/Services/DynamoDBv2/Custom/DocumentModel/DocumentBatchGet.cs
```diff
@@ -427,15 +427,22 @@ private Results GetAttributeItems()
...
-#else
         private static void CallUntilCompletion(IAmazonDynamoDB client, BatchGetItemRequest request, Results allResults)
-#endif
         {
             do
             {
-                var serviceResponse = client.BatchGetItem(request);
+#if AWS_ASYNC_API
+                // Cast the IAmazonDynamoDB to the concrete client instead, so we can access the internal sync-over-async methods
+                var internalClient = client as AmazonDynamoDBClient;
```
normj, line 436: nit: You can move the casting logic outside of the loop and only do it once.

sdk/src/Services/DynamoDBv2/Custom/DocumentModel/DocumentBatchWrite.cs
```diff
@@ -469,15 +469,22 @@ private static int GetNumberOfWrites(BatchWriteItemRequest request)
...
-        private void CallUntilCompletion(BatchWriteItemRequest request, Dictionary<string, Dictionary<Key, Document>> documentMap, AmazonDynamoDBClient client)
-#else
         private void CallUntilCompletion(BatchWriteItemRequest request, Dictionary<string, Dictionary<Key, Document>> documentMap, IAmazonDynamoDB client)
-#endif
         {
             do
             {
-                var result = client.BatchWriteItem(request);
+#if AWS_ASYNC_API
+                // Cast the IAmazonDynamoDB to the concrete client instead, so we can access the internal sync-over-async methods
```
normj, line 477: nit: Move the casting logic above the loop.

normj (approved):
A couple minor nits but change looks good.

sdk/src/Services/DynamoDBv2/Custom/DocumentModel/DocumentBatchGet.cs
ashovlin: Fixed in https://github.com/aws/aws-sdk-net/pull/3388/commits/aa1e3273fae0ba8826c15a91902728713e06a1d4

sdk/src/Services/DynamoDBv2/Custom/DocumentModel/DocumentBatchWrite.cs
ashovlin: Fixed in https://github.com/aws/aws-sdk-net/pull/3388/commits/aa1e3273fae0ba8826c15a91902728713e06a1d4

ashovlin:
Realized that I was using too wide of a condition `AWS_ASYNC_API`, which was casting unnecessarily for .NET Framework 4.7.2 even though it still has sync APIs we can use. Switched to `NETSTANDARD` in https://github.com/aws/aws-sdk-net/pull/3388/commits/aa1e3273fae0ba8826c15a91902728713e06a1d4

This also made some .NET Framework tests invalid, so I removed them.

normj (approved):

philasmar (approved):

normj (approved):

ashovlin:
Dry-run e8acaff2-7604-4ade-a913-8ea4cc934cfe passed after adding https://github.com/aws/aws-sdk-net/pull/3388/commits/2f8a80ea5900768e8fe24d18d513fed4f61a085e

PR #3390, opened 2024-07-17
Add basic <inheritdoc/> support to the API reference generator

ashovlin:
_7/16/24 note: Pushing up a draft before I'm out on vacation. I still need to test against the full SDK (see below for current testing), but here it is in case anybody wants to take an early look._

### Before
The [API Reference]() for the AWS SDK for .NET uses a homegrown doc generator, which does not follow `<inheritdoc/>` tags.

For example this block in `DefaultConfiguration`: https://github.com/aws/aws-sdk-net/blob/364ebb9f592dfc8b315e826430b0b9a6ad182f08/sdk/src/Core/Amazon.Runtime/DefaultConfiguration.cs#L88-L103

Results in this at https://docs.aws.amazon.com/sdkfornet/v3/apidocs/items/Runtime/TDefaultConfiguration.html:
image

### After
Now the generator will follow plain `<inheritdoc/>` tags for 
1. Types
2. Properties
3. Methods

Same page from above:
image

Since this was quick unplanned work, I did not add support for the following (will submit an internal backlog item):
1. Constructors
2. Fields
3. Following something like `<inheritdoc cref="..."/>` or `<inheritdoc path="...."/>`
    * Because of this, I replaced some of these more comlex `<inheritdoc>` tags in core with copied+pasted docs.

## Motivation and Context
1. We have some API ref pages today that are missing docs
2. While doing DynamoDB V4 work, we're expanding the high-level interfaces and clients, and it's nice to not paste all of the docs twice.

## Testing
1. Ran the generator locally against AWSSDK.Core and AWSSDK.DynamoDBv2 via `dotnet msbuild .\buildtools\doc-build.proj -t:doc-build`. Confirmed the following local pages now have docs that are missing in live:
    * https://docs.aws.amazon.com/sdkfornet/v3/apidocs/items/DynamoDBv2/TTableBuilder.html - methods
    * https://docs.aws.amazon.com/sdkfornet/v3/apidocs/items/Runtime/TDefaultConfiguration.html - properties
2. Dry-run in progress
3. I still want to run a full generation to verify that we're not regressing or generating unintended changes for other assemblies.

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/DefaultConfiguration.cs
+++ b/sdk/src/Core/Amazon.Runtime/DefaultConfiguration.cs
@@ -78,12 +78,8 @@ namespace Amazon.Runtime
     /// <para />
     /// All options above can be configured by users, and the overridden value will take precedence.
     /// <para />
-    /// <b>Note:</b> for any mode other than <see cref="DefaultConfigurationMode.Legacy"/>, the vended default values
-    /// might change as best practices may evolve. As a result, it is encouraged to perform testing when upgrading the SDK
-    /// if you are using a mode other than <see cref="DefaultConfigurationMode.Legacy"/>.
-    /// <para />
-    /// While the <see cref="DefaultConfigurationMode.Legacy"/> defaults mode is specific to .NET,
-    /// other modes are standardized across all of the AWS SDKs.
+    /// <b>Note:</b> the vended default values might change as best practices may evolve. As a result, it is encouraged to perform 
+    /// testing when upgrading the SDK.
     /// <para />
     /// The defaults mode can be configured:
     /// <list type="number">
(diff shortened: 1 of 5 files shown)


muhammad-othman (approved):

sdk/src/Core/Amazon.Runtime/DefaultConfiguration.cs
```diff
@@ -81,7 +81,26 @@ public interface IDefaultConfiguration
...
     }
 
-    /// <inheritdoc cref="IDefaultConfiguration"/>
+    /// <summary>
+    /// A defaults mode determines how certain default configuration options are resolved in the SDK. Based on the provided
+    /// mode, the SDK will vend sensible default values tailored to the specific <see cref="DefaultConfigurationMode"/>.
+    /// <para />
+    /// All options above can be configured by users, and the overridden value will take precedence.
+    /// <para />
+    /// <b>Note:</b> for any mode other than <see cref="DefaultConfigurationMode.Legacy"/>, the vended default values
```
dscpinheiro, line 90: For later when we merge this to `v4`, we should remove the `Legacy` references here. The updated docs are in: https://github.com/aws/aws-sdk-net/blob/v4-development/sdk/src/Core/Amazon.Runtime/DefaultConfiguration.cs#L21-L36

dscpinheiro (approved):

PR #3392, opened 2024-07-18
[Draft] Modeled authentication resolvers

dscpinheiro:
This PR won't be merged as is, I'm trying to get feedback on the proposed implementation of the modeled authentication resolvers (for the interfaces introduced in https://github.com/aws/aws-sdk-net/pull/3382).

For the vast majority of services, the generated auto resolvers will be similar to the example included in this PR (with the exception of services configured for SigV4A at the moment - such as EventBridge and S3).

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs
+++ b/sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs
@@ -1,0 +1,73 @@
+/*
+* Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+* 
+* Licensed under the Apache License, Version 2.0 (the "License").
+* You may not use this file except in compliance with the License.
+* A copy of the License is located at
+* 
+*  http://aws.amazon.com/apache2.0
+* 
+* or in the "license" file accompanying this file. This file is distributed
+* on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+* express or implied. See the License for the specific language governing
+* permissions and limitations under the License.
+*/
+
+using Amazon.Runtime.Credentials.Internal;
+using Smithy.Identity.Abstractions;
+using System.Collections.Generic;
+using System.Linq;
+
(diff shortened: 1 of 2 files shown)


sdk/src/Services/AutoScaling/Generated/Internal/AutoScalingAuthSchemeResolver.cs
```diff
@@ -0,0 +1,63 @@
...
+    // Auth parameters and scheme resolvers will be generated for each service;
+    // The logic for S3 and EventBridge (where authentication rules are in the endpoint-rule-set file) will be different.
+
+    public class AutoScalingAuthSchemeParameters : IAuthSchemeParameters
+    {
+        public string Operation { get; set; }
+        public string Region { get; set; }
+    }
+
+    public class AutoScalingAuthSchemeResolver :
```
normj, line 21: This class is a both an SDK pipeline handler and a Smithy RA auth resolver which I think is a bit odd. The class should be one thing not both. I would suggest creating `AutoScalingAuthSchemeHandler` that would be generated to convert the `IRequestContext` into `AutoScalingAuthSchemeParameters`. Then have the handler call the `AutoScalingAuthSchemeResolver`.

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs
```diff
@@ -0,0 +1,73 @@
...
+                // var identityResolver = scheme.IdentityResolver();
+                // var identity = identityResolver.GetIdentity();
+                // var signer = scheme.Signer();
+
+                // This code would also need to handle existing scenarios, such as:
+                // - Customer explicitly set credentials on their service client (i.e. new AmazonS3Client(new BasicAWSCredentials))
+            }
+        }
+
+        protected virtual List<IAuthSchemeOption> ServiceSpecificHandler(IRequestContext requestContext)
```
normj, line 68: Should this be abstract? Do we have a scenario where null would ever be valid to return?

dscpinheiro:
I'm closing this as the SRA is on hold due to other priorities. Will address the feedback once we start working on it again.

Issue #858, opened 2018-02-06
Exception in ErrorHandler.ProcessException()


xakep139:
If ErrorHandler get an exception of type `System.Exception` there will be a `System.ArgumentNullException` with message:
```
Value cannot be null.
Parameter name: key
```

Here's the stacktrace:
```
   at System.ThrowHelper.ThrowArgumentNullException(ExceptionArgument argument)
   at System.Collections.Generic.Dictionary`2.FindEntry(TKey key)
   at System.Collections.Generic.Dictionary`2.TryGetValue(TKey key, TValue& value)
   at Amazon.Runtime.Internal.ErrorHandler.ProcessException(IExecutionContext executionContext, Exception exception) in E:\JenkinsWorkspaces\v3-trebuchet-release\AWSDotNetPublic\sdk\src\Core\Amazon.Runtime\Pipeline\ErrorHandler\ErrorHandler.cs:line 200
   at Amazon.Runtime.Internal.ErrorHandler.<InvokeAsync>d__5`1.MoveNext() in E:\JenkinsWorkspaces\v3-trebuchet-release\AWSDotNetPublic\sdk\src\Core\Amazon.Runtime\Pipeline\ErrorHandler\ErrorHandler.cs:line 104
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Amazon.Runtime.Internal.CallbackHandler.<InvokeAsync>d__9`1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Amazon.Runtime.Internal.CredentialsRetriever.<InvokeAsync>d__7`1.MoveNext() in E:\JenkinsWorkspaces\v3-trebuchet-release\AWSDotNetPublic\sdk\src\Core\Amazon.Runtime\Pipeline\Handlers\CredentialsRetriever.cs:line 98
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[...]

PR #3398, opened 2024-07-23
Fixes the loop logic in ErrorHandler.ProcessException() and ErrorHandler.ProcessExceptionAsync() in case the initial exception is of type System.Exception.

ashishdhingra:
## Description
Fixes the loop logic in `ErrorHandler.ProcessException()` and `ErrorHandler.ProcessExceptionAsync()` in case the initial exception is of type `System.Exception`.

**Root Cause:**
In the below current code:
```C#
private bool ProcessException(IExecutionContext executionContext, Exception exception)
{
    // Log the exception
    this.Logger.Error(exception, "An exception of type {0} was handled in ErrorHandler.", exception.GetType().Name);
    executionContext.RequestContext.Metrics.AddProperty(Metric.Exception, exception);
    // Find the matching handler which can process the exception
    // Start by checking if there is a matching handler for the specific exception type,
    // if not check for handlers for it's base type till we find a match.
    var exceptionType = exception.GetType();
    do
    {
        IExceptionHandler exceptionHandler = null;
        if (this.ExceptionHandlers.TryGetValue(exceptionType, out exceptionHandler))
        {
            return exceptionHandler.Handle(executionContext, exception);
        }
        exceptionType = exceptionType.BaseType;

    } while (exceptionType != typeof(Exception));

    // No match found, rethrow the original exception.
    return true;
}
```
for exception of type `System.Exception`, the statement `exceptionType = exceptionType.BaseType;` returns:
- `System.Object` on 1st loop execution.
- `null` on 2nd loop execution. This fails at `if (this.ExceptionHandlers.TryGetValue(exceptionType, out exceptionHandler))` throwing error reported by customer.

## Motivation and Context
Issue #858

## Testing
Debugged locally. Dry-run completed successfully.

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Pipeline/ErrorHandler/ErrorHandler.cs
+++ b/sdk/src/Core/Amazon.Runtime/Pipeline/ErrorHandler/ErrorHandler.cs
@@ -112,7 +112,7 @@ namespace Amazon.Runtime.Internal
             // E.g. S3 GetLifecycleConfiguration, GetBucket policy and few other operations
             // return a 404 which is not returned back as an exception but as a empty response with 
             // error code.
-            if(executionContext.ResponseContext != null && executionContext.ResponseContext.Response != null)
+            if (executionContext.ResponseContext != null && executionContext.ResponseContext.Response != null)
             {
                 return executionContext.ResponseContext.Response as T;
             }
@@ -202,9 +202,9 @@ namespace Amazon.Runtime.Internal
                 {
                     return exceptionHandler.Handle(executionContext, exception);
                 }
-                exceptionType = exceptionType.BaseType;
 
-            } while (exceptionType != typeof(Exception));
+                exceptionType = exceptionType.BaseType;
+            } while (exceptionType != typeof(Exception) && exceptionType != typeof(object));
 
             // No match found, rethrow the original exception.
             return true;
@@ -242,9 +242,9 @@ namespace Amazon.Runtime.Internal
                 {
                     return await exceptionHandler.HandleAsync(executionContext, exception).ConfigureAwait(false);
                 }
-                exceptionType = exceptionType.BaseType;
 
-            } while (exceptionType != typeof(Exception));
+                exceptionType = exceptionType.BaseType;
+            } while (exceptionType != typeof(Exception) && exceptionType != typeof(object));
 
             // No match found, rethrow the original exception.
             return true;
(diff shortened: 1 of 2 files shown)


sdk/src/Core/Amazon.Runtime/Pipeline/ErrorHandler/ErrorHandler.cs
```diff
@@ -202,7 +202,11 @@ private bool ProcessException(IExecutionContext executionContext, Exception exce
                 {
                     return exceptionHandler.Handle(executionContext, exception);
                 }
-                exceptionType = exceptionType.BaseType;
+
+                if (exceptionType != typeof(Exception))
```
normj, line 206: I think this logic would be easier to follow if you replace this `if` statement and added ` && exceptionType != typeof(object)` into the while condition. Otherwise it looks weird to me to have the `exceptionType != typeof(Exception)` twice.

sdk/src/Core/Amazon.Runtime/Pipeline/ErrorHandler/ErrorHandler.cs
```diff
@@ -242,7 +246,11 @@ private async System.Threading.Tasks.Task<bool> ProcessExceptionAsync(IExecution
                 {
                     return await exceptionHandler.HandleAsync(executionContext, exception).ConfigureAwait(false);
                 }
-                exceptionType = exceptionType.BaseType;
+
+                if (exceptionType != typeof(Exception))
```
normj, line 250: Same comment as above.

normj (changes requested):

sdk/src/Core/Amazon.Runtime/Pipeline/ErrorHandler/ErrorHandler.cs
ashishdhingra: Done.

sdk/src/Core/Amazon.Runtime/Pipeline/ErrorHandler/ErrorHandler.cs
ashishdhingra: Done

dscpinheiro (approved):

peterrsongg (approved):

normj (approved):

PR #3406, opened 2024-07-29
Remove obsolete classes/ methods from core pt.2. Refactor

peterrsongg:
## Testing

I ran the core + custom unit test sln, which all passed. I built all the services + core in netstandard + netframework. I ran all the integ tests in `sdk/test/IntegrationTests` for both netstandard and netframework.

This PR mainly focuses on the removal of obsolete credential types and credential  management classes like ProfileManager.

Credential Types Removed
1. `ECSTaskCredential`
2. `StoredProfileAWSCredentials`
3. `StoredProfileCredentials`
4. `EnvironmentAWSCredentials`
5. `StoredProfileFederatedCredentials`
6. `StoredProfileSAMLCredentials`

This also removes `UseSigV4` from the request type. `SignatureVersion` should be used directly instead. 

`Parameters` property is removed from `WebServiceRequestEventArgs`, not be confused with `Parameters` on the request object.

## Types of changes

## Checklist

## License


normj (approved):

sdk/src/Core/Amazon.Runtime/Credentials/ECSTaskCredentials.cs
```diff
@@ -1,131 +0,0 @@
-﻿/*
```
dscpinheiro, line 1: For later, but these removals impact PowerShell too (I commented on this class because @afroz429 noticed it was used in https://github.com/aws/aws-tools-for-powershell/blob/fe860a680057dfcbccdb95cafd8dd1a361b3f0e0/modules/AWSPowerShell/Common/CredentialsArguments.cs#L212-L213 - but I wouldn't be surprised if there are others as well).

dscpinheiro (approved):

sdk/src/Core/Amazon.Runtime/Credentials/ECSTaskCredentials.cs
peterrsongg: okay, thanks for the callout

PR #3408, opened 2024-07-30
Split GetACL into GetBucketACL and GetObject ACL and the same for PutACL

peterrsongg:
## Description
`GetACL` and `PutACL` in the SDK is a combination of `GetBucketACL `and `PutBucketACL` and there is a customization made to rename the method and exclude the `GetObjectACL `methods. This PR obsoletes the existing `GetACL`/`PutACL `operations and adds the `GetBucketACL `and `GetObjectACL `/ `PutBucketACL `and `PutObjectACL `operations. It also adds the handwritten custom marshallers/request classes etc. and refactors any existing unit/integ test to use the new operations.

I put the GetACL and PutACL into a custom partial `AmazonS3Client` and did the same thing for `IAmazonS3`.

## Motivation and Context

1. Currently GetACL and PutACL do not read the contextParams from the model, so it is possible that in the future if these contextParams are used in the endpoint ruleset then we could resolve an incorrect endpoint. Splitting the operations ensures that the contextParams are used in the EndpointResolver.
2. Parity with other SDKs + maintanability.
## Testing

Dry run passes.
Added an integ test.
Also did a manual test and made sure that the old method is still callable.
## Types of changes

## Checklist

## License


muhammad-othman (commented):
This approach seems a bit complicated IMO. What about removing `GetACL` and `PutACL` from the generator and the models, and add them and add them manually?

peterrsongg:
> This approach seems a bit complicated IMO. What about removing `GetACL` and `PutACL` from the generator and the models, and add them and add them manually?

I agree that it is a bit complicated, but unfortunately it is just the way that customizations work in the generator, especially renames. All it does is point to the original json object in the model but renames the operation. We need to create a copy of this operation basically with the original name to be able to generate both GetACL and GetBucketACL in the client file.

I was thinking maybe we could use this functionality in the future? but you're probably right, it is better to manually add them, and delete this added complexity to the generator. I will see what the team thinks

sdk/test/Services/S3/IntegrationTests/PutObjectTests.cs
```diff
@@ -940,6 +940,96 @@ public void PutObject_StreamChecksumEnabled()
...
+
+                getObjectAclResponse.Grants.ForEach(grant =>
+                {
+                    Assert.IsTrue(grant.Grantee.URI == "http://acs.amazonaws.com/groups/global/AuthenticatedUsers");
+                    Assert.IsTrue(grant.Permission == S3Permission.READ);
+                });
+            }
+            finally 
+            {                 
+                AmazonS3Util.DeleteS3BucketWithObjects(Client, bucketName);
```
dscpinheiro, line 1029: I don't think this finally is correct, the `bucketName` variable here doesn't refer to the bucket you created but the `private static` variable with the same name: https://github.com/aws/aws-sdk-net/blob/e83c9f26e7fc2a45f3ce10a940ba64b8ffb7ff1d/sdk/test/Services/S3/IntegrationTests/PutObjectTests.cs#L33

I'd either use a different variable name in this method or update it not to create a new bucket at all (you'd need to confirm changing the ACLs doesn't impact the other tests here - as they all share the same bucket).

peterrsongg: thank you for pointing that out. I moved the variable outside the `try` block and renamed it. I also just ran the integ test and confirmed it is deleting the right bucket.

muhammad-othman:
Do we have to create this new custom S3 client? What about using the existing extension S3 client file instead?

peterrsongg:
> Do we have to create this new custom S3 client? What about using the existing extension S3 client file instead?

Nope, I just moved them to the extension file instead

muhammad-othman (approved):

generator/ServiceClientGeneratorLib/Operation.cs
```diff
@@ -66,6 +66,8 @@ public string Name
                 return ShapeName;
             }
         }
+
+
```
dscpinheiro, line 70: What are these extra lines for?

generator/ServiceClientGeneratorLib/Generators/Endpoints/EndpointResolver.tt
```diff
@@ -96,6 +96,10 @@ namespace <#=Config.Namespace#>.Internal
             }
 
 <#}#>
+<#
```
dscpinheiro, line 99: I wouldn't mention a potential V5. What about: `GetACL and PutACL are deprecated in V4 and may be removed in the future`?

sdk/src/Services/S3/Custom/Model/PutACLRequest.cs
```diff
@@ -271,6 +271,7 @@ namespace Amazon.S3.Model
     /// </para>
     ///  </li> </ul>
     /// </summary>
+    ///
```
dscpinheiro, line 274: Remove this extra line.

sdk/src/Services/S3/AWSSDK.S3.NetStandard.csproj
```diff
@@ -82,8 +82,6 @@
             <PrivateAssets>all</PrivateAssets>
         </PackageReference>
   </ItemGroup>
-  <ItemGroup>
```
dscpinheiro, line 85: This `<ItemGroup>` removal seems odd, I don't see any generator changes included in this PR that could've caused this.

sdk/src/Services/S3/Custom/_bcl/IAmazonS3.cs
```diff
@@ -0,0 +1,333 @@
...
+
+using Amazon.S3.Model;
+namespace Amazon.S3
+{
+#pragma warning disable CS1591
+    public partial interface IAmazonS3
+    {
+        #region GetACL
+        [Obsolete("GetACL combines both GetBucketAcl and GetObjectAcl and is deprecated. Please use the separated GetObjectAcl and GetBucketAcl operations.")]
+        Task<GetACLResponse> GetACLAsync(string bucketName, System.Threading.CancellationToken cancellationToken = default(CancellationToken));
```
dscpinheiro, line 15: This method is missing documentation. Also, since this is the `bcl` folder, shouldn't you also have to include the non-async methods?

generator/.DevConfigs/4f49cfc6-d4a8-477c-a685-6b96d52d73ab.json
```diff
@@ -0,0 +1,11 @@
+{
+    "services": [
+        {
+            "serviceName": "S3",
+            "type": "patch",
+            "changeLogMessages": [
+                "Split GetACL into GetBucketACL and GetObjectACL. Split PutACL into PutObjectACL and PutBucketACL. Obsolete PutACL and GetACL"
```
dscpinheiro, line 7: I'd expand this message a bit more (for example, if this goes into the V4 tracker issue, is there enough information for customers to understand why we're doing this?)

dscpinheiro (commented):
I have one general question too: Is there a reason why we're using lowercase for the operation names / request objects? For example, the changelog (and PR title) mention `GetACL into GetBucketACL and GetObjectACL` but then the classes are named `GetBucketAcl` and `GetObjectAcl`.

peterrsongg:
> I have one general question too: Is there a reason why we're using lowercase for the operation names / request objects? For example, the changelog (and PR title) mention `GetACL` into `GetBucketACL` and `GetObjectACL` but then the classes are named `GetBucketAcl` and `GetObjectAcl`.

Yeah so when we did the customization for `GetACL` and `PutACL` we renamed it to use capital ACL, but generating from the model the operation is generated as `GetAcl` and `PutAcl`. I think the name of the files should reflect what is modeled so I will change those to `Acl`

peterrsongg:
Addressed all comments and the PR is ready for another pass (dry-run also succeeded)

sdk/src/Services/S3/AWSSDK.S3.NetStandard.csproj
peterrsongg: this was a manual change I made on accident😅, rerunning the generator removed this.

sdk/src/Services/S3/Custom/_bcl/IAmazonS3.cs
peterrsongg: good call, i did a manual test and the non-async methods were callable so I thought I was in the clear, but based on the pattern the other methods follow it looks like I missed adding it here.

generator/.DevConfigs/4f49cfc6-d4a8-477c-a685-6b96d52d73ab.json
```diff
@@ -0,0 +1,11 @@
+{
+    "services": [
+        {
+            "serviceName": "S3",
+            "type": "patch",
+            "changeLogMessages": [
```
dscpinheiro, line 6: No need to change, but since this is an array you could also split this into multiple messages so it's easier to read. Like this:
```json
"changeLogMessages": [
    "Split `GetACL` into `GetBucketAcl` and `GetObjectAcl`",
    "Split `PutACL` into `PutObjectACL` and `PutBucketACL`",
    "Obsolete `PutACL` and `GetACL`: These operations are being split ..."
]
```

Issue #1448, opened 2019-11-07
DynamoDB OverrideTableName doesn't work


vickyrathee:
I am using `Table1` to override using `DynamoDBOperationConfig.OverrideTableName` but the SDK doesn't override the tableName and throwing error with class name `Documents `. So the override request not honored and getting error.

**Error : Requested resource not found: Table: Documents not found**

My requirement is to change the table name dynamically on business logic to insert documents in different tables. So I can't use the `[DynamoDBTable("Table1")]`

```
async Task Main()
{
	var document = new Documents {doc_id = "abc", name = "test", created = DateTime.UtcNow};
	
	string tableName = "Table1";
	var dynamo = new DynamoDbService(tableName);
	await dynamo.InsertDocumentAsync<Documents>(document);
	
}

public class DynamoDbService
{
	private static readonly string awsAccessKey = "access key here";
	private static readonly string awsSecretKey = "secret here";
	
	private static DynamoDBOperationConfig _operationConfig;

	private static BasicAWSCredentials awsCredentials = new BasicAWSCredentials(awsAccessKey, awsSecretKey);
	private static AmazonDynamoDBClient client = new AmazonDynamoDBClient(awsCredentials, RegionEndpoint.USEast1);
	private static DynamoDBContext context = new DynamoDBContext(client);

	public DynamoDbService(string tableName)
	{
		_operationConfig = new DynamoDBOperationConfig()
		{
			OverrideTableName = tableName
		};
	}
	
	public async Task InsertDocumentAsync<T>(T document)
	{
		await context.SaveAsync<T>(document);
	}

	public async Task<List<T>> GetDocumentsAsync<T>(string id, int limit = 10)
	{
		QueryFilter filter = new QueryFilter();
		filter.AddCondition("doc_id", QueryOperator.Equal, id);
		QueryOperationConfig queryConfig = new QueryOperationConfig
		{
			Filter = filter,
			Limit = limit,
			BackwardSearch = true
		};
		var documents = await context.FromQueryAsync<T>(queryConfig, _operationConfig).GetNextSetAsync();
		return documents;
	}
}

public class Documents
{
	public string doc_id { get; set; }
	public string name {get; set;}
	public DateTime created {get; set;}
}

```

![image](https://user-images.githubusercontent.com/6106479/68362871-ca388500-014e-11ea-9c21-a1d2fe7a12f3.png)

klaytaybai:
[...]

Issue #1421, opened 2019-10-07
DynamoDB dynamodb:DescribeTable Error when using OverrideTableName


ayongbannayan:
I'm using the object persistent model, the name of the DynamoDB table is also dynamic.  My queries  currently function however when using X-Ray I'm noticing an exception that is occurring after the query.

In my object I am not using the DynamoDBTable attribute because the table name is dynamic. I'm using the DynamoDBOperationConfig and setting the OverrideTableName property.

When excuting QueryAsync<T> the query returns results but X-Ray shows the following

![Screen Shot 2019-10-07 at 1 44 25 PM](https://user-images.githubusercontent.com/49486285/66335759-aee91880-e909-11e9-8769-cccc2f8d4271.png)

The error that is occurring is ...not authorized to perform: dynamodb:DescribeTable on resource: arn:aws:dynamodb:ca-central-1:xxxxxxxxx:table/T where T is the name of the object and not the table.
```
using (DynamoDBContext context = new DynamoDBContext(AmazonDynamoDB))
{
    DynamoDBOperationConfig config = new DynamoDBOperationConfig
    {
        OverrideTableName = “TableName”,
        IndexName = “IndexName”
    };

    AsyncSearch<ObjectType> query = context.QueryAsync<ObjectType>(“HashKey”, config);
    IEnumerable<ObjectType> objects = await query.GetRemainingAsync();
}
```
I'm not sure how to handle this exception. I've tried to set the DynamoDBTable however the error simply changes to the name I provide instead of the objects Name. Not sure what else I'm missing.

Regards
Artin

klaytaybai:
Hi @Artin-ST, thanks for reporting this.  I think it is root caused by a known bug.  I'll look into this more soon and add more details as I go along.

harshghorpade:
Hello, any update on this bug ? I can still reproduce this exact same issue with AWSSDK.DynamoDBv2 version 3.3.105.26 on .NET Core 3.1 Project.

AdalbertKhoolsaat:
Hey there, in the exact same manner as the OP I found out that I suffer from this bug as well.
At the company I work for we are considering using X-Ray (as part of CloudWatch ServiceLens), and we found this bug with the very first project we started monitoring (Using v3.3.105.34 on .NET Core 3.1).

Considering that the OPM is used quite generally around here , this issue is definitely pretty widespread.

ashishdhingra:
[...]

Issue #1218, opened 2019-02-27
OverrideTableName is not applied when passed in DynamoDBContext over DynamoDBOperationConfig


michaelmatveev:
## Expected Behavior
Suppose I create DynamoDBContext like follows
```
        private readonly string _tableName;
        private DynamoDBContext GetDbContext()
        {
            var dynamoDbOperationConfig = new DynamoDBOperationConfig
            {
                OverrideTableName = _tableName            
            };
            return new DynamoDBContext(_client, dynamoDbOperationConfig);            
        }
```
then I try to use context expecting to save an item with a class name different then a table name
```
       public async Task SaveItemAsync<T>(T item)
        {
            using (var context = GetDbContext())
            {
                await context.SaveAsync(item);
            }
        }
```
## Current Behavior
Get an error 
```
Result StackTrace:	
at Amazon.Runtime.Internal.HttpErrorResponseExceptionHandler.HandleException(IExecutionContext executionContext, HttpErrorResponseException exception) in D:\JenkinsWorkspaces\trebuchet-stage-release\AWSDotNetPublic\sdk\src\Core\Amazon.Runtime\Pipeline\ErrorHandler\HttpErrorResponseExceptionHandler.cs:line 60
   at Amazon.Runtime.Internal.ErrorHandler.ProcessException(IExecutionContext executionContext, Exception exception) in D:\JenkinsWorkspaces\trebuchet-stage-release\AWSDotNetPublic\sdk\src\Core\Amazon.Runtime\Pipeline\ErrorHandler\ErrorHandler.cs:line 212
   at Amazon.Runtime.Internal.ErrorHandler.InvokeSync(IExecutionContext executionContext) in D:\JenkinsWorkspaces\trebuchet-stage-release\AWSDotNetPublic\sdk\src\Core\Amazon.Runtime\Pipeline\ErrorHandler\ErrorHandler.cs:line 78
   at Amazon.Runtime.Internal.CallbackHandler.InvokeSync(IExecutionContext executionContext) in D:\JenkinsWorkspaces\trebuchet-stage-release\AWSDotNetPublic\sdk\src\Core\Amazon.Runtime\Pipeline\Handlers\CallbackHandler.cs:line 46
[...]

PR #3421, opened 2024-08-05
Feature/v4 ddb separate config

ashovlin:
## Description
Merging the feature branch into `v4-development`. This combines #3380, #3387 , #3383, and #3376 

## Motivation and Context
#1748 - `DynamoDBOperationConfig` is shared amongst the object-persistence methods, but has become overloaded over time, with some properties only applying to some operations.

This moves us towards new, operation-specific config objects. We've marked methods that take `DynamoDBOperationConfig` as deprecated in V4, but are not planning to remove them entirely yet. We may do so in a future major version.

See the description of #3376 for a longer explanation of which properties are relevant to which operations.

## Testing
PENDING DRY-RUN

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [X] My code follows the code style of this project
- [X] My change requires a change to the documentation
- [X] I have updated the documentation accordingly
- [X] I have read the **README** document
- [X] I have added tests to cover my changes
- [ ] All new and existing tests passed

## License

- [X] I confirm that this pull request can be released under the Apache 2 license

[issues]: https://github.com/aws/aws-sdk-net/issues
[license]: http://aws.amazon.com/apache2.0/
[cla]: http://en.wikipedia.org/wiki/Contributor_License_Agreement

PR #3422, opened 2024-08-05
Remove the Context/Operation Config Hierarchy, and Remove Table-level Configurations from DynamoDBOperationConfig

ashovlin:
## Description
1. ~This allows users to set `OverrideTableName` on the `DynamoDBContextConfig`, not just the `DynamoDBOperationConfig`~
   * During review we decided against this. 
3. ~While I was here I noticed that I didn't follow the hierarchy when adding `DisableFetchingTableMetadata` last year, so backfilled that too.~

New changes after PR feedback:

1. Separate the hierarchy between `DynamoDBContextConfig` and `DynamoDBOperationConfig`, to remove the appearance that you can pass an _operation config_ into the _context_. This came up several times with `OverrideTableName`.
2. Removed two table-level configurations from `DynamoDBOperationConfig` (and thus all of the new operation-specific config objects). These control how the context object caches table metadata, and should be specified there instead of at the operation level.

## Motivation and Context
We've gotten a few requests over the years for the ability to override the DynamoDB table name via the object-persistence context config, instead of needing to do so on every operation config.
* #1421 
*  #1448 
*  #1218

This is complicated by the fact that `DynamoDBOperationConfig` inherits from `DynamoDBContextConfig`. One could be passing an _operation_ config into the _context_ constructor, thinking that `OverrideTableName` would apply.
https://github.com/aws/aws-sdk-net/blob/2b0190e05c1787d2530d4c1a94beb3208b2b9f8e/sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs#L153

## Testing
1. Ran unit and integration tests in DynamoDBv2 sln locally.
3. Dry-run in progress.

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs
+++ b/sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs
@@ -48,8 +48,8 @@ namespace Amazon.DynamoDBv2.DataModel
     }
 
     /// <summary>
-    /// Configuration object for setting options on the DynamoDBContext.
-    /// and individual operations.
+    /// Configuration object for setting options on the <see cref="DynamoDBContext"/> that
+    /// will apply to all operations that use the context object.
     /// </summary>
 #if NET8_0_OR_GREATER
     [System.Diagnostics.CodeAnalysis.RequiresUnreferencedCode(Amazon.DynamoDBv2.Custom.Internal.InternalConstants.RequiresUnreferencedCodeMessage)]
@@ -80,6 +80,16 @@ namespace Amazon.DynamoDBv2.DataModel
         /// If property is not set, version checks are performed.
         /// </summary>
         public bool? SkipVersionCheck { get; set; }
+
+        /// <summary>
+        /// Indicates which DynamoDB table to use. This overrides the table specified 
+        /// by the <see cref="DynamoDBTableAttribute"/> on the .NET objects that you're saving or loading.
+        /// </summary>
+        /// <remarks>
+        /// If you specify this on <see cref="DynamoDBContextConfig"/>, then it will apply to all 
+        /// objects/tables used with that <see cref="DynamoDBContext"/> object unless overriden by an operation-specific config.
+        /// </remarks>
+        public string OverrideTableName { get; set; }
 
         /// <summary>
         /// Property that directs DynamoDBContext to prefix all table names
@@ -152,12 +162,6 @@ namespace Amazon.DynamoDBv2.DataModel
 #endif
     public class DynamoDBOperationConfig : DynamoDBContextConfig
     {
-        /// <summary>
-        /// Property that indicates the table to save an object to overriding the DynamoDBTable attribute 
-        /// declared for the type.
-        /// </summary>
-        public string OverrideTableName { get; set; }
-
         /// <summary>
         /// Property that indicates a query should traverse the index backward.
         /// If the property is false (or not set), traversal shall be forward.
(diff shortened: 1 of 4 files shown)


sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs
```diff
@@ -152,12 +162,6 @@ public DynamoDBContextConfig()
 #endif
     public class DynamoDBOperationConfig : DynamoDBContextConfig
```
ashovlin, line 163: Should we break this hierarchy while we're here?

We're discouraging the use of `DynamoDBOperationConfig` in V4 via `[Obsolete]` #3421, but this did lead to the confusion that this is addressing in the first place.

sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs
```diff
@@ -81,6 +81,16 @@ public DynamoDBContextConfig()
...
 
+        /// <summary>
+        /// Indicates which DynamoDB table to use. This overrides the table specified 
+        /// by the <see cref="DynamoDBTableAttribute"/> on the .NET objects that you're saving or loading.
+        /// </summary>
+        /// <remarks>
+        /// If you specify this on <see cref="DynamoDBContextConfig"/>, then it will apply to all 
+        /// objects/tables used with that <see cref="DynamoDBContext"/> object unless overriden by an operation-specific config.
+        /// </remarks>
+        public string OverrideTableName { get; set; }
```
normj, line 92: Having this property on the context's config doesn't make sense to me since the context works with multiple tables. I know you call that out in the doc that it replaces all table mappings. I'm thinking it was a failure in our response to the GitHub issues to not point out that could accomplish what they wanted using the `AWSConfigsDynamoDB.Context.TypeMappings` property. Here is how I configured one of my application to use the table that was created during deployment as part of CloudFormation with CloudFormation naming the table.

```
var tableName = System.Environment.GetEnvironmentVariable(ENV_TABLE_NAME_POLL_DEFINITION);
if (!string.IsNullOrEmpty(tableName))
{
    PollDefinitionTableName = tableName;
    AWSConfigsDynamoDB.Context.TypeMappings[typeof(PollDefinition)] = new Amazon.Util.TypeMapping(typeof(PollDefinition), tableName);
    Console.WriteLine($"Using table {tableName} for PollDefinition");
}
```

If we really want this capability on the context object I think we would have to make it a map of types to table names.

sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs
normj: I think it make sense to break the hierarchy. I assume we would copy over the properties from DynamoDBContextConfig to DynamoDBOperationConfig that made sense at the operation level.

normj (changes requested):

sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs
ashovlin: I wasn't too attached to this yet either, was mostly throwing it out there since there seems to be interest via the three issues. The only counterargument I can think of is if an application is only working with a single table, the current appraoch is simpler to override.

Should we just point users to `AWSConfigsDynamoDB.Context.TypeMappings`? and then break the hierarchy so it stops looking like you could be able to do this? Or maybe we can leave this PR out for a bit for feedback, I could go comment on the original issues.

normj: Yeah I think we should point users to `AWSConfigsDynamoDB.Context.TypeMappings`. You are right that the single property would be useful in a single table design but then confusing in the other case. I don't have a feeling how common single table design is but I don't believe it would be the overwhelming majority to warrant putting in this confusing property for non single table users.

normj: You could add a comment on the `OverrideTableName` for the operational config that if users want to make the override global they should use the `AWSConfigsDynamoDB.Context.TypeMappings`.

sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs
ashovlin: Done in https://github.com/aws/aws-sdk-net/pull/3422/commits/f478a1b33f0b6c2c4c12e525aaf0aaf25318d2c5

sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs
ashovlin: Done in https://github.com/aws/aws-sdk-net/pull/3422/commits/f478a1b33f0b6c2c4c12e525aaf0aaf25318d2c5.

Plus as long as we do the breaking change to separate the hierarchy in V4, if we did want to add a context-level table mapping, that'd be an additive change in the future.

sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs
```diff
@@ -150,17 +153,101 @@ public DynamoDBContextConfig()
...
+
+        /// <summary>
+        /// Indicates which DynamoDB table to use. This overrides the table specified 
+        /// by the <see cref="DynamoDBTableAttribute"/> on the .NET objects that you're saving or loading.
+        /// </summary>
+        /// <remarks>
+        /// If you want to specify this globally instead of for each operation, you can use 
+        /// the <see cref="TableAlias"/> or <see cref="TypeMapping"/> collections 
+        /// on <see cref="AWSConfigsDynamoDB.Context"/>.
+        /// </remarks>
```
ashovlin, line 183: New note the suggests an alternative to the operation-level override.

sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs
```diff
@@ -150,17 +153,101 @@ public DynamoDBContextConfig()
...
+        /// If true disables fetching table metadata automatically from DynamoDB. Table metadata must be 
+        /// defined by <see cref="DynamoDBAttribute"/> attributes and/or in <see cref = "AWSConfigsDynamoDB"/>.
+        /// </summary>
+        /// <remarks>
+        /// Setting this to true can avoid latency and thread starvation due to blocking asynchronous 
+        /// IAmazonDynamoDB.DescribeTable(string) calls that are used to populate the SDK's cache of 
+        /// table metadata. It requires that the table's index schema be accurately described via the above methods, 
+        /// otherwise exceptions may be thrown and/or the results of certain DynamoDB operations may change.
+        /// </remarks>
+        public bool? DisableFetchingTableMetadata { get; set; }
```
normj, line 238: Does this make sense at the operational level? Since only one operation invoke is going to do the cache load for a table would a user even now when the load is happening or not.

ashovlin: Maybe that's why I left it off originally... Agreed that you wouldn't need to set this past the first one.

I can remove it again, especially in parallel with #3430

sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs
```diff
@@ -150,17 +153,101 @@ public DynamoDBContextConfig()
...
+        /// The object persistence model API relies on an internal cache of the DynamoDB table's metadata to construct and validate 
+        /// requests. This controls how the cache key is derived, which influences when the SDK will call 
+        /// IAmazonDynamoDB.DescribeTable(string) internally to populate the cache.
+        /// </summary>
+        /// <remarks>
+        /// For <see cref="MetadataCachingMode.Default"/> the cache key will be a combination of the table name, credentials, region and service URL. 
+        /// For <see cref="MetadataCachingMode.TableNameOnly"/> the cache key will only consist of the table name. This reduces cache misses in contexts
+        /// where you are accessing tables with identical structure but using different credentials or endpoints (such as a multi-tenant application).
+        /// </remarks>
+        public MetadataCachingMode? MetadataCachingMode { get; set; }
```
normj, line 204: Like `DisableFetchingTableMetadata` this property doesn't make sense at the operational level.

ashovlin: Done in https://github.com/aws/aws-sdk-net/pull/3422/commits/295e52e6c8b2b2a6d0c4e025c3ac0bf453eb39b5

sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs
ashovlin: Done in https://github.com/aws/aws-sdk-net/pull/3422/commits/295e52e6c8b2b2a6d0c4e025c3ac0bf453eb39b5

normj (approved):

philasmar (approved):

Issue #3412, opened 2024-08-01
Avoid StringBuilder.AppendFormat


paulomorgado:
### Describe the feature

`Amazon.Runtime.Internal.Marshaller.SetUserAgentHeader` has this code:

```csharp
private static void SetUserAgentHeader(IRequestContext requestContext)
{
    var sb = new StringBuilder(requestContext.ClientConfig.UserAgent);

    var clientAppId = requestContext.ClientConfig.ClientAppId;
    if (!string.IsNullOrEmpty(clientAppId))
        sb.AppendFormat(" app/{0}", clientAppId);

    var retryMode = requestContext.ClientConfig.RetryMode.ToString().ToLower();
    sb.AppendFormat(" cfg/retry-mode#{0}", retryMode);

    sb.AppendFormat(" md/{0}", requestContext.IsAsync ? "ClientAsync" : "ClientSync");

    sb.AppendFormat(" cfg/init-coll#{0}", AWSConfigs.InitializeCollections ? "1" : "0");

    var userAgentAddition = requestContext.OriginalRequest.UserAgentAddition;
    if (!string.IsNullOrEmpty(userAgentAddition))
    {
        sb.AppendFormat(" {0}", userAgentAddition);
    }

    var userAgent = sb.ToString();

    userAgent = InternalSDKUtils.ReplaceInvalidUserAgentCharacters(userAgent);

    if (requestContext.ClientConfig.UseAlternateUserAgentHeader)
    {
        requestContext.Request.Headers[HeaderKeys.XAmzUserAgentHeader] = userAgent;
    }
    else
    {
        requestContext.Request.Headers[HeaderKeys.UserAgentHeader] = userAgent;
    }
}
```

There's no practical use of `StringBuilder.AppendFormat here.

### Use Case

Creating objects that do not need to be created, not only uses memory and CPU to create them, but also causes GC work.

Interpreting a format and formatting is unnecessary work when appending to a `StringBuilder`.

### Proposed Solution

Consider using:

```csharp
private static void SetUserAgentHeader(IRequestContext requestContext)
{
    var sb = new StringBuilder(requestContext.ClientConfig.UserAgent);

    var clientAppId = requestContext.ClientConfig.ClientAppId;
    if (!string.IsNullOrEmpty(clientAppId))
        sb.Append(" app/").Append(clientAppId);

    sb.Append(" cfg/retry-mode#}");
    foreach (var c in requestContext.ClientConfig.RetryMode.ToString())
[...]

PR #3423, opened 2024-08-05
Refactor User-Agent construction in Marshaller.cs

paulomorgado:
Updated `SetUserAgentHeader` to improve User-Agent string construction:
- Initialize `StringBuilder` with a capacity of 256.
- Append `UserAgent` after replacing invalid characters.
- Append `ClientAppId` after replacing invalid characters.
- Use `ToUserAgentHeaderString` for retry mode conversion.
- Append `IsAsync` and initialization config directly.
- Append `UserAgentAddition` after replacing invalid characters.
- Removed redundant call to `ReplaceInvalidUserAgentCharacters`.
- Added `ToUserAgentHeaderString` method for `RequestRetryMode`.

## Description
See #3412

## Motivation and Context
See #3412

## Testing
There should be no changes to the result.

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/Marshaller.cs
+++ b/sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/Marshaller.cs
@@ -15,6 +15,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Globalization;
 using System.Text;
 using Amazon.Runtime.Telemetry;
@@ -114,11 +115,13 @@ namespace Amazon.Runtime.Internal
             if (!string.IsNullOrEmpty(clientAppId))
                 sb.Append(" app/").Append(InternalSDKUtils.ReplaceInvalidUserAgentCharacters(clientAppId));
 
-            sb.Append(" cfg/retry-mode#}").Append(ToUserAgentHeaderString(requestContext.ClientConfig.RetryMode));
+            var retryMode = ToUserAgentHeaderString(requestContext.ClientConfig.RetryMode);
+            Debug.Assert(retryMode != requestContext.ClientConfig.RetryMode.ToString().ToLower(), "Invalid RetryMode string.");
+            sb.Append(" cfg/retry-mode#}").Append(retryMode);
 
-            sb.Append(" md/").Append(IsAsync ? "ClientAsync" : "ClientSync");
+            sb.Append(" md/").Append(requestContext.IsAsync ? "ClientAsync" : "ClientSync");
 
-            sb.Append(" cfg/init-coll#").Append(InitializeCollections ? '1' : '0');
+            sb.Append(" cfg/init-coll#").Append(AWSConfigs.InitializeCollections ? '1' : '0');
 
             var userAgentAddition = requestContext.OriginalRequest.UserAgentAddition;
             if (!string.IsNullOrEmpty(userAgentAddition))
--- a/sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/Marshaller.cs
+++ b/sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/Marshaller.cs
@@ -115,9 +115,7 @@ namespace Amazon.Runtime.Internal
             if (!string.IsNullOrEmpty(clientAppId))
                 sb.Append(" app/").Append(InternalSDKUtils.ReplaceInvalidUserAgentCharacters(clientAppId));
 
-            var retryMode = ToUserAgentHeaderString(requestContext.ClientConfig.RetryMode);
-            Debug.Assert(retryMode != requestContext.ClientConfig.RetryMode.ToString().ToLower(), "Invalid RetryMode string.");
-            sb.Append(" cfg/retry-mode#}").Append(retryMode);
+            sb.Append(" cfg/retry-mode#}").Append(ToUserAgentHeaderString(requestContext.ClientConfig.RetryMode));
 
             sb.Append(" md/").Append(requestContext.IsAsync ? "ClientAsync" : "ClientSync");
 
@@ -143,10 +141,15 @@ namespace Amazon.Runtime.Internal
 
         private static string ToUserAgentHeaderString(RequestRetryMode requestRetryMode)
         {
-            if (requestRetryMode == RequestRetryMode.Adaptive)
-                return "adaptive";
-            else
-                return "standard";
+            switch (requestRetryMode)
+            {
+                case RequestRetryMode.Standard:
+                    return "standard";
+                case RequestRetryMode.Adaptive:
+                    return "adaptive";
+                default:
+                    return requestRetryMode.ToString().ToLowerInvariant();
+            }
         }
     }
 }
--- a/sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/Marshaller.cs
+++ b/sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/Marshaller.cs
@@ -106,28 +106,27 @@ namespace Amazon.Runtime.Internal
 
         private static void SetUserAgentHeader(IRequestContext requestContext)
         {
-            var sb = new StringBuilder(requestContext.ClientConfig.UserAgent);
+            var sb = new StringBuilder(256);
+
+            sb.Append(InternalSDKUtils.ReplaceInvalidUserAgentCharacters(requestContext.ClientConfig.UserAgent));
 
             var clientAppId = requestContext.ClientConfig.ClientAppId;
             if (!string.IsNullOrEmpty(clientAppId))
-                sb.AppendFormat(" app/{0}", clientAppId);
+                sb.Append(" app/").Append(InternalSDKUtils.ReplaceInvalidUserAgentCharacters(clientAppId));
 
-            var retryMode = requestContext.ClientConfig.RetryMode.ToString().ToLower();
-            sb.AppendFormat(" cfg/retry-mode#{0}", retryMode);
+            sb.Append(" cfg/retry-mode#}").Append(ToUserAgentHeaderString(requestContext.ClientConfig.RetryMode));
 
-            sb.AppendFormat(" md/{0}", requestContext.IsAsync ? "ClientAsync" : "ClientSync");
+            sb.Append(" md/").Append(IsAsync ? "ClientAsync" : "ClientSync");
 
-            sb.AppendFormat(" cfg/init-coll#{0}", AWSConfigs.InitializeCollections ? "1" : "0");
+            sb.Append(" cfg/init-coll#").Append(InitializeCollections ? '1' : '0');
 
             var userAgentAddition = requestContext.OriginalRequest.UserAgentAddition;
             if (!string.IsNullOrEmpty(userAgentAddition))
             {
-                sb.AppendFormat(" {0}", userAgentAddition);
+                sb.Append(' ').Append(InternalSDKUtils.ReplaceInvalidUserAgentCharacters(userAgentAddition));
             }
 
             var userAgent = sb.ToString();
-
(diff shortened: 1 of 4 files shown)


sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/Marshaller.cs
```diff
@@ -138,5 +137,13 @@ private static void SetUserAgentHeader(IRequestContext requestContext)
                 requestContext.Request.Headers[HeaderKeys.UserAgentHeader] = userAgent;
             }
         }
+
+        private static string ToUserAgentHeaderString(RequestRetryMode requestRetryMode)
+        {
+            if (requestRetryMode == RequestRetryMode.Adaptive)
```
muhammad-othman, line 143: This change seems a bit limited since we have `Legacy` mode and we may add other retry modes

muhammad-othman (changes requested):

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/Marshaller.cs
paulomorgado: I'm sorry, @muhammad-othman, but I don't understand your point.

`RequestRetryMode` has only 2 values: `Adaptive` and `Standard`. There's no `Legacy` value.

This code can be generated using something like [NetEscapades.EnumGenerators](https://andrewlock.net/netescapades-enumgenerators-a-source-generator-for-enum-performance/). This might me a bit more specific, but nothing that cannot be easily done. I just didn't want to taggle that in this PR.

paulomorgado: I've added a `Debug.Assert` to assert the value is the expected one, when in DEBUG.

paulomorgado: You need something that catches that at build-time, not at run-time.

Having `Debug.Assert` catches that at debug-time.

Comprehensive unit tests will also catch this kind of errors.

I don't mind pushing this to some utility class adding a comprehensive of unit tests, if you point me in the right direction.

muhammad-othman: I was just thinking that `Debug.Assert` might be lost and someone adding a new retry mode will run atleast one request to test their changes, but I agree unit tests are much better.

You can use this utility method to make sure that the enum didn't change with a note to update `ToUserAgentHeaderString` if it changed.
https://github.com/aws/aws-sdk-net/blob/v4-development/sdk/test/Common/Utils/AssertExtensions.cs#L137
And here is an example of its usage 
https://github.com/aws/aws-sdk-net/blob/8e74d19cbc090d66def761fd5ae9bf01d76e6329/sdk/test/Services/SecurityToken/UnitTests/Custom/StaticCheckTests.cs#L44


And you can add the tests here
https://github.com/aws/aws-sdk-net/blob/v4-development/sdk/test/UnitTests/Custom/Runtime

paulomorgado: @muhammad-othman,

> And you can add the tests here https://github.com/aws/aws-sdk-net/blob/v4-development/sdk/test/UnitTests/Custom/Runtime

What solution would that be? There are 405 `.sln` files in the repo.

muhammad-othman: Basically any service solution will include the core UnitTestUtilities porject, here is an example from S3

image

Please include the newly add test file the following projects as we don't include all the files added to the Runtime folder by default
```
sdk\test\UnitTests\Custom\AWSSDK.UnitTestUtilities.NetFramework.csproj
sdk\test\UnitTests\Custom\AWSSDK.UnitTestUtilities.NetStandard.csproj
```

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/Marshaller.cs
```diff
@@ -106,29 +107,30 @@ private static void SetRecursionDetectionHeader(IDictionary<string, string> head
...
 
             var clientAppId = requestContext.ClientConfig.ClientAppId;
             if (!string.IsNullOrEmpty(clientAppId))
-                sb.AppendFormat(" app/{0}", clientAppId);
+                sb.Append(" app/").Append(InternalSDKUtils.ReplaceInvalidUserAgentCharacters(clientAppId));
 
-            var retryMode = requestContext.ClientConfig.RetryMode.ToString().ToLower();
-            sb.AppendFormat(" cfg/retry-mode#{0}", retryMode);
+            var retryMode = ToUserAgentHeaderString(requestContext.ClientConfig.RetryMode);
+            Debug.Assert(retryMode != requestContext.ClientConfig.RetryMode.ToString().ToLower(), "Invalid RetryMode string.");
```
normj, line 119: I think doing the whole `requestContext.ClientConfig.RetryMode.ToString().ToLower()` defeats the purpose of having the `ToUserAgentHeaderString` method to remove allocation. The `requestContext.ClientConfig.RetryMode.ToString().ToLower()` is still going to be evaluated in `release` mode. The `Debug.Assert` is always called whether in release or debug mode it is just the evaluated Boolean past into the `Debug.Assert` won't trigger an assert in `release` mode.

normj (changes requested):

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/Marshaller.cs
paulomorgado: Have you checked, @normj?

Invocations of methods decorated with `ConditionalAttribute` should not have IL generated if the condition is false.

Check the difference between [this](https://sharplab.io/#v2:C4LgTgrgdgNAJiA1AHwAICYCMBYAUKgBgAJVMA6AEQEsBDAcygHsBnYKgY2YG488A3GmCJUiAXiIEeuPBQCmAIwh0yAQWbNZYYAApEiEQD5xBAJRS8pAJzaqZvEA) and [this](https://sharplab.io/#v2:C4LgTgrgdgPgAgJgIwFgBQcAMACOSB0AIgJYCGA5lAPYDOwxAxjQNzroBupY2x2AvNkys06QgFMARhHL4AgjRpiwwABQBqNbwB8AzAEph6PAE4VxA+iA).

normj: @paulomorgado learn something new everyday. Thanks for pointing out my misunderstanding.

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/Marshaller.cs
paulomorgado: After digging a little bit into the code, I realized that there is no validation when `IRequestContext.ClientConfig.RetryMode`, so I changed `ToUserAgentHeaderString` to have a fast path for know values and a fallback of `requestRetryMode.ToString().ToLowerInvariant()`:

```csharp
private static string ToUserAgentHeaderString(RequestRetryMode requestRetryMode)
{
    switch (requestRetryMode)
    {
        case RequestRetryMode.Standard:
            return "standard";
        case RequestRetryMode.Adaptive:
            return "adaptive";
        default:
            return requestRetryMode.ToString().ToLowerInvariant();
    }
}
```

Added this to sdk/test/Services/SecurityToken/UnitTests/Custom/StaticCheckTests.cs:

```csharp
[TestMethod]
public void LookForRequestRetryModeChanges()
{
    var expectedHash = "2FC699DB6A4284C5D53F3B7FA4E64165576FD1F6AE458BC737AE503F06DFBE4C";
    AssertExtensions.AssertEnumUnchanged(
        typeof(RequestRetryMode),
        expectedHash,
        "The Amazon.Runtime.Internal.Marshaller.ToUserAgentHeaderString method implementation may need to be updated.");
}
```

normj (approved):

muhammad-othman (approved):

muhammad-othman:
Thank you for making the requested changes.
Running this PR through our internal dry run system and will merge it after that.

paulomorgado:
@normj, @muhammad-othman,

I noticed that `RequestRetryMode` is changing from:

```csharp
/// <summary>
/// RetryMode determines which request retry mode is used for requests that do 
/// not complete successfully.
/// </summary>
public enum RequestRetryMode
{
    /// <summary>
    /// Legacy request retry strategy.
    /// </summary>
    Legacy,
    /// <summary>
    /// Standardized request retry strategy that is consistent across all SDKs.
    /// </summary>
    Standard,
    /// <summary>
    /// An experimental request retry strategy that builds on the Standard strategy
    /// and introduces congestion control through client side rate limiting.
    /// </summary>
    Adaptive
}
```

```csharp
/// <summary>
/// RetryMode determines which request retry mode is used for requests that do 
/// not complete successfully.
/// </summary>
public enum RequestRetryMode
{
    /// <summary>
    /// Standardized request retry strategy that is consistent across all SDKs.
    /// </summary>
    Standard,
    /// <summary>
    /// An experimental request retry strategy that builds on the Standard strategy
    /// and introduces congestion control through client side rate limiting.
    /// </summary>
    Adaptive
}
```

This might be an issue for anyone saving these settings as integers because it will be changing from:

`Legacy` = `0`
`Standard` = `1`
`Adaptive` = `2`

to:

`Standard` = `0`
`Adaptive` = `1`

So, `Legacy` becomes `Standard`, `Standard` becomes `Adaptive` and `Adaptive` becomes invalid.

normj:
For the way this enum is used I don't see anybody relying on the integer value. It it would be set via the property on the config or via an environment variable which uses the string value. But it wouldn't be bad thing to assign `standard` and `adaptive` the same integer value as v3 in case there is some scenario I can't think of.

muhammad-othman:
@paulomorgado  Our dry run failed because there conflicts between this branch and the v4-development, can you check these conflicts?

paulomorgado:
Hi @muhammad-othman,

> @paulomorgado Our dry run failed because there conflicts between this branch and the v4-development, can you check these conflicts?

I don't know what those conflicts are, but I rebased on the latest commit of `v4-development`.

paulomorgado:
Hi @normj,

> For the way this enum is used I don't see anybody relying on the integer value. It it would be set via the property on the config or via an environment variable which uses the string value. But it wouldn't be bad thing to assign `standard` and `adaptive` the same integer value as v3 in case there is some scenario I can't think of.

I'm using here an in-memory collection as a simple example, but it can be any configuration source:

```csharp
var configuration = new ConfigurationBuilder()
    .AddInMemoryCollection(new Dictionary<string, string?>
    {
        ["Amazon:KeyManagementService:RetryMode"] = "2",
    })
    .Build();

var keyManagmentServiceConfig = new AmazonKeyManagementServiceConfig();
configuration.GetSection("Amazon:KeyManagementService").Bind(keyManagmentServiceConfig);

Console.WriteLine($"The retry mode is: {keyManagmentServiceConfig.RetryMode}");

// The retry mode is: Adaptive
```

The value of an `enum` is always a scalar value: `byte`, `short`, `int`, `long`. Adding to that, the literal `0` is implicitly convertible to any `enum` type.

muhammad-othman (approved):

muhammad-othman:
Hi @paulomorgado,
I believe the `RequestRetryMode` discussion doesn't have to be resolved in this PR and this PR is ready to be merged. Do you have any changes you want to add before merging?

paulomorgado:
> Hi @paulomorgado, I believe the `RequestRetryMode` discussion doesn't have to be resolved in this PR and this PR is ready to be merged. Do you have any changes you want to add before merging?

That discussion was a general heads up on the issue.

I don't have anything to add to this PR.

PR #3427, opened 2024-08-06
fix: Update DownloadDirectoryCommand to handle non existent directories gracefully

muhammad-othman:
## Description
Update `DownloadDirectoryCommand` to handle non existent directories gracefully

With the V4 change of no initializing the collections by default, `DownloadDirectoryCommand` throws null reference exception when trying to download a non existent directory since `listResponse.S3Objects` collection is set to null.
This PR fixes that by checking if `listResponse.S3Objects` is not null before enumeration.

## Motivation and Context
#DOTNET-7682

## Testing
* Ran S3 unit and integration tests locally.
* Added `DownloadNonExistentS3Directory` test to validate this fix.
* `DRY_RUN-97fefe47-1e53-4477-bcdb-33d5013e92d3`

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Util/PaginatedResourceFactory.cs
+++ b/sdk/src/Core/Amazon.Util/PaginatedResourceFactory.cs
@@ -70,6 +70,9 @@ namespace Amazon.Util
                     nextToken = GetPropertyValueFromPath<string>(nextSet, tokenResponsePropertyPath);
                     currentItems = GetPropertyValueFromPath<List<ItemType>>(nextSet, itemListPropertyPath);
 
+                    if (currentItems == null)
+                        currentItems = new List<ItemType>();
+
                     return new Marker<ItemType>(currentItems, nextToken);
                 };
             return new PaginatedResource<ItemType>(fetcher);
--- a/sdk/src/Services/S3/Custom/Transfer/Internal/_bcl+netstandard/DownloadDirectoryCommand.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/Internal/_bcl+netstandard/DownloadDirectoryCommand.cs
@@ -127,12 +127,16 @@ namespace Amazon.S3.Transfer.Internal
             {
                 ListObjectsResponse listResponse = await this._s3Client.ListObjectsAsync(listRequest, cancellationToken)
                     .ConfigureAwait(continueOnCapturedContext: false);
-                foreach (S3Object s3o in listResponse.S3Objects)
+
+                if (listResponse.S3Objects != null) 
                 {
-                    if (ShouldDownload(s3o))
+                    foreach (S3Object s3o in listResponse.S3Objects)
                     {
-                        this._totalBytes += s3o.Size.GetValueOrDefault();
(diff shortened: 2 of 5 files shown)


afroz429 (approved):

dscpinheiro:
I don't think we have unit tests, but doesn't this also impact the `S3.IO` classes (https://github.com/aws/aws-sdk-net/tree/main/sdk/src/Services/S3/Custom/_bcl/IO)? There are calls to `ListObjects` and `ListBuckets` there too.

muhammad-othman:
> I don't think we have unit tests, but doesn't this also impact the `S3.IO` classes (https://github.com/aws/aws-sdk-net/tree/main/sdk/src/Services/S3/Custom/_bcl/IO)? There are calls to `ListObjects` and `ListBuckets` there too.

Looked at `S3.IO` classes and it looks like they handle null lists properly, however I've found an issue in `GetAllObjectKeys` which I fixed in  https://github.com/aws/aws-sdk-net/pull/3427/commits/6b55776827d392602a5e82879166f8cb8c4e727f.

dscpinheiro:
What about the `PaginatedResourceFactory` (e.g. in https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Services/S3/Custom/_bcl/IO/S3DirectoryInfo.cs#L277)? Does it handle null collections in the response properly?

normj (approved):

afroz429 (approved):

muhammad-othman:
> What about the `PaginatedResourceFactory` (e.g. in https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Services/S3/Custom/_bcl/IO/S3DirectoryInfo.cs#L277)? Does it handle null collections in the response properly?

Nope it didn't, I had make this change to fix it https://github.com/aws/aws-sdk-net/compare/6b55776827d392602a5e82879166f8cb8c4e727f..f289bbbd39a792ab73f03717d04accf3ae1be7f5 
I'm not sure if this is the right approach though since we are trying to eliminate the empty lists creation, but the problem here is that by default we return an enumerator that isn't null but the underline list is null, which will require many changes in the `PaginatedResource` to fix and I might miss some corner cases, also the `PaginatedResource` isn't widely used in the SDK anyway.

normj:
> Nope it didn't, I had make this change to fix it https://github.com/aws/aws-sdk-net/compare/6b55776827d392602a5e82879166f8cb8c4e727f..f289bbbd39a792ab73f03717d04accf3ae1be7f5 I'm not sure if this is the right approach though since we are trying to eliminate the empty lists creation, but the problem here is that by default we return an enumerator that isn't null but the underline list is null, which will require many changes in the `PaginatedResource` to fix and I might miss some corner cases, also the `PaginatedResource` isn't widely used in the SDK anyway.

I think your change makes sense. It is only in creating the empty collection as part of creating the paginator. I expect users of paginators to be wanting to simple iterate through the response not use paginators to check for existence or lack of any existence of resources.

normj (approved):

PR #3430, opened 2024-08-07
feat: add a DynamoDBContextBuilder to construct a DynamoDBContext

philasmar:
To build the low level DynamoDB requests from both high-level programming models, the SDK requires knowledge of the table and index key structure. It generally relies on an internal call to `DescribeTable`. This call may invoked from synchronous public APIs, which relies on the sync-over-async antipattern for versions of .NET that only have an async HTTP client. This can lead to performance issues and deadlocks.

In October 2023 we released new initializers for users to provide the table and index structure on the client and skip the `DescribeTable` call.

* Design: https://quip-amazon.com/H6OSA2nu5EQx/DynamoDB-Threadpool-Deadlock-Current-Library
* Launch blog post: https://aws.amazon.com/blogs/developer/improved-dynamodb-initialization-patterns-for-the-aws-sdk-for-net/

## Description
The aim of this PR is to mark the methods that reply on sync-over-async calls as `[Obsolete]` in order to encourage users to use the new `TableBuilder` for the Document model, as well as the new `DynamoDBContextBuilder`, which is introduced in this PR, for the object-persistence model.

For the document model, this PR obsoletes all the `LoadTable` and `TryLoadTable` calls.
For the object-persistence model, this PR adds a new `DynamoDBContextBuilder` which constructs a `DynamoDBContext` that sets `DisableFetchingTableMetadata` to `true` by default.

## Motivation and Context
This change is required to mode customers away from sync-over-async calls which have lead to performance issues and deadlocks in the past.

## Testing
For testing, I have added new tests that use the new `DynamoDBContextBuilder` to create the `DynamoDBContext` and I left some tests that create the `DynamoDBContext` using the old way.
I have ran all the tests locally and they have passed.

## Types of changes

## Checklist

## License


philasmar:
> I would like to see unit test coverage for the different methods that build the context based upon a provided client, region, and configuration. Are these somewhere that I missed?

The constructed `DynamoDBContext` does not expose any of the underlying objects that were configured by the builder, so there's nothing for me to assert on. There is a unit test that asserts an exception is thrown when 2 methods are called together, but that is the extent of it.

sdk/src/Services/DynamoDBv2/Custom/DataModel/IDynamoDBContextBuilder.cs
```diff
@@ -0,0 +1,49 @@
...
+
+namespace Amazon.DynamoDBv2.DataModel
+{
+    /// <summary>
+    /// Interface for a builder that constructs a <see cref="DynamoDBContext"/>
+    /// </summary>
+    public interface IDynamoDBContextBuilder : IDisposable
+    {
+        /// <summary>
+        /// Sets the DynamoDB client to be used by the <see cref="DynamoDBContext"/> that is constructed
```
normj, line 26: Add a comment that if a client is not provided then a client will be created for users using the environmental to search for credentials and region configuration.

sdk/src/Services/DynamoDBv2/Custom/DataModel/IDynamoDBContextBuilder.cs
```diff
@@ -0,0 +1,49 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+using System;
+
+namespace Amazon.DynamoDBv2.DataModel
+{
+    /// <summary>
+    /// Interface for a builder that constructs a <see cref="DynamoDBContext"/>
```
normj, line 21: Add would suggest giving more context for users upgrading about how this newer approach disables the `DescribeTable` mechanism. And explain how the can use the `ConfigureContext` method to revert to pervious behavior if they need it.

sdk/src/Services/DynamoDBv2/Custom/DataModel/IDynamoDBContextBuilder.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// Configures the <see cref="DynamoDBContext"/> that is being constructed
+        /// </summary>
+        /// <param name="configure">The configuration applied to the constructed <see cref="DynamoDBContext"/></param>
+        IDynamoDBContextBuilder ConfigureContext(Action<DynamoDBContextConfig> configure);
+
+        /// <summary>
+        /// Call at the end to retrieve the new <see cref="DynamoDBContext"/>
+        /// </summary>
+        /// <returns>Built <see cref="DynamoDBContext"/></returns>
+        DynamoDBContext Build();
```
normj, line 47: Shouldn't this be returning back the `IDynamoDBContext`?

sdk/src/Services/DynamoDBv2/Custom/DataModel/IDynamoDBContextBuilder.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// <summary>
+    /// Interface for a builder that constructs a <see cref="DynamoDBContext"/>
+    /// </summary>
+    public interface IDynamoDBContextBuilder : IDisposable
+    {
+        /// <summary>
+        /// Sets the DynamoDB client to be used by the <see cref="DynamoDBContext"/> that is constructed
+        /// </summary>
+        /// <param name="client"><see cref="IAmazonDynamoDB"/> to use to access DynamoDB.</param>
+        IDynamoDBContextBuilder SetDynamoDBClient(IAmazonDynamoDB client);
```
normj, line 29: Instead of these 2 set methods for client and region it might be better to have method to add a client factory method. Having to methods makes me think I could set both which would be undefined behavior especially of the region being set is different then the client past in. Also if we have these 2 do we need to add a third for setting the Client config.

Something like:
```
IDynamoDBContextBuilder WithDynamoDBClient(Func<IAmazonDynamoDBClient> factory);
```

If they don't ever call this then you use the internal factory of creating a client like we do today when a client is not provided. Call that out in the docs.

sdk/src/Services/DynamoDBv2/Custom/DataModel/IDynamoDBContextBuilder.cs
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+using System;
+
+namespace Amazon.DynamoDBv2.DataModel
+{
+    /// <summary>
+    /// Interface for a builder that constructs a <see cref="DynamoDBContext"/>
+    /// </summary>
+    public interface IDynamoDBContextBuilder : IDisposable
```
normj, line 23: Style preference: When creating a fluent interface where each method returns back the instance of the builder the common verb names I see for methods are `Add`, `With` and `Configure`. Using `Set` implies to me the return is `void` and I don't think I can use fluent style. I would recommend for your `Set` methods use `With`.

sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextBuilder.cs
```diff
@@ -0,0 +1,153 @@
...
+        /// <summary>
+        /// Implements the Dispose pattern
+        /// </summary>
+        /// <param name="disposing">Whether this object is being disposed via a call to Dispose
+        /// or garbage collected.</param>
+        protected virtual void Dispose(bool disposing)
+        {
+            if (_disposed) return;
+
+            if (disposing && _context != null)
```
normj, line 134: I don't understand why we are disposing the context here. The builder gave the context back to the user and it is up to the user to decide when that should be disposed. Generally a builder is used once during the init of an application and the context is left around for the life time of the application. 

The builder should not be implementing `IDisposable` and it doesn't need a `_context` member variable. The context is a local variable creating in the `Build` method and returned to the user.

normj (changes requested):

sdk/src/Services/DynamoDBv2/Custom/DataModel/IDynamoDBContextBuilder.cs
philasmar: Added more docs in the interface and inheritor.

sdk/src/Services/DynamoDBv2/Custom/DataModel/IDynamoDBContextBuilder.cs
philasmar: We are expected to return a `DynamoDBContext` if we are replacing the old approach of using the `DynamoDBContext` constructor. This is also similar to what the `TableBuilder` is  https://github.com/aws/aws-sdk-net/blob/97320c917081225082ffb8b67f6c43e726c47dfb/sdk/src/Services/DynamoDBv2/Custom/DocumentModel/ITableBuilder.cs#L27

sdk/src/Services/DynamoDBv2/Custom/DataModel/IDynamoDBContextBuilder.cs
philasmar: Removed the 2 set methods, but I added this comment to the new `WithDynamoDBClient`

sdk/src/Services/DynamoDBv2/Custom/DataModel/IDynamoDBContextBuilder.cs
philasmar: Understood, thx

sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextBuilder.cs
philasmar: I needed to implement `IDisposable` because I had `DynamoDBContext` and `IAmazonDBClient` as global members which required me to dispose them, but as you mentioned, I didn't need to be doing that.

sdk/src/Services/DynamoDBv2/Custom/DataModel/IDynamoDBContextBuilder.cs
normj: Ok I agree we should follow the pattern we already setup with `ITableBuilder`. I suppose there isn't much reason to mock the builder and you would probably just mock the instance of `IDynamoDBContext`.

sdk/src/Services/DynamoDBv2/Custom/DataModel/IDynamoDBContextBuilder.cs
normj: Add a line that says alternatively to attributes on the class you can register the table definition using the `RegisterTableDefinition` method on the context object.

sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextBuilder.cs
```diff
@@ -0,0 +1,96 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+using System;
+
+namespace Amazon.DynamoDBv2.DataModel
+{
+    /// <summary>
```
normj, line 20: Do you want to use `<inheritdoc/>` instead of duplicating this large comment block?

normj (changes requested):

sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextBuilder.cs
philasmar: Done

sdk/src/Services/DynamoDBv2/Custom/DataModel/IDynamoDBContextBuilder.cs
philasmar: Done

cklim0815is:
> To build the low level DynamoDB requests from both high-level programming models, the SDK requires knowledge of the table and index key structure. It generally relies on an internal call to `DescribeTable`. This call may invoked from synchronous public APIs, which relies on the sync-over-async antipattern for versions of .NET that only have an async HTTP client. This can lead to performance issues and deadlocks.
> 
> In October 2023 we released new initializers for users to provide the table and index structure on the client and skip the `DescribeTable` call.
> 
> * Design: https://quip-amazon.com/H6OSA2nu5EQx/DynamoDB-Threadpool-Deadlock-Current-Library
> * Launch blog post: https://aws.amazon.com/blogs/developer/improved-dynamodb-initialization-patterns-for-the-aws-sdk-for-net/
> 
> ## Description
> The aim of this PR is to mark the methods that reply on sync-over-async calls as `[Obsolete]` in order to encourage users to use the new `TableBuilder` for the Document model, as well as the new `DynamoDBContextBuilder`, which is introduced in this PR, for the object-persistence model.
> 
> For the document model, this PR obsoletes all the `LoadTable` and `TryLoadTable` calls.
> For the object-persistence model, this PR adds a new `DynamoDBContextBuilder` which constructs a `DynamoDBContext` that sets `DisableFetchingTableMetadata` to `true` by default.
> 
> ## Motivation and Context
> This change is required to mode customers away from sync-over-async calls which have lead to performance issues and deadlocks in the past.
> 
> ## Testing
> For testing, I have added new tests that use the new `DynamoDBContextBuilder` to create the `DynamoDBContext` and I left some tests that create the `DynamoDBContext` using the old way.
> I have ran all the tests locally and they have passed.
> 
> ## Types of changes
> 
> - [ ] Bug fix (non-breaking change which fixes an issue)
> - [ ] New feature (non-breaking change which adds functionality)
> - [x] Breaking change (fix or feature that would cause existing functionality to change)
> 
> ## Checklist
> 
> 
> - [x] My code follows the code style of this project
> - [x] My change requires a change to the documentation
> - [ ] I have updated the documentation accordingly
> - [ ] I have read the **README** document
> - [x] I have added tests to cover my changes
> - [x] All new and existing tests passed
> 
> ## License
> 
> 
> 
> - [x] I confirm that this pull request can be released under the Apache 2 license
> 
> [issues]: https://github.com/aws/aws-sdk-net/issues
> [license]: http://aws.amazon.com/apache2.0/
> [cla]: http://en.wikipedia.org/wiki/Contributor_License_Agreement

normj (approved):

sdk/test/Services/DynamoDBv2/IntegrationTests/TTLTests.cs
```diff
@@ -88,6 +88,66 @@ public void TestStoreAsEpoch()
             Assert.IsNotNull(epochMap["NonEpochDate2"].S);
         }
 
+        public void TestStoreAsAnnotatedEpoch()
```
ashovlin, line 91: A few places where we duplicated tests to use with annotated classes, is there a way to parameterize the type or make it generic to avoid the duplication?

We did that for `TestHashRangeObjects` in https://github.com/aws/aws-sdk-net/commit/e69fd9481c8312887b32790414424901ae6e1e54#diff-73d1ee8f0f06d29cfa60986f66dc2fe72a23143a7390ba33cd125d66d5034d56

ashovlin: We discussed internally, the new annotated types don't all descend from the base `Employee`, so it's challenging to reuse the test harness.

ashovlin (approved):

philasmar:
I rebased on top of `v4-development`. I'll run a dry run and merge if it succeeds.

Issue #3432, opened 2024-08-08
Using the SDK in web application testing broken (.NET 8 URI canonicalization)


tskarman:
The change from https://github.com/aws/aws-sdk-net/pull/3077 breaks tests execution for me.
This is in a scenario were a limited server implementation is verified via the AWS SDK.

One way to solve this for me would be an option to disable your .NET 8 behavior, i.e. allow me to specify that `UriCreationOptions.DangerousDisablePathAndQueryCanonicalization` should not be used in System.Uri instance creation on the SDK's side.

The error arising being: **System.InvalidOperationException: 'GetComponents() may not be used for Path/Query on a Uri instance created with UriCreationOptions.DangerousDisablePathAndQueryCanonicalization.'**

Or in more general terms, I am guessing this breaks scenarios where custom `HttpClient` instances are provided through the `AmazonS3Config.HttpClientFactory` setting.

A quick description of my setup:

```csharp
// In-memory application for functional end to end tests (i.e. ASP.NET Core HTTP infrastructure)
WebApplicationFactory<Program> factory = ...;
// Create a client off that.
var httpClient = factory.CreateClient();

// That server has base address http://localhost/
// Resulting in URLs like this in the course of test execution: http://localhost/test-6e4c3a03-5508-47ba-9daa-422d9d55cd31/
// where the path is a bucket name

var configuration = new AmazonS3Config
{
    ServiceURL = factory.Server.BaseAddress.OriginalString,
    ForcePathStyle = true,
    LogResponse = false,
    DisableLogging = false,
    LogMetrics = false,
    MaxErrorRetry = 0,
    // This factory will always return the same instance, namely the client passed into it at construction.
    HttpClientFactory = new TestHttpClientFactory(httpClient),
};
```

This is where the `HttpClient` of `Microsoft.AspNetCore.Mvc.Testing` is configured: https://github.com/dotnet/aspnetcore/blob/4966b6c/src/Mvc/Mvc.Testing/src/WebApplicationFactoryClientOptions.cs#L64

So pretty vanilla.

The exception arises inside this call: https://github.com/dotnet/aspnetcore/blob/c096dbb/src/Hosting/TestHost/src/ClientHandler.cs#L174

This is the call stack in my scenario:

```
>	System.Private.Uri.dll!System.Uri.GetComponents(System.UriComponents components, System.UriFormat format) Line 335	C#
[...]

PR #3434, opened 2024-08-10
Add ability to turn off the Uri option DangerousDisablePathAndQueryCanonicalization

normj:
## Description
In .NET 8 we started using the `DangerousDisablePathAndQueryCanonicalization` when creating the `Uri` class to avoid .NET changing resource paths which would interfere with S3 object keys that look like paths that should be canonicalized. This is correct behavior we want when making service calls.

A user has found in there testing environment where they are putting in a fake endpoint using the `WebApplicationFactory` that the tests are failing because `WebApplicationFactory` calls the `GetComponents` method on the Uri class which is not allowed for a Uri class using `DangerousDisablePathAndQueryCanonicalization`.

This PR gives users using the `WebApplicationFactory` a back door to turn off the usage of `DangerousDisablePathAndQueryCanonicalization` so they can continue using it with .NET 8.

## Motivation and Context
https://github.com/aws/aws-sdk-net/issues/3432

## Testing
Added a new unit test. Dry run pending.


dscpinheiro (approved):
FYI @normj: I added a commit updating the DevConfig (otherwise service packages wouldn't pick up the change automatically).

boblodgett (approved):

Issue #1555, opened 2020-03-13
Document.FromJson() changes the decimal value


simon10says:
When using `Document.FromJson()`, the decimal value always increment/decrement by 1 when the number of digits are more than 18.
## Expected Behavior
The `FromJson()` conversion should not alter the value and should support up to 38 digits of precision as supported by DynamoDB.

## Current Behavior
1. If I've about 18 digits, the last value always increment/decrement by 1. E.g. xxx.1234, will become xxx.1233 (where xxx are the first 14 digits)

## Steps to Reproduce (for bugs)
Create a blank Console application in Visual Studio and run the code below:
```cs
static void Main(string[] args)
{
    var doc = Document.FromJson("{\"decimal\":1584097275961.1234}");
    Console.WriteLine($"doc: {doc["decimal"]}");
}
```
You will see the result `doc: 1584097275961.1233`; it should be xxx.**1234**

## Your Environment

* AWSSKD.dynamoDBv2: 3.3.105.9
* Visual Studio version: Microsoft Visual Studio Community 2019 Version 16.4.5

simon10says:
Related to #1039

klaytaybai:
Hi @simon10says, thanks for reviving interest in this issue.  It is concerning to me.  However, due to C# limitations with number precision at high digits and our current implementation, a decision was made to wait until a next major version for a fix.  We recommend that anybody using large numbers that require greater precision than offered by the decimal class use strings to store these numbers instead so as to not lose any precision.

Since the previous issue is so old and might not get as much attention, I'm going to mark that one as a duplicate and keep this one open going forward.

simon10says:
Hi @klaytaybai , our teams had already used the workaround you suggested, so, we are good for now. However, IMHO, I urge you to add the workaround into the next release asap because this is the kind of bug that causes a **'ghost'** effect in the code and the result, which go undetected by the developers and end-user, but caused **tainted data**; and I wonder how many have gone undetected with this bug. For us,

1. We capped our precision to 28 digits as defined by `C# decimal` even thought dynamoDB is limited to 38 digits
2.  One of our developer decided to add in a test case for 18 digits precision because after our long internal testing, we always get the 'good' result and have never hit the 'bad' result' and it is the 18th digital precision that is the key
3. We spent days debugging this and the very last thing we could have suspect is this `FromJson()` 

Bugs could be fixed; but if we had went live with tainted data, it will be catastrophe

NimaAra:
[...]

Issue #1039, opened 2018-08-16
Decimal values in JSON lose precision when converted using Document.FromJson()


empathicqubit:
In DynamoDB, Document.FromJson produces incorrect values when converting a decimal larger than a double.

## Expected Behavior

It returns an N-typed AttributeValue, preserving the level of precision that DynamoDB supports.

## Current Behavior

It truncates the value to the size of a double.
## Possible Solution

Change ThirdParty.Json.LitJson.JsonReader.ProcessNumber(),  aws/aws-sdk-net:sdk/src/Core/ThirdParty/Json/JsonReader.cs:132 to either throw or convert to decimal.

## Steps to Reproduce (for bugs)

```cs
using Amazon.DynamoDBv2.DocumentModel;
var document = Document.FromJson("{\"decimal\":171717171717171717171.0}");
```

## Context

Store large numbers in DynamoDB correctly.
## Your Environment

* AWSSDK.Core version used: 3.3.10.3
* Service assembly and version used: AWSSDK.DynamoDB (3.3.10.3)
* Operating System and version: Windows 10 1709
* Visual Studio version: 15.7.4
* Targeted .NET platform: netcoreapp2.0

## .NET Core Info

* .NET Core version used for development: 
* .NET Core version installed in the environment where application runs: 
* Output of `dotnet --info`: 
```.NET Core SDK (reflecting any global.json):
 Version:   2.1.302
 Commit:    9048955601

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.16299
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.1.302\

Host (useful for support):
  Version: 2.1.2
  Commit:  811c3ce6c0

.NET Core SDKs installed:
  1.0.0-preview2-003131 [C:\PROGRA~1\dotnet\sdk]
  1.0.0-preview2-003131 [C:\Program Files\dotnet\sdk]
  1.0.0-rc4-004771 [C:\PROGRA~1\dotnet\sdk]
  1.0.0-rc4-004771 [C:\Program Files\dotnet\sdk]
  1.0.0-rc4-004842 [C:\PROGRA~1\dotnet\sdk]
  1.0.0-rc4-004842 [C:\Program Files\dotnet\sdk]
  1.0.0 [C:\PROGRA~1\dotnet\sdk]
  1.0.0 [C:\Program Files\dotnet\sdk]
[...]

PR #3447, opened 2024-08-19
Support decimal precision for Document.FromJson and ToJson by using System.Text.Json instead of LitJson

ashovlin:
### Before
Currently, `Document.FromJson` and `Document.ToJson` in DynamoDBv2 use LitJson for JSON handling.

This does not support `decimal`, as the first numeric type it matches to here is `double`.

https://github.com/aws/aws-sdk-net/blob/c88243bd458afadccab6282e97675141d94047ed/sdk/src/Core/ThirdParty/Json/JsonReader.cs#L132-L144

This is leading to a loss of precision for decimal values, see the linked issues.

### After
This proposes to address this by replacing LitJson in these code paths with System.TextJson.
* We want to do this anyway as long term we'd like to remove the embedded copy of LitJson from AWSSDK.Core.

While this addresses the reported issue, we still need to look at:
* Is this causing a regression for any other numeric values or types, or even non-numeric types?
* Is it handling comments, malformed JSON, etc. the same way?

## Motivation and Context
#1555 , #1039 

## Testing
* Ran DynamoDBv2 sln tests locally
* PENDING dry-run

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/DynamoDBv2/Custom/DocumentModel/JsonUtils.cs
+++ b/sdk/src/Services/DynamoDBv2/Custom/DocumentModel/JsonUtils.cs
@@ -16,12 +16,9 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
-using System.Text;
-
-using Amazon.DynamoDBv2.Model;
-using Amazon.Util;
-using ThirdParty.Json.LitJson;
 using System.IO;
+using System.Text;
+using System.Text.Json;
 
 namespace Amazon.DynamoDBv2.DocumentModel
 {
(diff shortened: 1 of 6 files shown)


sdk/src/Services/DynamoDBv2/Custom/DocumentModel/JsonUtils.cs
```diff
@@ -83,19 +87,16 @@ public static IEnumerable<Document> FromJsonArray(string jsonText)
...
 
             WriteJson(document, writer, DynamoDBEntryConversion.V2);
 
-            // Trim everything before the first '{' character
-            var jsonIndex = FirstIndex(sb, '{');
-            if (jsonIndex > 0)
-                sb.Remove(0, jsonIndex);
-
-            var jsonText = sb.ToString();
-            return jsonText;
```
ashovlin, line 98: Not sure why this is included, didn't port it over for now.

sdk/test/Services/DynamoDBv2/IntegrationTests/JSONTests.cs
```diff
@@ -6,23 +6,20 @@
...
 {
     public partial class DynamoDBTests : TestBase<AmazonDynamoDBClient>
     {
-        const string sampleJson = @"{
+        const string _sampleJson = @"{
 ""Name"" : ""Alan"" ,
 ""Age"" : 31,
 ""CompanyName"" : ""Big River"" ,
-""CurrentStatus"" : ""Active""
+""CurrentStatus"" : ""Active"",
```
ashovlin, line 22: This test was passing before despite the missing comma, but STJ throws an exception now.

dscpinheiro:
Closing in favour of https://github.com/aws/aws-sdk-net/pull/3534

PR #3458, opened 2024-08-27
Add ToString methods for better enum string conversion

paulomorgado:
## Description

- Added private `ToString` methods for various comparison `enum`s in `ConditionFactory.cs` to improve readability and maintainability.
- Updated `NewCondition` methods to use these new `ToString` methods for consistent string representation.
- Optimized `boolean` conversion to `string` using a ternary operator.

## Motivation and Context

The result of invoking `ToString` on an `enum` is cached in a limited space and there's no guarantee that the value will be cached for the same value in consecutive invocations.

Having code that returns compile-time generated static `string`s will guarantee that that `string` is always available and easy to access.

Invoking `ToString` in `bool` will return the `string`literals `"True"` or `"False"`, depending on its value. However, invoking `ToLowerInvariant()` on those `String`s will incur in the generation/allocation of a new `string` every time.

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Auth/AccessControlPolicy/ConditionFactory.cs
+++ b/sdk/src/Core/Amazon.Auth/AccessControlPolicy/ConditionFactory.cs
@@ -102,7 +102,7 @@ namespace Amazon.Auth.AccessControlPolicy
         /// <summary>
         /// Enumeration of the supported ways an ARN comparison can be evaluated.
         /// </summary>
-        public enum ArnComparisonType 
+        public enum ArnComparisonType
         {
             /// <summary>Exact matching</summary>
             ArnEquals,
@@ -177,7 +177,7 @@ namespace Amazon.Auth.AccessControlPolicy
         /// <summary>
         /// Enumeration of the supported ways a string comparison can be evaluated.
         /// </summary>
-        public enum StringComparisonType 
+        public enum StringComparisonType
         {
             /// <summary>
             /// Case-sensitive exact string matching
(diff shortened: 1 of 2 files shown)


paulomorgado:
Tests invoking `AssertExtensions.AssertEnumUnchanged` should be added to validate changes to these `enum`s, but I haven't figured out what the best place for it would be.

danielmarbach:
Nice change @paulomorgado

paulomorgado:
> Nice change @paulomorgado

Thanks, @danielmarbach.

Maybe introducing [NetEscapades.EnumGenerators: a source generator for enum performance](https://andrewlock.net/netescapades-enumgenerators-a-source-generator-for-enum-performance/) would be a good option to avoid hand-written code.

peterrsongg (changes requested):
The change itself looks good, but these need unit tests
I think [this](https://github.com/aws/aws-sdk-net/blob/main/sdk/test/UnitTests/Custom/PolicyTests.cs) is a good place to put the tests. 

I think just testing a few is sufficient, probably no reason to test all of them but I would feel more comfortable knowing we had unit tests.

I wrote a method to get the expected hash of the original type

```
    public static class HashingUtils
    {
        private static SHA256 SHA256 = SHA256.Create();
        public static string GetHash(Type type)
        {
            return BitConverter.ToString(SHA256.ComputeHash(Encoding.Default.GetBytes(type.ToString()))).Replace("-","");
        }
    }
```

which you could use to get the expected hash and then pass that in to AssertEnumUnchanged. You may have to tweak it a little

paulomorgado:
> The change itself looks good, but these need unit tests I think [this](https://github.com/aws/aws-sdk-net/blob/main/sdk/test/UnitTests/Custom/PolicyTests.cs) is a good place to put the tests.
> 
> I think just testing a few is sufficient, probably no reason to test all of them but I would feel more comfortable knowing we had unit tests.
> 
> I wrote a method to get the expected hash of the original type
> 
> ```
>     public static class HashingUtils
>     {
>         private static SHA256 SHA256 = SHA256.Create();
>         public static string GetHash(Type type)
>         {
>             return BitConverter.ToString(SHA256.ComputeHash(Encoding.Default.GetBytes(type.ToString()))).Replace("-","");
>         }
>     }
> ```
> 
> which you could use to get the expected hash and then pass that in to AssertEnumUnchanged. You may have to tweak it a little

On it!

normj (approved):

peterrsongg (approved):

Issue #3463, opened 2024-08-30
Clock skew correction causes repeated retries


cgriebel:
### Describe the bug

The clock skew mitigation makes multiple retries when the API responds with any Error Code in `clockSkewErrorCodes` even if no clock skew is detected.
https://github.com/aws/aws-sdk-net/blob/2fd15245e35206c5063a90a033a4be897da7931a/sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs#L399-L407

This seems to be caused by a recent change in https://github.com/aws/aws-sdk-net/pull/3378 to `IsClockSkew` in the RetryPolicy:
https://github.com/aws/aws-sdk-net/blob/2fd15245e35206c5063a90a033a4be897da7931a/sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs#L456-L462

In cases where there is a legitimate failure (in our case an AuthFailure) unrelated to clock skew, the code now retries until `MaxErrorRetry`; in our case, this causes significant performance degradation.

### Expected Behavior

If there is no clock skew, the client shouldn't retry.

### Current Behavior

The client assumes clock skew and always retries for all `clockSkewErrorCodes`.

### Reproduction Steps

Sample Repository: https://github.com/cgriebel/aws-clock-skew-issue

```csharp
using Amazon;
using Amazon.EC2;
using Amazon.Runtime;


AWSConfigs.LoggingConfig.LogTo = LoggingOptions.Console;
AWSConfigs.LoggingConfig.LogResponses = ResponseLoggingOption.OnError;

var credentials = new BasicAWSCredentials("access-key", "secret-key");

var config = new AmazonEC2Config()
{
    RegionEndpoint = RegionEndpoint.APSoutheast5,

    // 10 isn't required, but a higher retry count makes the performance impact more obvious
    MaxErrorRetry = 10,
};

try
{
    // We need a legitimate failure unrelated to clock skew to demonstrate the behavior, in our case
    // we used a region that was not enabled for the account so that an "AuthFailure" error code is returned.
    var client = new AmazonEC2Client(credentials, config);
    var response = await client.DescribeAccountAttributesAsync();
}
catch (Exception ex)
{
    Console.WriteLine(ex.Message);
}
```

### AWS .NET SDK and/or Package version used

AWSSDK.EC2 3.7.405

### Targeted .NET Platform

.NET 8

### Operating System and version

Windows 11, Debian 12

[...]

PR #3465, opened 2024-09-03
Fixes issue #3463 where repeated retries were happening for error cod…

peterrsongg:
…es that weren't always clockskew related.

In a previous PR to fix a clockskew related threading issue, we updated the list of error codes to retry on. This list was too exhaustive and included error codes such as `AuthFailure` which are not always clockskew related. To not break the previous threading related fix, I added a new property called `SignedAt` to the `IRequest` class. Whenever a request is signed, this value is also populated. 

This way we store the original time that the request was signed at, instead of calling `GetCorrectedUtcNowForEndpoint` inside the `IsClockSkew` method. The previous threading issue revolved around the fact that one thread may update the datetime returned by `GetCorrectedUtcNowForEndpoint` if 2 threads enter `IsClockSkew` simultaneously, leading to the request not being retried on one of the threads. By keeping track of the original time the request was signed at, we ensure that when we calculate the diff, we are using the time which that request was signed at and not one that was overwritten by a separate request in a separate thread.

The second change I made was to separate the `IsClockSkewErrorCode` into 2 lists, where one is guaranteed to be a clockskew error and another which might be a clockskew error. If it is a definite clockskew error we adjust the clockskew and retry. If it is a possible clockskew error we check the difference in `request.SignedAt` and the server time to determine if it is indeed clockskew related and only retry and update the clockskew in that scenario.

**Note**: When updating the clockskew we use the `servertime - correctedUtcNow()` because at this point we want to use the latest time, regardless of which thread we're on.

## Motivation and Context

Fixes issue #3463 
## Testing

I've added unit tests for the exact scenario the user reports in the issue.
DRY_RUN PASSED
## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4Signer.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4Signer.cs
@@ -1229,6 +1229,7 @@ namespace Amazon.Runtime.Internal.Auth
             }
 
             var signedAt = CorrectClockSkew.GetCorrectedUtcNowForEndpoint(request.Endpoint.ToString());
+            request.SignedAt = signedAt;
             var region = overrideSigningRegion ?? DetermineSigningRegion(clientConfig, clientConfig.RegionEndpointServiceName, request.AlternateEndpoint, request);
 
             // AWS4 presigned urls got S3 are expected to contain a 'UNSIGNED-PAYLOAD' magic string
--- a/sdk/src/Core/Amazon.Runtime/Internal/Auth/CloudFrontSigner.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Auth/CloudFrontSigner.cs
@@ -39,11 +39,12 @@ namespace Amazon.Runtime.Internal.Auth
                 throw new ArgumentOutOfRangeException("awsAccessKeyId", "The AWS Access Key ID cannot be NULL or a Zero length string");
             }
 
-            string dateTime = AWSSDKUtils.GetFormattedTimestampRFC822(0);
-            request.SignedAt = AWSSDKUtils.CorrectedUtcNow;
-            request.Headers.Add(HeaderKeys.XAmzDateHeader, dateTime);
+            DateTime dateTime = CorrectClockSkew.GetCorrectedUtcNowForEndpoint(request.Endpoint.ToString());
+            request.SignedAt = dateTime;
+            string dateTimeString = dateTime.ToString(AWSSDKUtils.RFC822DateFormat, CultureInfo.InvariantCulture);
+            request.Headers.Add(HeaderKeys.XAmzDateHeader, dateTimeString);
 
-            string signature = ComputeHash(dateTime, awsSecretAccessKey, SigningAlgorithm.HmacSHA1);
+            string signature = ComputeHash(dateTimeString, awsSecretAccessKey, SigningAlgorithm.HmacSHA1);
 
             request.Headers.Add(HeaderKeys.AuthorizationHeader, "AWS " + awsAccessKeyId + ":" + signature);
         }
--- a/sdk/src/Core/Amazon.Runtime/Internal/Auth/CloudFrontSigner.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Auth/CloudFrontSigner.cs
@@ -40,6 +40,7 @@ namespace Amazon.Runtime.Internal.Auth
             }
 
             string dateTime = AWSSDKUtils.GetFormattedTimestampRFC822(0);
+            request.SignedAt = AWSSDKUtils.CorrectedUtcNow;
             request.Headers.Add(HeaderKeys.XAmzDateHeader, dateTime);
 
             string signature = ComputeHash(dateTime, awsSecretAccessKey, SigningAlgorithm.HmacSHA1);
--- a/sdk/src/Core/Amazon.Runtime/Internal/Auth/QueryStringSigner.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Auth/QueryStringSigner.cs
@@ -57,7 +57,7 @@ namespace Amazon.Runtime.Internal.Auth
             request.Parameters["SignatureVersion"] = SignatureVersion2;
             request.Parameters["SignatureMethod"] = clientConfig.SignatureMethod.ToString();
             request.Parameters["Timestamp"] = AWSSDKUtils.GetFormattedTimestampISO8601(clientConfig);
-            request.SignedAt = clientConfig.CorrectedUtcNow;
+            request.SignedAt = CorrectClockSkew.GetCorrectedUtcNowForEndpoint(request.Endpoint.ToString());
             // remove Signature parameter, in case this is a retry
             request.Parameters.Remove("Signature");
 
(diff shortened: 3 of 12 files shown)


sdk/src/Core/Amazon.Runtime/Internal/Auth/CloudFrontSigner.cs
```diff
@@ -40,6 +40,7 @@ public override void Sign(IRequest request, IClientConfig clientConfig, RequestM
             }
 
             string dateTime = AWSSDKUtils.GetFormattedTimestampRFC822(0);
```
peterrsongg, line 42: GetFormattedTimestampRFC822 calls AWSSDKUtils.CorrectedUtcNow internally, so instead of trying to parse the datetime just to get the time I just call it here.

sdk/test/UnitTests/Custom/Runtime/RetryHandlerTests.cs
```diff
@@ -224,13 +224,82 @@ public void AlwaysRetryOnClockSkewErrorCode()
...
+            Assert.AreEqual(2, Tester.CallCount);
+
+            // RetryPolicy should see that the clock skew for bucketname.s3.amazonaws.com is zero and change it to ~ -7
+            Assert.IsTrue(CorrectClockSkew.GetClockCorrectionForEndpoint(requestEndpoint.ToString()) < TimeSpan.FromMinutes(-7));
+        }
+
+        // Tests that if the error code is a possible clockskew error code such as  AuthFailure, the retry handler
+        // checks the diff to see whether to retry or not. In this case since the diff < 5 minutes we don't retry since this 
+        // is not a clockskew error.
+        [TestMethod]
```
peterrsongg, line 271: the user's scenario in the bug report

sdk/src/Core/Amazon.Runtime/Internal/Auth/CloudFrontSigner.cs
normj: Why not do `CorrectClockSkew.GetClockCorrectionForEndpoint(request.Endpoint.ToString())` like the AWS4Signer does?

sdk/src/Core/Amazon.Runtime/Internal/Auth/QueryStringSigner.cs
```diff
@@ -57,6 +57,7 @@ public override void Sign(IRequest request, IClientConfig clientConfig, RequestM
             request.Parameters["SignatureVersion"] = SignatureVersion2;
             request.Parameters["SignatureMethod"] = clientConfig.SignatureMethod.ToString();
             request.Parameters["Timestamp"] = AWSSDKUtils.GetFormattedTimestampISO8601(clientConfig);
+            request.SignedAt = clientConfig.CorrectedUtcNow;
```
normj, line 60: Seems funny now there is a third way being done to get the corrected UTC time between the AWS4Signer and CloudFrontSigner.

sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4Signer.cs
```diff
@@ -1229,6 +1229,7 @@ public static AWS4SigningResult SignRequest(IRequest request,
             }
 
             var signedAt = CorrectClockSkew.GetCorrectedUtcNowForEndpoint(request.Endpoint.ToString());
+            request.SignedAt = signedAt;
```
normj, line 1232: Does our Sig V4a signer not handle clock skew today?

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs
```diff
@@ -421,50 +425,59 @@ private bool IsClockskew(IExecutionContext executionContext, Exception exception
...
+                }
+            }
+            // Some services such as S3 will return a generic 403 error for clock skew errors on HEAD requests
+            // with no error code. In this case we will always retry the request if the clock skew is greater than
+            // the threshold.
+            if ((isPossibleClockSkewErrorCode || isHead) && serverTimeDetermined)
+            {
+                DateTime clientTime = executionContext.RequestContext.Request.SignedAt ?? correctedNow;
+                var diff = clientTime - serverTime;
+                var absDiff = diff.Ticks < 0 ? -diff : diff;
```
normj, line 467: Is this a micro performance optimization instead of calling `Math.Abs`?

sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4Signer.cs
peterrsongg: It does handle clockskew today
https://github.com/aws/aws-sdk-net/blob/main/extensions/src/AWSSDK.Extensions.CrtIntegration/CrtAWS4aSigner.cs#L104
The above code eventually calls `CorrectClockSkew.GetCorrectedUtcNowForEndpoint`  to get the time it was signed at. Would you like me to populate `request.SignedAt` in the crt extensions package as well?

sdk/src/Core/Amazon.Runtime/Internal/Auth/CloudFrontSigner.cs
peterrsongg: Good catch, it actually seems like the `CloudFrontSigner` is not accounting for clockskew. `AWSSDKUtils.GetFormattedTimestampRFC822(0)` doesn't check `CorrectClockSkew`'s dictionary at all. I could change this to something like `CorrectClockSkew.GetCorrectedUtcNowForEndpoint(request.Endpoint.ToString()).ToString(AWSSDKUtils.RFC822DateFormat, CultureInfo.InvariantCulture)` but that would be a behavior change (even though it is the correct one).

sdk/src/Core/Amazon.Runtime/Internal/Auth/QueryStringSigner.cs
peterrsongg: I went with this because `AWSSDKUtils.GetFormattedTimestampISO8601` uses `clientConfig.CorrectedUtcNow` which ends up calling  `CorrectClockSkew.GetCorrectedUtcNowForEndpoint`. However, I agree it is a bit confusing and will change it.

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs
peterrsongg: I'm not sure if this gives a performance benefit, but I'd rather leave it this way since it was here before I made this change.

normj: Ok, I see it was in the old code as well.

sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4Signer.cs
normj: Why would we not set `request.SignedAt` for `CrtAWS4aSigner`? Without it wouldn't SigV4a requests rely on the global clockskew value versus the actual SDK request time which is what we are trying to fix with this PR and the previous PR.

peterrsongg: Just added it to crt as well

normj (approved):

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs
```diff
@@ -421,50 +425,59 @@ private bool IsClockskew(IExecutionContext executionContext, Exception exception
             var isHead =
                 executionContext.RequestContext.Request != null &&
                 string.Equals(executionContext.RequestContext.Request.HttpMethod, "HEAD", StringComparison.Ordinal);
-            var isClockskewErrorCode =
+            var isPossibleClockSkewErrorCode =
                 ase != null &&
-                (ase.ErrorCode == null || clockSkewErrorCodes.Contains(ase.ErrorCode));
-
-            if (isHead || isClockskewErrorCode)
+                (ase.ErrorCode == null || possibleClockSkewErrorCodes.Contains(ase.ErrorCode));
```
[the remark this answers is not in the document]
jeremysimmons: this should be a comment in the code. you'll forget in a month...

PR #3474, opened 2024-09-12
Address Protocol Tests that are breaking changes

peterrsongg:
## Description

This addresses all the Protocol Tests which we deemed as breaking changes
1. `SimpleScalarPropertiesPureWhiteSpace`: Xml responses that returned pure whitespace in the body of an xml element were being trimmed. For example `     `=>` `. Now we will return the whitespace as is without trimming it.
2. `RestJsonHttpPayloadTraitsWithNoBlobBody` `HttpPayloadTraitsWithNoBlobBody`: If the payload of a response has no data, instead of returning an empty memory stream as the payload we will return null.
3. Unmarshalling an attribute with `@xsi:` was not working, since `@xsi` is a prefix. We will now successfully unmarshall an attribute with a prefix.

Some refactoring of the tt files.
Some other tests were request level changes that weren't breaking but were fixed with the nullability changes (see empty lists/maps tests)

**Note** The other tests in the VNext list can't be addressed until we pull in the latest smithy version, since I added some additional tests for nestedXmlMapsWithXmlName. This would also cause us to pull in a lot of other protocol tests, so should be addressed in a separate PR. Those tests are not breaking changes so it is not critical to ship them in V4.
## Motivation and Context

Need to address in major version as there are breaking changes.
DOTNET_7435

## Testing

DRY_RUN passed in v4.
All protocol tests passed.
## Types of changes

## Checklist

## License

### DIFF
--- a/generator/ServiceClientGeneratorLib/Generators/Marshallers/BaseResponseUnmarshaller.cs
+++ b/generator/ServiceClientGeneratorLib/Generators/Marshallers/BaseResponseUnmarshaller.cs
@@ -1683,7 +1683,8 @@ this.Write(" = (int)context.ResponseData.StatusCode;\r\n");
 		if (!withPrefix)
 		{
 			var locationNameString = locationName.ToString();
-			return locationNameString.StartsWith("xsi:") ? locationNameString.Substring(4) : locationNameString;
+			int colonIndex = locationNameString.IndexOf(":");
+			return colonIndex != -1 ? locationNameString.Substring(colonIndex + 1) : locationNameString;
 		}
         // the locationName and modeled name must only be different for those members which are marshalled on the body.
         if (!string.Equals(locationName.ToString(), member.ModeledName, StringComparison.Ordinal) || member.MarshallLocation != MarshallLocation.Body)
--- a/generator/ServiceClientGeneratorLib/Generators/Marshallers/BaseResponseUnmarshaller.cs
+++ b/generator/ServiceClientGeneratorLib/Generators/Marshallers/BaseResponseUnmarshaller.cs
@@ -1669,7 +1669,7 @@ this.Write(" = (int)context.ResponseData.StatusCode;\r\n");
         2. Use the value of the locationName on the member's target if present
         3. Use the name of the member's target
     */
-    protected string DetermineXmlMarshallName(Member member)
+    protected string DetermineXmlMarshallName(Member member, bool withPrefix = true)
     {
         var locationName = member.data[ServiceModel.LocationNameKey];
         if (locationName == null)
@@ -1680,6 +1680,11 @@ this.Write(" = (int)context.ResponseData.StatusCode;\r\n");
                 return member.Shape.Name;
             return member.MarshallName;
         }
+		if (!withPrefix)
+		{
+			var locationNameString = locationName.ToString();
+			return locationNameString.StartsWith("xsi:") ? locationNameString.Substring(4) : locationNameString;
+		}
         // the locationName and modeled name must only be different for those members which are marshalled on the body.
         if (!string.Equals(locationName.ToString(), member.ModeledName, StringComparison.Ordinal) || member.MarshallLocation != MarshallLocation.Body)
         {
--- a/generator/ServiceClientGeneratorLib/Generators/Marshallers/RestXmlRequestMarshaller.tt
+++ b/generator/ServiceClientGeneratorLib/Generators/Marshallers/RestXmlRequestMarshaller.tt
@@ -124,7 +124,7 @@ namespace <#=this.Config.Namespace #>.Model.Internal.MarshallTransformations
 #>
 				if(<#=variableName + ".IsSet" + member.PropertyName#>())
 <#+
-WriteXmlAttributeString(level + 1,member,variableName, isPayload: false);
+WriteXmlAttributeString(level + 1, member, variableName, isPayload: false);
 #>
 <#+
 			}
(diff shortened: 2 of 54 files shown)


generator/ServiceClientGeneratorLib/Generators/Marshallers/BaseResponseUnmarshaller.cs
```diff
@@ -1680,6 +1680,11 @@ protected string DetermineXmlMarshallName(Member member)
                 return member.Shape.Name;
             return member.MarshallName;
         }
+		if (!withPrefix)
+		{
+			var locationNameString = locationName.ToString();
+			return locationNameString.StartsWith("xsi:") ? locationNameString.Substring(4) : locationNameString;
```
boblodgett, line 1686: Is the prefix always going to be xsi or should we be doing a more generic prefix strip?

generator/ServiceClientGeneratorLib/Generators/Marshallers/RestXmlRequestMarshaller.tt
```diff
@@ -123,7 +123,9 @@ namespace <#=this.Config.Namespace #>.Model.Internal.MarshallTransformations
 			{
 #>
 				if(<#=variableName + ".IsSet" + member.PropertyName#>())
-					xmlWriter.WriteAttributeString("<#=member.MarshallName#>", "<#=member.XmlNamespace#>",<#=member.PrimitiveMarshaller#>(<#=variableName#>.<#=member.PropertyName#>));
+<#+
+WriteXmlAttributeString(level + 1,member,variableName, isPayload: false);
```
boblodgett, line 127: readability: missing spaces after the `,`s

generator/ServiceClientGeneratorLib/Generators/Marshallers/RestXmlRequestMarshaller.tt
```diff
@@ -736,4 +750,39 @@ namespace <#=this.Config.Namespace #>.Model.Internal.MarshallTransformations
...
+<#+
+	// xsi is a common prefix in attributes, but it cannot be included in the local name itself as the xmlWriter class will throw an exception. 
+	// Some services may model the xmlName, with the xsi prefix included like it is here https://github.com/smithy-lang/smithy/blob/main/smithy-aws-protocol-tests/model/restXmlWithNamespace/main.smithy#L147
+	// In this case, we add xsi to the prefix argument of WriteAttributeString and extract the local name.
+	protected void WriteXmlAttributeString(int level, Member member, string variableName,  bool isPayload, Operation operation = null)
+	{
+		var prefix = "";
+		var localName = "";
+		string ns = member.XmlNamespace;
+		if (member.MarshallName.StartsWith("xsi"))
```
boblodgett, line 763: Should this be generic prefix parsing instead?

peterrsongg: Just for some context 

Normally a service would model it like this where they tell us the prefix. In that case we use the prefix provided https://github.com/aws/aws-sdk-net/blob/806efa402e867e444e309960fbb688d6e1bda966/generator/TestServiceModels/restxmlwithnamespace-tests-client/rest-xml-protocol-namespace-2019-12-16.normal.json#L100

But in some cases the service will actually attach `xsi` to the locationName itself like this https://github.com/aws/aws-sdk-net/blob/806efa402e867e444e309960fbb688d6e1bda966/generator/TestServiceModels/restxmlwithnamespace-tests-client/rest-xml-protocol-namespace-2019-12-16.normal.json#L60

Although no actual service does this, the protocol tests want to make sure we can handle that. The only other valid prefix from what I've seen is `xsd` (other than `xsi`), but I don't see a problem with just assuming everything before the first `:` is a prefix, since we can't unmarshall an attribute with a `:` in it anyways. Will make it more generic.

boblodgett (approved):

sdk/test/Services/RestXmlProtocol/Generated/Model/Internal/MarshallTransformations/XmlAttributesRequestMarshaller.cs
```diff
@@ -64,7 +64,7 @@ public IRequest Marshall(XmlAttributesRequest publicRequest)
             {   
                 xmlWriter.WriteStartElement("XmlAttributesRequest", "");
                 if(publicRequest.IsSetAttr())
-                    xmlWriter.WriteAttributeString("test", "",StringUtils.FromString(publicRequest.Attr));
+                        xmlWriter.WriteAttributeString("","test", "",StringUtils.FromString(publicRequest.Attr));
```
dscpinheiro, line 67: Is the extra space before `xmlWriter` intentional? If you're making a change, it'd also be good to add spaces between the `,`, e.g.: 
```csharp
xmlWriter.WriteAttributeString("", "test", "", StringUtils.FromString(publicRequest.Attr))
```

generator/ServiceClientGeneratorLib/Generators/Marshallers/RestXmlRequestMarshaller.tt
```diff
@@ -736,4 +750,40 @@ namespace <#=this.Config.Namespace #>.Model.Internal.MarshallTransformations
 <#=new string(' ', (level) * 4)#>					}
 <#+
 	}
+#>
+<#+
+	// xsi is a common prefix in attributes, but it cannot be included in the local name itself as the xmlWriter class will throw an exception. 
+	// Some services may model the xmlName, with the xsi prefix included like it is here https://github.com/smithy-lang/smithy/blob/main/smithy-aws-protocol-tests/model/restXmlWithNamespace/main.smithy#L147
```
dscpinheiro, line 756: Small nit: In the GitHub UI, you can right click that line and select `Copy permalink`. 

I don't know how often those protocol tests change, but if they do line 147 on the `main` branch may not be the same.

peterrsongg: cool tip! didn't know about that. just updated

sdk/test/Services/RestXmlProtocol/Generated/Model/Internal/MarshallTransformations/XmlAttributesRequestMarshaller.cs
peterrsongg: ah good catch, removed the extra indentation

dscpinheiro (approved):

PR #3477, opened 2024-09-18
Separate DynamoDBStreams from DynamoDB

peterrsongg:
## Description

This removes customizations in the generator to generate DynamoDBStreams in the same package as DynamoDB. This is has caused a lot of special case logic and operational burden for the .NET SDK team over the years, and we are the last SDK to maintain this behavior of combining DynamoDB and DynamoDB stream into the same package.

The user migration experience should just be adding the new package and then adding the namespace used by the DynamoDB stream package.

I added an entry in sdk-versions.json for ddstreams. I ran the generator with the generator customizations removed to generate ddb streams. I also refactored existing tests to use the new package.
## Motivation and Context

DOTNET-7688
## Testing

E2E tests pass in v4. 
## Types of changes

## Checklist

## License


generator/ServiceModels/_sdk-versions.json
```diff
@@ -6,3091 +6,3099 @@
...
+    },
+    "DynamoDBv2": {
+      "Version": "4.0.0.0",
+      "AssemblyVersionOverride": "4.0",
+      "Dependencies": {
+        "Core": "4.0.0.0"
+      },
+      "InPreview": false
+    },
+    "DynamoDBStreams": {
```
peterrsongg, line 156: the only addition i made.

generator/ServiceModels/_sdk-versions.json
```diff
@@ -6,3091 +6,3099 @@
     "OverrideCoreVersion" : "4.0",
     "DefaultToPreview"    : true,
     "PreviewLabel"        : "preview.2",
-    "ServiceVersions"     : {
```
dscpinheiro, line 9: I would not change the indentation of this file (it's hard to see what changed without your comment and will make the next V3 -> V4 sync harder).

What about adding the package at the end (similar to what the build system already does for new packages):
```json
        "SSMQuickSetup"                               : {
            "Version" : "4.0.0.0",
            "AssemblyVersionOverride" : "4.0",
            "Dependencies"            : {
                "Core" : "4.0.0.0"
            },
            "InPreview"               : false
        },
        "DynamoDBStreams"                       : {
            "Version" : "4.0.0.0",
            "AssemblyVersionOverride" : "4.0",
            "Dependencies"            : {
                "Core" : "4.0.0.0"
            },
            "InPreview"               : false
        }
    },
    "DefaultAssemblyVersion" : "4.0"
```

peterrsongg: fixed! VS auto-formatted it which made it look weird. and added to the end

generator/ServiceClientGeneratorLib/GeneratorDriver.cs
```diff
@@ -871,8 +871,7 @@ public static void UpdateNuGetPackagesInReadme(GenerationManifest manifest, Gene
             var nugetPackages = new Dictionary<string, string>();
             foreach (var service in manifest.ServiceConfigurations.OrderBy(x => x.ClassName))
             {
-                // Service like DynamoDB streams are included in a parent service.
-                if (service.ParentConfig != null || service.IsTestService)
```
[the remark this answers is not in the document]
peterrsongg: Yup ddbstreams is the only service using parent-base-name. I ran the generator just to be sure, but yeah if a service were to use it in the future this would break them.

I'm okay with just being safe for now and removing the comment but reverting the if statement

dscpinheiro (approved):

boblodgett (approved):

PR #3490, opened 2024-09-27
Check for null collections in AmazonEC2ResponseHandler

dscpinheiro:
## Description
Updates the `AmazonEC2ResponseHandler` class to handle null collections when populating legacy fields (e.g. `Ipv4Ranges` to `IpRanges`). As discussed internally, we can even remove these before GA, but that'll be addressed in a separate task / PR.

## Testing
- Dry-run: `DRY_RUN-efd5cd23-acb5-41cb-b31e-1f99acde78b2`
- Ran example and confirmed exception is not thrown:
```csharp
var request = new DescribeSecurityGroupsRequest();
var client = new AmazonEC2Client(RegionEndpoint.USEast1);
var groups = await client.Paginators
    .DescribeSecurityGroups(request)
    .SecurityGroups
    .ToListAsync();
```

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/EC2/Custom/Internal/AmazonEC2ResponseHandler.cs
+++ b/sdk/src/Services/EC2/Custom/Internal/AmazonEC2ResponseHandler.cs
@@ -204,12 +204,18 @@ namespace Amazon.EC2.Internal
         {
             foreach (var ipPermission in ipPermissions)
             {
+                if (ipPermission.Ipv4Ranges == null)
+                {
+                    continue;
+                }
+
 #pragma warning disable CS0612,CS0618
                 ipPermission.IpRanges = ipPermission.Ipv4Ranges.Select(i => i.CidrIp).ToList();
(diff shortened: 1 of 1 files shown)


sdk/src/Services/EC2/Custom/Internal/AmazonEC2ResponseHandler.cs
```diff
@@ -219,39 +225,42 @@ private static void RestoreRequestIpPermissions(List<IpPermission> IpPermissions
         {
             foreach (var ipPermission in IpPermissions)
             {
-                if(ipPermission.RestoreOldIpV4Range)
+                if (ipPermission.RestoreOldIpV4Range)
```
dscpinheiro, line 228: No null check here since `RestoreRequestIpPermissions` is only invoked when `IpPermissions` is not null.

normj (approved):

afroz429 (approved):

PR #3501, opened 2024-10-07
Optimize `Metrics.IncrementCounter()` for .NET 8+

stevenaw:
## Description
Optimize the performance of `Metrics.IncrementCounter()` on .NET 8+ by using the ref-returning `CollectionsMarshal.GetValueRefOrAddDefault()`. To my knowledge, the `lock` statement is still needed owing to the [following remark in the documentation](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.collectionsmarshal.getvaluereforadddefault?view=net-8.0)

> Items should not be added to or removed from the [Dictionary<TKey,TValue>](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2?view=net-8.0) while the ref TValue is in use.

## Motivation and Context
Metrics counters tend to be called a lot. The optimized version is also less code.

## Testing
Unit tests covering both sequential and concurrent scenarios.

image

## Benchmarks

<details>
  
  ```csharp
public class DictionaryIncrement
{
    private object metricsLock = new object();
    private readonly Dictionary<string,long> Counters = new Dictionary<string,long>();
    private string metric;

    [GlobalSetup]
    public void Setup()
    {
        metric = "foo";
    }

    [Benchmark(Baseline = true)]
    public void Original()
    {
        lock (metricsLock)
        {
            long value;
            if (!Counters.TryGetValue(metric, out value))
            {
                value = 1;
            }
            else
            {
                value++;
            }
            Counters[metric] = value;
        }
    }

    [Benchmark]
    public void Optimized()
    {
        lock (metricsLock)
        {
            ref long value = ref CollectionsMarshal.GetValueRefOrAddDefault(Counters, metric, out _);
            value++;
        }
    }
}
  ```
</details>

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Internal/Util/Metrics.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Util/Metrics.cs
@@ -22,6 +22,7 @@ using System.Text;
 using Amazon.Util;
 using Amazon.Runtime.Internal.Util;
 using ThirdParty.Json.LitJson;
+using System.Threading;
 
 namespace Amazon.Runtime.Internal.Util
 {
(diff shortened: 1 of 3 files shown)


stevenaw:
Actually, I can't seem to find this in use anywhere. Perhaps this is old code?

muhammad-othman:
@stevenaw It looks like we use `SetCounter` rather than `IncrementCounter`, it is mainly used for retries which is already being tracked at the request pipeline and we get the latest value at `SetCounter`, maybe this was used before or was added for completeness.

PR #3506, opened 2024-10-09
Fix #3497. Make internalsdkutils trim compatible

### LINKED ISSUE #3497: InternalSDKUtils trim warning
### Describe the bug

Usage of AWSSDK.Core 3.7.400.22 with native aot causes the following trimmer warning.

ILC : Trim analysis warning IL2075: Amazon.Util.Internal.InternalSDKUtils.ApplyValues(Object,IDictionary`2<String,Object>): 'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicProperties' in call to 'System.Type.GetProperty(String)'. The return value of method 'System.Object.GetType()' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.

### Regression Issue

### Expected Behavior

No trim warnings.

### Current Behavior

Trim warnings raised.

### Reproduction Steps

Reference AWSSDK.Core and try to publish with PublishAot true...  Various parts of my csproj are below.

```
  <PropertyGroup>
    <TrimmerSingleWarn>false</TrimmerSingleWarn> 
  </PropertyGroup>

  <ItemGroup> 
    <PackageReference Include="Amazon.Lambda.RuntimeSupport" Version="1.11.0" />
    <PackageReference Include="Amazon.Lambda.Core" Version="2.3.0" />
    <PackageReference Include="Amazon.Lambda.Serialization.SystemTextJson" Version="2.4.3" />
    <PackageReference Include="Amazon.Lambda.APIGatewayEvents" Version="2.7.1" />
  </ItemGroup>

  <ItemGroup>
    <!-- <TrimmerRootAssembly Include="AWSSDK.Core" /> -->
    <TrimmerRootAssembly Include="Amazon.Lambda.APIGatewayEvents" />
    <TrimmerRootAssembly Include="bootstrap" />
  </ItemGroup>
```

### Possible Solution

_No response_

Also, setting the AWSSDK.Core reference as a trimmer root assembly results in even more warnings.

AWSSDK.Core 3.7.400.22

### Targeted .NET Platform

.NET 8

OSX Arm64

peterrsongg:
## Description

Since the `ApplyValues` method accepts a generic object, we cannot use the `DynamicallyAccessedMembersAttributes` on the `target` parameter. This is because the attribute can only be used on method return values or method parameters whose type is `Type` or `String`.

We could refactor this method to accept a `Type`, and then instead of calling `target.GetType` we would just use target... but that would mean we would have to change the calling code everywhere and that isn't possible since this is a public method that customers can call (breaking change).The only viable solution is to add the attribute to the class.
## Motivation and Context

Fixes #3497 
## Testing

Full dry run passes
## Types of changes

## Checklist

## License

### DIFF
--- a/generator/.DevConfigs/12af73ee-4e14-4b81-af61-e4fd810df074.json
+++ b/generator/.DevConfigs/12af73ee-4e14-4b81-af61-e4fd810df074.json
@@ -1,0 +1,9 @@
+{
+    "core": {
+      "changeLogMessages": [
+        "Fix #3497, make InternalSDKUtils aot compatible."
+      ],
+      "type": "patch",
+      "updateMinimum": true
+    }
+  }
--- a/sdk/src/Core/Amazon.Util/Internal/InternalSDKUtils.cs
+++ b/sdk/src/Core/Amazon.Util/Internal/InternalSDKUtils.cs
@@ -23,9 +23,13 @@ using System.Text.RegularExpressions;
 using Amazon.Runtime.Internal.Util;
 using Amazon.Util.Internal.PlatformServices;
 using System.Text;
+using System.Diagnostics.CodeAnalysis;
 
 namespace Amazon.Util.Internal
 {
+#if NET8_0_OR_GREATER
+    [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicProperties)]
+#endif
     public static partial class InternalSDKUtils
     {
         #region UserAgent


muhammad-othman (approved):

Dreamescaper:
I don't really understand. Why does this attribute helps with the trimming? Per my understanding, this attribute instructs the compiler not to trim members of InternalSDKUtils. Why does it help to preserve ApplyValues argument members?...

Dreamescaper:
I suggest to make this method generic instead. It allows to add DynamicallyAccessedMembers, and there it is invoked in the same way as previous non-generic one.

```csharp
public static void ApplyValues
    <[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicProperties)] T>
    (T target, IDictionary<string, object> propertyValues)
```

peterrsongg:
> I suggest to make this method generic instead. It allows to add DynamicallyAccessedMembers, and there it is invoked in the same way as previous non-generic one.
> 
> ```cs
> public static void ApplyValues
>     <[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicProperties)] T>
>     (T target, IDictionary<string, object> propertyValues)
> ```



> I don't really understand. Why does this attribute helps with the trimming? Per my understanding, this attribute instructs the compiler not to trim members of InternalSDKUtils. Why does it help to preserve ApplyValues argument members?...

Thanks for taking a look. That understanding is correct, I must've been looking at a subsequent build that had no warnings because upon recompilation the warning pops up again.. I did try your suggestion though and I get the same warning. I tried a lot of different things as well, but the warning is persistent. Given that this is a static class and the parameter is not a member of the class i don't think there's a way to guarantee that the object passed into the method will have its public properties un-trimmed. I'm starting to wonder if we should just annotate this method with `RequiresUnreferencedCode`.

peterrsongg:
closing for now to reinvestigate

Dreamescaper:
That's strange, adding an attribute to a type parameter solved the issue for me.
What runtime version do you use? dotnet 8/9?

Adding RequiresUnreferencedCode will make S3 client non-AOT compatible, not great.

peterrsongg:
@Dreamescaper I'm using dotnet 8. do you mind pasting your csproj here? maybe we have different settings. And do you mind pasting the output from your terminal as well?

Dreamescaper:
Here's my project zip:
[ConsoleApp1.zip](https://github.com/user-attachments/files/17332999/ConsoleApp1.zip)

```
dotnet publish

Restore complete (1,4s)
  ConsoleApp1 succeeded (18,0s) → bin\Release\net8.0\win-x64\publish\

Build succeeded in 19,9s
```


For comparision - here's same when non-generic version is used:

```
dotnet publish

Restore complete (0,6s)
  ConsoleApp1 succeeded with 1 warning(s) (8,6s) → bin\Release\net8.0\win-x64\publish\
    C:\Users\liakh\source\repos\aws-sdk-net\sdk\src\ConsoleApp1\InternalSDKUtils.cs(39): Trim analysis warning IL2075: Amazon.Util.Internal.InternalSDKUtils.ApplyValues(Object,IDictionary`2<String,Object>): 'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicProperties' in call to 'System.Type.GetProperty(String)'. The return value of method 'System.Object.GetType()' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.

Build succeeded with 1 warning(s) in 9,7s
```

Dreamescaper:
Oh, I probably know. In addition to using generic version, I've replaced `target.GetType()` with `typeof(T)`. That's probably why the warning went away.

peterrsongg:
> Oh, I probably know. In addition to using generic version, I've replaced `target.GetType()` with `typeof(T)`. That's probably why the warning went away.

Ah that was the difference, and the warning went away. Thank you for the suggestion and for catching that the warning was still there! I must've looked at the second cached build which didn't output warnings.  Going to run this through our system to make sure the change is safe.

Issue #3017, opened 2023-08-02
TransferUtility.DownloadDirectoryAsync throws an exception under certain specific conditions


rbluff:
### Describe the bug

`TransferUtility.DownloadDirectoryAsync(...)` throws a AmazonClientException if one of the files inside the S3 directory has a filename that contains a tilde ~ character and is in 8:3 format AND the case of the local destination directory path supplied to `TransferUtilityDownloadDirectoryRequest.LocalDirectory` does not match the actual case of the directory path on the system.

### Expected Behavior

I expected TransferUtility.DownloadDirectoryAsync(...) to successfully download the S3 directory to the local directory without errors.

### Current Behavior

Amazon.Runtime.AmazonClientException: The file C:\\Windows\\TEMP\\abc\\Imports\\01_4791\\Job52\\files\\Audio\\vc~ended.caf is not allowed outside of the target directory C:\\Windows\\TEMP\\abc\\Imports\\01_4791\\Job52.\r\n   at Amazon.S3.Transfer.Internal.DownloadDirectoryCommand.ConstructTransferUtilityDownloadRequest(S3Object s3Object, Int32 prefixLength) in D:\\JenkinsWorkspaces\\trebuchet-stage-release\\AWSDotNetPublic\\sdk\\src\\Services\\S3\\Custom\\Transfer\\Internal\\DownloadDirectoryCommand.cs:line 105\r\n   at Amazon.S3.Transfer.Internal.DownloadDirectoryCommand.<ExecuteAsync>d__22.MoveNext() in D:\\JenkinsWorkspaces\\trebuchet-stage-release\\AWSDotNetPublic\\sdk\\src\\Services\\S3\\Custom\\Transfer\\Internal\\_bcl45+netstandard\\DownloadDirectoryCommand.cs:line 105\r\n--- End of stack trace from previous location where exception was thrown ---

Note that the filename of the file in S3 (`vc~ended.caf`) contains 8 characters in the filename and 3 characters in the extension, and also contains a tilde `~` character.

The directory path supplied to `TransferUtilityDownloadDirectoryRequest.LocalDirectory` was `C:\Windows\TEMP\abc\Imports\01_4791\Job52` but the actual case of the temp folder on the system is `C:\Windows\Temp`. ie. the supplied path contained `TEMP` but the actual path on the system was `Temp`.

### Reproduction Steps

```
public async Task DownloadDirectoryAsync(string sourceDirectoryKey, string destinationDirectoryPath)
{
[...]

PR #3509, opened 2024-10-11
chore: updated the documentation for TransferUtilityDownloadDirectoryRequest.LocalDirectory mentioning that it is case-sensitive.

ashishdhingra:
## Description
Updated the documentation for `TransferUtilityDownloadDirectoryRequest.LocalDirectory` mentioning that it is case-sensitive.

## Motivation and Context
Issue #3017

**Repro Steps:**
- In S3 bucket, create a folder named `DownloadDirectory-Windows83-Test`.
- Upload 2 text files named `basic.txt` and `vc~ended.txt` (this one has Windows 8:3 name) to the above folder.
- Using code below, try to download file to where actual directory on Windows file system is `D:\temp`.
  ```
  TransferUtility directoryTransferUtility = new TransferUtility(amazonS3Client);
  TransferUtilityDownloadDirectoryRequest request = new TransferUtilityDownloadDirectoryRequest
  {
      BucketName = bucketName,
      S3Directory = "DownloadDirectory-Windows83-Test",
      LocalDirectory = @"D:\TEMP\abc\Imports\01_4791\Job52",
      DownloadFilesConcurrently = true,
  };
  await directoryTransferUtility.DownloadDirectoryAsync(request).ConfigureAwait(false);
  ```

**Findings:**
The issue appears to be with `System.IO.FileInfo` at [InternalSDKUtils.IsFilePathRootedWithDirectoryPath()](https://github.com/aws/aws-sdk-net/blob/ddb9b60e8b804f561ffb5415eb729ecc3ca2582a/sdk/src/Core/Amazon.Util/Internal/InternalSDKUtils.cs#L240) where:
- For normal file (e.g. Somefile.txt), it returns `C:\Windows\TEMP\SomeDirectory\Somefile.txt` in `FullName` property.
- For file containing `~` character (e.g. `vc~ended.caf`), it returns `C:\Windows\Temp\SomeDirectory\vc~ended.caf` in `FullName` property (the `OriginalPath` property of `FileInfo` object has casing supplied by user).

We should not use `StringComparison.InvariantCultureIgnoreCase` (as proposed in issue description) since on Linux, the file system is case sensitive, and we would want to match the exact directory name.

We cannot access `OriginalPath` property since it is `protected` in base [FileSystemInfo](https://learn.microsoft.com/en-us/dotnet/api/system.io.filesysteminfo.originalpath?view=netframework-4.8.1#system-io-filesysteminfo-originalpath) class.

On Windows, there is a registry setting which when set, makes file system case-sensitive (just like Linux). Refer [Enabling Case Sensitivity on Windows 10](https://www.ortussolutions.com/blog/how-to-make-windows-folders-case-sensitive) for example, which has example command `fsutil.exe file SetCaseSensitiveInfo C:\folder\path enable`.

So it's better to update API documentation for `TransferUtilityDownloadDirectoryRequest.LocalDirectory` mentioning that it is case-sensitive for certain platforms/scenerios.

## Testing
Refer repro steps above. No dry-run is necessary for API doc only change.

## Types of changes

## Checklist

## License

### DIFF
--- a/generator/.DevConfigs/5a43ed01-13a4-4f03-afb5-209de16e31c3.json
+++ b/generator/.DevConfigs/5a43ed01-13a4-4f03-afb5-209de16e31c3.json
@@ -1,0 +1,9 @@
+{
+  "services": [
+    {
+      "serviceName": "S3",
+      "type": "patch",
+      "changeLogMessages": [ "Updated the documentation for TransferUtilityDownloadDirectoryRequest.LocalDirectory mentioning that it is case-sensitive." ]
+    }
+  ]
+}
--- a/sdk/src/Services/S3/Custom/Transfer/TransferUtilityDownloadDirectoryRequest.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/TransferUtilityDownloadDirectoryRequest.cs
@@ -82,6 +82,7 @@ namespace Amazon.S3.Transfer
         /// <summary>
         /// 	Gets or sets the local directory where objects from Amazon S3 will be downloaded.  
 		/// 	If the directory doesn't exist, it will be created.
+        /// 	For some platforms like Linux, the file system is case-sensitive. The correct casing to the actual path must be used.
         /// </summary>
         /// <value>
         /// 	The local directory where objects from Amazon S3 will be downloaded.


generator/.DevConfigs/5a43ed01-13a4-4f03-afb5-209de16e31c3.json
```diff
@@ -0,0 +1,9 @@
+{
+  "services": [
+    {
+      "serviceName": "S3",
+      "type": "patch",
+      "changeLogMessages": [ "chore: updated the documentation for TransferUtilityDownloadDirectoryRequest.LocalDirectory mentioning that it is case-sensitive." ]
```
dscpinheiro, line 6: I'd remove `chore: ` from here (we don't normally use the conventional commit format in the changelog).

ashishdhingra: Done

sdk/src/Services/S3/Custom/Transfer/TransferUtilityDownloadDirectoryRequest.cs
```diff
@@ -82,6 +82,8 @@ internal bool IsSetBucketName()
         /// <summary>
         /// 	Gets or sets the local directory where objects from Amazon S3 will be downloaded.  
 		/// 	If the directory doesn't exist, it will be created.
+        /// 	Kindly note that this is case-sensitive for some platforms/scenarios, so this should
```
normj, line 85: I appreciate your politeness but "Kindly note" doesn't really fit the tone of API docs.

```
For some platforms like Linux the file system is case-sensitive. The correct casing to the actual path must be used.
```

normj (changes requested):

sdk/src/Services/S3/Custom/Transfer/TransferUtilityDownloadDirectoryRequest.cs
ashishdhingra: Updated

normj (approved):

dscpinheiro (approved):

PR #3513, opened 2024-10-15
fix: update AWS_S3_USE_ARN_REGION to take precedence over value set i…

peterrsongg:
…n config file

## Description

This updates AmazonS3Config use-arn-region to take the value set in environment variable over that set in the ~/.aws/config file. This follows other env var + config file precedence.

## Motivation and Context

dotnet-7073
## Testing

dry run in v4 passes
Added unit tests
## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/AmazonS3Config.cs
+++ b/sdk/src/Services/S3/Custom/AmazonS3Config.cs
@@ -35,13 +35,14 @@ namespace Amazon.S3
         private const string DefaultProfileName = "default";
         private const string AwsS3UsEast1RegionalEndpointsEnvironmentVariable = "AWS_S3_US_EAST_1_REGIONAL_ENDPOINT";
         private const string DisableMRAPEnvName = "AWS_S3_DISABLE_MULTIREGION_ACCESS_POINTS";
+        private const string AwsConfigFileEnvName = "AWS_CONFIG_FILE";
 
         private bool forcePathStyle = false;
         private bool useAccelerateEndpoint = false;
         private S3UsEast1RegionalEndpointValue? s3UsEast1RegionalEndpointValue;
         private readonly string legacyUSEast1GlobalRegionSystemName = RegionEndpoint.USEast1.SystemName;
 
-        private static CredentialProfileStoreChain credentialProfileChain = new CredentialProfileStoreChain();
+        private static CredentialProfileStoreChain credentialProfileChain = new CredentialProfileStoreChain(Environment.GetEnvironmentVariable(AwsConfigFileEnvName));
 
         // we cache this per execution process to avoid excessive file I/O
         private static CredentialProfile _profile;
@@ -114,14 +115,17 @@ namespace Amazon.S3
                         return _useArnRegion.Value;
                     }
 
-                    _useArnRegion = _profile?.S3UseArnRegion;
-
-                    if (!_useArnRegion.HasValue && !string.IsNullOrEmpty(Environment.GetEnvironmentVariable(UseArnRegionEnvName)))
+                    if (!string.IsNullOrEmpty(Environment.GetEnvironmentVariable(UseArnRegionEnvName)))
                     {
                         if (bool.TryParse(Environment.GetEnvironmentVariable(UseArnRegionEnvName), out var value))
                         {
                             _useArnRegion = value;
                         }
+                    }
+
+                    if (!_useArnRegion.HasValue)
+                    {
+                        _useArnRegion = _profile?.S3UseArnRegion;
                     }
 
                     if (!_useArnRegion.HasValue)
--- a/sdk/src/Services/S3/Custom/AmazonS3Config.cs
+++ b/sdk/src/Services/S3/Custom/AmazonS3Config.cs
@@ -42,7 +42,7 @@ namespace Amazon.S3
         private S3UsEast1RegionalEndpointValue? s3UsEast1RegionalEndpointValue;
         private readonly string legacyUSEast1GlobalRegionSystemName = RegionEndpoint.USEast1.SystemName;
 
-        private static CredentialProfileStoreChain credentialProfileChain = new CredentialProfileStoreChain(Environment.GetEnvironmentVariable(AwsConfigFileEnvName));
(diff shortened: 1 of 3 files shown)


sdk/src/Services/S3/Custom/AmazonS3Config.cs
```diff
@@ -35,13 +35,14 @@ public partial class AmazonS3Config : ClientConfig
...
         private const string DisableMRAPEnvName = "AWS_S3_DISABLE_MULTIREGION_ACCESS_POINTS";
+        private const string AwsConfigFileEnvName = "AWS_CONFIG_FILE";
 
         private bool forcePathStyle = false;
         private bool useAccelerateEndpoint = false;
         private S3UsEast1RegionalEndpointValue? s3UsEast1RegionalEndpointValue;
         private readonly string legacyUSEast1GlobalRegionSystemName = RegionEndpoint.USEast1.SystemName;
 
-        private static CredentialProfileStoreChain credentialProfileChain = new CredentialProfileStoreChain();
+        private static CredentialProfileStoreChain credentialProfileChain = new CredentialProfileStoreChain(Environment.GetEnvironmentVariable(AwsConfigFileEnvName));
```
peterrsongg, line 45: I'm open to leaving this as is. We added support for `AWS_CONFIG_FILE` a couple years ago and when this class was created, this environment variable wasn't available. Now that it is, it makes sense for the `CredentialProfileStoreChain` to check this value to see if it should look somewhere else for the config file. 

Given that the constructor accepts a `ProfileLocation`, I thought it made sense to add it here. If it isn't set, then it will just continue to work as it did previously.

sdk/test/Services/S3/UnitTests/Custom/UseArnRegionTests.cs
```diff
@@ -68,90 +96,73 @@ public void ExplicitlyEnable()
         [TestCategory("S3")]
         public void EnvironmentVariableEnable()
         {
-            Environment.SetEnvironmentVariable("AWS_S3_USE_ARN_REGION", "true");
-            try
-            {
-                var config = new AmazonS3Config
-                {
-                };
+            Environment.SetEnvironmentVariable(AWS_S3_USE_ARN_REGION_ENVIRONMENT_VARIABLE, "true");
```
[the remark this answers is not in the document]
peterrsongg: The reason I removed the `try / finally` is because even if a test fails, the next test will just reset the environment variable before continuing with its logic. But I will add it back to be more defensive. Doesn't hurt.

For example we don't do this in the `DisableS3ExpressSessionAuthTests`.

sdk/src/Services/S3/Custom/AmazonS3Config.cs
peterrsongg: Okay, given there are other services that need the update, we can do it all in a separate pr👍

sdk/test/Services/S3/UnitTests/Custom/UseArnRegionTests.cs
```diff
@@ -39,18 +39,49 @@ namespace AWSSDK.UnitTests
...
+            _beginningAWSProfileEnvironmentValue = Environment.GetEnvironmentVariable(AWS_PROFILE_ENVIRONMENT_VARIABLE);
+            // Then clear the current value so every test is starting from a clean slate
+            Environment.SetEnvironmentVariable(AWS_PROFILE_ENVIRONMENT_VARIABLE, "");
+            // set credentials file and use it to load CredentialProfileStoreChain
+
+            // set credentials file and use it to load CredentialProfileStoreChain
+            _tempCredentialsFilePath = Path.GetTempFileName();
+            File.WriteAllText(_tempCredentialsFilePath, ProfileText);
+            ReflectionHelpers.Invoke(typeof(AmazonS3Config), "credentialProfileChain", new CredentialProfileStoreChain(_tempCredentialsFilePath));
+            ReflectionHelpers.Invoke(typeof(AmazonS3Config), "_triedToResolveProfile", false);
```
dscpinheiro, line 75: Sorry I didn't notice this during my initial review, but why is using reflection to set a private property required? Is it because you uncommented the last 2 tests in this file?

peterrsongg: This wasn't here in the initial review. This is something I added after removing the `AWS_CONFIG_FILE` option in the CredentialsProfileChain constructor in `AmazonS3Config`. Now that we don't have the option to specify via environment variable, we must use reflection so the `CredentialProfileChain` knows where to look

sdk/src/Services/S3/Custom/AmazonS3Config.cs
```diff
@@ -35,6 +35,7 @@ public partial class AmazonS3Config : ClientConfig
         private const string DefaultProfileName = "default";
         private const string AwsS3UsEast1RegionalEndpointsEnvironmentVariable = "AWS_S3_US_EAST_1_REGIONAL_ENDPOINT";
         private const string DisableMRAPEnvName = "AWS_S3_DISABLE_MULTIREGION_ACCESS_POINTS";
+        private const string AwsConfigFileEnvName = "AWS_CONFIG_FILE";
```
dscpinheiro, line 38: Remove this since it's not used anymore.

sdk/test/Services/S3/UnitTests/Custom/UseArnRegionTests.cs
dscpinheiro: Discussed offline, but this is the same pattern used in https://github.com/aws/aws-sdk-net/blob/68f6a0520854f18fd991fd87cfd847995b075ab3/sdk/test/Services/S3/UnitTests/Custom/DisableS3ExpressSessionAuthTests.cs (another S3 option that can be set either via environment variable or config).

We should follow the same pattern for the tests here (just verify the precedence is correct), and when support for `AWS_CONFIG_FILE` is added remove the reflection usages.

dscpinheiro (approved):

normj (approved):
Be sure to update the V4 tracker issue

PR #3514, opened 2024-10-16
Add Observability to DynamoDBContext operations is DynamoDB HLL

96malhar:
## Description

This PR adds observability to to `DynamoDBContext` operations is DynamoDB HLL by creating traces using the SDK observability APIs.
* Traces are only created for public APIs that make a network call.
* I have updated the Async public APIs to await the `Task` instead of returning it. This is done to ensure that the trace is not disposed before the function completes its execution.

## Motivation and Context

This change is required as part of the Observability initiative for the SDK and HLLs
DOTNET-7679

## Testing

* All existing Unit and Integrations tests pass locally
* **Pending a full dry-run**. Since I am targeting a feature branch, I'll perform a full dry run on the feature branch itself after observability updates are merged in for both the data model and document model operations

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/DynamoDBv2/Custom/DataModel/_async/Context.Async.cs
+++ b/sdk/src/Services/DynamoDBv2/Custom/DataModel/_async/Context.Async.cs
@@ -41,9 +41,12 @@ namespace Amazon.DynamoDBv2.DataModel
         /// <param name="value">Object to save.</param>
         /// <param name="cancellationToken">Token which can be used to cancel the task.</param>
         /// <returns>A Task that can be used to poll or wait for results, or both.</returns>
-        public Task SaveAsync<T>(T value, CancellationToken cancellationToken = default(CancellationToken))
+        public async Task SaveAsync<T>(T value, CancellationToken cancellationToken = default(CancellationToken))
         {
-            return SaveHelperAsync(value, null, cancellationToken);
+            using (DynamoDBTelemetry.CreateSpan(this, nameof(SaveAsync)))
+            {
+                await SaveHelperAsync(value, null, cancellationToken).ConfigureAwait(false);
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


sdk/src/Services/DynamoDBv2/Custom/DataModel/_async/Context.Async.cs
```diff
@@ -41,9 +41,12 @@ public partial class DynamoDBContext : IDynamoDBContext
         /// <param name="value">Object to save.</param>
         /// <param name="cancellationToken">Token which can be used to cancel the task.</param>
         /// <returns>A Task that can be used to poll or wait for results, or both.</returns>
-        public Task SaveAsync<T>(T value, CancellationToken cancellationToken = default(CancellationToken))
+        public async Task SaveAsync<T>(T value, CancellationToken cancellationToken = default(CancellationToken))
         {
-            return SaveHelperAsync(value, null, cancellationToken);
+            using (DynamoDBTelemetry.CreateSpan(this, nameof(SaveAsync)))
+            {
+                await SaveHelperAsync(value, null, cancellationToken).ConfigureAwait(false);
```
philasmar, line 48: Wouldn't it be better if you just create the span inside `SaveHelperAsync` instead of duplicating the code in the 4 `SaveAsync` overloads?

philasmar: This comment applies to all the other operations

96malhar: I though about this too but I am worried that in the future when we expand this library, we may add a new public composite operation that uses all these private methods and we'd end up setting the wrong operation name in the traces.

For example, if we create the trace inside `SaveHelperAsync` we'd still want to set the operation name to `SaveAsync` since that is the public API that customers use. 

In the future if we create a new composite operation called `Foo` that calls `LoadHelpersAsync`, manipulates the returned value and then again calls `SaveHelpersAsync`, we'd end up creating traces for `LoadAsync` and `SaveAsync` even though these operations were not called directly.

normj (approved):

philasmar (approved):

PR #3531, opened 2024-10-30
[SRA] Create Identity Resolvers

muhammad-othman:
## Description
* Added `IIdentityResolverConfiguration` with its default implementation and used it to retrieve the appropriate `IdentityResolver` in `AuthScheme.GetIdentityResolver`.
* Added `DefaultAWSCredentialsIdentityResolver` and `AnonymousIdentityResolver`.

## Motivation and Context
#DOTNET-7539

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Credentials/Internal/IdentityResolvers/AnonymousIdentityResolver.cs
+++ b/sdk/src/Core/Amazon.Runtime/Credentials/Internal/IdentityResolvers/AnonymousIdentityResolver.cs
@@ -15,14 +15,19 @@
 
 using Smithy.Identity.Abstractions;
 
-namespace Amazon.Runtime.Credentials.Internal
+namespace Amazon.Runtime.Credentials.Internal.IdentityResolvers
 {
-    public class DefaultAWSCredentialsIdentityResolver : IIdentityResolver
+    /// <summary>
+    /// A resolver that provides an anonymous identity. 
+    /// </summary>
+    public class AnonymousIdentityResolver : IIdentityResolver
     {
+        /// <summary>
+        /// Resolves the identity by returning an instance of <see cref="AnonymousAWSCredentials"/>.
+        /// </summary>
         public BaseIdentity ResolveIdentity()
         {
-            // TODO: This solves the same problem as the FallbackCredentialsFactory.
-            throw new System.NotImplementedException();
+            return new AnonymousAWSCredentials();
         }
     }
 }
--- a/sdk/src/Core/Amazon.Runtime/Credentials/Internal/IdentityResolvers/DefaultAWSCredentialsIdentityResolver.cs
+++ b/sdk/src/Core/Amazon.Runtime/Credentials/Internal/IdentityResolvers/DefaultAWSCredentialsIdentityResolver.cs
@@ -1,0 +1,190 @@
+﻿/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+using Amazon.Runtime.CredentialManagement;
+using Amazon.Runtime.Internal.Util;
+using Smithy.Identity.Abstractions;
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.IO;
+using System.Security;
+using System.Threading;
+
+namespace Amazon.Runtime.Credentials.Internal.IdentityResolvers
+{
+    /// <summary>
+    /// A resolver that provides an AWSCredentials identity. 
+    /// </summary>
+    public class DefaultAWSCredentialsIdentityResolver : IIdentityResolver
+    {
+        private const string AWS_PROFILE_ENVIRONMENT_VARIABLE = "AWS_PROFILE";
+        private const string DEFAULT_PROFILE_NAME = "default";
+
+        private static ReaderWriterLockSlim _cachedCredentialsLock = new ReaderWriterLockSlim();
+        private delegate AWSCredentials CredentialsGenerator();
+        private AWSCredentials _cachedCredentials;
+        private List<CredentialsGenerator> _credentialsGenerators { get; set; }
+        private readonly CredentialProfileStoreChain _credentialProfileChain = new CredentialProfileStoreChain();
+
+        public DefaultAWSCredentialsIdentityResolver()
+        {
+            try
+            {
+                _cachedCredentialsLock.EnterWriteLock();
+                _cachedCredentials = null;
+                _credentialsGenerators = new List<CredentialsGenerator>
+                {
+#if BCL
+                () => new AppConfigAWSCredentials(), // Test explicit keys/profile name first.
+#endif
+                    () => AssumeRoleWithWebIdentityCredentials.FromEnvironmentVariables(),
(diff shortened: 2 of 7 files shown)


sdk/src/Core/Amazon.Runtime/Credentials/Internal/IdentityResolvers/DefaultAWSCredentialsIdentityResolver.cs
```diff
@@ -0,0 +1,191 @@
...
+                cachedCredentialsLock.ExitWriteLock();
+            }
+        }
+
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Design", "CA1031:Do not catch general exception types", Justification = "We need to catch all exceptions to be able to move the the next generator.")]
+        public BaseIdentity ResolveIdentity()
+        {
+            try
+            {
+                cachedCredentialsLock.EnterReadLock();
```
muhammad-othman, line 71: I will need to double check if locking is required with this singleton implementation, probably I will only need to add locking when creating the singleton instance.

sdk/src/Core/Amazon.Runtime/Credentials/Internal/AnonymousAuthScheme.cs
```diff
@@ -25,8 +25,9 @@ public class AnonymousAuthScheme : IAuthScheme<AnonymousAWSCredentials>
         /// <inheritdoc/>
         public string SchemeId => "smithy.api#noAuth";
 
-#pragma warning disable CA1065
-        public IIdentityResolver IdentityResolver => throw new System.NotImplementedException();
-#pragma warning restore CA1065
+        public IIdentityResolver GetIdentityResolver(IIdentityResolverConfiguration configuration)
```
dscpinheiro, line 28: Missing `<inheritdoc/>`? (applies to all auth schemes)

sdk/src/Core/Amazon.Runtime/Credentials/Internal/IdentityResolvers/IdentityResolverConfiguration.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        public static readonly IIdentityResolverConfiguration Instance = new DefaultIdentityResolverConfiguration();
+        private static readonly Dictionary<Type, IIdentityResolver> identityResolvers = new Dictionary<Type, IIdentityResolver>()
+            {
+                { typeof(AnonymousAWSCredentials), new AnonymousIdentityResolver() },
+                { typeof(AWSCredentials), new DefaultAWSCredentialsIdentityResolver() },
+            };
+
+        /// <inheritdoc/>
+        public IIdentityResolver GetIdentityResolver<T>() where T : BaseIdentity
```
dscpinheiro, line 33: Does this need to be annotated for AOT?

sdk/src/Core/Amazon.Runtime/Credentials/Internal/IdentityResolvers/AnonymousIdentityResolver.cs
```diff
@@ -15,14 +15,20 @@
 
 using Smithy.Identity.Abstractions;
 
-namespace Amazon.Runtime.Credentials.Internal
+namespace Amazon.Runtime.Credentials.Internal.IdentityResolvers
 {
-    public class DefaultAWSCredentialsIdentityResolver : IIdentityResolver
+    /// <summary>
+    /// A resolver that provides an anonymous identity. 
+    /// This implementation is used when scheme ID is noAuth,
```
dscpinheiro, line 22: I don't think this line is necessary here (same applies to line 30 of `DefaultAWSCredentialsIdentityResolver`); the resolvers don't care about auth schemes (and I feel like if they ever change we will forget to update this comment).

sdk/src/Core/Amazon.Runtime/Credentials/Internal/IdentityResolvers/DefaultAWSCredentialsIdentityResolver.cs
muhammad-othman: nvm, I don't think we should remove the locking.

sdk/src/Core/Amazon.Runtime/Credentials/Internal/IdentityResolvers/IdentityResolverConfiguration.cs
muhammad-othman: The identities classes shouldn't get trimmed since they are used by other types, but it might cause a warning, let me check.

muhammad-othman: Created a console app with `PublishTrimmed` flag set to `true`, and used the resolver to get an `AnonymousAWSCredentials`
```
 var config = DefaultIdentityResolverConfiguration.Instance;

 var resolver = config.GetIdentityResolver<AnonymousAWSCredentials>();

 var crd = resolver.ResolveIdentity();

 Console.WriteLine(crd.ToString());
 Console.WriteLine("Expiration");
 Console.WriteLine(crd.Expiration);
 ```
 
No trimming warnings appeared.

File: sdk/src/Core/Amazon.Runtime/Credentials/Internal/IdentityResolvers/DefaultAWSCredentialsIdentityResolver.cs [hunk 5]
```diff
@@ -0,0 +1,190 @@
...
+                }
+
+                List<Exception> errors = new List<Exception>();
+                foreach (CredentialsGenerator generator in _credentialsGenerators)
+                {
+                    try
+                    {
+                        _cachedCredentials = generator();
+                    }
+                    // Breaking the FallbackCredentialFactory chain in case a ProcessAWSCredentialException exception
```
dscpinheiro, line 96:
```suggestion
                    // Breaking the chain in case a ProcessAWSCredentialException exception 
```

sdk/src/Core/Amazon.Runtime/Credentials/Internal/IdentityResolvers/DefaultAWSTokenIdentityResolver.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        public DefaultAWSTokenIdentityResolver()
+        {
+            AWSTokenProvider = new AWSTokenProviderChain(new ProfileTokenProvider());
+        }
+
+        public BaseIdentity ResolveIdentity()
+        {
+            AWSToken token;
+#if BCL
```
muhammad-othman, line 40: This kinda pushes us to create an `async` version of `ResolveIdentity`, we can always add it later if we don't want to diverge allot from the SRA.

dscpinheiro: We should not do this (look how long it took to fix the blocking call we had in the DynamoDB HLL). Honestly I'd prefer if we did all SSO related changes in its PR (since we may have to adapt the existing components we have in the SDK).

sdk/src/Core/Amazon.Runtime/Credentials/Internal/IdentityResolvers/DefaultAWSTokenIdentityResolver.cs
```diff
@@ -0,0 +1,60 @@
...
+            if (!AWSTokenProvider.TryResolveToken(out token))
+            {
+                throw new AmazonClientException("Failed to resolve AWSToken using the configured AWSTokenProvider in DefaultAWSTokenIdentityResolver.");
+            }
+#endif
+
+#if AWS_ASYNC_API
+            var tokenResponse = AWSTokenProvider.TryResolveTokenAsync().Result;
```
dscpinheiro, line 48: We should not do this (look how long it took to fix the blocking call we had in the DynamoDB HLL). Honestly I'd prefer if we did all SSO related changes in its PR (since we may have to adapt the existing components we have in the SDK).

normj (approved):

dscpinheiro (approved):

PR #3532, opened 2024-10-30
Do not trim leading slashes for copy object and copy part operations

peterrsongg:
…perations

## Description

The .NET SDK trims leading slashes on copy object and copy part operations. This behavior can be turned off by setting `DisableTrimmingLeadingSlash` to true. This change removes that flag so that the .NET SDK will not trim leading slashes.

## Motivation and Context

This is something we can only fix in v4
## Testing

dry run in v4 passes.

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/test/Services/S3/IntegrationTests/CopyObjectTests.cs
+++ b/sdk/test/Services/S3/IntegrationTests/CopyObjectTests.cs
@@ -157,12 +157,12 @@ namespace AWSSDK_DotNet.IntegrationTests.Tests.S3
             Assert.IsTrue(taggingMetadata.Tagging.Any(tag => tag.Key == "newtag1" && tag.Value == "1"));
         }
 
-        [DataRow(false, testKey, "/destinationTestKey1.txt", "destinationTestKey1.txt")]
-        [DataRow(true, testKeyWithSlash, "/destinationTestKey2.txt", "/destinationTestKey2.txt")]
-        [DataRow(true, testKeyWithSlash, "/", "/")]
+        [DataRow(testKey, "/destinationTestKey1.txt", "/destinationTestKey1.txt")]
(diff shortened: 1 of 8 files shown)


dscpinheiro (approved):

sdk/test/Services/S3/IntegrationTests/CopyObjectTests.cs
```diff
@@ -157,12 +157,12 @@ public void TestCopyObjectWithTagsReplace()
             Assert.IsTrue(taggingMetadata.Tagging.Any(tag => tag.Key == "newtag1" && tag.Value == "1"));
         }
 
-        [DataRow(false, testKey, "/destinationTestKey1.txt", "destinationTestKey1.txt")]
-        [DataRow(true, testKeyWithSlash, "/destinationTestKey2.txt", "/destinationTestKey2.txt")]
-        [DataRow(true, testKeyWithSlash, "/", "/")]
+        [DataRow(testKey, "/destinationTestKey1.txt", "/destinationTestKey1.txt")]
```
muhammad-othman, line 160: nit: I think it is better to remove the slash from the `destinationKey` instead of adding it to the `expectedKey` to make this test case different than the second one, or remove the slashes from the next test case.

muhammad-othman (approved):

sdk/test/Services/S3/IntegrationTests/CopyObjectTests.cs
peterrsongg: done

dscpinheiro: I think this test could be even better if the data row was:
```
[DataRow(testKeyWithSlash, "destinationTestKey1.txt", "destinationTestKey1.txt")]
```

There's already another test case in this class copying `testKey` somewhere else. The code above will validate the leading slash is handled properly.

PR #3534, opened 2024-10-31
Support decimal precision for Document.FromJson and ToJson by using System.Text.Json instead of LitJson

normj:
## Description
Taking over @ashovlin previous PR switching the `ToJson` and `FromJson` methods in DynamoDB to use System.Text.json instead of LitJson. https://github.com/aws/aws-sdk-net/pull/3447

I have double check data type conversions like doubles, dates, bools and binary data comparing with V3 behavior and the behavior was consistent except for the expect use of `Decimal` for larger precision and the values being cast to `Decimal`. 

## Motivation and Context
Avoid losing precision for floating point numbers with a lot of digits. It doesn't solve all use cases because .NET's `Decimal` type doesn't support as high an exponent as `double` does. If the floating point number is using a high exponent it will use a double and possibly lose precision. This is limitation of floating point numbers in .NET because .NET doesn't have a type like Java's BigDecimal.

## Testing
Had a successful dry run. Tests were added in @ashovlin initial work. Also did spot checking of data types.

### DIFF
--- a/sdk/src/Services/DynamoDBv2/Custom/DocumentModel/JsonUtils.cs
+++ b/sdk/src/Services/DynamoDBv2/Custom/DocumentModel/JsonUtils.cs
@@ -16,12 +16,9 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
-using System.Text;
-
-using Amazon.DynamoDBv2.Model;
-using Amazon.Util;
-using ThirdParty.Json.LitJson;
 using System.IO;
+using System.Text;
+using System.Text.Json;
 
 namespace Amazon.DynamoDBv2.DocumentModel
 {
@@ -40,11 +37,18 @@ namespace Amazon.DynamoDBv2.DocumentModel
         /// <returns></returns>
         public static Document FromJson(string jsonText)
         {
-            var json = JsonMapper.ToObject(jsonText);
-            if (!json.IsObject)
-                throw new InvalidOperationException("Expected object at JSON root.");
+            var json = JsonDocument.Parse(jsonText, new JsonDocumentOptions
+            {
+                AllowTrailingCommas = true,
+                CommentHandling = JsonCommentHandling.Skip
+            });
 
-            var document = ToEntry(json, DynamoDBEntryConversion.V2) as Document;
+            if (json.RootElement.ValueKind != JsonValueKind.Object)
+            {
+                throw new InvalidOperationException("Expected object at JSON root.");
+            }
+
+            var document = ToEntry(json.RootElement, DynamoDBEntryConversion.V2) as Document;
             if (document == null)
                 throw new InvalidOperationException();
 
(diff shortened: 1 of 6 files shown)


sdk/src/Services/DynamoDBv2/Custom/DocumentModel/JsonUtils.cs
```diff
@@ -320,71 +330,81 @@ internal static void WriteJson(DynamoDBEntry entry, JsonWriter writer, DynamoDBE
...
         {
             var stringValue = value as string;
 
             switch (type)
             {
                 case DynamoDBEntryType.Numeric:
-                    writer.WriteRaw(stringValue);
-                    break;
+                    {
+#if NETCOREAPP3_1  // WriteRawValue was added in .NET 6, but we need to write out Number values without quotes
```
dscpinheiro, line 384: Question: Does this mean for the .NET Framework target we'd try to use `WriteRawValue` too (which is only available in .NET 6)? Or is it fine because we have `System.Text.Json` as a dependency for `net472`?

ashovlin: It's [available in 4.7.2](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonwriter.writerawvalue?view=net-8.0#system-text-json-utf8jsonwriter-writerawvalue(system-string-system-boolean). At the time 3.1 was the only target we needed this workaround for.

dscpinheiro (approved):

sdk/src/Services/DynamoDBv2/Custom/DocumentModel/JsonUtils.cs
```diff
@@ -40,11 +37,18 @@ internal static class JsonUtils
         /// <returns></returns>
         public static Document FromJson(string jsonText)
         {
-            var json = JsonMapper.ToObject(jsonText);
-            if (!json.IsObject)
+            var json = JsonDocument.Parse(jsonText, new JsonDocumentOptions
+            {
+                AllowTrailingCommas = true,
```
peterrsongg, line 42: In your description you say 
"A slight change in behavior is the JSON being past into the FromJson is more strictly parsed. Meaning it doesn't allow extra , for the last property in an object. LitJson would allow that. Given that isn't valid JSON this seems an acceptable change for V4."

but here you are allowing trailing commas. Did you set `AllowTrailingCommas` to true on purpose to keep the same behavior as litJson or did you mean to set this to `false`?

sdk/src/Services/DynamoDBv2/Custom/DocumentModel/JsonUtils.cs
```diff
@@ -320,71 +330,81 @@ internal static void WriteJson(DynamoDBEntry entry, JsonWriter writer, DynamoDBE
...
             var stringValue = value as string;
 
             switch (type)
             {
                 case DynamoDBEntryType.Numeric:
-                    writer.WriteRaw(stringValue);
-                    break;
+                    {
+#if NETCOREAPP3_1  // WriteRawValue was added in .NET 6, but we need to write out Number values without quotes
+                        using var document = JsonDocument.Parse(stringValue);
```
peterrsongg, line 385: I think there is an alternative, simpler solution here which is to use something like:
```
var number = Convert.ToDouble(stringValue);
writer.WriteNumberValue(number);
```

This way, we don't need to have a preprocessor directive. But maybe I'm missing something and there is a reason why it was done this way.😅

peterrsongg (commented):
one minor comment and one optimization suggestion

sdk/src/Services/DynamoDBv2/Custom/DocumentModel/JsonUtils.cs
peterrsongg: Is this what the user was complaining about for losing decimal precision?

Lanayx:
Wow, looking forward to this PR. Just stumbled upon JSON value "0.0000087" being converted to "8.7e-06" after applying Document.FromJson

sdk/src/Services/DynamoDBv2/Custom/DocumentModel/JsonUtils.cs
normj: Oh yeah. I copied the comment from Alex's original PR he must have added setting that property afterwards.

sdk/src/Services/DynamoDBv2/Custom/DocumentModel/JsonUtils.cs
normj: We would potentially lose precision trying to convert the string version of the number to the best fit .NET numeric type. Overall I don't think that will simplify the logic.

peterrsongg (approved):

normj:
@Lanayx Glad this change will help you out. This is merged into our next major version which we are are currently on preview 4. We are hoping to a preview 5 within the next couple weeks that will include this change.

PR #3540, opened 2024-11-01
[SRA] Rules-based authentication resolvers

dscpinheiro:
## Description
This PR focuses on rules-based resolvers (which delegate authentication resolution to their EP2.0 resolvers). Only a handful of services (such as `S3` and `EventBridge`) will use this code path.

## Motivation and Context
`DOTNET-7651`

## Testing
Built the `AutoScaling` and `S3` solutions locally (the handlers are not connected to the request pipeline yet).

## License

### DIFF
--- a/generator/ServiceClientGeneratorLib/Generators/AuthResolvers/ModeledResolver.tt
+++ b/generator/ServiceClientGeneratorLib/Generators/AuthResolvers/ModeledResolver.tt
@@ -35,14 +35,33 @@ namespace <#=this.Config.Namespace#>.Internal
     /// </summary>
     public class Amazon<#=this.Config.ClassName#>AuthSchemeHandler : BaseAuthResolverHandler
     {
+<# if (IsServiceAllowListedForRulesBasedResolver()) { #>
+        private readonly Amazon<#=this.Config.ClassName#>EndpointResolver _endpointResolver = new();
+
+<# } #>
         /// <summary>
-        /// Auth scheme resolver for AutoScaling.
+        /// Modeled auth scheme resolver for <#=this.Config.ClassName#>.
         /// </summary>
         public Amazon<#=this.Config.ClassName#>AuthSchemeResolver AuthSchemeResolver { get; } = new();
 
         /// <inheritdoc/>
-        protected override List<IAuthSchemeOption> ResolveAuthOptions(IRequestContext requestContext)
+        protected override List<IAuthSchemeOption> ResolveAuthOptions(IExecutionContext executionContext)
         {
+<# if (IsServiceAllowListedForRulesBasedResolver()) { #>
+            // Since <#=this.Config.ClassName#> includes auth schemes in its endpoint rules, we'll attempt to delegate resolution to the endpoint
+            // resolver first (falling back to the modeled resolver if no options are returned).
+            var endpoint = _endpointResolver.GetEndpoint(executionContext);
+
+            // This means the endpoints resolver is executed twice intentionally (at this point and then later in the pipeline
+            // to determine which endpoint the SDK should use for the request).
+            var endpointAuthSchemes = RetrieveSchemesFromEndpoint(endpoint);
+            if (endpointAuthSchemes != null)
+            {
+                return endpointAuthSchemes;
+            }
+
+<# } #>
+            var requestContext = executionContext.RequestContext;
             var mappedParameters = new Amazon<#=this.Config.ClassName#>AuthSchemeParameters
             {
                 Operation = requestContext.Request.RequestName,
--- a/sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs
+++ b/sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs
@@ -14,7 +14,9 @@
 */
 
 using Amazon.Runtime.Credentials.Internal;
+using Amazon.Runtime.Endpoints;
 using Smithy.Identity.Abstractions;
+using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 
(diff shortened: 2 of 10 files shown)


generator/ServiceClientGeneratorLib/Generators/AuthResolvers/ModeledResolver.tt
```diff
@@ -35,14 +35,33 @@ namespace <#=this.Config.Namespace#>.Internal
...
+<# } #>
         /// <summary>
-        /// Auth scheme resolver for AutoScaling.
+        /// Modeled auth scheme resolver for <#=this.Config.ClassName#>.
         /// </summary>
         public Amazon<#=this.Config.ClassName#>AuthSchemeResolver AuthSchemeResolver { get; } = new();
 
         /// <inheritdoc/>
-        protected override List<IAuthSchemeOption> ResolveAuthOptions(IRequestContext requestContext)
+        protected override List<IAuthSchemeOption> ResolveAuthOptions(IExecutionContext executionContext)
```
dscpinheiro, line 48: I wanted to get feedback on this approach I took: The SRA docs mention we _should code generate the mapping from auth scheme parameters to endpoint resolver parameters_, however that was a lot of duplication for S3. By calling the generated resolver instead, we'll use all endpoint parameters anyways (including new ones we might add in the future): https://github.com/aws/aws-sdk-net/blob/36af0432b27eda1efb55c8f3ed6c8c11326f7ddd/sdk/src/Services/S3/Generated/Internal/AmazonS3EndpointResolver.cs#L70-L82

One benefit is that it allowed me to keep the implementation to retrieve auth schemes from an endpoint (in `sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs`) simpler.

normj: How would rerunning the endpoint resolver twice work when the SDK supports account id based endpoint rules? Because we would need the identity to get the account id for the endpoint rules I believe.

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs
```diff
@@ -65,9 +67,66 @@ protected void PreInvoke(IExecutionContext executionContext)
...
+        /// retrieve the auth options from the attributes of the resolved endpoint.
+        /// </summary>
+        protected static List<IAuthSchemeOption> RetrieveSchemesFromEndpoint(Endpoint endpoint)
+        {
+            if (endpoint == null || endpoint.Attributes == null)
+            {
+                return null;
+            }
+
+            var authSchemes = (IList)endpoint.Attributes["authSchemes"];
```
normj, line 81: nit: Can we use `as IList` for defensive purposes.

generator/ServiceClientGeneratorLib/Generators/AuthResolvers/ModeledResolver.tt
normj: If we do need to run endpoint rules twice and it will use the exact same parameters we could stash the results in `executionContext.RequestContext.ContextAttributes` and then have the endpoint resolver check for the existence and if so reuse those results.

dscpinheiro: In the internal design doc (that recommended running the EP2.0 resolver twice), there was a comment on not allowing S3 to use the `AWS::Auth::AccountId` parameter in rules that have authentication schemes.

I'll follow up to see if that was ever built or if it's something we should do in our generator.

dscpinheiro: For the other comment, there's already an `EndpointAttributes` property in the request context: https://github.com/aws/aws-sdk-net/blob/ad13fd46500b624f3e39190a8ed9f32633ca0559/sdk/src/Core/Amazon.Runtime/Internal/IRequest.cs#L385-L388

I guess we could populate that? (although the same concern with running the resolver twice applies). That property is used in other S3 classes to check if it's running on S3 Express: https://github.com/aws/aws-sdk-net/blob/ad13fd46500b624f3e39190a8ed9f32633ca0559/sdk/src/Services/S3/Custom/Util/AmazonS3Util.cs#L371-L378

dscpinheiro: For the other comment, I thought about populating the existing `EndpointAttributes` property: https://github.com/aws/aws-sdk-net/blob/ad13fd46500b624f3e39190a8ed9f32633ca0559/sdk/src/Core/Amazon.Runtime/Internal/IRequest.cs#L388

But it looks like when the endpoints resolver calls `GetEndpoint` it calculates them again anyway (this constructor is called by the generated code): https://github.com/aws/aws-sdk-net/blob/ad13fd46500b624f3e39190a8ed9f32633ca0559/sdk/src/Core/Amazon.Runtime/Endpoints/Endpoint.cs#L45-L49

Or did you mean I should cache the entire endpoint? (I don't think that's the case since we'd need to change whenever we add support for account-ID based endpoints).

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs
```diff
@@ -65,9 +67,66 @@ protected void PreInvoke(IExecutionContext executionContext)
...
+            {
+                var schemeName = schemeNames.First();
+                switch (schemeName)
+                {
+                    case "sigv4":
+                    case "sigv4-s3express":
+                        return AuthSchemeOption.DEFAULT_SIGV4;
+                    case "sigv4a":
+                        return AuthSchemeOption.DEFAULT_SIGV4A;
+                    default:
```
muhammad-othman, line 100: Can we add the other auth scheme options preemptively even if they aren't used now? maybe atleast ad the noAuth version?

generator/ServiceClientGeneratorLib/Generators/AuthResolvers/ModeledResolver.tt
normj: My understanding was running the endpoint rules twice would only work if we the service didn't use account id as part of the rule. In that case would we need to change this when the SDK supports account id based rules? So yes I was suggesting caching the results within the scope of the current request in stead of running the endpoint rules twice. But maybe I'm misunderstanding something and the second time could cause a different results in some scenarios.

muhammad-othman (approved):

Issue #3542, opened 2024-11-04
Bedrock ConverseStreamResponse provides no easy way to read the streamed response without blocking the calling .NET thread


tlecomte:
### Describe the bug

The common way to read from Bedrock `ConverseStreamResponse` is apparently to call `response.Stread.AsEnumerable()`. This achieves the goal of streaming the response completions as they are received, but at the cost of keeping the calling .NET thread busy. So it can lead to thread exhaustion.

Ideally we need a new variant that does not block the thread, so maybe an async enumerator exposed as `response.Stread.AsAsyncEnumerable()`.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

I expect to have an easy method that gives an `IAsyncEnumerable` from `ConverseStreamResponse`, and which is implemented without blocking the calling thread.

### Current Behavior

The current API surface in `ConverseStreamResponse` encourages the use of `response.Stread.AsEnumerable()` which blocks the calling thread.

### Reproduction Steps

N/A

### Possible Solution

This problem was mentioned in https://github.com/aws/aws-sdk-net/issues/3360#issuecomment-2230161648 for a further improvement, with a suggested implementation. I believe the suggested implementation is imperfect, but it can be a basis for a solution.

### AWS .NET SDK and/or Package version used

AWSSDK.BedrockRuntime 3.7.404.8

### Targeted .NET Platform

.NET 8

### Operating System and version

Docker

tlecomte:
This is in principle how it could be implemented, using an extension, and with System.Threading.Channels:

```csharp
public static class ConverseStreamOutputExtensions
{
        public async IAsyncEnumerable<IEventStreamEvent> AsAsyncEnumerable(ConverseStreamOutput output, [EnumeratorCancellation] CancellationToken cancellationToken)
        {
            var channel = Channel.CreateUnbounded<IEventStreamEvent>();

            void OnEventReceived(object? sender, EventStreamEventReceivedArgs<IEventStreamEvent> args)
            {
                var _ = channel.Writer.TryWrite(args.EventStreamEvent);

                // the metadata event is the last one of the event stream
                if (args.EventStreamEvent is ConverseStreamMetadataEvent)
                {
                    channel.Writer.Complete();
                }
            }

            void OnExceptionReceived(object? sender, EventStreamExceptionReceivedArgs<BedrockRuntimeEventStreamException> args)
            {
                var _ = channel.Writer.TryComplete(args.EventStreamException);
[...]

PR #3543, opened 2024-11-04
feat(Event streams): add support for cancellation and IAsyncEnumerable

tlecomte:
## Description

Adjust `EnumerableEventStream` to add support for cancellation and for `IAsyncEnumerable`:

- `IEnumerableEventStream` inherits from `IAsyncEnumerable`. That means that we can use `await foreach` on it, for example.
- `EnumerableEventStream` implements it with a new `GetAsyncEnumerator` method that mimics the existing `GetEnumerator`, but uses `ReadFromStreamAsync` so that it does not block the calling thread while waiting on the incoming network stream.

For example, to process the response stream from Bedrock `ConverseStreamRequest`, before this change we had to do:

```csharp
var response = await bedrockRuntimeClient.ConverseStreamAsync(converseStreamRequest);

foreach (var item in response.Stream.AsEnumerable())
{
      // do something with the item
}
```

The problem above is that `foreach (var item in response.Stream.AsEnumerable())` is blocking the current .NET thread while waiting for the events to arrive, so it could cause thread pool exhaustion.

With the changes in this pull request, we can write instead:

```csharp
var response = await bedrockRuntimeClient.ConverseStreamAsync(converseStreamRequest);

await foreach (var item in response.Stream)
{
      // do something with the item
}
```

Notice the `await foreach` that is possible now that `IEnumerableEventStream` implements `IAsyncEnumerable`. The benefit is that the .NET thread is not blocked.

It is also easy to add support for cancellation. For example:

```csharp
await foreach (var item in response.Stream.WithCancellation(cancellationToken))
{
      // do something with the item
}
```

## Motivation and Context
The semantics of AWS event streams map well to .NET `IAsyncEnumerable`, but it's hard to build an `IAsyncEnumerable` from such a stream. `EnumerableEventStream` does provide an async `StartProcessingAsync` method, but only a synchronous enumerator. So the happy path with the current API is to use a synchronous enumerator, which has the downside of blocking the current .NET thread while enumerating the events. This can lead to the thread pool starvation.

This addresses https://github.com/aws/aws-sdk-net/issues/3542

## Testing
Added an integration test. To run it, the test code needs to be adjusted to comment out the `Ignore` attribute, give credentials and a region.

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/EventStreams/Internal/EnumerableEventStream.cs
+++ b/sdk/src/Core/Amazon.Runtime/EventStreams/Internal/EnumerableEventStream.cs
@@ -1,4 +1,4 @@
-﻿/*******************************************************************************
+/*******************************************************************************
  *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *  Licensed under the Apache License, Version 2.0 (the "License"). You may not use
  *  this file except in compliance with the License. A copy of the License is located at
@@ -25,6 +25,7 @@ using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
 #if AWS_ASYNC_API
+using System.Threading;
 using System.Threading.Tasks;
 #endif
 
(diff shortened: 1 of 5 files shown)


tlecomte:
The issue checklist mentions updating the documentation. I think that it may apply here, and I'm happy to help, but I would need a pointer to know where the source for the doc is maintained.

tlecomte:
Removed the unrelated changes in BedrockRuntime.sln.

peterrsongg:
@tlecomte Thanks for the contribution, we just happened to receive another PR shortly after yours that also addresses the same problem you're running into. We'll review that PR instead as it targets the next major version of the SDK and it allows us to use newer features in the .NET Framework (your PR would need to account for .NET 3.5 as well, for example)." This means that this feature will be available in V4 of the SDK.

https://github.com/aws/aws-sdk-net/pull/3545

Issue #3479, opened 2024-09-20
Unable to configure S3 Client when using DynamoDB S3Link


Danny-UKDM:
### Describe the bug

When implementing `S3Link` for both `S3Link.Create()` and `myLinkedProperty.UploadStreamAsync()` there appears to be no method for configuring the `AmazonS3Client` which is created under the hood, or for providing your own configured instance of an `AmazonS3Client`. 

This appears fine for a production environment, but prevents you from being able to ensure the underlying `AmazonS3Client` is correctly configured for end-to-end integration testing via a service like localstack.

Despite best efforts to ensure the `DynamoDBContext` passed into `S3Link.Create()` is configured for localstack in a `"Development"` environment (which functions as expected elsewhere), the `AmazonS3Client` ultimately throws an error of:
```
"The AWS Access Key Id you provided does not exist in our records"
```

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

Either: 

1 - When correctly configuring the DI registration of `DynamoDBContext` for `Development`, which is passed into `S3Link.Create()`, the underlying functionality correctly clones the configuration and creates the correctly `Development`-configured `AmazonS3Client` under the hood when performing operations.

2 - `S3Link` allows the caller to pass their own configured instance(s) of `AmazonS3Client` which is known to be correctly configured for `Development` and end-to-end integration testing via a service like localstack.

### Current Behavior

1 - Once calling `myLinkedProperty.UploadStreamAsync()` after using `S3Link.Create()` with a localstack-configured instance of `DynamoDBContext`, the following exception is ultimately thrown: 

```csharp
Amazon.S3.AmazonS3Exception: The AWS Access Key Id you provided does not exist in our records.
 ---> Amazon.Runtime.Internal.HttpErrorResponseException: Exception of type 'Amazon.Runtime.Internal.HttpErrorResponseException' was thrown.
   at Amazon.Runtime.HttpWebRequestMessage.ProcessHttpResponseMessage(HttpResponseMessage responseMessage)
   at Amazon.Runtime.HttpWebRequestMessage.GetResponseAsync(CancellationToken cancellationToken)
   at Amazon.Runtime.Internal.HttpHandler`1.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.Runtime.Internal.RedirectHandler.InvokeAsync[T](IExecutionContext executionContext)
[...]

PR #3544, opened 2024-11-05
Fixes issue #3479 where S3ForcePathStyle is unable to be read from a config file

peterrsongg:
## Description

A user can set ForcePathStyle on the `AmazonS3Config` object but it doesn't get read from the config file which makes it difficult to set when using something like the AWSSDK.Extensions.NetCoreSetup library, since that creates the client under the hood. This just adds the ability to read this value from the config file.

## Motivation and Context

Fixes issue #3479 

## Testing

Dry run in v3 passes
Added unit tests which pass.
## Types of changes

## Checklist

## License

### DIFF
--- a/generator/.DevConfigs/666f52b5-8ab6-4817-b54c-f42d2d41bb46.json
+++ b/generator/.DevConfigs/666f52b5-8ab6-4817-b54c-f42d2d41bb46.json
@@ -1,0 +1,18 @@
+{
+    "core": {
+        "changeLogMessages": [
+          "Add S3ForcePathStyle as a config option for credential profiles."
+        ],
+        "type": "patch",
+        "updateMinimum": true
+      },
+    "services": [
+        {
+            "serviceName": "S3",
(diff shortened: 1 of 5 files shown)


dscpinheiro:
For reference, it doesn't look like other SDKs are reading (or expecting?) a `s3_force_path_style` property in the config file. The closest one I found was the CLI, which has a S3 specific section: https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-files.html#cli-configure-files-s3
```
[profile development]
s3 =
  max_concurrent_requests = 20
  max_queue_size = 10000
  multipart_threshold = 64MB
  multipart_chunksize = 16MB
  max_bandwidth = 50MB/s
  use_accelerate_endpoint = true
  addressing_style = path
```

Where `addressing_style` is:
> Specifies which addressing style to use. This controls whether the bucket name is in the hostname or is part of the URL. Valid values are: path, virtual, and auto. The default value is auto.

PR #3546, opened 2024-11-06
[SRA] Update service clients to remove usage of fallback credentials factory

muhammad-othman:
## Description
* Remove the usage of `FallbackCredentialFactory` in service client constructors (limited to `AutoScaling` service for now to avoid huge merge conflicts later).
* Rename `AmazonServiceClient.Credentials` to be `DefaultAWSCredentials`.
* Added new `IdentityResolverConfiguration` property to the client config.
* Marked the `ClientConfig.AWSTokenProvider` property as obsolete.

TODO: Remove the creation of the Signer property (will work on this part while updating the signer handler).

## Motivation and Context
DOTNET-7538

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/IClientConfig.cs
+++ b/sdk/src/Core/Amazon.Runtime/IClientConfig.cs
@@ -60,6 +60,7 @@ namespace Amazon.Runtime
         /// <summary>
         /// Gets the configuration for identity resolvers used by the service client, which manages
         /// the resolvers used to obtain various types of identities.
+        /// Use this to retrieve and manage specific identity resolvers based on the required identity type.
         /// <para />
         /// See <see cref="DefaultIdentityResolverConfiguration"/> for additional information.
         /// </summary>
--- a/sdk/src/Core/Amazon.Runtime/IClientConfig.cs
+++ b/sdk/src/Core/Amazon.Runtime/IClientConfig.cs
@@ -14,11 +14,13 @@
  */
 using System;
 using System.Net;
+using Smithy.Identity.Abstractions;
 using Amazon.Runtime.Internal;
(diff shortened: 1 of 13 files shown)


sdk/src/Core/Amazon.Runtime/IClientConfig.cs
```diff
@@ -55,6 +57,14 @@ public partial interface IClientConfig
         /// </summary>
         Profile Profile { get; }
 
+        /// <summary>
+        /// Gets the configuration for identity resolvers used by the service client, which manages
+        /// the resolvers used to obtain various types of identities.
```
dscpinheiro, line 62:
```suggestion
        /// Gets the configuration for the identity resolvers configured in the service client.
```

I think the documentation could be improved, but I don't have a better suggestion at the moment (looks like Java ended up not using `*Configuration` at all: https://github.com/aws/aws-sdk-java-v2/pull/4423).

dscpinheiro (approved):

sdk/src/Core/Amazon.Runtime/IClientConfig.cs
muhammad-othman: I did see that and it took me a while to figure out that the `IdentityResolvers` is the `IdentityResolverConfiguration` mentioned in the SRA, so I would prefer to use the SRA names to make it easier to find what it means in the SRA unless we have a good reason not to do so.

dscpinheiro: Oh, for sure. I wasn't suggesting renaming the type, just mentioning we can probably improve the documentation for this property (since it's visible in the config and may not be clear to customers at first).

normj (approved):

PR #3548, opened 2024-11-12
Remove CreateSigner from serviceClient and requests with authtype

muhammad-othman:
## Description
* Removed the creation of the Signer property and the logic to create the instance for it in ServiceClients and Requests with AuthType.
* Updated the serviceClient generator to include all the services otherwise we will have compilation errors, but only commited AutoScaling service for now to avoid huge merge conflicts later.
* Renamed credentials `constructorInput` in model customizations to `DefaultAWSCredentials`.
* Marked `FallbackCredentialsFactory` as deprecated and removed it from ServiceClient extensions.
* Refactored the other usages of `FallbackCredentialsFactory` to use`DefaultIdentityResolverConfiguration`.

## Motivation and Context
* DOTNET-7538
* DOTNET-7649

### DIFF
--- a/sdk/src/Services/RDS/Custom/Util/RDSAuthTokenGenerator.cs
+++ b/sdk/src/Services/RDS/Custom/Util/RDSAuthTokenGenerator.cs
@@ -18,6 +18,7 @@ using Amazon.Runtime.Internal;
 using Amazon.Runtime.Internal.Auth;
 using Amazon.Runtime.Internal.Transform;
 using Amazon.Runtime.Internal.Util;
+using Amazon.Runtime.Credentials.Internal.IdentityResolvers;
 using System;
 using System.Globalization;
 
@@ -85,7 +86,7 @@ namespace Amazon.RDS.Util
         /// <returns></returns>
         public static string GenerateAuthToken(RegionEndpoint region, string hostname, int port, string dbUser)
         {
-            AWSCredentials credentials = FallbackCredentialsFactory.GetCredentials();
+            AWSCredentials credentials = DefaultIdentityResolverConfiguration.ResolveDefaultIdentity<AWSCredentials>();
             return GenerateAuthToken(credentials, region, hostname, port, dbUser);
         }
 
--- a/sdk/test/Services/RDS/UnitTests/Custom/RDSAuthTokenGeneratorTest.cs
+++ b/sdk/test/Services/RDS/UnitTests/Custom/RDSAuthTokenGeneratorTest.cs
@@ -17,12 +17,16 @@ using Amazon;
 using Amazon.RDS;
 using Amazon.RDS.Util;
 using Amazon.Runtime;
+using Amazon.Runtime.Credentials.Internal.IdentityResolvers;
 using Amazon.Util;
 using AWSSDK_DotNet.IntegrationTests.Utils;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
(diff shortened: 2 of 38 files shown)


sdk/src/Services/RDS/Custom/Util/RDSAuthTokenGenerator.cs
```diff
@@ -85,7 +86,7 @@ public static string GenerateAuthToken(string hostname, int port, string dbUser)
         /// <returns></returns>
         public static string GenerateAuthToken(RegionEndpoint region, string hostname, int port, string dbUser)
         {
-            AWSCredentials credentials = FallbackCredentialsFactory.GetCredentials();
+            AWSCredentials credentials = DefaultIdentityResolverConfiguration.ResolveDefaultIdentity<AWSCredentials>();
```
dscpinheiro, line 89: Could this be `DefaultAWSCredentialsIdentityResolver` instead? Then customers could call get credentials directly (and I assume most people wouldn't know about the `AWSCredentials` abstract class).

(This comment isn't about this specific helper, but thinking about how we'd tell customers to stop using the `FallbackCredentialsFactory` after it's deprecated).

normj (approved):

sdk/src/Services/RDS/Custom/Util/RDSAuthTokenGenerator.cs
muhammad-othman: I think replacing this call with `DefaultAWSCredentialsIdentityResolver.ResolveAWSCredentials()` makes sense. I initially thought we might need to resolve other types of identities, but so far I've only used it to resolve `AWSCredentials`.

sdk/test/Services/RDS/UnitTests/Custom/RDSAuthTokenGeneratorTest.cs
```diff
@@ -41,26 +45,34 @@ public class RDSAuthTokenGeneratorTest
...
+        private Dictionary<Type, IIdentityResolver> originalIdentityResolvers;
 
         [TestInitialize]
         public void Initialize()
         {
-            originalFallbackList = FallbackCredentialsFactory.CredentialsGenerators;
+            AWSConfigs.AWSRegion = AWSRegion.SystemName;
+
+            FieldInfo field = typeof(DefaultIdentityResolverConfiguration).GetField
+                ("identityResolvers", BindingFlags.Static | BindingFlags.NonPublic);
```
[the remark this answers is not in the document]
muhammad-othman: Adding `Reset` method will not allow us to change the resolvers, the only way to do that is to make them public similar to `FallbackCredentialsFactory.CredentialsGenerators`, this made sense for `FallbackCredentialsFactory` since we use it directly in the serviceClients and there is no way of overriding it, however we do allow users to override the identity resolvers and I don't think we should change make a field public just for the tests.

sdk/src/Services/RDS/Custom/Util/RDSAuthTokenGenerator.cs
dscpinheiro: Talked offline, but OK not to change this for now (although for the migration guide we should check if this is the best approach, there are customers using the `FallbackCredentialsFactory` directly: https://github.com/search?q=FallbackCredentialsFactory.GetCredentials+NOT+repo%3Aaws%2Faws-sdk-net&type=code)

dscpinheiro (approved):

PR #3552, opened 2024-11-18
Update request pipeline to use the AuthSchemeResolver to resolve the identity and signer

muhammad-othman:
## Description
* Added `AuthScheme.Signer()` method to retrieve the signer associated with this authentication scheme.
*  Retrieved identity resolver and signer for the current scheme and attach them to request context in `BaseAuthResolverHandler`.
* Updated `BaseAuthResolverHandler` to allow children to override the signer.
* Added `AuthSchemeHandler` to the pipeline (limited to S3 and AutoScaling for now).

## Motivation and Context
* DOTNET-7802

## Testing
* Ran the unit tests and integration tests for `S3` and `AutoScaling` services.

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Credentials/Internal/AnonymousAuthScheme.cs
+++ b/sdk/src/Core/Amazon.Runtime/Credentials/Internal/AnonymousAuthScheme.cs
@@ -14,6 +14,7 @@
  */
 
 using Smithy.Identity.Abstractions;
+using Amazon.Runtime.Internal.Auth;
 
 namespace Amazon.Runtime.Credentials.Internal
 {
@@ -29,6 +30,12 @@ namespace Amazon.Runtime.Credentials.Internal
         public IIdentityResolver GetIdentityResolver(IIdentityResolverConfiguration configuration)
         {
             return configuration.GetIdentityResolver<AnonymousAWSCredentials>();       
+        }
+
+        /// <inheritdoc/>
+        public ISigner Signer()
+        {
+            return new NullSigner();
         }
     }
 }
--- a/sdk/src/Core/Amazon.Runtime/Credentials/Internal/AnonymousAuthScheme.cs
+++ b/sdk/src/Core/Amazon.Runtime/Credentials/Internal/AnonymousAuthScheme.cs
@@ -23,6 +23,7 @@ namespace Amazon.Runtime.Credentials.Internal
     /// </summary>
     public class AnonymousAuthScheme : IAuthScheme<AnonymousAWSCredentials>
     {
+        private static readonly ISigner _signer = new NullSigner();
         /// <inheritdoc/>
         public string SchemeId => AuthSchemeOption.NoAuth;
 
@@ -35,7 +36,7 @@ namespace Amazon.Runtime.Credentials.Internal
         /// <inheritdoc/>
         public ISigner Signer()
         {
-            return new NullSigner();
+            return _signer;
         }
     }
 }
--- a/sdk/src/Core/Amazon.Runtime/Credentials/Internal/AwsV4aAuthScheme.cs
+++ b/sdk/src/Core/Amazon.Runtime/Credentials/Internal/AwsV4aAuthScheme.cs
@@ -13,6 +13,7 @@
 * permissions and limitations under the License.
 */
 
+using Amazon.Runtime.Internal.Auth;
 using Smithy.Identity.Abstractions;
 
 namespace Amazon.Runtime.Credentials.Internal
@@ -30,6 +31,12 @@ namespace Amazon.Runtime.Credentials.Internal
         public IIdentityResolver GetIdentityResolver(IIdentityResolverConfiguration configuration)
         {
             return configuration.GetIdentityResolver<AWSCredentials>();
+        }
+
+        /// <inheritdoc/>
+        public ISigner Signer()
+        {
+            return new AWS4aSignerCRTWrapper();
         }
     }
 }
--- a/sdk/src/Core/Amazon.Runtime/Credentials/Internal/AwsV4aAuthScheme.cs
+++ b/sdk/src/Core/Amazon.Runtime/Credentials/Internal/AwsV4aAuthScheme.cs
@@ -13,6 +13,7 @@
 * permissions and limitations under the License.
 */
 
+using System.Threading;
 using Amazon.Runtime.Internal.Auth;
 using Smithy.Identity.Abstractions;
 
@@ -38,7 +39,7 @@ namespace Amazon.Runtime.Credentials.Internal
         public ISigner Signer()
         {
             if (_signer == null)
-                _signer = new AWS4aSignerCRTWrapper();
+                Interlocked.Exchange(ref _signer, new AWS4aSignerCRTWrapper());
             return _signer;
         }
     }
--- a/sdk/src/Core/Amazon.Runtime/Credentials/Internal/AwsV4aAuthScheme.cs
+++ b/sdk/src/Core/Amazon.Runtime/Credentials/Internal/AwsV4aAuthScheme.cs
@@ -24,6 +24,7 @@ namespace Amazon.Runtime.Credentials.Internal
     /// </summary>
     public class AwsV4aAuthScheme : IAuthScheme<AWSCredentials>
     {
+        private static ISigner _signer;
         /// <inheritdoc/>
         public string SchemeId => AuthSchemeOption.SigV4A;
 
@@ -36,7 +37,9 @@ namespace Amazon.Runtime.Credentials.Internal
         /// <inheritdoc/>
         public ISigner Signer()
         {
-            return new AWS4aSignerCRTWrapper();
+            if (_signer == null)
+                _signer = new AWS4aSignerCRTWrapper();
+            return _signer;
         }
     }
 }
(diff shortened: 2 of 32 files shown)


sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs
```diff
@@ -56,17 +76,22 @@ protected void PreInvoke(IExecutionContext executionContext)
...
+                    executionContext.RequestContext.Identity = identityResolver.ResolveIdentity();
+                }
 
-                // This code would also need to handle existing scenarios, such as:
-                // - Customer explicitly set credentials on their service client (i.e. new AmazonS3Client(new BasicAWSCredentials))
+                executionContext.RequestContext.Signer = GetSigner(scheme);
             }
         }
 
+        protected virtual AbstractAWSSigner GetSigner(IAuthScheme<BaseIdentity> scheme)
```
normj, line 90: Should this return `ISigner`. `AbstractAWSSigner` seems tired to a SigV4(a) signer.

sdk/src/Core/Amazon.Runtime/Credentials/Internal/AnonymousAuthScheme.cs
```diff
@@ -30,5 +31,11 @@ public IIdentityResolver GetIdentityResolver(IIdentityResolverConfiguration conf
         {
             return configuration.GetIdentityResolver<AnonymousAWSCredentials>();       
         }
+
+        /// <inheritdoc/>
+        public ISigner Signer()
+        {
+            return new NullSigner();
```
normj, line 38: For each of these schemes should the signer be a member variable so we can reuse the same instance instead of recreating the signer for every request?

normj (changes requested):

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs
```diff
@@ -56,17 +76,22 @@ protected void PreInvoke(IExecutionContext executionContext)
                     continue;
                 }
 
-                // TODO: Retrieve identity resolver and signer for the current scheme and attach them to request context.
-                // Interfaces haven't been defined yet, but code will be similar to:
-                // var identityResolver = scheme.IdentityResolver();
-                // var identity = identityResolver.GetIdentity();
-                // var signer = scheme.Signer();
+                executionContext.RequestContext.Identity = this.Identity;
```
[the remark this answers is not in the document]
muhammad-othman: Yep I did miss that, will rename `this.Identity` to `DefaultAWSCredentials`  and make sure that we use it for SigV4/SigV4a.

For the other note, we didn't allow the users to provide a Token when creating the client, however the could provide `IAWSTokenProvider` implementation, which should be respected when implementing SSO, right @dscpinheiro?

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs
```diff
@@ -56,17 +76,22 @@ protected void PreInvoke(IExecutionContext executionContext)
...
+                if (executionContext.RequestContext.Identity == null)
+                {
+                    var identityResolver = scheme.GetIdentityResolver(executionContext.RequestContext.ClientConfig.IdentityResolverConfiguration);
+                    executionContext.RequestContext.Identity = identityResolver.ResolveIdentity();
```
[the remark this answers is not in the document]
muhammad-othman: This shouldn't happen since we do cache the credentials in `DefaultAWSCredentialsIdentityResolver`.

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs
muhammad-othman: All the signers do implement `AbstractAWSSigner`, and this is the type of `executionContext.RequestContext.Signer` 
so we will have to cast it again anyway.
image

Changing `executionContext.RequestContext.Signer` to `ISigner` wont be very simple since we do depend on signer referencing core internal types that need to be extracted / refactored (e.g. IRequest).

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/CredentialsRetriever.cs
```diff
@@ -89,13 +76,15 @@ public override void InvokeSync(IExecutionContext executionContext)
         public override async System.Threading.Tasks.Task<T> InvokeAsync<T>(IExecutionContext executionContext)
         {
             ImmutableCredentials ic = null;
-            if (Credentials != null && !(Credentials is AnonymousAWSCredentials))
+            var identity = executionContext.RequestContext.Identity as AWSCredentials;
+
+            if (identity != null && !(identity is AnonymousAWSCredentials))
```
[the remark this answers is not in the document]
muhammad-othman: It should be even simpler than that, we can just return `null` for `GetCredentials` and `GetCredentialsAsync` has an implementation in the base class that just uses `GetCredentials` and we don't need to change that.

Now this made me think that we shouldn't need the `BearerTokenSigner` special case too that we have above, since the signer cannot be `BearerTokenSigner` when the Identity is `AWSCredentials`.
image

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs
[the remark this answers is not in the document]
dscpinheiro: Yes, they could provide an implementation for the token provider property. 

Although I'm adding (in a separate PR) a default token identity resolver that should always be available. So my understanding is if an operation specifies both token and SigV4 as supported auth schemes we'll be able to resolve them automatically.

sdk/src/Core/Amazon.Runtime/Credentials/Internal/AwsV4aAuthScheme.cs
```diff
@@ -36,7 +37,9 @@ public IIdentityResolver GetIdentityResolver(IIdentityResolverConfiguration conf
         /// <inheritdoc/>
         public ISigner Signer()
         {
-            return new AWS4aSignerCRTWrapper();
+            if (_signer == null)
```
dscpinheiro, line 40: Any reason why this one doesn't assign the value where the field is declared? Is it because it'll try to load the CRT dependency?

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs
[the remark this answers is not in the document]
muhammad-othman: @dscpinheiro  I mean we should check if the user defined a token provider and use it before using the new default token identity resolver.

sdk/src/Core/Amazon.Runtime/Credentials/Internal/AwsV4aAuthScheme.cs
dscpinheiro: Hm, that makes me think we should update the `BaseAuthResolverHandler` to also check if the CRT is available when there are multiple auth schemes and one of them is SigV4A (in that case, we should skip SigV4A and try to use one of the other auth schemes).

This can be done in a separate PR; Here's what the endpoints resolver does: https://github.com/aws/aws-sdk-net/blob/816c44e7c6a19639cc65065c0f8fd533420638b0/sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseEndpointResolver.cs#L166-L176

dscpinheiro (approved):

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs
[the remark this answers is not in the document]
dscpinheiro: > @dscpinheiro I mean we should check if the user defined a token provider and use it before using the new default token identity resolver.

Sorry, the second sentence was replying to Norm's question (_On that note, if a service does use both SigV4 and Token how do you configure the service client for both?_)

sdk/src/Core/Amazon.Runtime/Credentials/Internal/AwsV4aAuthScheme.cs
```diff
@@ -31,5 +33,13 @@ public IIdentityResolver GetIdentityResolver(IIdentityResolverConfiguration conf
         {
             return configuration.GetIdentityResolver<AWSCredentials>();
         }
+
+        /// <inheritdoc/>
+        public ISigner Signer()
+        {
+            if (_signer == null)
+                _signer = new AWS4aSignerCRTWrapper();
```
normj, line 41: To avoid weird threading issues can we use the Interlock class for this lazy setting.

```
Interlocked.Exchange(ref _signer, new AWS4aSignerCRTWrapper());
```

normj (changes requested):
One small comment and then I'm good.

sdk/src/Core/Amazon.Runtime/Credentials/Internal/AwsV4aAuthScheme.cs
muhammad-othman: Updated.

PR #3553, opened 2024-11-19
Update unmarshallers to use system text json.

peterrsongg:
## Description
Updates the unmarshallers to use System Text Json for deserialization. 

There are a lot of generated service changes, you can look at SQS as an example as they are all the same change. The main thing that should be reviewed is any change in Core.

5ccb4c66ce3c - generated SQS changes as an example
3654c5dac176 - manual s3 changes
18c64c55bb43 - core and unit tests changes for deserialization
e2072fa856ca - generator changes

**NOTE** To run the unit tests I added you need to run the generator first. Use this command to run the services required to run the unit tests in CoreAndCustomUnitTests.Netframework.sln
```
-sm autoscaling;bearer-token-auth-test;cloudfront;monitoring;dynamodb;ec2;elasticmapreduce;elastictranscoder;iot-data;kms;redshift;s3;sts;sso;sso-oidc -sro
```

All protocol tests pass

Some things I want to point out.

1. Since `Utf8JsonReader` cannot accept a stream in its constructor, I created a wrapper around `Utf8JsonReader` that accepts a stream and fills the buffer to pass into the reader. Since the wrapper is also a ref struct it can hold `Utf8JsonReader` as a field.
2. Added AWSConfigs.StreamingUtf8JsonReaderBufferSize as a config option if a user wants to tweak the default buffer size used by the reader.
3. Each unmarshaller is implementing only the interface for its respective protocol now, which means I had to split `IUnmarshaller` => `IXmlUnmarshaller` and `IJsonUnmarshaller`. This gives the added benefit that each structure unmarshaller no longer throws `NotImplementedException` for the other protocol. 
This is necessary for when we add new protocols down the line. We don't want each structure unmarshaller to implement all protocol interfaces just to throw `NotImplemented` for the others.
4. Split the Dictionary, List, KeyValue unmarshallers to separate Json / XML unmarshallers since each protocol only implements its own interface.
5. Split the error unmarshallers to Xml and Json error response unmarshaller classes since the json error response unmarshallers require a reader to be passed in.
6. Since S3 is handcoded I had to make some custom changes, specifically for the XmlErrorResponse instantiations.
7. Due to an issue with aws tools for Powershell compatibility netstandard2.0 and netframework targets must use system.text.json 6.0.11
## Testing

I tested a couple real API calls. I added unit tests for the streamingUtf8JsonReader,  and all the protocol tests pass.

Full dry run will be done in the feature branch.
Overall, I saw around a 77% performance gain in unmarshalling (note this is on my machine not an EC2 instance). I will attach specific numbers from an ec2 instance as a follow up.

## Types of changes

## Checklist

## License


slang25:
this looks _fantastic_, very excited for this change

sdk/src/Core/Amazon.Runtime/Internal/Transform/StreamingUtf8JsonReader.cs
```diff
@@ -0,0 +1,134 @@
...
+                }
+
+                leftover.CopyTo(buffer);
+                bytesRead = stream.Read(buffer, leftover.Length, buffer.Length - leftover.Length);
+            }
+            else
+            {
+                bytesRead = stream.Read(buffer, 0, buffer.Length);
+            }
+            if (bytesRead == 0)
```
slang25, line 128: should the method return here, I worry that the follow line sets the reader using the `buffer` that is being returned to the pool

sdk/src/Services/SQS/Generated/Model/Internal/MarshallTransformations/AddPermissionRequestMarshaller.cs
```diff
@@ -63,54 +63,58 @@ public IRequest Marshall(AddPermissionRequest publicRequest)
             request.HttpMethod = "POST";
 
             request.ResourcePath = "/";
-            using (MemoryStream memoryStream = new MemoryStream())
+#if NETCOREAPP3_1_OR_GREATER
```
slang25, line 66: This might be stretching the scope of this change too much, but would it be possible to make the source of `MemoryStream`/`IBufferWriter<T>` pluggable.

I'd really love to add my own pipeline middleware that allows for injecting a [RecyclableMemoryStream](https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream) (which implements both `MemoryStream` and `IBufferWriter<T>`)

normj: My interpretation of this is we would have a property on `AWSConfigs` called `StreamFactory` of type `IStreamFactory` defined as:
```
public interface IStreamFactory
{
    Stream GetStream();
}
```

For .NET Core 3.1 and above we would default this to an instance of ArrayBufferWriterFactory:
```
internal class ArrayBufferWriterFactory : IStreamFactory
{
    public Stream GetStream()
    {
        return ArrayBufferWriter<byte>();
    }
}
```

For older targets we would use a MemoryStream based factory:
```
internal class MemoryStreamFactory : IStreamFactory
{
    public Stream GetStream()
    {
        throw new MemoryStream();
    }
}
```

In our unmarshallers where we need a stream we could get rid of the conditional check and just call `AWSConfigs.StreamFactory.GetManager()`.

Then @slang25 can take a dependency on `Microsoft.IO.RecyclableMemoryStream` and right his own factory and assign it to `AWSConfigs.StreamFactory`.
```
internal class RecyclableMemoryStreamFactory : IStreamFactory
{
    private static readonly RecyclableMemoryStreamManager manager = new RecyclableMemoryStreamManager();

    public Stream GetStream()
    {
        return manager.GetStream();
    }
}
```

Does that capture your thoughts @slang25 ?

sdk/src/Core/Amazon.Runtime/Internal/Transform/StreamingUtf8JsonReader.cs
peterrsongg: ah, yeah good suggestion the reader would be using a buffer returned to the pool in that scenario. If I return there though, it will return to the `Read()` method and the `_reader` won't have a buffer to work with, and I don't want to set buffer to null. I think instead I'll just refactor a bit and add something like the following in the `Read()` method
```
            if (!hasMoreData)
            {
                GetMoreBytesFromStream(_stream, ref _buffer, ref _reader);
                hasMoreData = _reader.Read();
                if (_reader.IsFinalBlock)
                   ArrayPool<byte>.Shared.Return(_buffer);
            }
```

peterrsongg: while fixing a bug in the streamingutf8 reader I decided to remove the pooling due to multiple resizings of the array. I put a justification in a comment, but basically the byte array returned by `Array.Resize` isn't owned by the pool so keeping track of that didn't seem worth the Renting + copying I would have to do in order to do that

peterrsongg: flipped back to pooling

sdk/src/Services/SQS/Generated/Model/Internal/MarshallTransformations/AddPermissionRequestMarshaller.cs
peterrsongg: Interesting, I just took a look at the class linked that does something similar, which allows the a wrapper over IBufferWriter and at first glance it doesn't seem to bad, but i'm not sure why they are overriding `Write` and `ReadAsync` etc.. IBufferWriter only defines 3 methods to implement.

IF we don't have to override all those methods listed [here](https://github.com/CommunityToolkit/dotnet/blob/c23e1cc6e918f86c589facf98f70ca122495c385/src/CommunityToolkit.HighPerformance/Streams/IBufferWriterStream%7BTWriter%7D.cs) I think it is doable. I'll hold off until I hear back though.

slang25: What I'm looking for is a way to get the best perf out of the AWS SDK, so having the ability to pool these buffers that are going to copy all the bytes into memory is going to reduce memory allocations significantly.

My understanding is that `IBufferWriter<T>` is the fast modern to do [synchronous buffered writing](https://learn.microsoft.com/en-us/dotnet/standard/io/buffers#ibufferwritert), vs `Stream` which is more ubiquitous in the .NET world today.

That said, I think a `Stream GetStream()` signature would work, but you could additional do a type check to see if it implements a `IBufferWriter<T>`, and then use the appropriate overloads. That makes it convenient for use with Streams, and has a fast path for things like `RecyclableMemoryStreamManager`. To be honest, this isn't a deal breaker, I haven't ran benchmarks but I'm guessing the gains are not tremendous.

The problem with a static `IStreamFactory` is that I want to be able to pool the buffers and therefore want a way to return them. So the `AWSConfigs` static property would need an interface to support that for it to achieve what I'm looking for.

I was thinking about having a `PipelineHandler` that could be added into the pipeline, and setting the stream on some sort of context object that passes through. That way it could own the creation/pooling of buffers, and returning them. However a static property would be more ergonomic.

sdk/test/UnitTests/Custom/Marshalling/StreamingUtf8JsonReaderTests.cs
```diff
@@ -0,0 +1,111 @@
...
+    /// large payloads, which would trigger the logic for getting more bytes from the stream. 
+    /// This class just tests the wrapper class StreamingUtf8JsonReader. 
+    /// </summary>
+    [TestClass]
+    public class StreamingUtf8JsonReaderTests
+    {
+        [TestMethod]
+        public void HandlesUtf8BOM()
+        {
+            // we can't use reflection to access the private fields of StreamingUtf8JsonReader since it is a ref struct so we have to test it this way.
```
normj, line 26: You could make the methods `internal` in StreamingUtf7JsonReader` and then since the unit test project has access via the `InternalsVisibleToAttribute` test the handle method directly. Following this pattern you could also have tests directly on `FillBuffer` and `GetMoreBytesFromStream`,

sdk/test/UnitTests/Custom/Marshalling/StreamingUtf8JsonReaderTests.cs
```diff
@@ -0,0 +1,111 @@
...
+        // This method tests that if the json token starts in one buffer but continues into the next buffer
+        // the reader can handle parsing it correctly.
+        [TestMethod]
+        public void StreamingUtf8JsonReaderHandlesJsonTokenThatSpansMultipleBuffers()
+        {
+            // Arrange
+            // here we're creating a json string that is greater than the default buffer size to test the GetMoreBytesFromStream logic
+            var sb = new StringBuilder();
+            sb.Append("{ \"key\": \"");
+            sb.Append(new string('x', 7500));
```
normj, line 52: In case we increase the default size but don't change the value here you might want to reference the `AWSConfigs` value plus some size bigger.

sdk/test/UnitTests/Custom/Marshalling/StreamingUtf8JsonReaderTests.cs
```diff
@@ -0,0 +1,111 @@
...
+using Amazon.EC2.Model;
+namespace AWSSDK.UnitTests
+{
+    /// <summary>
+    /// Protocol tests already exists to test the marshalling and unmarshalling of request and responses in json, but they don't test very
+    /// large payloads, which would trigger the logic for getting more bytes from the stream. 
+    /// This class just tests the wrapper class StreamingUtf8JsonReader. 
+    /// </summary>
+    [TestClass]
+    public class StreamingUtf8JsonReaderTests
```
normj, line 21: We should have a test for the scenario where the `PassReaderByRef` is used for converting the JsonDocument and the the document doesn't fit in the buffer.

sdk/test/UnitTests/Custom/Marshalling/StreamingUtf8JsonReaderTests.cs
```diff
@@ -0,0 +1,111 @@
...
+            // Arrange
+            // here we're creating a json string that is greater than the default buffer size to test the GetMoreBytesFromStream logic
+            var sb = new StringBuilder();
+            sb.Append("{ \"key\": \"");
+            sb.Append(new string('x', 7500));
+            sb.Append("\" }");
+            string largeJson = sb.ToString();
+
+            byte[] payload = Encoding.UTF8.GetBytes(largeJson);
+            using (var stream = new MemoryStream(payload))
```
normj, line 57: To help verification you might want to add an internal property on the `StreamingUtf8JsonReader` for the current length of the buffer. Then you can check before and after that the buffer size increased.

normj (changes requested):

sdk/src/Core/Amazon.Runtime/EventStreams/Internal/EventStreamUtils.cs
```diff
@@ -46,7 +47,7 @@ public static XmlUnmarshallerContext ConvertMessageToXmlContext(IEventStreamMess
         /// Converts an <seealso cref="EventStreamMessage"/> to a <seealso cref="JsonUnmarshallerContext"/>
         /// </summary>
         /// <param name="message">The event stream message</param>
-        /// <returns>The JsonUnmarshallerContext</returns>
+        /// <returns>The JsonUnmarshallerContext and Utf8JsonReader</returns>
```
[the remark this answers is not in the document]
peterrsongg: Ah, i had an iteration of this where I tried to return a tuple. just a miss removing it.

sdk/src/Core/Amazon.Runtime/Internal/Transform/UnmarshallerContext.cs
```diff
@@ -194,6 +195,17 @@ public bool ReadAtDepth(int targetDepth)
             return Read() && this.CurrentDepth >= targetDepth;
         }
 
+        /// <summary>
+        /// Reads the next token at depth greater than or equal to target depth.
+        /// </summary>
+        /// <param name="targetDepth">Tokens are read at depth greater than or equal to target depth.</param>
+        /// <param name="reader">The Utf8JsonReader used to read the document</param>
+        /// <returns>True if a token was read and current depth is greater than or equal to target depth.</returns>
+        public bool ReadAtDepth(int targetDepth, ref StreamingUtf8JsonReader reader)
```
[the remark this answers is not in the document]
peterrsongg: good catch, this can be moved

sdk/src/Core/Amazon.Runtime/Internal/Util/JsonConstants.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Text;
+
+namespace Amazon.Runtime.Internal.Util
+{
+    /// <summary>
+    /// Constants used in JSON unmarshalling.
+    /// </summary>
+    public static class JsonConstants
+    {
+        public static ReadOnlySpan<byte> Utf8Bom => new byte[] { 0xEF, 0xBB, 0xBF };
```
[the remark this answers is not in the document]
peterrsongg: I can't do this because `ReadOnlySpan<byte>` is a ref struct and can't be used as an auto-implemented property unless the class itself is a ref struct :(

sdk/src/Core/Amazon.Runtime/Internal/Util/StreamingUtf8JsonReader.cs
```diff
@@ -0,0 +1,183 @@
...
+
+            _stream = stream;
+            _buffer = ArrayPool<byte>.Shared.Rent(AWSConfigs.StreamingUtf8JsonReaderBufferSize ?? 4096);
+            // need to initialize the reader even if the buffer is empty because auto-default of unassigned fields is only 
+            // supported in C# 11+
+            _reader = new Utf8JsonReader(_buffer);
+            HandleUtf8Bom(ref _buffer);
+        }
+
+        private void HandleUtf8Bom(ref byte[] buffer)
```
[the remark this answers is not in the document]
peterrsongg: makes sense will make the change

sdk/src/Core/Amazon.Runtime/Internal/Util/StreamingUtf8JsonReader.cs
```diff
@@ -0,0 +1,183 @@
...
+            bool resized = false;
+            int previousBufferLength = buffer.Length;
+
+            if (reader.BytesConsumed < buffer.Length)
+            {
+                // if Read() returned false and we are here inside this if block that means that we couldn't fully parse the JSON token 
+                // because it was too large to fit in the remainder of the buffer.
+                ReadOnlySpan<byte> leftover = buffer.AsSpan().Slice((int)reader.BytesConsumed);
+
+                if (leftover.Length == buffer.Length)
```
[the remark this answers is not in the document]
peterrsongg: So this basically covers the case where on one read  BytesConsumed is less than buffer.Length, meaning that there is a JSON token that can't be fully read b/c it ends in the next part of the stream. So we copy the leftover to the current buffer and fill the buffer with the remainder of the stream. Then on the NEXT read if leftover.Length == buffer.Length that means 0 data was read, which means even after copying over the leftover from the previous read and reading more data in, we still can't read the full JSON token so we must double the buffer length to see if that would give the reader enough space to read the JSON token.

sdk/src/Core/Amazon.Runtime/Internal/Transform/JsonUnmarshallerContext.cs
```diff
@@ -409,14 +418,19 @@ private void UpdateContext()
...
         #endregion
 
-        public JsonData ToJsonData()
+        public JsonDocument ToJsonDocument(ref StreamingUtf8JsonReader streamingReader)
         {
-            var data = JsonMapper.ToObject(jsonReader);
+            JsonDocument document = null;
+            streamingReader.PassReaderByRef((ref Utf8JsonReader reader) =>
+            {
+                document = JsonDocument.ParseValue(ref reader);
```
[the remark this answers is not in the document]
peterrsongg: That is a great point. Then we would have to actually read all the data from the stream and fill the buffer and then pass in the reader, instead.

sdk/test/UnitTests/Custom/Marshalling/StreamingUtf8JsonReaderTests.cs
peterrsongg: I see the value in testing FillBuffer directly, but I don't really see the value in testing `GetMoreBytesFromStream` directly since this method is only called in the context of `Read()`. The test where I make one Json token = defaultvalueBufferSize + 500 covers the array resizing logic, and the smaller buffer size covers the `reader.BytesConsumed < buffer.Length` logic. However, I could add one additional test where the stream is exactly the size of the buffer.

So I' like to keep as many methods private as I can, but if you feel strongly about this I can change it to internal. I just don't see what additional testing I would do other than exact buffer size.

sdk/test/UnitTests/Custom/Marshalling/StreamingUtf8JsonReaderTests.cs
peterrsongg: I tried to do this, but the problem is that once we hit the final block we set buffer to null and then getting the buffer's size leads to a null pointer

sdk/src/Core/Amazon.Runtime/Internal/Util/StreamingUtf8JsonReader.cs
```diff
@@ -0,0 +1,189 @@
...
+            {
+                AllowTrailingCommas = true
+            };
+            // need to initialize the reader even if the buffer is empty because auto-default of unassigned fields is only 
+            // supported in C# 11+
+            _reader = new Utf8JsonReader(_buffer, _options);
+            SkipUtf8Bom(ref _buffer);
+        }
+
+        private void SkipUtf8Bom(ref byte[] buffer)
```
normj, line 64: Realizing this method does more then just Skip the Ut8Bom it is also what is in charge of providing the initial content for the buffer I think you should just collapse this method with the constructor. That also gets rid of the awkward `_reader` creating in the constructor that gets recreated here.  Something like this:
```
public StreamingUtf8JsonReader(Stream stream)
{
    if (stream is null)
        throw new ArgumentException("Stream must not be null. Please initialize a stream and pass it into the constructor.");

    _stream = stream;
    _buffer = ArrayPool<byte>.Shared.Rent(4096);
    _options = new JsonReaderOptions
    {
        AllowTrailingCommas = true
    };

    // Read the initial buffer from the stream and skip over UTF8 BOM if it exists.
    int bytesRead = _stream.Read(_buffer, 0, _buffer.Length);
    int start = 0;
    if (_buffer.AsSpan().StartsWith(JsonConstants.Utf8Bom))
    {
        start += JsonConstants.Utf8Bom.Length;
        bytesRead -= Utf8Bom.Length;
    }

    _reader = new Utf8JsonReader(_buffer.AsSpan(start, bytesRead), isFinalBlock: bytesRead == 0, new JsonReaderState(_options));
}
```

sdk/src/Core/Amazon.Runtime/Internal/Util/StreamingUtf8JsonReader.cs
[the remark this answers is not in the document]
normj: Okay but the comment about the `if` block is over the `leftover` and logic about going into the if block is more about if `BytesConsumed` is equal to 0. It is an extra layer of logic to understand that if `BytesConsumed` is equal to 0 then `leftover.Length == buffer.Length` is true. The logic flow is also hard to read in general because the conditions are spread out across the function with more variables being created that are only used sometimes. For example if you are recreating the buffer you go in the first `if` block then skip some code then go in another blocking continue the resize logic if the `resized` variable is true.

This is my attempt trying to keep the related logic close to each other and get rid of excess variables. I haven't test this though.

```csharp
int bytesRead = 0;

if (reader.BytesConsumed < buffer.Length)
{
    // If BytesConsumed is 0 then the previous reader.Read() method failed because
    // the next JSON token to read from the stream is too large to fit in the buffer. 
    // That means we need to resize the buffer and try again to read the next JSON token.
    if (reader.BytesConsumed == 0)
    {                        
        // Rent a new buffer twice the size of the previous and fill the new resized buffer with data from the stream
        var resizedBuffer = ArrayPool<byte>.Shared.Rent(Math.Min(int.MaxValue, (buffer.Length * 2)));
        bytesRead = FillBuffer(stream, ref resizedBuffer, 0, resizedBuffer.Length);

        // Recreate ref parameter reader using a span of the data from the new resized buffer.
        var resizedSpan = resizedBuffer.AsSpan(0, bytesRead + buffer.Length);
        reader = new Utf8JsonReader(resizedSpan, isFinalBlock: bytesRead == 0, reader.CurrentState);

        // Return old buffer and update the ref buffer parameter to new resized buffer.
        Logger.GetLogger(typeof(StreamingUtf8JsonReader)).DebugFormat("Resizing buffer from {0} to {1}", buffer.Length, resizedBuffer.Length);
        ArrayPool<byte>.Shared.Return(buffer);
        buffer = resizedBuffer;

        // Do an early return since we have a new Utf8JsonReader with a filled buffer.
        return;
    }
    else
    {
        // Move the unprocessed data from the buffer to the start and the fill
        // remaining space in the buffer with new content from the stream.
        ReadOnlySpan<byte> leftover = buffer.AsSpan().Slice((int)reader.BytesConsumed);
        leftover.CopyTo(buffer);
        bytesRead = FillBuffer(stream, ref buffer, leftover.Length, buffer.Length - leftover.Length);
    }
}

else
{
    bytesRead = FillBuffer(stream, ref buffer, 0, buffer.Length);
}
```

sdk/src/Core/Amazon.Runtime/Internal/Util/JsonConstants.cs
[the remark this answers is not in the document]
peterrsongg: got it 👍

sdk/src/Core/Amazon.Runtime/Internal/Transform/JsonUnmarshallerContext.cs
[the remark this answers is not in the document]
peterrsongg: So originally this method was called `ToJsonData`
https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Core/Amazon.Runtime/Internal/Transform/JsonUnmarshallerContext.cs#L412

and since the reader was part of the context, it just used the context's reader. But I don't see it being called anywhere either actually 

https://github.com/search?q=repo%3Aaws%2Faws-sdk-net+ToJsonData&type=code

So I'm okay with just removing this method and then also the PassReaderByRef, method since this was the only reason I created this method in the first place

sdk/src/Core/Amazon.Runtime/Internal/Transform/UnmarshallerContext.cs
```diff
@@ -243,6 +244,14 @@ private static bool TestExpression(string expression, int startingStackDepth, st
         /// </returns>
         public abstract bool Read();
 
+        /// <summary>
+        /// Reads the next node in the document and updates the context accordingly.
+        /// This is specifically used for JSON protocols which require a reader to be passed in.
+        /// </summary>
+        /// <param name="reader">The Utf8JsonReader</param>
+        /// <returns></returns>
+        public abstract bool Read(ref StreamingUtf8JsonReader reader);
```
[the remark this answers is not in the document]
peterrsongg: Good point, I removed both the abstract `Read()` with no parameters and abstract `Read(ref streamingUtf8JsonReader reader)` and moved each one to its unmarshaller context class. I also moved `ReadAtDepth(int targetDepth)`  since the Json unmarshaller context's `ReadAtDepth` accepts an additional argument anyways.

sdk/src/Core/Amazon.Runtime/Internal/Util/StreamingUtf8JsonReader.cs
peterrsongg: that's a good point about the awkward `_reader` recreation. I will  just collapse it at this point.

peterrsongg:
I've addressed your feedback, thanks for the detailed look. I updated the logic in StreamingUtf8JsonReader as well so that it is easier to follow

sdk/src/Core/Amazon.Runtime/Internal/Transform/JsonUnmarshallerContext.cs
```diff
@@ -231,74 +242,74 @@ public override bool Read()
...
             return text;
         }
 
+        /// <summary>
+        ///     Returns the text contents of the current token being parsed.
+        /// </summary>
+        /// <returns>
+        ///     The text contents of the current token being parsed.
+        /// </returns>
+        public override string ReadText()
```
normj, line 298: See if you have to have this method defined on the base class. Any time we have to code like this check to see if the methods should really be on the base type.

sdk/src/Core/Amazon.Runtime/Internal/Util/StreamingUtf8JsonReader.cs
```diff
@@ -0,0 +1,158 @@
...
+        private Stream _stream;
+        private byte[] _buffer;
+        public StreamingUtf8JsonReader(Stream stream)
+        {
+            if (stream is null)
+                throw new ArgumentException("Stream must not be null. Please initialize a stream and pass it into the constructor.");
+
+            _stream = stream;
+            _buffer = ArrayPool<byte>.Shared.Rent(AWSConfigs.StreamingUtf8JsonReaderBufferSize ?? 4096);
+            // need to initialize the reader even if the buffer is empty because auto-default of unassigned fields is only
```
normj, line 56: Delete this orphan comment.

sdk/src/Core/Amazon.Runtime/Internal/Util/StreamingUtf8JsonReader.cs
```diff
@@ -0,0 +1,158 @@
...
+                // In that case we need to resize the buffer and try again to read the JSON token.
+                if (reader.BytesConsumed == 0)
+                {
+                    var resizedBuffer = ArrayPool<byte>.Shared.Rent(Math.Min(int.MaxValue, (buffer.Length * 2)));
+                    Logger.GetLogger(typeof(StreamingUtf8JsonReader)).DebugFormat("Resizing buffer from {0} to {1}", buffer.Length, resizedBuffer.Length);
+
+                    buffer.AsSpan().CopyTo(resizedBuffer);
+                    ArrayPool<byte>.Shared.Return(buffer);
+                    buffer = resizedBuffer;
+                    bytesRead = FillBuffer(stream, ref buffer, leftover.Length, buffer.Length - leftover.Length);
```
normj, line 112: Since this is a brand new empty buffer why would would the offset ever be anything then `0` and bytes to read be `buffer.Length`? Then you can move the `leftover` variable to in the else block the only place it is used.

sdk/src/Core/Amazon.Runtime/Internal/Util/StreamingUtf8JsonReader.cs
```diff
@@ -0,0 +1,158 @@
...
+                    var resizedSpan = buffer.AsSpan(0, bytesRead + previousBufferLength);
```
normj, line 113: Again this is a new buffer empty buffer we filled up with previous `FillBuffer` call. The span should be from 0 to `bytesRead`. We don't care about the `previousBufferLength`.

sdk/src/Core/Amazon.Runtime/Internal/Util/StreamingUtf8JsonReader.cs
```diff
@@ -0,0 +1,158 @@
...
+        private static void GetMoreBytesFromStream(Stream stream, ref byte[] buffer, ref Utf8JsonReader reader)
+        {
+            int bytesRead = 0;
+            int previousBufferLength = buffer.Length;
+            if (reader.BytesConsumed < buffer.Length)
+            {
+                ReadOnlySpan<byte> leftover = buffer.AsSpan().Slice((int)reader.BytesConsumed);
+                // If BytesConsumed is 0 that means that the previous Read failed because the JSON token was too large to fit in the buffer.
+                // In that case we need to resize the buffer and try again to read the JSON token.
+                if (reader.BytesConsumed == 0)
```
normj, line 104: If I'm misunderstanding the code in here with the `leftover` and `previousBufferLength` variables that is why this code needs some code. I can read the code but it is hard to understand the whys. That was why in my sample I gave you I broke the logic into spaced logic with comments for that blocks so you could understand the stages of the logic.

sdk/src/Core/Amazon.Runtime/Internal/Util/StreamingUtf8JsonReader.cs
peterrsongg: It isn't a brand new empty buffer since we do 
```
buffer = resizedBuffer;
```
The buffer is basically resizedBuffer with its first half filled with buffer's data. So, the offset needs to be `leftover.Length` in this case because we don't want to overwrite that data, which is what would happen if we set the offset to 0.

leftover's length is also important because when we do
```
buffer = resizedBuffer
```

we lose the length of the buffer (since it is now buffer.Length * 2). So that is how we keep track of the offset as well as how much data to read. And since we use leftover.Length in the other else block as well, I'd rather just keep it there. It doesn't allocate a new buffer, since we're calling `AsSpan()`, and 

I guess since reader.BytesConsumed = 0. I can get rid of perviousBufferLength though, since leftover.Length == previousBufferLength at this point.

I will add more comments, as it is nuanced

sdk/src/Core/Amazon.Runtime/Internal/Util/StreamingUtf8JsonReader.cs
peterrsongg: I explain in the comment above. however, I will remove `previousBufferLength` because that is just `leftover.Length` in this case.

sdk/src/Core/Amazon.Runtime/Internal/Transform/JsonUnmarshallerContext.cs
peterrsongg: Oh good catch. This one I just forgot to remove. Since the method signatures are different, I removed it from `UnmarshallerContext`.

sdk/src/Core/Amazon.Runtime/Internal/Util/StreamingUtf8JsonReader.cs
peterrsongg: Gotcha, I've added much more comments

peterrsongg:
Based on our conversation internally I made some updates. I also added a comment why we do `buffer.AsSpan(0, bytesRead)` instead of `buffer.AsSpan(0, bytesRead + leftover.Length)`.

sdk/src/Core/Amazon.Runtime/Internal/Util/StreamingUtf8JsonReader.cs
```diff
@@ -0,0 +1,169 @@
...
+                    leftover.CopyTo(buffer);
+                    bytesRead = FillBuffer(stream, ref buffer, leftover.Length, buffer.Length - leftover.Length);
+                }
+            }
+            else
+            {
+                bytesRead = FillBuffer(stream, ref buffer, 0, buffer.Length);
+            }
+
+            if (bytesRead == 0)
```
normj, line 140: I'm thinking this is getting complicated because we are depending on if the code goes into the `else` block up above dealing with leftover data we also expect this code to handle how to create the reader.  What do you think of this version where whenever we go into any condition block it does everything it needs to do and not rely on logic somewhere else to finish things up.

I also updated how the `Utf8JsonReader` was created in the resize condition to say if it is the final block if we did not fill up the entire new buffer.

```csharp
private static void GetMoreBytesFromStream(Stream stream, ref byte[] buffer, ref Utf8JsonReader reader)
{
    if (reader.BytesConsumed < buffer.Length)
    {
        ReadOnlySpan<byte> leftover = buffer.AsSpan().Slice((int)reader.BytesConsumed);

        // If BytesConsumed is 0 that means that the previous Read failed because the JSON token was too large to fit in the buffer.
        // In that case we need to resize the buffer and try again to read the JSON token.
        if (reader.BytesConsumed == 0)
        {
            var resizedBuffer = ArrayPool<byte>.Shared.Rent(Math.Min(int.MaxValue, (buffer.Length * 2)));
            Logger.GetLogger(typeof(StreamingUtf8JsonReader)).DebugFormat("Resizing buffer from {0} to {1}", buffer.Length, resizedBuffer.Length);
            // copy over the data from the previous read's buffer to the newly resized buffer.
            buffer.AsSpan().CopyTo(resizedBuffer);
            // return the previous buffer to the pool and set the new buffer to equal the resized buffer.
            ArrayPool<byte>.Shared.Return(buffer);
            buffer = resizedBuffer;
            // fill the new resized buffer with data from the stream. the offset MUST be leftover.Length 
            // so we don't overwrite the data that was copied from the previous buffer, and the number of bytes
            // we read must be buffer.Length - leftover.Length which is just the second half of the buffer.
            var bytesRead = FillBuffer(stream, ref buffer, leftover.Length, buffer.Length - leftover.Length);
            var resizedSpan = buffer.AsSpan(0, bytesRead + leftover.Length);
            reader = new Utf8JsonReader(resizedSpan, isFinalBlock: bytesRead == 0 || resizedSpan.Length != resizedBuffer.Length, reader.CurrentState);
            // early return since we have a reader
            return;
        }
        // The buffer has had some data processed but there is leftover data unprocessed.
        // In this case we move the unprocessed data to the start of the buffer and fill the rest with new data.
        else
        {
            leftover.CopyTo(buffer);
            var bytesRead = FillBuffer(stream, ref buffer, leftover.Length, buffer.Length - leftover.Length);
            reader = new Utf8JsonReader(buffer.AsSpan(0, bytesRead + leftover.Length), isFinalBlock: ((bytesRead + leftover.Length) != buffer.Length || bytesRead == 0), reader.CurrentState);
            return;
        }
    }
    // In this case enture buffer has been processed so attempt to refill the enture
    // buffer.
    else
    {
        var bytesRead = FillBuffer(stream, ref buffer, 0, buffer.Length);
        reader = new Utf8JsonReader(buffer.AsSpan(0, bytesRead), isFinalBlock: bytesRead == 0, reader.CurrentState);
        return;
    }
}
```

sdk/test/UnitTests/Custom/Marshalling/StreamingUtf8JsonReaderTests.cs
```diff
@@ -0,0 +1,229 @@
...
+    /// </summary>
+    [TestClass]
+    public class StreamingUtf8JsonReaderTests
+    {
+        private static int originalBufferSize;
+        [ClassInitialize]
+        public static void Initialize(TestContext testContext)
+        {
+            originalBufferSize = AWSConfigs.StreamingUtf8JsonReaderBufferSize.GetValueOrDefault();
+            AWSConfigs.StreamingUtf8JsonReaderBufferSize = 4096;
```
normj, line 29: I would recommend adding a second constructor on the `StreamingUtf8JsonReader` that takes in the buffer size so you don't have to toggle the `AWSConfigs.StreamingUtf8JsonReaderBufferSize` global setting.

sdk/src/Core/Amazon.Runtime/Internal/Transform/JsonErrorResponseUnmarshaller.cs
```diff
@@ -14,115 +14,131 @@
...
         /// </summary>
         /// <param name="context">The json parsing context. 
+        /// <param name="reader">The Utf8JsonReader</param>
         /// Usually an <c>Amazon.Runtime.Internal.JsonUnmarshallerContext</c>.</param>
         /// <returns>An <c>ErrorResponse</c> object.</returns>
-        public ErrorResponse Unmarshall(JsonUnmarshallerContext context)
+        public ErrorResponse Unmarshall(JsonUnmarshallerContext context, ref StreamingUtf8JsonReader reader)
         {
             ErrorResponse response;
+            bool firstByte = false;
```
normj, line 42: I'm getting compile errors due to `firstByte` never being used.

sdk/src/Core/Amazon.Runtime/Internal/Util/StreamingUtf8JsonReader.cs
```diff
@@ -0,0 +1,181 @@
...
+            {
+                // empty buffer will be returned 
+                reader = new Utf8JsonReader(buffer.AsSpan(0, bytesRead), isFinalBlock: true, reader.CurrentState);
+                return;
+            }
+
+            // passing in 0, bytesRead +leftover.Length is safe here because even if we are on the last buffer block, and leftover is junk, that means bytesRead would've been zero
```
normj, line 159: nit: Can you give context for the "passing in 0". Just something like "Passing in 0 for the start position and the span length to (bytesRead + leftover.Length) is safe here ..."

sdk/src/Core/Amazon.Runtime/Internal/Transform/JsonErrorResponseUnmarshaller.cs
peterrsongg: 😓 yeah that variable isn't needed. made sure core compiles

normj (approved):

muhammad-othman (approved):

PR #3556, opened 2024-11-21
Merge V3 into V4

dscpinheiro:
## Description
Merges V3 into V4 (as of last Friday - 11/15) since dry-runs in V4 are currently broken: In addition to some service changes (Elastic Transcoder being discontinued), there are some tests that are failing in the HEAD of the `v4-development` branch (see inline comments for more details).

These are the relevant commits:
- Merge commit: https://github.com/aws/aws-sdk-net/pull/3556/commits/3ff1cd930d360f17b70789e91cbb862454ecdf4c
- Update to generated files: https://github.com/aws/aws-sdk-net/pull/3556/commits/c44be2c32a9c224ab0426bfde7aa0070b34a01d9
- Manual changes to tests / merge conflicts: https://github.com/aws/aws-sdk-net/pull/3556/commits/5aee938a09bfa5f659c3edc07382d7ea70107d55

## Testing
Dry-run: `DRY_RUN-7c729c82-ad66-47f3-a65d-7e2a8eada339`

## Checklist

## License


sdk/test/IntegrationTests/Tests/General.cs
```diff
@@ -680,6 +680,7 @@ public void TestManualClockCorrection()
         // By default it only tests a small subset of services.
         [TestMethod]
         [TestCategory("General")]
+        [Ignore]
```
dscpinheiro, line 683: This test always fails at least once before succeeding, and it's currently commented out in the NetStandard project. @peterrsongg is working on a SEP for clock skew, this test should be revisited after that.

sdk/test/Services/DynamoDBv2/IntegrationTests/JSONTests.cs
```diff
@@ -95,7 +95,7 @@ public void TestDocumentPutGet()
 
         [TestMethod]
         [TestCategory("DynamoDBv2")]
-        public static void TestDocumentJsonConversions()
```
dscpinheiro, line 98: @normj These tests weren't running. It looks like the methods cannot be static, this was the message on our build system:
> [MSTest][Discovery][AWSSDK.IntegrationTests.NetFramework.dll] UTA007: Method TestDocumentBinaryDecoding defined in class AWSSDK_DotNet.IntegrationTests.Tests.DynamoDB.DynamoDBTests does not have correct signature. Test method marked with the [TestMethod] attribute must be non-static, public, return-type as void  and should not take any parameter. Example: public void Test.Class1.Test(). Additionally, if you are using async-await in test method then return-type must be Task. Example: public async Task Test.Class1.Test2()

I also had to fix the sample JSON used as it was complaining about trailing commas.

sdk/test/NetStandard/UnitTests/Services/DynamoDBTests.cs
```diff
@@ -203,6 +203,7 @@ public async Task TestMockingTableClient_TransactWriteAsync()
         public void TestMockingTableClient_LoadTable_Throws()
         {
             var mockClient = new Mock<IAmazonDynamoDB>();
+            mockClient.Setup(x => x.Config).Returns(new AmazonDynamoDBConfig());
```
dscpinheiro, line 206: This test was failing even before the merge (i.e. on the `HEAD` of `v4-development`). The issue was `Table.LoadTable` (which is tested below) now tries to apply our observability wrapper but the config object on our mock client was null (so we'd get a NPE instead of invalid operation).

sdk/src/Services/DynamoDBv2/Custom/DataModel/AsyncSearch.cs
```diff
@@ -15,7 +15,6 @@
 
 using Amazon.DynamoDBv2.DocumentModel;
 using Amazon.Runtime.Telemetry.Tracing;
-using Amazon.Runtime.Telemetry.Tracing;
```
dscpinheiro: For all DynamoDB changes here: Looks like some lines were added twice during the merge commit. I removed them and confirmed the contents of the `sdk/src/Services/DynamoDBv2/Custom` folder match what we had in `v4-development` before.

muhammad-othman (approved):

sdk/test/IntegrationTests/Tests/General.cs
dscpinheiro: Yes, updated the reason for skipping the tests in https://github.com/aws/aws-sdk-net/pull/3556/commits/042cf27049f1eb325c11853e0b26ff20b15dc9b0

boblodgett (approved):

PR #3559, opened 2024-11-26
Fix failed tests and Move DefaultAWSCredentials to the service config

muhammad-othman:
## Description
This PR contains the following commits:
* https://github.com/aws/aws-sdk-net/commit/e1a5773d7088e4df3e0209ad9104ffbc95797875 This commit allows all the services to generate the `AuthResolver` to be able to run the dry run, but will only commit `AutoScaling` and `S3` changes to keep the PR small.
    * Also in the generator I've removed the condition around creating `CustomizeRuntimePipeline` method to allow test clients to add the `AuthResolverHandler` to the pipeline.
* https://github.com/aws/aws-sdk-net/commit/50f00465507f613b6158d668341069ed8c548968 `CognitoSync` service was failing because we added a customization that removes `CredentialsRetriever` then adds a custom implementation after the Marshaller, this mess up the order of execution for the pipeline and causes the `CredentialsRetriever` to be executed before the auth handler, which causes it to fail since at this point the credentials weren't added yet.
This commit updates the customization to use `replace` to keep `CredentialsRetriever` right order.
* https://github.com/aws/aws-sdk-net/commit/9a1cf7ea908f386b66f7f8685687afe776833ad6 We've removed the constant signer for `SimpleDB` which causes its authentication fails because it was using `QueryStringSigner` since its `signatureVersion` is v2 in the models, this commit reverts this change.
* https://github.com/aws/aws-sdk-net/commit/f19f5eca8bedf66719490d6287547e9e8fa477c8 FallbackCredentialsFactorySTSTests were failing because we removed the credentials from `CredentialsRetriever` ctor, here I'm removing passing the credentials to the tests.
* https://github.com/aws/aws-sdk-net/commit/a70501fd3d7b4cbc7d450c0560b818ca51ee05c9 Moving `DefaultAWSCredentials` from the service client to the service config.
* https://github.com/aws/aws-sdk-net/commit/ddfbf2ce03f380c0c5efd4d143db70ae627def98 Skipped `BearerTokenServiceTests` untill `smithy.api#httpBearerAuth` schemeId is fully implemented.

## Motivation and Context
* `DOTNET-7829`
* Fix failing tests

## Testing
* `DRY_RUN-d4fb9bde-2fa3-48e7-b57b-9e90a226da7d`.

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/ClientConfig.cs
+++ b/sdk/src/Core/Amazon.Runtime/ClientConfig.cs
@@ -90,6 +90,7 @@ namespace Amazon.Runtime
         private const int MaxRetriesDefault = 2;
         private const long DefaultMinCompressionSizeBytes = 10240;
         private bool didProcessServiceURL = false;
+        private AWSCredentials _defaultAWSCredentials = null;
         private IIdentityResolverConfiguration _identityResolverConfiguration = DefaultIdentityResolverConfiguration.Instance;
         private IAWSTokenProvider _awsTokenProvider = new DefaultAWSTokenProviderChain();
         private TelemetryProvider telemetryProvider = AWSConfigs.TelemetryProvider;
@@ -182,6 +183,14 @@ namespace Amazon.Runtime
 
             return null;
         }
+
+        /// <inheritdoc />
+        public AWSCredentials DefaultAWSCredentials
+        {
+            get { return this._defaultAWSCredentials; }
+            set { this._defaultAWSCredentials = value; }
+        }
+
         /// <inheritdoc />
         public IIdentityResolverConfiguration IdentityResolverConfiguration
         {
--- a/sdk/src/Core/Amazon.Runtime/IClientConfig.cs
+++ b/sdk/src/Core/Amazon.Runtime/IClientConfig.cs
@@ -58,6 +58,17 @@ namespace Amazon.Runtime
         Profile Profile { get; }
 
         /// <summary>
+        /// Gets the default <see cref="AWSCredentials"/> by the service client.
+        /// <remarks>
+        /// <para>
+        /// Credentials provided directly to <see cref="AmazonServiceClient"/> constructor 
+        /// will take precedence over these default credentials.
+        /// </para>
+        /// </remarks>
+        /// </summary>
+        AWSCredentials DefaultAWSCredentials { get; }
+
+        /// <summary>
         /// Gets the configuration for identity resolvers used by the service client, which manages
         /// the resolvers used to obtain various types of identities.
         /// Use this to retrieve and manage specific identity resolvers based on the required identity type.
--- a/sdk/src/Core/Amazon.Runtime/IClientConfig.cs
+++ b/sdk/src/Core/Amazon.Runtime/IClientConfig.cs
@@ -57,14 +57,19 @@ namespace Amazon.Runtime
         /// </summary>
         Profile Profile { get; }
 
-        /// <summary>
-        /// Gets the default <see cref="AWSCredentials"/> by the service client.
-        /// <remarks>
-        /// <para>
-        /// Credentials provided directly to <see cref="AmazonServiceClient"/> constructor 
-        /// will take precedence over these default credentials.
-        /// </para>
-        /// </remarks>
+        /// <summary> 
+        /// <para> 
+        /// The AWS credentials used for authenticating calls to AWS for services using AWS signature version 4 (SigV4). 
+        /// SigV4 is the most common authentication mechanism used for AWS service calls. If AWSCredentials are used as a 
+        /// parameter to the service client's constructor the value will be set on this property. 
+        /// </para> 
+        /// <para> 
+        /// Common instances of AWSCredentials are <see cref="Amazon.Runtime.BasicAWSCredentials" /> for static credentials and 
+        /// <see cref="Amazon.Runtime.AssumeRoleAWSCredentials" /> for getting credentials by assuming an IAM role. 
+        /// </para> 
+        /// <para> 
+        /// If null, the SDK will determine which credentials to use at request time using information from the source service model. 
+        /// </para> 
         /// </summary>
         AWSCredentials DefaultAWSCredentials { get; }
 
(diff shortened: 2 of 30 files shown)


sdk/src/Core/Amazon.Runtime/IClientConfig.cs
```diff
@@ -57,6 +57,17 @@ public partial interface IClientConfig
         /// </summary>
         Profile Profile { get; }
 
+        /// <summary>
+        /// Gets the default <see cref="AWSCredentials"/> by the service client.
```
dscpinheiro, line 61: Is the summary here accurate? My understanding was these credentials would only be set if passed when creating the service client (e.g. `new AmazonEventBridgeClient(new AssumeRoleAWSCredentials(...))`

Or are we also populating them during the auth resolver?

dscpinheiro: My question was more related to the contents of this property. If I create a client as most customers would:
```
using var s3Client = new AmazonS3Client(new AmazonS3Config
{
    RegionEndpoint = RegionEndpoint.USEast2,
});
```

Would `s3Client.Config.DefaultAWSCredentials` be populated after I call any service operation (e.g. `await s3Client.PutObjectAsync(...)`) and the auth resolver runs? The current comment doesn't make that clear.

sdk/src/Core/Amazon.Runtime/ClientConfig.cs
dscpinheiro, line 1: Just thought of another question: Now that the credentials are available in the config, do we still need them in the auth resolver pipeline handler? https://github.com/aws/aws-sdk-net/blob/sra-identity-auth/sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs#L41

The `PreInvoke` method should have access to the config directly, right?

muhammad-othman: Yep, I think we can safely remove this property and use the one in the clientconfig.
Gonna work on this change.

sdk/src/Core/Amazon.Runtime/IClientConfig.cs
muhammad-othman: 1. If the user didn't set any credentials, we wont set the `DefaultAWSCredentials` anywhere and we will use the IdentityResolver to get the credentials with every request.
2. If the user set the `DefaultAWSCredentials` at the clientConfig, we will use it and we wont check the IdentityResolver 
https://github.com/aws/aws-sdk-net/blob/ddfbf2ce03f380c0c5efd4d143db70ae627def98/sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs#L82
3. If the user passed the credentials to the service client, we will set the config credentials to what the user passed here
https://github.com/aws/aws-sdk-net/blob/muhamoth/DOTNET-7829-sra-Move-DefaultAWSCredentials-to-service-config/sdk/src/Core/Amazon.Runtime/AmazonServiceClient.cs#L164

dscpinheiro: Got it. What do you think of:
```csharp
/// <summary>
/// <para>
/// Credentials specified when creating the service client, if any (for example, via <see cref="AssumeRoleAWSCredentials" />).
/// When set, the SDK will use these credentials for all API calls of the corresponding service client.
/// </para>
/// <para>
/// If null, the SDK will determine which credentials to use at request time using information from the source service model.
/// </para>
/// </summary>
AWSCredentials DefaultAWSCredentials { get; }
```

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs
```diff
@@ -79,7 +67,7 @@ protected void PreInvoke(IExecutionContext executionContext)
                 if (scheme is AwsV4aAuthScheme || scheme is AwsV4AuthScheme)
```
normj, line 67: Since we still have SimpleDB using query string signer do we need to update this condition so we use the credentials from the config object for SimpleDB?

sdk/src/Core/Amazon.Runtime/IClientConfig.cs
normj: Here is my tweak to the comment.
```
/// <summary>
/// <para>
/// The AWS credentials used for authenticating calls to AWS for services using AWS signature version 4 (SigV4).
/// SigV4 is the most common authentication mechanism used for AWS service calls. If AWSCredentials are used as a 
/// parameter to the service client's constructor the value will be set on this property.
/// </para>
/// <para>
/// Common instances of AWSCredentials are <see cref="Amazon.Runtime.BasicAWSCredentials" /> for static credentials and
/// <see cref="Amazon.Runtime.AssumeRoleAWSCredentials" /> for getting credentials by assuming an IAM role.
/// </para>
/// <para>
/// If null, the SDK will determine which credentials to use at request time using information from the source service model.
/// </para>
/// </summary>
AWSCredentials DefaultAWSCredentials { get; }
```

normj (changes requested):

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs
muhammad-othman: Nope, SimpleDB does use `AwsV4AuthScheme` and this condition will evaluate to true, however when it reaches the `GetSinger` call, it wont call this method https://github.com/aws/aws-sdk-net/blob/46cbc7042a7b3e78a3aaa6d339176d143539cfb9/sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs#L85 
It will call this overload instead
image

Similar to how S3 calls this overload to get the `S3Signer`
https://github.com/aws/aws-sdk-net/blob/46cbc7042a7b3e78a3aaa6d339176d143539cfb9/sdk/src/Services/S3/Generated/Internal/AmazonS3AuthResolver.cs#L81

This method is added if `DetermineSignerOverride` doesn't return null for the `SignatureVersion`
https://github.com/aws/aws-sdk-net/blob/46cbc7042a7b3e78a3aaa6d339176d143539cfb9/generator/ServiceClientGeneratorLib/Generators/AuthResolvers/ModeledResolver.tt#L76

dscpinheiro (approved):

normj (approved):

PR #3562, opened 2024-11-30
[SRA] Add identity resolver for bearer authentication scheme

dscpinheiro:
## Description
Updates the `IdentityResolverConfiguration` to resolve the bearer authentication scheme (i.e. the `smithy.api#httpBearerAuth` trait). I also had to modify other existing components (such as the bearer token signer) since their implementation is different than SigV4 / SigV4A credentials.

## Motivation and Context
`DOTNET-7652`

## Testing
- Dry-run: `DRY_RUN-23d6c74a-9d68-484f-b185-acd1787f844f`
- Ran a local application specifying a profile using SSO (note: while testing I noticed we're not including the Smithy project DLL in the Core `.nupkg` file - will address that in a separate PR).

## Types of changes

## Checklist

## License

### DIFF
--- a/extensions/src/Smithy.Identity.Abstractions/Identity/IIdentityResolver.cs
+++ b/extensions/src/Smithy.Identity.Abstractions/Identity/IIdentityResolver.cs
@@ -1,4 +1,7 @@
-﻿namespace Smithy.Identity.Abstractions
+﻿using System.Threading;
+using System.Threading.Tasks;
+
+namespace Smithy.Identity.Abstractions
 {
     /// <summary>
     /// An identity resolver is a component that is able to resolve a customer's <see cref="BaseIdentity"/>.
@@ -17,5 +20,11 @@
         /// If the identity cannot be resolved an exception will be thrown.
         /// </summary>
         BaseIdentity ResolveIdentity();
+
+        /// <summary>
+        /// Loads the customer's identity for this resolver. 
+        /// If the identity cannot be resolved an exception will be thrown.
+        /// </summary>
+        Task<BaseIdentity> ResolveIdentityAsync(CancellationToken cancellationToken);
     }
 }
--- a/sdk/src/Core/Amazon.Runtime/BearerTokenSigner.cs
+++ b/sdk/src/Core/Amazon.Runtime/BearerTokenSigner.cs
@@ -28,6 +28,11 @@ namespace Amazon.Runtime
     {
         public override bool RequiresCredentials { get; } = false;
 
+        /// <summary>
+        /// The resolved bearer token, determined during the auth resolver handler.
+        /// </summary>
+        public string ResolvedToken { get; internal set; }
+
 #if BCL
         public override void Sign(
             IRequest request, 
@@ -43,16 +48,12 @@ namespace Amazon.Runtime
                     "Endpoint must not use 'http'.");
             }
 
-            if (null == clientConfig.AWSTokenProvider)
-                return;
-
-            if (!clientConfig.AWSTokenProvider.TryResolveToken(out var token) ||
-                string.IsNullOrEmpty(token.Token))
+            if (string.IsNullOrEmpty(ResolvedToken))
             {
                 throw new AmazonClientException("No Token found.  Operation requires a Bearer token.");
             }
 
-            request.Headers["Authorization"] = $"Bearer {token}";
+            request.Headers["Authorization"] = $"Bearer {ResolvedToken}";
         }
 #else
         public override void Sign(
@@ -67,7 +68,7 @@ namespace Amazon.Runtime
 #endif
 
 #if AWS_ASYNC_API
-        public override async Task SignAsync(
+        public override Task SignAsync(
             IRequest request, 
             IClientConfig clientConfig, 
             RequestMetrics metrics, 
@@ -81,15 +82,13 @@ namespace Amazon.Runtime
                     "Endpoint must not use 'http'.");
             }
 
-            if (null == clientConfig.AWSTokenProvider)
-                return;
-
-            var result = await clientConfig.AWSTokenProvider.TryResolveTokenAsync(token).ConfigureAwait(false);
-
-            if (true != result.Success || string.IsNullOrEmpty(result.Value.Token))
+            if (string.IsNullOrEmpty(ResolvedToken))
+            {
                 throw new AmazonClientException("No Token found.  Operation requires a Bearer token.");
+            }
 
-            request.Headers["Authorization"] = $"Bearer {result.Value.Token}";
+            request.Headers["Authorization"] = $"Bearer {ResolvedToken}";
+            return Task.CompletedTask;
         }
 #endif
 
--- a/sdk/src/Core/Amazon.Runtime/Credentials/Internal/IdentityResolvers/DefaultAWSTokenIdentityResolver.cs
+++ b/sdk/src/Core/Amazon.Runtime/Credentials/Internal/IdentityResolvers/DefaultAWSTokenIdentityResolver.cs
@@ -1,0 +1,61 @@
+﻿/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+using Smithy.Identity.Abstractions;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Amazon.Runtime.Credentials.Internal.IdentityResolvers
+{
+    /// <summary>
+    /// A resolver that provides a bearer token identity.
+    /// </summary>
+    public class DefaultAWSTokenIdentityResolver : IIdentityResolver
+    {
+        private readonly IAWSTokenProvider _tokenProvider;
+
+        public DefaultAWSTokenIdentityResolver() : this(null) { }
+
+        public DefaultAWSTokenIdentityResolver(string profileName)
+            => _tokenProvider = new AWSTokenProviderChain(new ProfileTokenProvider(profileName));
+
+        /// <inheritdoc/>
+        public BaseIdentity ResolveIdentity()
+        {
+#if NETFRAMEWORK
+            if (_tokenProvider.TryResolveToken(out var token))
+            {
+                return token;
+            }
+
+            throw new AmazonClientException($"Failed to resolve bearer token in {nameof(DefaultAWSTokenIdentityResolver)}");
+#else
+            throw new System.NotSupportedException($"{nameof(ResolveIdentityAsync)} must be used in cross-platform .NET");
+#endif
+        }
+
+        /// <inheritdoc/>
+        public async Task<BaseIdentity> ResolveIdentityAsync(CancellationToken cancellationToken)
+        {
+            var tokenResponse = await _tokenProvider.TryResolveTokenAsync(cancellationToken).ConfigureAwait(false);
+            if (tokenResponse.Success)
+            {
+                return tokenResponse.Value;
+            }
+
+            throw new AmazonClientException($"Failed to resolve bearer token in {nameof(DefaultAWSTokenIdentityResolver)}");
+        }
+    }
+}
--- a/sdk/src/Core/Amazon.Runtime/Credentials/Internal/IdentityResolvers/IdentityResolverConfiguration.cs
+++ b/sdk/src/Core/Amazon.Runtime/Credentials/Internal/IdentityResolvers/IdentityResolverConfiguration.cs
@@ -23,17 +23,20 @@ namespace Amazon.Runtime.Credentials.Internal.IdentityResolvers
     public class DefaultIdentityResolverConfiguration : IIdentityResolverConfiguration
     {
         public static readonly IIdentityResolverConfiguration Instance = new DefaultIdentityResolverConfiguration();
-        private static readonly Dictionary<Type, IIdentityResolver> identityResolvers = new Dictionary<Type, IIdentityResolver>()
-            {
-                { typeof(AnonymousAWSCredentials), new AnonymousIdentityResolver() },
-                { typeof(AWSCredentials), new DefaultAWSCredentialsIdentityResolver() },
-            };
+        private static readonly Dictionary<Type, IIdentityResolver> identityResolvers = new()
+        {
+            { typeof(AnonymousAWSCredentials), new AnonymousIdentityResolver() },
+            { typeof(AWSCredentials), new DefaultAWSCredentialsIdentityResolver() },
+            { typeof(AWSToken), new DefaultAWSTokenIdentityResolver() },
+        };
 
         /// <inheritdoc/>
         public IIdentityResolver GetIdentityResolver<T>() where T : BaseIdentity
         {
-            if(identityResolvers.TryGetValue(typeof(T), out var identityResolver))
+            if (identityResolvers.TryGetValue(typeof(T), out var identityResolver))
+            {
                 return identityResolver;
+            }
 
             throw new NotImplementedException($"{typeof(T).Name} is not supported");
         }
--- a/sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs
+++ b/sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs
@@ -20,17 +20,18 @@ using Smithy.Identity.Abstractions;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading.Tasks;
 
 namespace Amazon.Runtime.Internal
 {
     public abstract class BaseAuthResolverHandler : PipelineHandler
     {
-        // These are the client supported schemes, included here for now.
         private readonly HashSet<IAuthScheme<BaseIdentity>> _supportedSchemes = new()
         {
             new AnonymousAuthScheme(),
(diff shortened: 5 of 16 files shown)


extensions/src/Smithy.Identity.Abstractions/Identity/IIdentityResolver.cs
```diff
@@ -17,5 +20,11 @@ public interface IIdentityResolver
         /// If the identity cannot be resolved an exception will be thrown.
         /// </summary>
         BaseIdentity ResolveIdentity();
+
+        /// <summary>
+        /// Loads the customer's identity for this resolver. 
+        /// If the identity cannot be resolved an exception will be thrown.
```
normj, line 26: Can we say what exception will be thrown? I'm guessing we are going to get into issues with the exception being AWS exception but just want to ask.

sdk/src/Core/Amazon.Runtime/Credentials/Internal/IdentityResolvers/DefaultAWSTokenIdentityResolver.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+#if NETFRAMEWORK
+            if (_tokenProvider.TryResolveToken(out var token))
+            {
+                return token;
+            }
+
+            throw new AmazonClientException($"Failed to resolve bearer token in {nameof(DefaultAWSTokenIdentityResolver)}");
+#else
+            throw new System.NotSupportedException($"{nameof(ResolveIdentityAsync)} must be used in cross-platform .NET");
```
normj, line 45: Not liking this pattern of having methods exposed for cross platform .NET that then throw a runtime exception. We should either not have them or accept they have to do blocking async calls.

sdk/src/Core/Amazon.Runtime/Credentials/Internal/IdentityResolvers/IdentityResolverConfiguration.cs
```diff
@@ -23,17 +23,20 @@ namespace Amazon.Runtime.Credentials.Internal.IdentityResolvers
...
+            { typeof(AnonymousAWSCredentials), new AnonymousIdentityResolver() },
+            { typeof(AWSCredentials), new DefaultAWSCredentialsIdentityResolver() },
+            { typeof(AWSToken), new DefaultAWSTokenIdentityResolver() },
+        };
 
         /// <inheritdoc/>
         public IIdentityResolver GetIdentityResolver<T>() where T : BaseIdentity
         {
-            if(identityResolvers.TryGetValue(typeof(T), out var identityResolver))
+            if (identityResolvers.TryGetValue(typeof(T), out var identityResolver))
```
normj, line 36: Seeing this `typeof` calls and working with type properties reminds we should a NativeAOT compilation test. I don't want to rely only on the warnings as that has burned us in the past.

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/Signer.cs
```diff
@@ -106,7 +106,19 @@ public static void SignRequest(IRequestContext requestContext)
...
                 return;
+            }
+
+            // The signer interface expects immutable credentials, which does not match the AWSToken structure.
+            // We'll manually set a property so that the bearer token still works as expected.
+            if (requestContext.Signer is BearerTokenSigner tokenSigner)
+            {
+                if (requestContext.Identity is AWSToken resolvedToken)
+                {
+                    tokenSigner.ResolvedToken = resolvedToken.Token;
```
normj, line 119: We are mutating state on the bearer token signer which is being use across threads. We need a safer way to do this.

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/Signer.cs
```diff
@@ -143,7 +155,19 @@ private static async Task SignRequestAsync(IRequestContext requestContext)
...
```
normj, line 168: We are mutating state on the bearer token signer which is being use across threads. We need a safer way to do this.

normj (changes requested):

sdk/src/Core/Amazon.Runtime/BearerTokenSigner.cs
```diff
@@ -28,6 +28,11 @@ public class BearerTokenSigner : AbstractAWSSigner
     {
         public override bool RequiresCredentials { get; } = false;
 
+        /// <summary>
```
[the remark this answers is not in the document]
dscpinheiro: Yes, I think that makes sense. I hadn't noticed its location but I'll move it.

sdk/src/Core/Amazon.Runtime/Credentials/Internal/IdentityResolvers/IdentityResolverConfiguration.cs
dscpinheiro: I asked about this on a previous PR: https://github.com/aws/aws-sdk-net/pull/3531#discussion_r1823088083

sdk/src/Core/Amazon.Runtime/Credentials/Internal/IdentityResolvers/DefaultAWSTokenIdentityResolver.cs
dscpinheiro: Not a fan either. The SSO components only include the sync APIs when using the .NET Framework: https://github.com/aws/aws-sdk-net/blob/6cb3b06458449566cd85c4c79bb7849bcd4ceb48/sdk/src/Services/SSOOIDC/Custom/_bcl45%2Bnetstandard/AmazonSSOOIDCClient.Extension.cs#L26-L31

I'll update this to use a blocking call even though it's a bad practice.

extensions/src/Smithy.Identity.Abstractions/Identity/IIdentityResolver.cs
dscpinheiro: Yes, it's an Amazon exception although I noticed the fallback credentials factory currently uses `AmazonServiceException` but I used `AmazonClientException` here. They should be the same, right?

sdk/src/Core/Amazon.Runtime/BearerTokenSigner.cs
```diff
@@ -43,16 +48,12 @@ public override void Sign(
                     "Endpoint must not use 'http'.");
             }
 
-            if (null == clientConfig.AWSTokenProvider)
-                return;
-
-            if (!clientConfig.AWSTokenProvider.TryResolveToken(out var token) ||
-                string.IsNullOrEmpty(token.Token))
+            if (string.IsNullOrEmpty(ResolvedToken))
```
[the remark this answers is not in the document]
dscpinheiro: Yes, I moved the common logic to a private method in https://github.com/aws/aws-sdk-net/pull/3562/commits/8f86f7c0a8b212e51a0dfd29fe586b6fa8d34ae6#diff-113f1e4a9c0bdc1d09dba3df83ee6c023433efb2aa1aca37f862650c5fd043f5

sdk/src/Core/Amazon.Runtime/BearerTokenSigner.cs
```diff
@@ -43,16 +48,12 @@ public override void Sign(
...
             {
                 throw new AmazonClientException("No Token found.  Operation requires a Bearer token.");
             }
 
-            request.Headers["Authorization"] = $"Bearer {token}";
+            request.Headers["Authorization"] = $"Bearer {ResolvedToken}";
         }
 #else
         public override void Sign(
```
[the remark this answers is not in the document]
dscpinheiro: Yes, this was a miss on my part. The conditional compilations were required before because this class was using the token provider implementation directly. It's not needed anymore now that the token is provided differently.

sdk/src/Core/Amazon.Runtime/Credentials/Internal/IdentityResolvers/DefaultAWSTokenIdentityResolver.cs
dscpinheiro: Added the blocking call in https://github.com/aws/aws-sdk-net/pull/3562/commits/8f86f7c0a8b212e51a0dfd29fe586b6fa8d34ae6#diff-96434be2538b1000e07b1470f39a00fcb6c772b0fb91ba06c74a65e277b24cce

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/Signer.cs
dscpinheiro: I updated the signer to use a lock here: https://github.com/aws/aws-sdk-net/pull/3562/commits/8f86f7c0a8b212e51a0dfd29fe586b6fa8d34ae6#diff-4f67ae8bdcfede242a71fccc1853340fa9affa50ebf705fdff16b52438253293

Let me know if there's a better way to do this.

extensions/src/Smithy.Identity.Abstractions/Identity/IIdentityResolver.cs
normj: Actually it looks like `FallbackCredentialsFactory` has two spots it throws `AmazonClientException` and one spot where throws `AmazonServiceException`. Using `AmazonClientException` makes more sense to me.

sdk/src/Core/Amazon.Runtime/Credentials/Internal/IdentityResolvers/DefaultAWSTokenIdentityResolver.cs
```diff
@@ -0,0 +1,68 @@
...
+#if NETFRAMEWORK
+            if (_tokenProvider.TryResolveToken(out var token))
+            {
+                return token;
+            }
+
+            throw new AmazonClientException($"Failed to resolve bearer token in {nameof(DefaultAWSTokenIdentityResolver)}");
+#else
+            // This is bad, but the token provider does not have a synchronous implementation for the .NET Framework.
+            var tokenResponse = _tokenProvider.TryResolveTokenAsync().Result;
```
normj, line 46: Instead of using `.Result` we should be using `.GetAwaiter().GetResult()` when we have to make these blocking async calls. It handles exceptions better. Using `.GetAwaiter().GetResult()` you will get the real exception and `.Result` will get an AggregateException.

sdk/src/Core/Amazon.Runtime/Credentials/Internal/IdentityResolvers/IdentityResolverConfiguration.cs
normj: I'm pretty sure you are good with NativeAOT because you aren't asking for anything that won't be known at compile time. I just think as a sanity check at the end of SRA when we are doing bug bashing we should include some NativeAOT testing.

sdk/src/Core/Amazon.Runtime/Internal/Auth/BearerTokenSigner.cs
```diff
@@ -13,66 +13,47 @@
...
         }
-#endif
 
 #if AWS_ASYNC_API
-        public override async Task SignAsync(
+        public override Task SignAsync(
             IRequest request, 
             IClientConfig clientConfig, 
             RequestMetrics metrics, 
             ImmutableCredentials credentials,
```
normj, line 48: How much would that break things if we change the `credentials` parameter to `BaseIdentity` and then each signer cast the credentials to the type it expects. I really don't like having this `ResolvedToken` instance state for an object that is going to be used multi threaded.

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs
```diff
@@ -67,19 +68,86 @@ protected void PreInvoke(IExecutionContext executionContext)
...
+            var authOptions = ResolveAuthOptions(executionContext);
+            if (authOptions == null || authOptions.Count == 0)
+            {
+                throw new AmazonClientException($"No valid authentication schemes defined for ${executionContext.RequestContext.RequestName}");
+            }
+
+            var clientConfig = executionContext.RequestContext.ClientConfig;
+            var cancellationToken = executionContext.RequestContext.CancellationToken;
+
+            foreach (var authOption in authOptions)
```
normj, line 109: Out of curiosity of no auth option is picked after we exhaust the `authOptions` list should we throw an error that there is no auth compatible? Especially if `authOptions` had values in it.

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/Signer.cs
normj: See my comment above about switching the `Sign` method to take an an `BaseIdentity` for `credentials`.

sdk/src/Core/Amazon.Runtime/Credentials/Internal/IdentityResolvers/DefaultAWSTokenIdentityResolver.cs
dscpinheiro: Fixed in https://github.com/aws/aws-sdk-net/pull/3562/commits/32ab9122a184a2403d6176c675b3389b00b107ce

sdk/src/Core/Amazon.Runtime/Internal/Auth/BearerTokenSigner.cs
dscpinheiro: I updated the signer method signature in https://github.com/aws/aws-sdk-net/pull/3562/commits/32ab9122a184a2403d6176c675b3389b00b107ce (it did require changing more code than I expected, but I removed this instance state).

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs
dscpinheiro: Yes, I think we should (although I'd prefer doing it in a separate PR at this point). 

One other scenario we need to account for is if a service has multiple auth schemes but we can't resolve the first one for some reason (e.g. `"auth": [ "aws.auth#sigv4a", "aws.auth#sigv4"]`). We "know" how to do SigV4A but will fail if the CRT isn't available (and that's the best customer experience when there's another scheme there).

muhammad-othman (approved):

sdk/src/Core/Amazon.Runtime/Internal/Auth/EventBridgeSigner.cs
```diff
@@ -33,27 +34,29 @@ public override ClientProtocol Protocol
 
         public override void Sign(IRequest request, IClientConfig clientConfig, RequestMetrics metrics, string awsAccessKeyId, string awsSecretAccessKey)
         {
-            Sign(request, clientConfig, metrics, new ImmutableCredentials(awsAccessKeyId, awsSecretAccessKey, ""));
+            Sign(request, clientConfig, metrics, new BasicAWSCredentials(awsAccessKeyId, awsSecretAccessKey));
```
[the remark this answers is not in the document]
dscpinheiro: Bo refactored this section a couple of months ago and this signer is not used anymore: https://github.com/aws/aws-sdk-net/commit/33e75593caa43bc01e3b0dd49ce32bc3d5c7f5ba

Agree we should remove before V4 goes GA.

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs
```diff
@@ -64,19 +65,84 @@ protected void PreInvoke(IExecutionContext executionContext)
...
+                if ((scheme is AwsV4aAuthScheme || scheme is AwsV4AuthScheme) && clientConfig.DefaultAWSCredentials != null)
                 {
                     // We can use DefaultAWSCredentials if it was set by the user for these schemes.
-                    executionContext.RequestContext.Identity = executionContext.RequestContext.ClientConfig.DefaultAWSCredentials;
+                    executionContext.RequestContext.Identity = clientConfig.DefaultAWSCredentials;
+                    break;
                 }
 
-                if (executionContext.RequestContext.Identity == null)
+#if NETFRAMEWORK
```
muhammad-othman, line 77: I think we should add an else statement that calls the async version in a blocking way, similar what was done in `DefaultAWSTokenIdentityResolver.ResolveIdentity`, to account for the internal sync actions being called.

sdk/src/Core/Amazon.Runtime/Credentials/Internal/IdentityResolvers/DefaultAWSTokenIdentityResolver.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+#if NETFRAMEWORK
+            if (_tokenProvider.TryResolveToken(out var token))
+            {
+                return token;
+            }
+
+            throw new AmazonClientException($"Failed to resolve bearer token in {nameof(DefaultAWSTokenIdentityResolver)}");
+#else
+            // This is bad, but the token provider does not have a synchronous implementation for the .NET Framework.
```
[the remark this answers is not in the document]
dscpinheiro: Fixed in https://github.com/aws/aws-sdk-net/pull/3562/commits/2b8461c9fb031e589c047f211ac47e7b88d8c412

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs
dscpinheiro: Updated in https://github.com/aws/aws-sdk-net/pull/3562/commits/f2b2126357d7095d72c3cf889bb3c6411db788a9

normj (commented):
In the `AbstractAWSSigner` we need to get rid of the abstract `Sign` operation that takes in a access and secret key so we are always dealing with the `BaseIdentity`. I created a PR to this PR that does that work and I'm currently running the dry run. Assuming the dry run is successful and @muhammad-othman approves my PR we can merge that PR into this PR and then I'll approve this PR so we can merge it.

https://github.com/aws/aws-sdk-net/pull/3575

normj (approved):

PR #3571, opened 2024-12-09
Add DSQL auth token generator

normj:
## Description
Add utility code for generating auth token used for connect to the new DSQL service. This code is very similar to the RDS auth token generator feature already in the SDK. https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Services/RDS/Custom/Util/RDSAuthTokenGenerator.cs#L29

Well adding this I fixed up the documentation in the RDS auth generator to match what I was saying in the DSQL version. Not sure why the RDS implied it would only search for credentials through EC2 instance metadata.

## Motivation and Context
Make it easier to generate the DSQL auth token for .NET users.

## Testing
Added new unit tests and ran a successful dry run. I also created a DSQL cluster and confirmed I was able to generate an auth token and connect to the cluster with the auth token.

### DIFF
--- a/generator/.DevConfigs/5B1CABC0-FA39-48C5-B8BB-D4DBBFF4F45D.json
+++ b/generator/.DevConfigs/5B1CABC0-FA39-48C5-B8BB-D4DBBFF4F45D.json
@@ -1,0 +1,18 @@
+{
+    "services": [
+        {
+            "serviceName": "DSQL",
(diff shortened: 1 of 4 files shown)


boblodgett (approved):

eddiemcs3 (approved):

mullermp:
Nice!

PR #3572, opened 2024-12-09
[V4] DateTime handling adjustments for inconsistencies between local and UTC times.

boblodgett:
## Description

This PR addresses the inconsistencies of using LocalTime and UniversalTime within the SDK to consistently use UniversalTime (UTC) whenever possible as a first choice and only using LocalTime when appropriate to do so.

DevConfig commit: 8884a550237eb5f854fe495ea75f80fcb5d3ab00
Manual changes are in these commits only: cb4f74d744253439422feb57ba53d79666b276d1,  5f8866085fb055d8333432804a2ff5cb2e8730e8, 3f93a6d739a92a7011de5b04a96061891ee43576, and c072d6d4b2fcf0707565a5f3e830549161dd324a
Generated changes commit: 9611cbaae076a25c258b6e341f762942adbf9ef9

Note for reviewers:

* Please pay special attention to things like expiration times, data sent to and from DynamoDb, and credentials.
* DynamoDB (TestContext test [.\sdk\test\NetStandard\IntegrationTests\IntegrationTests\DynamoDB\DataModelTests.cs] gave me some problems and unspecified times can still be sent to and from DynamoDb. Testing may need to expand here).
* Pay attention to the impact of the breaking change list and how that could impact developers.
* Cosmetic changes exist in this PR. See all notes below.
* One TODO needs some discussion for SimpleDB. (Decided to leave as is for now using localtime)

## Motivation and Context

**Breaking changes:** Changes where the behavior is now different and may or may not produce a compile time error.

* Removed obsolete properties such `EndTime` then changed `EndTimeUtc` to `EndTime`. This could lead to offset times if anyone was still using the marked obsolete original `EndTime` for example. A compile time error will occur for anyone using `EndTimeUtc`. Removing BackwardCompatibility properties and code in the generator is part of this change. Also id this to the manual S3 models.
* Response unmarshallers for TimeStamps and list TimeStamps for formats `TimestampFormat.ISO8601 ||  TimestampFormat.RFC822` datetimes were being parsed into local times. Adjusted DateTime parsing to return UTC times. (BaseResponseUnmarshaller.tt, MultiValueHeaderParser.cs)
* Fixed the DateTimeUnmarshaller which was parsing datetime strings into and returning them as local time which in some cases were still getting converted back to UTC on a prior bug fix but not always. DateTime strings unmarshalled are assumed to be UTC time and will be specified and unmarshalled as UTC. (SimpleTypeUnmarshaller.cs)
* ConvertFromUnixEpochSeconds/ConvertFromUnixEpochMilliseconds incorrectly returning the Unix Epoch time as localtime instead of by definition is a UTC time. This changes the behavior where these methods were used. These were assigned to headers typically: https://github.com/search?q=repo%3Aaws%2Faws-sdk-net+ConvertFromUnixEpoch&type=code (AWSSDKUtils.cs)
* DynamoDB RetrieveDateTimeInUtc has been switched to `true` as the default.

**Cosmetic changes:** These changes were mainly so that I could keep track and ensure I looked at and ensured all code using DateTime.Now / .ToUniversalTime, .ToLocalTime, new DateTime, etc, was for a correct and valid reason. Updating to UTC will assist with validation errors at a later time.

* Timing loops such as:  `var elapsed = DateTime.UtcNow - StartTime;`
* Ticks usage: `DateTime.Now.Ticks` and `DateTime.UtcNow.Ticks` produces the same long.
* ToFileTime usage `DateTime.Now.ToFileTime` => `DateTime.UtcNow.ToFileTime`
* Removal of extra `.ToUniversalTime()` calls. Example: `string epochSeconds = AWSSDKUtils.ConvertToUnixEpochSecondsString(expiresOn.ToUniversalTime());` -- AWSSDKUtils.ConvertToUnixEpochSecondsString already ensures the DateTime is UTC.
* In generator changed `IsDateTime` to `IsTimeStamp`. This is because the shape is actually modeled as a `TimeStamp` not a C# language specific `DateTime`.
* Switch service specific DateTime utility methods over to runtime internal methods in StringUtils. Example: `Amazon.EC2.Internal.CustomMarshallTransformations.ConvertDateTimeISOWithoutMillisecondsUtc` => `Amazon.Runtime.Internal.Util.StringUtils.FromDateTimeToISO8601NoMs`. Deleted the service specific DateTime utility methods.
* Removed `#pragma warning disable CS0612, CS0618 // Type or member is obsolete` around `AWSSDKUtils.CorrectedUtcNow` because the property isn't marked obsolete.
* Internal Cache class internally uses UTC time. (SdkCache.cs)

**Minor bug fixes:** These are not breaking changes but fix slight issues in the SDK.

* Fixed Epoch date to UTC per definition where the epoch date was created in local time. (TokenBucket.cs)
* Stopped using expiry times as localtime. Changed to UTC. (InstanceProfileAWSCredentials.cs, ProcessAWSCredentials.cs, RefreshingAWSCredentials.cs)
* Ensured DateTime.Max and DateTime.Min are marked with a DateTime.Kind DateTimeKind.Utc for proper calculations.
* Instad of assuming SAML credentails are localtime then converting to UTC assume that the time given is UTC to workproperly with credential expiration being in UTC time for other credential providers. (SAMLImutableCredentials.cs)
* Console logger outputs timestamps as a UTC date incase output is sent off the local machine and for easier comparison with other UTC dates. We could add a preference flag here though. (Logger.Console.cs)
* RetryPolicies return UTC server time instead of a UTC time converted to local time. (RetryPolicy.cs)
* AWSPublicIpAddressRanges mixing UTC and local time. (AWSPublicIpAddressRanges.cs)
* GetFormattedTimestampISO8601 incorrectly creating a DateTime object as local time even though it is passed in as UTC. Then formatting it as a UTC string. Although incorrect the DaeTimeKind.Local flag didn't actually harm anything and the behavior is the same by not using it.

## Testing

DryRun: Passed (https://tiny.amazon.com/b78l9mvc/IsenLink)

Manual Testing: In progress -- Example (PresignedURLs)
  * S3 Presigned Urls: SigV4 and SigV2 --OK

Adding more tests where appropriate: Pending

## Types of changes

## Checklist

## License


sdk/src/Core/Amazon.Runtime/Credentials/RefreshingAWSCredentials.cs
```diff
@@ -175,7 +175,7 @@ private static void UpdateToGeneratedCredentials(CredentialsRefreshState state,
                     errorMessage = string.Format(CultureInfo.InvariantCulture,
                         "The retrieved credentials have already expired: Now = {0}, Credentials expiration = {1}",
 #pragma warning disable CS0612,CS0618 // Type or member is obsolete
-                        AWSSDKUtils.CorrectedUtcNow.ToLocalTime(), state.Expiration);
+                        AWSSDKUtils.CorrectedUtcNow, state.Expiration);
```
peterrsongg, line 178: we can remove the #pragma here too

sdk/src/Core/Amazon.Runtime/Credentials/RefreshingAWSCredentials.cs
```diff
@@ -194,7 +194,7 @@ private static void UpdateToGeneratedCredentials(CredentialsRefreshState state,
                 logger.InfoFormat(
                     "The preempt expiry time is set too high: Current time = {0}, Credentials expiry time = {1}, Preempt expiry time = {2}.",
 #pragma warning disable CS0612,CS0618 // Type or member is obsolete
-                    AWSSDKUtils.CorrectedUtcNow.ToLocalTime(),
+                    AWSSDKUtils.CorrectedUtcNow,
```
peterrsongg, line 197: we can also remove the #pragma here

sdk/src/Core/Amazon.Runtime/Internal/Util/SdkCache.cs
```diff
@@ -507,14 +507,14 @@ private set
...
+                LastUseTime = GetCorrectedTime();
             }
         }
 
-        private static DateTime GetCorrectedLocalTime()
+        private static DateTime GetCorrectedTime()
         {
 #pragma warning disable CS0612,CS0618 // Type or member is obsolete
-            return AWSSDKUtils.CorrectedUtcNow.ToLocalTime();
+            return AWSSDKUtils.CorrectedUtcNow;
```
peterrsongg, line 517: Now I'm wondering, is there even a point of having this method if we can just call AWSSDKUtils.CorrectedUtcNow()? Seems redundant

sdk/src/Core/Amazon.Runtime/Credentials/ProcessAWSCredentials.cs
```diff
@@ -176,6 +176,15 @@ private CredentialsRefreshState SetCredentialsRefreshState(ProcessExecutionResul
                         try
                         {
                             processCredentialDataV1 = JsonSerializerHelper.Deserialize<ProcessCredentialVersion1>(processInfo.StandardOutput, ProcessCredentialVersion1JsonSerializerContexts.Default);
+                            if (processCredentialDataV1.Expiration == DateTime.MaxValue && processCredentialDataV1.Expiration.Kind != DateTimeKind.Utc)
```
peterrsongg, line 179: why do we need to check if the expiration is DateTime.MaxValue? wouldn't we want to convert the expiration to universal time regardless of if it is MaxValue?

sdk/src/Services/DynamoDBv2/Custom/Conversion/SchemaV1.cs
```diff
@@ -252,11 +252,11 @@ protected override bool TryTo(DateTime value, out Primitive p)
         }
         protected override bool TryFrom(Primitive p, Type targetType, out DateTime result)
         {
-            if (DateTime.TryParseExact(p.StringValue, AWSSDKUtils.ISO8601DateFormat, CultureInfo.InvariantCulture, DateTimeStyles.AssumeUniversal, out result))
+            if (DateTime.TryParseExact(p.StringValue, AWSSDKUtils.ISO8601DateFormat, CultureInfo.InvariantCulture, DateTimeStyles.AssumeUniversal | DateTimeStyles.AdjustToUniversal, out result))
```
peterrsongg, line 255: I know this is called out as a breaking change, but if a user was previously using this method and passing in a time zone identifier like "PST", with this change the outputted DateTime would be converted to UTC which is a sneaky runtime breaking change right? Could a customer unknowingly start storing Datetimes as utc when they were storing them as local time?

Or are we basically taking the stance that time should be stored as UTC and that this previous behavior was just incorrect?

sdk/test/NetStandard/IntegrationTests/IntegrationTests/DynamoDB/DataModelTests.cs
```diff
@@ -383,7 +383,7 @@ private async Task TestHashObjects()
                 Assert.Equal(product.KeySizes.Count, retrieved.KeySizes.Count);
                 Assert.NotNull(retrieved.CompanyInfo);
                 Assert.Equal(product.CompanyInfo.Name, retrieved.CompanyInfo.Name);
-                Assert.Equal(product.CompanyInfo.Founded, retrieved.CompanyInfo.Founded);
+                Assert.Equal(product.CompanyInfo.Founded.ToUniversalTime(), retrieved.CompanyInfo.Founded);
```
peterrsongg, line 386: is this one of the tests that were failing b/c CompanyInfo.Founded was returning a local time?

sdk/src/Services/S3/Custom/_bcl/IO/S3FileInfo.cs
```diff
@@ -184,18 +184,15 @@ public DateTime LastWriteTime
...
-                    //    .LastModified.ToLocalTime();
                     var request = new GetObjectMetadataRequest
                     {
                         BucketName = bucket,
                         Key = S3Helper.EncodeKey(key),
                     };
                     ((Amazon.Runtime.Internal.IAmazonWebServiceRequest)request).AddBeforeRequestHandler(S3Helper.FileIORequestEventHandler);
                     var response = s3Client.GetObjectMetadata(request);
+
+                    //This is a value use of .ToLocalTime because there is both LastWriteTime and LastWriteTimeUtc for S3FileInfo operations.
```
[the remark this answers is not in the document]
boblodgett: Fixed all of these in the next commit.

sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextInternal.cs
```diff
@@ -475,15 +475,15 @@ private object FromDynamoDBEntry(SimplePropertyStorage propertyStorage, DynamoDB
             if (conversion.HasConverter(targetType))
             {
                 var output = conversion.ConvertFromEntry(targetType, entry);
-                if (flatConfig.RetrieveDateTimeInUtc)
+                if (!flatConfig.RetrieveDateTimeInUtc)
                 {
                     if (targetType == typeof(DateTime))
                     {
-                        return ((DateTime)output).ToUniversalTime();
+                        return ((DateTime)output).ToLocalTime();
```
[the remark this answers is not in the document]
boblodgett: No, this is because the default for `flatConfig.RetrieveDateTimeInUtc` is true instead of false. Because it is true dates will automatically be coming back as UTC. If the flag is set to false we are explicitly converting it to localtime now. So basically the logic is flipped now in v4 on this.

I also added a comment on these in the next commit.

sdk/test/UnitTests/Custom/Marshalling/BackwardsCompatibilityForDateTimePropertiesTests.cs
```diff
@@ -36,91 +36,22 @@ public class BackwardsCompatibilityForDateTimePropertiesTests
...
-            DateTime timeUtc = new DateTime(2018, 7, 18, 0, 0, 0, DateTimeKind.Utc);
-            DateTime timeLocal = new DateTime(2018, 7, 18, 0, 0, 0, DateTimeKind.Utc).ToLocalTime();
-            DateTime timeUnspecified = new DateTime(timeLocal.Ticks);
-
-            request.StartTimeUtc = timeUtc;
-            request.EndTimeUtc = timeLocal;
-            request.TimeUtc = timeUnspecified;
-
-#pragma warning disable CS0618 // Type or member is obsolete
             Assert.AreEqual(request.StartTime, timeUtc);
```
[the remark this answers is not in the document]
boblodgett: It could if we broke the code and changed the property StartTime to convert to localtime for example on `set`. So it is just verifying the value set it not touched be it UTC, Local, or Unspecified. No conversions should be done at this time.

sdk/src/Core/Amazon.Runtime/Credentials/RefreshingAWSCredentials.cs
boblodgett: Fixed all of these in the next commit.

sdk/src/Core/Amazon.Runtime/Internal/Util/SdkCache.cs
boblodgett: No point in the private method anymore. I removed it changed all the places it was called to directly call `AWSSDKUtils.CorrectedUtcNow` in the next commit.

sdk/src/Core/Amazon.Runtime/Credentials/ProcessAWSCredentials.cs
boblodgett: The default for the ProcessCredentialVersion1 class is to set the expiration to DateTime.MaxValue. The problem with this is it will be an unspecified time. We want all times to be in UTC so we check if it is that default and if it isn't UTC we set the Kind as Utc. This retains the value of MaxValue and just updates the kind. We cannot do ToUniversal for example as it will go out of bounds of the DateTime type.

sdk/src/Services/DynamoDBv2/Custom/Conversion/SchemaV1.cs
boblodgett: It is a runtime breaking change. Norm and I discussed this and decided to return UTC by default. The good news is the datetimes are always stored as UTC in the database now. It it is UTC, LocalTime, or Unspecified it will convert `ToUniversal` before storing it in the DB so round tripping is fine.

The gotcha is if the user gets the UTC value back and does a datetime.Year or datetime.ToString("yyyy-mm-dd hh:MM;ss") for example where DateTime simply fills in the values. It will not use the Kind as part of the value fill. So the user could be formatting the string thinking it is local time still and it is actually the UTC values now.

sdk/test/NetStandard/IntegrationTests/IntegrationTests/DynamoDB/DataModelTests.cs
boblodgett: Yes, this was the failing test. However if you look at this commit: https://github.com/aws/aws-sdk-net/commit/5f8866085fb055d8333432804a2ff5cb2e8730e8 which is part of this PR you will see where I switched the default to UTC. Because of this UTC is being returned and the product.CompanyInfo.Founded is in unspecified time. So this is now updated to ToUniversal only for the assert to pass.

sdk/src/Core/Amazon.Runtime/Credentials/RefreshingAWSCredentials.cs
boblodgett: Fixed all of these in the next commit.

normj:
Excluding my minor comments I already added I think the change looks good. I have gone through the credential expiration a couple times and everything looks like it is doing comparisons by UTC. I have done some interop testing with V3 and this branch saving and retrieving data from DynamoDB. I tested storing dates as both strings and epochs. I can confirm the data in the table was persisted the same no matter if I used V3 or this branch. The `DateTime` objects coming back were also the except for the expected change that V4 has the `Kind` set to `Utc` and V3 has the `Kind` set to `Local`.

We should look Roslyn/CodeAnalyzers rules that we can use to enforce the correct date behavior going forward but that can be a separate task.

generator/.DevConfigs/460d1afb-6912-4244-82e2-cfb811a84743.json
```diff
@@ -0,0 +1,9 @@
+{
+    "core": {
+      "changeLogMessages": [
+        "Updated DateTime handling."
```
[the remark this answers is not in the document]
boblodgett: Updated. Let me know if it needs more.

sdk/src/Services/S3/Custom/Transfer/Internal/DownloadDirectoryCommand.cs
```diff
@@ -196,20 +196,12 @@ private bool IsInstructionFile(string key)
...
-            // skip objects based on ModifiedSinceDateUtc
-            if (this._request.IsSetModifiedSinceDateUtc() && s3o.LastModified.GetValueOrDefault().ToUniversalTime() <= this._request.ModifiedSinceDateUtc.ToUniversalTime())
-                return false;
-            // skip objects based on UnmodifiedSinceDateUtc
-            if (this._request.IsSetUnmodifiedSinceDateUtc() && s3o.LastModified.GetValueOrDefault().ToUniversalTime() > this._request.UnmodifiedSinceDateUtc.ToUniversalTime())
-                return false;
-#pragma warning disable CS0618 //A class member was marked with the Obsolete attribute
             // skip objects based on ModifiedSinceDate
-            if (this._request.IsSetModifiedSinceDate() && s3o.LastModified <= this._request.ModifiedSinceDate)
+            if (this._request.IsSetModifiedSinceDate() && s3o.LastModified.GetValueOrDefault().ToUniversalTime() <= this._request.ModifiedSinceDate.ToUniversalTime())
```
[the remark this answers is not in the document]
boblodgett: Removed for s3o.LastModified. It must remain on _request.ModifiedSinceDate.

sdk/src/Services/S3/Custom/Transfer/Internal/DownloadDirectoryCommand.cs
```diff
@@ -196,20 +196,12 @@ private bool IsInstructionFile(string key)
...
                 return false;
             // skip objects based on UnmodifiedSinceDate
-            if (this._request.IsSetUnmodifiedSinceDate() && s3o.LastModified > this._request.UnmodifiedSinceDate)
+            if (this._request.IsSetUnmodifiedSinceDate() && s3o.LastModified.GetValueOrDefault().ToUniversalTime() > this._request.UnmodifiedSinceDate.ToUniversalTime())
```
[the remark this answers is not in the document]
boblodgett: Removed for s3o.LastModified. It must remain on _request.ModifiedSinceDate.

sdk/src/Services/S3/Custom/_bcl/IO/S3DirectoryInfo.cs
```diff
@@ -299,7 +299,8 @@ public DateTime LastWriteTime
                     }
                 }
 
-                return ret;
+                //This is a valid use of .ToLocalTime because there is both LastWriteTime and LastWriteTimeUtc for S3DirectoryInfo operations.
+                return ret.ToLocalTime();
```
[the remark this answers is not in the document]
boblodgett: Added to the summary comment.

sdk/src/Services/S3/Custom/_bcl/IO/S3FileInfo.cs
```diff
@@ -184,18 +184,15 @@ public DateTime LastWriteTime
...
-                    //    .LastModified.ToLocalTime();
                     var request = new GetObjectMetadataRequest
                     {
                         BucketName = bucket,
                         Key = S3Helper.EncodeKey(key),
                     };
                     ((Amazon.Runtime.Internal.IAmazonWebServiceRequest)request).AddBeforeRequestHandler(S3Helper.FileIORequestEventHandler);
                     var response = s3Client.GetObjectMetadata(request);
+
+                    //This is a valid use of .ToLocalTime because there is both LastWriteTime and LastWriteTimeUtc for S3FileInfo operations.
```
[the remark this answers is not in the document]
boblodgett: Added to the summary comment.

boblodgett:
Adjustments applied and a new commit has been pushed up. I will do another dry-run but nothing changed that should impact that.

sdk/src/Services/DynamoDBv2/Custom/AWSConfigs.DynamoDB.cs
```diff
@@ -170,8 +170,9 @@ public partial class DynamoDBContextConfig
         /// <summary>
         /// If true, all <see cref="DateTime"/> properties are retrieved in UTC timezone while reading data from DynamoDB. Else, the local timezone is used.
         /// </summary>
-        /// <remarks>This setting is only applicable to the high-level library. Service calls made via <see cref="AmazonDynamoDBClient"/> will always return <see cref="DateTime"/> attributes in UTC.</remarks>
-        public bool? RetrieveDateTimeInUtc { get; set; }
+        /// <remarks>This setting is only applicable to the high-level library. Service calls made via <see cref="AmazonDynamoDBClient"/> will always
```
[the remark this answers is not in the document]
boblodgett: Fixed.

boblodgett:
TestFromJsonCanHandleAllDataTypes is the only integ test that failed. Investigating.

normj (approved):

peterrsongg (approved):

PR #3575, opened 2024-12-10
Remove abstract Sign override that uses access and secret key.

normj:
An addendum to @dscpinheiro PR https://github.com/aws/aws-sdk-net/pull/3562 that removes the abstract `Sign` method that uses access and secret key parameters in the `AbstractAWSSigner` and makes the Sign method that uses `BaseIdentity` abstract. This allows us to remove any overrides in signers that don't use access keys and forces them to implement the version that uses `BaseIdentity`.

I also added checks in the signer that the instance of `BaseIdentity` was the correct type for that signer.

Running a dry run now to confirm changes are good.

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Internal/Auth/AbstractAWSSigner.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Auth/AbstractAWSSigner.cs
@@ -108,26 +108,18 @@ namespace Amazon.Runtime.Internal.Auth
                 throw new Amazon.Runtime.SignatureException("Failed to generate signature: " + e.Message, e);
             }
         }
-        public abstract void Sign(IRequest request, IClientConfig clientConfig, RequestMetrics metrics, string awsAccessKeyId, string awsSecretAccessKey);
 
-        public virtual void Sign(IRequest request, IClientConfig clientConfig, RequestMetrics metrics, BaseIdentity baseIdentity)
-        {
-            if (baseIdentity is AWSCredentials credentials)
-            {
-                var immutableCredentials = credentials.GetCredentials();
-                Sign(request, clientConfig, metrics, immutableCredentials?.AccessKey, immutableCredentials?.SecretKey);
-            }
-        }
+        public abstract void Sign(IRequest request, IClientConfig clientConfig, RequestMetrics metrics, BaseIdentity identity);
 
 #if AWS_ASYNC_API
         public virtual System.Threading.Tasks.Task SignAsync(
             IRequest request, 
             IClientConfig clientConfig,
             RequestMetrics metrics,
-            BaseIdentity baseIdentity,
+            BaseIdentity identity,
             CancellationToken token = default)
         {
-            Sign(request, clientConfig, metrics, baseIdentity);
+            Sign(request, clientConfig, metrics, identity);
 #if NETSTANDARD
             return Task.CompletedTask;
 #else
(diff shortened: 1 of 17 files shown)


sdk/src/Core/Amazon.Runtime/Internal/Auth/BearerTokenSigner.cs
```diff
@@ -55,7 +52,7 @@ private void InternalSign(IRequest request, BaseIdentity baseIdentity)
 
             if (baseIdentity is not AWSToken awsToken || string.IsNullOrEmpty(awsToken.Token))
             {
-                throw new AmazonClientException("No Token found.  Operation requires a Bearer token.");
+                throw new AmazonClientException("No Token found. Operation requires a Bearer token.");
```
dscpinheiro, line 55: This is annoying, but the bearer token signer tests assert the message with the extra space...

https://github.com/aws/aws-sdk-net/blob/f2b2126357d7095d72c3cf889bb3c6411db788a9/sdk/test/UnitTests/Custom/Runtime/BearerTokenSignerTests.cs#L255

https://github.com/aws/aws-sdk-net/blob/f2b2126357d7095d72c3cf889bb3c6411db788a9/sdk/test/UnitTests/Custom/Runtime/BearerTokenSignerTests.cs#L285

dscpinheiro (approved):

sdk/src/Core/Amazon.Runtime/Internal/Auth/BearerTokenSigner.cs
normj: I updated the test to not have the extra space. The space just bugs me.

sdk/src/Core/Amazon.Runtime/Internal/Auth/AbstractAWSSigner.cs
```diff
@@ -108,16 +108,8 @@ protected static string ComputeHash(byte[] data, string secretkey, SigningAlgori
...
-            if (baseIdentity is AWSCredentials credentials)
-            {
-                var immutableCredentials = credentials.GetCredentials();
-                Sign(request, clientConfig, metrics, immutableCredentials?.AccessKey, immutableCredentials?.SecretKey);
-            }
-        }
+        public abstract void Sign(IRequest request, IClientConfig clientConfig, RequestMetrics metrics, BaseIdentity identity);
 
 #if AWS_ASYNC_API
         public virtual System.Threading.Tasks.Task SignAsync(
```
muhammad-othman, line 115: nit: we can also rename the `baseIdentity` to `identity` similar to the sync version.

muhammad-othman (approved):

sdk/src/Core/Amazon.Runtime/Internal/Auth/BearerTokenSigner.cs
```diff
@@ -28,18 +28,15 @@ public class BearerTokenSigner : AbstractAWSSigner
...
-        public override void Sign(IRequest request, IClientConfig clientConfig, RequestMetrics metrics, BaseIdentity baseIdentity)
+        public override void Sign(IRequest request, IClientConfig clientConfig, RequestMetrics metrics, BaseIdentity identity)
         {
-            InternalSign(request, baseIdentity);
+            InternalSign(request, identity);
         }
 
 #if AWS_ASYNC_API
-        public override Task SignAsync(IRequest request, IClientConfig clientConfig, RequestMetrics metrics, BaseIdentity baseIdentity, CancellationToken token = default)
+        public override Task SignAsync(IRequest request, IClientConfig clientConfig, RequestMetrics metrics, BaseIdentity identity, CancellationToken token = default)
```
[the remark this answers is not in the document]
normj: Good call, done.

sdk/src/Core/Amazon.Runtime/Internal/Auth/AbstractAWSSigner.cs
normj: Done

normj:
Dry run successful.

PR #3578, opened 2024-12-12
Generator implementation for V2 smoke tests for C2J files.

AlexDaines:
## Description
Implements the generator for smoke tests v2 format based on the new smoke-2.json schema. Key features:

- Supports all configuration options from the smoke-2.json schema
- Handles complex input types including arrays and nested objects
- Generates both success and failure test cases
- Includes proper test categorization
- Manages special cases for Tags and Includes objects

Questions for review:
1. Special case handling for Tags/Includes - should this be more generic?
2. Configuration defaults - should we explicitly set schema defaults?
3. Error handling - any additional checks needed beyond ErrorCode?
4. Input validation - should we validate against operation parameters?
5. Test categorization - any additional AWS-specific categories needed?

## Motivation and Context
This implementation supports the new smoke tests format (smoke-2.json) which provides more configuration options and better structure for test cases. The generator creates test files that:
- Allow per-test configuration
- Support all endpoint configuration options
- Handle complex input structures
- Provide proper error handling and validation

## Testing
- Verified generation with ACM service model
- Tested various scenarios including:
  - Success and failure cases
  - Complex input structures (arrays, nested objects)
  - Special cases (Tags, Includes)
  - Edge cases (null values, empty strings)
  - All configuration options
- Generated tests compile successfully

## Types of changes

## Checklist

## License

### DIFF
--- a/generator/ServiceClientGeneratorLib/Generators/TestFiles/SmokeTestsV2.tt
+++ b/generator/ServiceClientGeneratorLib/Generators/TestFiles/SmokeTestsV2.tt
@@ -8,6 +8,7 @@
 	AddLicenseHeader();
 #>
 using System;
+using System.Collections.Generic;
 using System.Threading.Tasks;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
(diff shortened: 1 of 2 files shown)


normj (changes requested):

generator/ServiceClientGeneratorLib/Generators/TestFiles/SmokeTestsV2.tt
```diff
@@ -8,6 +8,7 @@
 	AddLicenseHeader();
 #>
 using System;
+using System.Collections.Generic;
 using System.Threading.Tasks;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
```
dscpinheiro, line 13: (I'm not reviewing the PR, but...) Is there a reason why you're using MSTest as the test framework? 

The [hand-written files](https://github.com/aws/aws-sdk-net/blob/main/sdk/test/Services/SecretsManager/SmokeTests/SmokeTest.cs) we have right now use xUnit (meaning they can run in parallel) and using xUnit will allow you to test your PR without making any changes to our build system (as long as they're found by the projects in this folder: https://github.com/aws/aws-sdk-net/tree/v4-development/sdk/test/SmokeTests).

generator/ServiceClientGeneratorLib/Generators/TestFiles/SmokeTestsV2.tt
```diff
@@ -26,36 +27,223 @@ namespace AWSSDK_DotNet.IntegrationTests.SmokeTestsV2
...
+            {
+                foreach (string key in input.PropertyNames)
+                {
+                    var value = input[key];
+                    if (value != null)
+                    {
+                        if (value.IsString)
+                        {
+#>
+            request.<#=key#> = "<#=value.ToString()#>";
```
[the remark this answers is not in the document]
AlexDaines: Now using FindPropertyName, which should handles cases where the JSON key name doesn't exactly match the .NET property name.

generator/ServiceClientGeneratorLib/Generators/TestFiles/SmokeTestsV2.tt
```diff
@@ -26,36 +27,223 @@ namespace AWSSDK_DotNet.IntegrationTests.SmokeTestsV2
...
+                var errorId = GetExpectedErrorId(testCase);
+#>
+            try
+            {
+                var response = await serviceClient.<#=modeledOperation.Name#>Async(request);
+                Assert.Fail("Expected <#=errorId ?? "an error response"#> exception, but the call succeeded.");
+            }
+            catch (Amazon<#=this.Config.ClassName#>Exception ex)
+            {
+                Assert.AreEqual("<#=errorId#>", ex.ErrorCode);
```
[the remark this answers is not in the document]
AlexDaines: Null check for this added

generator/ServiceClientGeneratorLib/Generators/TestFiles/SmokeTestsV2.tt
AlexDaines: @normj

peterrsongg:
I tried running the generator and it seems like this needs to be changed as well

```
            if (this.Configuration.ServiceModel.HasSmokeTestsV2)
            {
                ExecuteIntegrationTestGenerator(new SmokeTestsV2(), "SmokeTestsV2.cs");
            }
```

with this logic, each file will be `SmokeTestsV2.cs` and one will overwrite the other. You should probably do something similar to the other generators and do something like

```
ExecuteIntegrationTestGenerator(new SmokeTestsV2(), serviceName + "SmokeTestsV2.cs");
```

peterrsongg:
^^ looking into the problem I mentioned above a bit more, it seems like it thinks these generated integ tests are orphaned files because they are not a part of the `generatedTestFiles` that get populated here
https://github.com/aws/aws-sdk-net/blob/main/generator/ServiceClientGenerator/Program.cs#L64-L79

since it is not part of that list, when it is passed into `GeneratorDriver.RemoveOrphanedShapesAndServices` it gets removed. This is probably the first instance of a generated integ tests so you should add these generated integ tests to the list so it isn't removed

generator/ServiceClientGeneratorLib/Generators/TestFiles/SmokeTestsV2.tt
```diff
@@ -26,36 +27,283 @@ namespace AWSSDK_DotNet.IntegrationTests.SmokeTestsV2
...
+                            }
+                        }
+                    }
+                }
+            }
+
+            if (IsSuccessExpected(testCase))
+            {
+#>
+            var response = await serviceClient.<#=modeledOperation.Name#>Async(request);
```
peterrsongg, line 278: Add `.ConfigureAwait(false)` at the end of this async call. We should always be adding that at the end of async calls, unless you really know what you're doing. See https://devblogs.microsoft.com/dotnet/configureawait-faq/#why-would-i-want-to-use-configureawait(false) for why we should use `configureAwait(false)`

generator/ServiceClientGeneratorLib/Generators/TestFiles/SmokeTestsV2.tt
```diff
@@ -26,36 +27,283 @@ namespace AWSSDK_DotNet.IntegrationTests.SmokeTestsV2
...
+<#
+            }
+            else
+            {
+                var errorId = GetExpectedErrorId(testCase);
+#>
+            try
+            {
+                var response = await serviceClient.<#=modeledOperation.Name#>Async(request);
```
peterrsongg, line 287: same comment as above

generator/ServiceClientGeneratorLib/Generators/TestFiles/SmokeTestsV2.tt
peterrsongg: unless you want to make a buildsystem change I'd probably listen to daniel here

peterrsongg (changes requested):

dscpinheiro:
Closing as the smoke tests work is on hold. Will revisit this later.

Issue #3475, opened 2024-09-13
Native AOT support for DynamoDB


dscpinheiro:
### Discussed in https://github.com/aws/aws-sdk-net/discussions/3194

<div type='discussions-op-text'>

<sup>Originally posted by **mrdaghestani** February 20, 2024</sup>
Hi,
I understand that NativeAOT is still a very new feature and probably most people prefer not to use it but, I wanted to suggest that we can talk about supporting it in DynamoDB SDK.

I was thinking it could be very useful if there was a source generator for mapping DynamoDB items into database models inside our code base so we can ignore using reflection and probably that would solve the support for NativeAOT.

What do you think? Is it something that you're thinking of implementing? If yes, what's the schedule?

Thank you</div>

ashishdhingra:
Needs to be reviewed by team for prioritization.

Dreamescaper:
Related to https://github.com/aws/aws-sdk-net/issues/2542

petro2050:
@ashishdhingra @bhoradc is there any update on this? DynamoDB is the only blocker for me. :(

normj:
I made some progress putting a PR out for the DocumentModel library to be Native AOT. The PR is targeting our upcoming V4 version. https://github.com/aws/aws-sdk-net/pull/3583

normj:
Second PR https://github.com/aws/aws-sdk-net/pull/3606 merged into V4 that updates the DataModel also known as Object Persistence Model for Native AOT support. This change will go out as part of preview 6.

normj:
Closing the issue now that Native AOT support as been released as part of V4 preview 6. Follow the this GitHub issue for tracking the status of V4: https://github.com/aws/aws-sdk-net/issues/3362

Dreamescaper:
@normj 
I've tested AOT integration. Apparently, it doesn't support type properties and collection properties.
1) Nested Types:
```csharp
var nestedTypeId = Random.Shared.Next().ToString();
await context.SaveAsync(new TypeWithNestedTypeProperty
{
    Id = nestedTypeId,
    Name = "Test",
    SubType = new SubType() { SubName = "Subname" }
});
Console.WriteLine("Item saved.");

var nestedItem = await context.LoadAsync<TypeWithNestedTypeProperty>(nestedTypeId);
Console.WriteLine($"Item loaded: {nestedItem}");
Console.WriteLine($"SubType.SubName: {nestedItem.SubType.SubName}");

[...]

PR #3583, opened 2024-12-19
[V4] Address Native AOT warnings for DynamoDB's Document Model library

normj:
## Description
**Note: This is just the Document Model. Not the DataModel (aka Object Persistence Model). The DataModel builds on top of the Document Model so getting the Document Model library Native AOT complaint is the prerequisite**

This PR removes the `RequiresUnreferencedCode` from the DocumentModel library and adds the appropriate `DynamicallyAccessedMembers` attributes and code adjustments to make Native AOT trim warnings go away when using the DocumentModel library with Native AOT.

## Motivation and Context
https://github.com/aws/aws-sdk-net/issues/3475
Add support for using the Document Model library using Native AOT. Using Native AOT for the following Lambda function reduced the cold start time from approximately 1,500 ms to 500 ms.

```csharp
public class Function
{
    static IAmazonDynamoDB _ddbClient;
    static Table _table;

    private static async Task Main()
    {
        _ddbClient = new AmazonDynamoDBClient(RegionEndpoint.USWest2);

        _table = new TableBuilder(_ddbClient, "ZipCodes")
                            .AddHashKey("Code", DynamoDBEntryType.String)
                            .Build();


        Func<string, ILambdaContext, Task<string>> handler = FunctionHandler;
        await LambdaBootstrapBuilder.Create(handler, new SourceGeneratorLambdaJsonSerializer<LambdaFunctionJsonSerializerContext>())
            .Build()
            .RunAsync();
    }


    public static async Task<string> FunctionHandler(string input, ILambdaContext context)
    {
        var document = await _table.GetItemAsync(input);
        if (document == null)
        {
            return $"No zip code found for {input}";
        }

        return $"{document["City"].AsString()}, {document["State"].AsString()}";
    }
}


[JsonSerializable(typeof(string))]
public partial class LambdaFunctionJsonSerializerContext : JsonSerializerContext
{
}
``` 

## Testing
Successful dry run
Build Native AOT Lambda function using Native AOT to ensure no trim warnings were produced.
Built a standalone console application that exercised more API then the Lambda function compiled and executed locally using Native AOT for Windows.

When I do Native AOT compile on Windows I do still get the following trim warning but it doesn't come from the AWS SDK and the warning does not happen when targeting Linux.
```
> dotnet publish -r win-x64 /p:PublishAot=true
Restore complete (1.0s)
  DynamoDBDocumentAotTest succeeded with 1 warning(s) (28.0s) → bin\Release\net8.0\win-x64\publish\
    /_/src/libraries/System.Net.Quic/src/System/Net/Quic/Internal/MsQuicApi.cs(90): warning IL3000: System.Net.Quic.MsQuicApi..cctor(): 'System.Reflection.Assembly.Location.get' always returns an empty string for assemblies embedded in a single-file app. If the path to the app directory is needed, consider calling 'System.AppContext.BaseDirectory'.
```

### DIFF
--- a/sdk/src/Services/DynamoDBv2/Custom/Conversion/DynamoDBEntryConversion.cs
+++ b/sdk/src/Services/DynamoDBv2/Custom/Conversion/DynamoDBEntryConversion.cs
@@ -20,6 +20,8 @@ using System.Globalization;
 using Amazon.DynamoDBv2.DataModel;
 using Amazon.DynamoDBv2.DocumentModel;
 using Amazon.Util.Internal;
+using System.Diagnostics.CodeAnalysis;
+
 
 #if NETSTANDARD
 using Amazon.Runtime.Internal.Util;
@@ -71,9 +73,6 @@ namespace Amazon.DynamoDBv2
     /// A collection of converters capable of converting between
     /// .NET and DynamoDB objects.
     /// </summary>
-#if NET8_0_OR_GREATER
-    [System.Diagnostics.CodeAnalysis.RequiresUnreferencedCode(Amazon.DynamoDBv2.Custom.Internal.InternalConstants.RequiresUnreferencedCodeMessage)]
-#endif
     public class DynamoDBEntryConversion
     {
         #region Static members
@@ -252,7 +251,11 @@ namespace Amazon.DynamoDBv2
         /// <typeparam name="TOutput"></typeparam>
         /// <param name="entry"></param>
         /// <returns></returns>
+#if NET8_0_OR_GREATER
+        public TOutput ConvertFromEntry<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods | DynamicallyAccessedMemberTypes.PublicConstructors)] TOutput>(DynamoDBEntry entry)
+#else
         public TOutput ConvertFromEntry<TOutput>(DynamoDBEntry entry)
+#endif
         {
             TOutput output;
             if (TryConvertFromEntry<TOutput>(entry, out output))
@@ -269,7 +272,11 @@ namespace Amazon.DynamoDBv2
         /// <param name="entry"></param>
         /// <returns></returns>
         /// <exception cref="ArgumentNullException"></exception>
+#if NET8_0_OR_GREATER
+        public object ConvertFromEntry([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods | DynamicallyAccessedMemberTypes.PublicConstructors)] Type outputType, DynamoDBEntry entry)
+#else
         public object ConvertFromEntry(Type outputType, DynamoDBEntry entry)
+#endif
         {
             if (outputType == null) throw new ArgumentNullException("outputType");
             if (entry == null) throw new ArgumentNullException("entry");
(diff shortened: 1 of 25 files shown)


sdk/src/Services/DynamoDBv2/Custom/Conversion/DynamoDBEntryConversion.cs
```diff
@@ -651,19 +696,6 @@ public virtual bool TryFrom(Document d, Type targetType, out object result)
 
     internal abstract class Converter<T> : Converter
     {
-        public override IEnumerable<Type> GetTargetTypes()
```
normj, line 654: I removed this override in the abstract class to instead have the final subclasses return their target types without using reflection.

sdk/src/Services/DynamoDBv2/Custom/Conversion/SchemaV1.cs
```diff
@@ -30,227 +31,323 @@ namespace Amazon.DynamoDBv2
 {
     #region Basic converters
 
-#if NET8_0_OR_GREATER
-    [System.Diagnostics.CodeAnalysis.RequiresUnreferencedCode(Amazon.DynamoDBv2.Custom.Internal.InternalConstants.RequiresUnreferencedCodeMessage)]
-#endif
     internal class ByteConverterV1 : Converter<byte>
     {
+        public override IEnumerable<Type> GetTargetTypes()
```
normj, line 36: These are the methods that were added after removing parent override that was using unbounded reflection.

sdk/src/Services/DynamoDBv2/Custom/Conversion/SchemaV1.cs
```diff
@@ -416,48 +544,25 @@ protected override bool TryFrom(DynamoDBBool b, Type targetType, out bool result
...
         }
     }
 
-#if NET8_0_OR_GREATER
-    [System.Diagnostics.CodeAnalysis.RequiresUnreferencedCode(Amazon.DynamoDBv2.Custom.Internal.InternalConstants.RequiresUnreferencedCodeMessage)]
-#endif
     internal abstract class CollectionConverter : Converter
     {
-        private IEnumerable<Type> targetTypes;
-        private static IEnumerable<Type> GetTargetTypes(IEnumerable<Type> memberTypes)
```
normj, line 432: This method was doing lots of reflection to expand all of the "primitive" types into the varies array, list and set variations. This unbounded reflection is not possible in reflection. I replaced the logic by creating a new static list of types with all of the "primitive" permeations.

sdk/src/Services/DynamoDBv2/Custom/Conversion/SchemaV1.cs
```diff
@@ -468,15 +573,17 @@ protected bool EntriesToCollection(Type targetType, Type elementType, IEnumerabl
...
     {
         public PrimitiveCollectionConverterV1()
-            : base(Utils.PrimitiveTypes)
+            : base()
         { }
-        
+
+        public override IEnumerable<Type> GetTargetTypes()
+        {
+            return Utils.PrimitiveTypesCollectionsAndArray;
```
normj, line 584: This new `PrimitiveTypesCollectionAndArray` field used here is the replacement for the deleted `GetTargetTypes` method above.

sdk/src/Services/DynamoDBv2/Custom/DataModel/Utils.cs
```diff
@@ -109,7 +174,17 @@ public static Type GetPrimitiveElementType(Type collectionType)
 
             throw new InvalidOperationException("Unable to determine element type");
         }
-        public static Type GetElementType(Type collectionType)
+
+#if NET8_0_OR_GREATER
+        [System.Diagnostics.CodeAnalysis.UnconditionalSuppressMessage("ReflectionAnalysis", "IL2073",
```
normj, line 179: I couldn't figure out how to remove the trim warning for the `collectionType.GetElementType()` type. I'm confident though at this point in using the library that type will already be known. The type would have been referenced in the application code.

To give me extra confidence that nobody was using these utilities methods out of context I changed all the visibility of these utility methods to `internal`.

dscpinheiro (approved):

muhammad-othman (approved):

AlexDaines (approved):

PR #3584, opened 2024-12-19
Revert the creation of Smithy.Identity.Abstractions project and move its interfaces to AWSSDK.Core

muhammad-othman:
## Description
* Removed `Smithy.Identity.Abstractions` project and update its references to use `AWSSDK.Core` instead.
* Moved `Smithy.Identity.Abstractions` interfaces to `AWSSDK.Core`
* Moved `DefaultAWSCredentialsIdentityResolver`, `DefaultAWSCredentialsIdentityResolver` and `AnonymousIdentityResolver` from internal namespace to `Amazon.Runtime`.
* Included the generated changes for `AutoScaling` and `S3` only to keep the PR small.

## Motivation and Context
* DOTNET-7844

## Testing
* `DRY_RUN-6c97e06c-7f6b-480b-98b0-9cc2369be5c0`


afroz429 (approved):

sdk/src/Core/Amazon.Runtime/Auth/IAuthScheme.cs
```diff
@@ -1,4 +1,19 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+namespace Amazon.Runtime
```
normj, line 16: For these new types have the namespace match the folder structure. So in this case the namespace is `Amazon.Runtime.Auth`. I know the types in the Credentials folder use the `Amazon.Runtime` namespace instead of `Amazon.Runtime.Credentials` but all of the other subnamespaces don't follow that pattern. I can remember why we are inconsistent with `Credentials` but we should not overload the `Amazon.Runtime` namespace`.

The other question are we comfortable with these interfaces being "public" to users or should we put them in `Amazon.Runtime.Internal.Auth` for now? I'm concern we still haven't defined the `ISigner` interface with it having a big TODO. We have more flexibility if we start with `Amazon.Runtime.Internal.Auth`.

muhammad-othman: I was trying to follow the same pattern that we used for the credentials, didn't know that it is the exception. Will update it in the next revision.

The reason I made these interfaces public is that some of them are needed for public configurations like `ClientConfig.AWSTokenProvider` & `ClientConfig.IdentityResolverConfiguration` or they types of these configurations properties. `ISigner` would be an exception to this since it is used by the `IAuthScheme` which shouldn't be created by the users. 
What do you think about only moving `ISigner`&`IAuthScheme` to an internal namespace since these are the only ones that we don't expect the users to interact with?

normj: If `ISigner` and `IAuthScheme` do need to be accessed by users I would move them to the internal namespace. Especially since currently are signers are in the internal namespace. We could potentially move the later when/if we decided to separate signers which is a customer ask.

normj (approved):

PR #3589, opened 2024-12-27
Remove ImmutableCredentials property from IRequestContext

muhammad-othman:
## Description
* Removed `ImmutableCredentials` property from `IRequestContext`.
* Existing usages of `requestContext.ImmutableCredentials` are updated to use `request.Identity` instead.
* Update the `ISigner` interface to include the methods and properties needed for its usages.
* Remove the references to `AbstractAWSSigner` start using the new `ISigner` interface instead.

## Motivation and Context

`DOTNET-7845`

## Testing

`DRY_RUN-93591562-46d9-4349-a75e-07822865d20a`


sdk/test/IntegrationTests/Tests/General.cs
```diff
@@ -201,26 +201,6 @@ public void TestSerializingExceptions()
             }
         }
 
-        [TestMethod]
-        [TestCategory("General")]
-        public void TestAnonymousCredentialsGetThroughPipeline()
```
muhammad-othman, line 206: This test case isn't valid anymore since `AnonymousAWSCredentials` cannot be used for actions that require credentials.

sdk/test/UnitTests/Custom/Runtime/SignerTests.cs
```diff
@@ -24,15 +24,14 @@ public void TestSignerWithAnonymousCredentials()
...
-            pipeline.AddHandler(new CredentialsRetriever());
 
             var signer = new MockSigner();
             var context = CreateTestContext(signer);
             context.RequestContext.Identity = new AnonymousAWSCredentials();
             pipeline.InvokeSync(context);
 
             Assert.IsTrue(context.RequestContext.IsSigned);
-            Assert.AreEqual(0, signer.SignCount);
+            Assert.AreEqual(1, signer.SignCount);
```
muhammad-othman, line 34: Calling any an action with `AnonymousAWSCredentials` now goes through the signing phase and calls the `NullSigner`.

generator/ServiceClientGeneratorLib/Generators/Endpoints/EndpointResolver.partial.cs
```diff
@@ -25,7 +25,7 @@ private string GetValueSource(Parameter parameter)
                 case "AWS::S3::DisableMultiRegionAccessPoints": return "config.DisableMultiregionAccessPoints";
                 case "AWS::S3::UseGlobalEndpoint": return "config.USEast1RegionalEndpointValue == S3UsEast1RegionalEndpointValue.Legacy";
                 case "AWS::STS::UseGlobalEndpoint": return "false";
-                case "AWS::Auth::AccountId": return "requestContext.ImmutableCredentials.AccountId";
+                case "AWS::Auth::AccountId": return "(requestContext.Identity as AWSCredentials)?.GetCredentials()?.AccountId";
```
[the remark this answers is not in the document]
dscpinheiro: Agree we don't need to fix it now. Most of the credential providers will also return cached values if available (so it'll only be expensive occasionally instead of every request).

sdk/src/Services/S3/Custom/Util/BucketRegionDetector.cs
```diff
@@ -113,7 +113,7 @@ private static string CheckRegionAndUpdateCache(AmazonS3Uri requestedBucketUri,
         private static string GetHeadBucketPreSignedUrl(string bucketName, IRequestContext requestContext)
         {
             // all buckets accessible via USEast1
-            using (var s3Client = GetUsEast1ClientFromCredentials(requestContext.ImmutableCredentials))
+            using (var s3Client = GetUsEast1ClientFromCredentials(requestContext.ClientConfig.DefaultAWSCredentials))
```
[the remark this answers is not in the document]
dscpinheiro: Ack. It'd be good to include a comment, it's easy to miss why `DefaultAWSCredentials` is being used if you don't have more context.

sdk/src/Core/Amazon.Runtime/Internal/Auth/ISigner.cs
```diff
@@ -24,11 +31,25 @@ namespace Amazon.Runtime.Internal.Auth
     /// the service to authenticate the SDK customer's identity.
     /// </para>
     /// </summary>
-#pragma warning disable CA1040 // Avoid empty interfaces
     public interface ISigner
```
boblodgett, line 34: Should add /// comments to the other public methods since you added them to RequiresCredentials.

boblodgett (approved):

sdk/src/Core/Amazon.Runtime/Internal/Auth/ISigner.cs
dscpinheiro: @muhammad-othman There's a blob in the SRA book if you want to use that: _Signs the provided HTTP request. The provided identity is used for signing, and an exception will be thrown if the identity is not compatible with this signer._

I also noticed the existing comment for `RequiresCredentials` mentions immutable credentials (which we're trying to remove). May as well remove the `<see cref="ImmutableCredentials"/>` from there.

muhammad-othman: Add comments to the `ISigner` members.

dscpinheiro (approved):

PR #3590, opened 2024-12-27
[SRA] Invalidate credentials caching when environment variables change

dscpinheiro:
## Description
This PR makes two changes to the `DefaultAWSCredentialsIdentityResolver` (which is the replacement for the `FallbackCredentialsFactory` in V4):
* If there are changes to environment variables used by our credentials provider (such as `AWS_PROFILE`), do not return cached credentials
* If a custom / non-default profile is specified _and_ the profile doesn't exist, throw an error instead of going to the next provider in the credential chain (which at the moment is IMDS). With the previous behavior, customers running outside of EC2 would get a different error (`Unable to get IAM security credentials from EC2 Instance Metadata Service`) that wouldn't reference the invalid profile at all.

## Motivation and Context
Internal issue: `DOTNET-7855`

## Testing
* Dry-run: `DRY_RUN-cf127d60-49ba-48a7-b0e9-04eecaa186c3`
* Console app:
```csharp
var s3Client = new AmazonS3Client();
var response1 = await s3Client.ListBucketsAsync();
Console.WriteLine($"Found {response1.Buckets.Count} buckets for default profile");

Environment.SetEnvironmentVariable("AWS_PROFILE", "test-runner");
var response2 = await s3Client.ListBucketsAsync();
Console.WriteLine($"Found {response2.Buckets.Count} buckets for test-runner profile");

try
{
    Environment.SetEnvironmentVariable("AWS_PROFILE", "non-existent");
    await s3Client.ListBucketsAsync();
}
catch (AmazonClientException)
{
    Console.WriteLine("Failed as expected locally since non-existent profile will cause the credential search to be short-circuited.");
}
```

Output:
```
Found 30 buckets for default profile
Found 16 buckets for test-runner profile
Failed as expected locally since non-existent profile will cause the credential search to be short-circuited.
```

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
+++ b/sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
@@ -111,6 +111,13 @@ namespace Amazon.Runtime.Credentials
                     {
                         throw;
                     }
+                    // Also breaking the chain in case a custom profile name was specified and the profile does not exist.
+                    // This differs from the FallbackCredentialsFactory, which would default to the IMDS provider (and throw
+                    // an error that may not be applicable when running outside of an EC2 environment).
+                    catch (ProfileNotFoundException)
+                    {
+                        throw;
+                    }
                     catch (Exception e)
                     {
                         _cachedCredentials = null;
@@ -163,7 +170,12 @@ namespace Amazon.Runtime.Credentials
                 return profile.GetAWSCredentials(source, true);
             }
 
-            throw new AmazonClientException($"Unable to find the \"{profileName}\" profile in CredentialProfileStoreChain.");
+            if (!profileName.Equals(DEFAULT_PROFILE_NAME, StringComparison.OrdinalIgnoreCase))
+            {
+                throw new ProfileNotFoundException($"Unable to find the \"{profileName}\" profile.");
+            }
+
+            throw new AmazonClientException($"Unable to find the \"{DEFAULT_PROFILE_NAME}\" profile.");
         }
 
         internal static string GetProfileName()
--- a/sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
+++ b/sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
@@ -34,11 +34,12 @@ namespace Amazon.Runtime.Credentials
         private const string AWS_PROFILE_ENVIRONMENT_VARIABLE = "AWS_PROFILE";
         private const string DEFAULT_PROFILE_NAME = "default";
 
-        private static ReaderWriterLockSlim _cachedCredentialsLock = new ReaderWriterLockSlim();
+        private static readonly ReaderWriterLockSlim _cachedCredentialsLock = new();
         private delegate AWSCredentials CredentialsGenerator();
         private AWSCredentials _cachedCredentials;
-        private List<CredentialsGenerator> _credentialsGenerators { get; set; }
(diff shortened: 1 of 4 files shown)


sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
```diff
@@ -51,8 +52,8 @@ public DefaultAWSCredentialsIdentityResolver()
 #if BCL
                     () => new AppConfigAWSCredentials(), // Test explicit keys/profile name first.
 #endif
-                    () => AssumeRoleWithWebIdentityCredentials.FromEnvironmentVariables(),
                     () => new EnvironmentVariablesAWSCredentials(), // Look for credentials set in environment vars.
+                    () => AssumeRoleWithWebIdentityCredentials.FromEnvironmentVariables(),
```
dscpinheiro, line 56: **Important**: This change is intentional! The search order now matches the `credentials-provider-chain-search-precedence` SEP. 

The order defined in the spec is: `Environment variables: credentials` -> `Environment variables: web identity token / STS` -> `Profile: credentials`.

sdk/test/UnitTests/Custom/Runtime/Credentials/InstanceProfileAWSCredentialsTest.cs
```diff
@@ -93,7 +93,6 @@ public void StaticStabilityWhenIMDSExperiencesAnOutageScenarioTest()
                 Expiration = currentTime.AddHours(6)
             };
 
-            using (new AWSConfigsDateFaker(() => currentTime.ToUniversalTime()))
```
dscpinheiro, line 96: This line was causing the test to fail locally for me. `currentTime` is already UTC after the DateTime changes that were recently merged into `v4-development`.

afroz429 (approved):

sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
```diff
@@ -104,19 +111,30 @@ public BaseIdentity ResolveIdentity()
                     {
                         throw;
                     }
+                    // Also breaking the chain in case a custom profile name was specified and the profile does not exist.
+                    // This differs from the FallbackCredentialsFactory, which would default to the IMDS provider (and throw
+                    // an error that may not be applicable when running outside of an EC2 environment).
+                    catch (ProfileNotFoundException)
```
muhammad-othman, line 117: Seems like we are resolved the `Unable to get IAM security credentials from EC2 Instance Metadata Service.` confusing error when users set the profile name somehow, but if they didn't set anything they will get the same confusing error, right?
Based on what I've seen so far, usually users who get confused by this error either forget to set the credentials or set it in a wrong way and then they don't know what to do.

dscpinheiro: Yes, if the `default` profile doesn't exist (or if the `AWS_PROFILE` variable isn't set), we'll still go to IMDS as a last resort. 

As we discussed internally, we may need to change the IMDS provider but that's out of scope of this PR since we need to make sure we don't break the IMDS SEP.

muhammad-othman (approved):

sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
muhammad-othman: Approved since we are planning to fix the message in a different PR.

PR #3597, opened 2025-01-07
Maintain copy of stream for exceptions during unmarshalling

peterrsongg:
## Description

This PR includes 2 changes. The first change is that if during unmarshalling we get an exception, we create a copy of the existing stream because when a `StreamingUtf8JsonReader` is instantiated it automatically reads from the stream and fills the buffer. This moves the position to the end of the streaming and when we try to get the body of the response via `GetResponseBodyBytes`, the `WrappingStream`'s position is at the end and the unmarshaller thinks there is nothing to read. To get around this we copy the stream to a copy stream and set the `WrappingStream` to the stream copy. 

The second change is that if the exception sent back by the service has no body and only a header, we do not try to read the response because `System.Text.Json`'s `Utf8JsonReader` will throw an exception if it cannot find a valid json token unlike `LitJson's` `JsonReader` which doesn't throw an exception if the body is empty.

I've included some of the generated changes as an example.

## Motivation and Context

All unit tests in CoreAndCustom.NetFramework sln pass with this change

## Testing

All tests pass

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Internal/Transform/JsonUnmarshallerContext.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Transform/JsonUnmarshallerContext.cs
@@ -20,6 +20,7 @@ using System.Text;
 using ThirdParty.Json.LitJson;
 using Amazon.Runtime.Internal.Util;
 using System.Text.Json;
+using Amazon.Util;
 
 namespace Amazon.Runtime.Internal.Transform
 {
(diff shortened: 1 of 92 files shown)


sdk/src/Core/Amazon.Runtime/Internal/Transform/JsonUnmarshallerContext.cs
```diff
@@ -88,13 +89,22 @@ public JsonUnmarshallerContext(
...
-                this.WrappingStream = new CachingWrapperStream(responseStream);
+                var copyStream = new MemoryStream();
+                AWSSDKUtils.CopyStreamAndSetPositionToZero(responseStream, copyStream);
+                this.WrappingStream = new CachingWrapperStream(copyStream);
             }
             else if (maintainResponseBody)
             {
-                this.WrappingStream = new CachingWrapperStream(responseStream, AWSConfigs.LoggingConfig.LogResponsesSizeLimit);
+                var copyStream = new MemoryStream();
+                AWSSDKUtils.CopyStreamAndSetPositionToZero(responseStream, copyStream);
```
normj, line 106: The point of the `AWSConfigs.LoggingConfig.LogResponsesSizeLimit` is we don't buffer too large of a response in memory as we stream the response into the marshaller. By coping the entire response into a separate `MemoryStream` we make the `AWSConfigs.LoggingConfig.LogResponsesSizeLimit` useless. Actually worse because we have the entire response buffered in the `MemoryStream` and then the cache in the `CachingWrapperStream` up to `AWSConfigs.LoggingConfig.LogResponsesSizeLimit`.

normj (commented):
I'm going to need you to walk me through how injecting yet another stream into the nesting of streams solves the problem. Does it have something to do with the HTTP response not being seekable?

Issue #3595, opened 2024-12-31
No way to execute dry-run EC2 operations (e.g. StartInstances)


sstragier-prosperops:
### Describe the bug

There doesn't appear to be a way to execute dry-run EC2 operations (e.g. StartInstances) since `DryRun` isn't a parameter of the `StartInstancesRequest` and the `DryRun` method in `AmazonEC2Client` (which isn't exposed on the interface, #3594) doesn't work because it appears to be looking for the old non-async request methods which are no longer public.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

I should be able to run EC2 operations that support a DryRun flag (e.g. StartInstances, StopInstances)

### Current Behavior

I am not able to run EC2 operations with the DryRun flag to check access to these operations

### Reproduction Steps

```c#
var ec2Client = new AmazonEC2Client();
var response = ec2Client.DryRun(new StartInstances { InstanceIds = new() { "i-xxxxxxx" } });
// response.Error = "Unrecognized request"
```

### AWS .NET SDK and/or Package version used

AWSSDK.EC2 3.7.405.7

### Targeted .NET Platform

.NET 6

### Operating System and version

Windows 10

ashishdhingra:
@sstragier-prosperops Good afternoon. Thanks for opening the issue. The reason why `IAmazonEC2` interface does not have access to the `DryRun()` method defined in `AmazonEC2Client` is because:
- [IAmazonEC2](https://github.com/aws/aws-sdk-net/blob/c3a8d89d4faad2febb0084a04c0e4ae66981ad63/sdk/src/Services/EC2/Generated/_netstandard/IAmazonEC2.cs#L42) (taking `netstandard` target as an example) interface is auto-generated from [ec2](https://github.com/aws/aws-sdk-net/blob/c3a8d89d4faad2febb0084a04c0e4ae66981ad63/generator/ServiceModels/ec2/ec2-2016-11-15.api.json#L1) service model (notice that it's in `Generated` folder).
- [AmazonEC2Client.DryRun()](https://github.com/aws/aws-sdk-net/blob/c3a8d89d4faad2febb0084a04c0e4ae66981ad63/sdk/src/Services/EC2/Custom/AmazonEC2Client.Extensions.cs#L50) is more kind of custom extension method (notice it as defined under `Custom` folder).

[...]

PR #3602, opened 2025-01-09
fix: Fixed an issue where custom AmazonEC2Client.DryRun() method was not working for non-BCL targets. Also included this method in IAmazonEC2 interface.

ashishdhingra:
## Description
fix: Fixed an issue where custom AmazonEC2Client.DryRun() method was not working for non-BCL targets. Also included this method in IAmazonEC2 interface.

This PR addressed following:
- Includes `DryRun()` method as part of `IAmazonEC2` interface for easier mock-ability.
- For non-BCL targets it includes NonPublic methods in reflection search. Kindly note that we need to explicitly use `BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic`.

Based on discussion with @normj, we should perhaps deprecate/obsolete this method in **V4** and remove customization in [ec2.customizations.json](https://github.com/aws/aws-sdk-net/blob/c3a8d89d4faad2febb0084a04c0e4ae66981ad63/generator/ServiceModels/ec2/ec2.customizations.json#L138) to not exclude `DryRun` property from request models (this was done due to legacy reasons when EC2 code was hand-coded). Created separate backlog item for it in Input queue.

## Motivation and Context
Issue #3595

## Testing
- Tested locally using below code:
  ```C#
  IAmazonEC2 ec2Client = new AmazonEC2Client();
  var ec2DryRunRespose = ec2Client.DryRun(new StartInstancesRequest { InstanceIds = new() { "<<EC2InstanceID>>" } });
  ```
  No unit test/integration test was added since we would need to spin up new EC2 instance. We would also deprecate this method in **V4**.

- Catapult  dry-run completed successfully.

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/EC2/Custom/AmazonEC2Client.Extensions.cs
+++ b/sdk/src/Services/EC2/Custom/AmazonEC2Client.Extensions.cs
@@ -34,7 +34,7 @@ namespace Amazon.EC2
     /// pay only for capacity that you actually use. Amazon EC2 provides developers the tools to build failure resilient applications and isolate
     /// themselves from common failure scenarios. </para> <para> Visit http://aws.amazon.com/ec2/ for more information. </para>
     /// </summary>
-    public partial class AmazonEC2Client
+    public partial class AmazonEC2Client : AmazonServiceClient, IAmazonEC2
     {
         #region DryRun
 
@@ -112,7 +112,7 @@ namespace Amazon.EC2
                     _methodCache = new Dictionary<Type, DryRunInfo>();
 
                     var ec2RequestType = typeof(AmazonEC2Request);
-                    var allMembers = typeof(AmazonEC2Client).GetMembers();
+                    var allMembers = typeof(AmazonEC2Client).GetMembers(BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic);
                     foreach (var member in allMembers)
                     {
                         MethodInfo method = member as MethodInfo;
--- a/generator/.DevConfigs/2a17f370-6558-4542-a38f-2361e83b84ea.json
+++ b/generator/.DevConfigs/2a17f370-6558-4542-a38f-2361e83b84ea.json
@@ -1,0 +1,11 @@
+{
+  "services": [
+    {
+      "serviceName": "EC2",
+      "type": "patch",
+      "changeLogMessages": [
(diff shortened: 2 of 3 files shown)


sdk/src/Services/EC2/AWSSDK.EC2.NetStandard.csproj
```diff
@@ -52,18 +52,18 @@
   </Choose>
```
dscpinheiro, line 52: This file shouldn't be changed, it'll get overwritten by the generator.

ashishdhingra: Agreed. It was a whitespace change. I will exclude it and commit without running a new dry run.

ashishdhingra: Done.

sdk/src/Services/EC2/Custom/AmazonEC2Client.Extensions.cs
```diff
@@ -112,7 +112,7 @@ private static Dictionary<Type, DryRunInfo> MethodCache
                     _methodCache = new Dictionary<Type, DryRunInfo>();
 
                     var ec2RequestType = typeof(AmazonEC2Request);
-                    var allMembers = typeof(AmazonEC2Client).GetMembers();
+                    var allMembers = typeof(AmazonEC2Client).GetMembers(BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic);
```
dscpinheiro, line 115: I don't quite understand why you had to add `BindingFlags.NonPublic`, there are no private methods added in this PR.

I wonder if this could break AOT applications (probably need @normj's input here).

ashishdhingra: We needed internal members to access synchronous versions of methods which are marked as internal. I had discussed this with @normj in person. 😊

normj: @dscpinheiro @ashishdhingra I built a Native AOT console application calling dry run with and with out this change and had no trim warnings and the app ran successful. I believe since the `Type` can be known at compile the Native AOT compiler knows not to trim the methods out.

dscpinheiro (approved):

normj (approved):

PR #3606, opened 2025-01-15
[V4] Support Native AOT trimming for DynamoDB DataModel

normj:
## Description
Building off of the previous https://github.com/aws/aws-sdk-net/pull/3583 that added Native AOT for DocumentModel this adds support for DataModel. 

The changes replaces the `RequiresUnreferencedCodeAttribute` with `DynamicallyAccessedMembersAttribute` with the `DynamicallyAccessedMemberTypes` parameter set to the required level of reflection needed. Any time the type in question is the POCO used by the callers to represent type the `DynamicallyAccessedMemberTypes` is set to `All` since full reflection is required by the library for loading and saving the type.

I got tired of adding all of the `#if NET8_0_OR_GREATER` conditional compilations throughout the code when using the trimming attributes like `DynamicallyAccessedMemberTypes`. I backported the trimming attributes into the `ThirdParty.RuntimeBackports` namespace so I could use the trimming attributes in all targets. As part of this I also removed   `#if NET8_0_OR_GREATER` added for trimming attributes in Core and the DocumentModel as part of a global search.

Because of the S3Link feature that dynamically creates an S3Client I had to extend the `GlobalRuntimeDependencyRegistry` to allow registering the S3 client.

Using the following Lambda function using the DataModel library and deployed to Lambda I got the following cold start performances:

Without Native AOT: **1,697 ms**
Native AOT: **509 ms**
```
static IDynamoDBContext _context;

private static async Task Main()
{
    var ddbClient = new AmazonDynamoDBClient(RegionEndpoint.USWest2);

    _context = new DynamoDBContextBuilder()
                        .WithDynamoDBClient(() => ddbClient)
                        .Build();

    Func<string, ILambdaContext, Task<DataModelAotTest>> handler = FunctionHandler;
    await LambdaBootstrapBuilder.Create(handler, new SourceGeneratorLambdaJsonSerializer<LambdaFunctionJsonSerializerContext>())
        .Build()
        .RunAsync();
}

public static async Task<DataModelAotTest> FunctionHandler(string input, ILambdaContext context)
{
    var obj1 = new DataModelAotTest { Id = "1", Name = "Name1", FavoriteHobby = new Hobby { Name = "Hiking" } };

    await _context.SaveAsync(obj1);
    var obj2 = await _context.LoadAsync<DataModelAotTest>("1");

    return obj2;
}
```

## Motivation and Context
https://github.com/aws/aws-sdk-net/issues/3475

## Testing
* Successful dry run
* Built and publish as Native AOT that exercised the library to confirm no trimming warnings were produced.
* Deployed and executed Native AOT Lambda function that used the library.

### DIFF
--- a/generator/.DevConfigs/7FE49E8F-93BF-419E-B3B2-265D9D6F35E3.json
+++ b/generator/.DevConfigs/7FE49E8F-93BF-419E-B3B2-265D9D6F35E3.json
@@ -1,0 +1,18 @@
+{
+  "core": {
+    "changeLogMessages": [
+      "Backport the .NET trimming attributes like DynamicallyAccessedMembersAttribute into ThirdParty.RuntimeBackports namespace for Target Frameworks before .NET 8. This simplifies the SDK's codebase by removing compilation conditional checks when using the attributes."
+    ],
+    "type": "patch",
+    "updateMinimum": true
+  },
+  "services": [
+    {
+      "serviceName": "S3",
+      "type": "patch",
+      "changeLogMessages": [
+        "Add NativeAOT support for the DataModel namespace also known as the Object Persistence high level library."
+      ]
+    }
+  ]
+}
--- a/generator/.DevConfigs/7FE49E8F-93BF-419E-B3B2-265D9D6F35E3.json
+++ b/generator/.DevConfigs/7FE49E8F-93BF-419E-B3B2-265D9D6F35E3.json
@@ -8,7 +8,7 @@
   },
   "services": [
     {
-      "serviceName": "S3",
+      "serviceName": "DynamoDBv2",
       "type": "patch",
       "changeLogMessages": [
         "Add NativeAOT support for the DataModel namespace also known as the Object Persistence high level library."
--- a/sdk/src/Services/DynamoDBv2/Custom/DataModel/InternalModel.cs
+++ b/sdk/src/Services/DynamoDBv2/Custom/DataModel/InternalModel.cs
@@ -36,9 +36,6 @@ namespace Amazon.DynamoDBv2.DataModel
     /// <summary>
     /// Basic property storage information
     /// </summary>
-#if NET8_0_OR_GREATER
-    [System.Diagnostics.CodeAnalysis.RequiresUnreferencedCode(Amazon.DynamoDBv2.Custom.Internal.InternalConstants.RequiresUnreferencedCodeMessage)]
-#endif
     internal class SimplePropertyStorage
     {
         // local property name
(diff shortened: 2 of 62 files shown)


sdk/src/Services/DynamoDBv2/Custom/DataModel/InternalModel.cs
```diff
@@ -50,22 +48,25 @@ internal class SimplePropertyStorage
         public MemberInfo Member { get; protected set; }
         // Type of the property
 
-#if NET8_0_OR_GREATER
-        [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)]
```
dscpinheiro, line 54: Is changing from `PublicParameterlessConstructor` to `DataModelModeledType` here intentional?

sdk/src/Services/DynamoDBv2/Custom/DataModel/S3ClientCache.cs
```diff
@@ -34,12 +37,38 @@ internal void UseClient(ICoreAmazonS3 client, RegionEndpoint region)
...
+                        output = ServiceClientHelpers.CreateServiceFromAssembly<ICoreAmazonS3>(ServiceClientHelpers.S3_ASSEMBLY_NAME, ServiceClientHelpers.S3_SERVICE_CLASS_NAME, this.ddbClient);
+                    }
+                    catch (Exception e)
+                    {
+                        if (InternalSDKUtils.IsRunningNativeAot())
+                        {
+                            throw new MissingRuntimeDependencyException(ServiceClientHelpers.S3_ASSEMBLY_NAME, ServiceClientHelpers.S3_SERVICE_CLASS_NAME, nameof(GlobalRuntimeDependencyRegistry.RegisterS3Client));
+                        }
+
+                        var msg = string.Format(CultureInfo.CurrentCulture,
```
dscpinheiro, line 64: Is this message correct? It says `STS_ASSEMBLY_NAME` and `AssumeRoleAWSCredentials`.

generator/.DevConfigs/7FE49E8F-93BF-419E-B3B2-265D9D6F35E3.json
```diff
@@ -0,0 +1,18 @@
...
+  "core": {
+    "changeLogMessages": [
+      "Backport the .NET trimming attributes like DynamicallyAccessedMembersAttribute into ThirdParty.RuntimeBackports namespace for Target Frameworks before .NET 8. This simplifies the SDK's codebase by removing compilation conditional checks when using the attributes."
+    ],
+    "type": "patch",
+    "updateMinimum": true
+  },
+  "services": [
+    {
+      "serviceName": "S3",
```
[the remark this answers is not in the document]
normj: Fixed

sdk/src/Services/DynamoDBv2/Custom/DataModel/S3ClientCache.cs
normj: Fixed, bad copy paste job.

sdk/src/Services/DynamoDBv2/Custom/DataModel/InternalModel.cs
normj: I can't remember it was set to `PublicParameterlessConstructor` before but since the `MemberType` can represent a type from the user and this method relies on methods that also use the `DataModelModeledType` I need to make it consistent here.

dscpinheiro (approved):

Dreamescaper:
Does is work for nested types? For something like this?
```csharp
```

peterrsongg (approved):

Issue #3095, opened 2023-11-07
TransactWrite<T>.AddSaveItem throws when T does NOT contain DynamoDbProperty Attributes


JCKortlang:
### Describe the bug

When using `TransactWrite<T>.AddSaveItem` and `T` only contains a `DynamoDBHashKey` / `DynamoDBRangeKey` executing the transaction results in an exception.

### Expected Behavior

Expect that the update / write succeeds

### Current Behavior

Exception

```
  {"Timestamp":"2023-11-07T12:50:21.7307220-08:00","Level":"Error","MessageTemplate":"An unhandled exception has occurred while executing the request.","Exception":"Amazon.DynamoDBv2.AmazonDynamoDBException: 3 validation errors detected: Value null at 'transactItems.1.member.update.updateExpression' failed to satisfy constraint: Member must not be null; Value null at 'transactItems.2.member.update.updateExpression' failed to satisfy constraint: Member must not be null; Value null at 'transactItems.3.member.update.updateExpression' failed to satisfy constraint: Member must not be null
    ---> Amazon.Runtime.Internal.HttpErrorResponseException: Exception of type 'Amazon.Runtime.Internal.HttpErrorResponseException' was thrown.
      at Amazon.Runtime.HttpWebRequestMessage.GetResponseAsync(CancellationToken cancellationToken)
      at Amazon.Runtime.Internal.HttpHandler`1.InvokeAsync[T](IExecutionContext executionContext)
      at Amazon.Runtime.Internal.Unmarshaller.InvokeAsync[T](IExecutionContext executionContext)
      at Amazon.Runtime.Internal.ErrorHandler.InvokeAsync[T](IExecutionContext executionContext)
      --- End of inner exception stack trace ---
      at Amazon.Runtime.Internal.HttpErrorResponseExceptionHandler.HandleExceptionStream(IRequestContext requestContext, IWebResponseData httpErrorResponse, HttpErrorResponseException exception, Stream responseStream)
      at Amazon.Runtime.Internal.HttpErrorResponseExceptionHandler.HandleExceptionAsync(IExecutionContext executionContext, HttpErrorResponseException exception)
      at Amazon.Runtime.Internal.ExceptionHandler`1.HandleAsync(IExecutionContext executionContext, Exception exception)
      at Amazon.Runtime.Internal.ErrorHandler.ProcessExceptionAsync(IExecutionContext executionContext, Exception exception)
[...]

PR #3614, opened 2025-01-20
Fix TransactWrite<T>.AddSaveItem when T does NOT contain DynamoDbProperty Attributes

irina-herciu:
## Description
fix: Fixed an issue where TransactWrite<T>.AddSaveItem throws when T does NOT contain DynamoDbProperty Attributes 

given that T does not contain any DynamoDbProperty, in order to use TransactWriteItem API to save the entity, it will require to perform PutItem operation instead of an UpdateItem operation.

## Motivation and Context
Issue https://github.com/aws/aws-sdk-net/issues/3095

## Testing
Tested locally using below code:

 [Fact]
 public async Task GivenCreateMultiTableTransactWrite_WhenAddSaveItem_ShouldSucceed()
 {
     DynamoDBOperationConfig TransactionDynamoDbOperationConfig = new DynamoDBOperationConfig()
     {
         SkipVersionCheck = true,
         ConditionalOperator = ConditionalOperatorValues.And
     };

     var table1 = new Table1() { Id = "Id2", RelatedId = "Id1"};
     var table2 = new Table2() { Id = "Id1", RelatedId = "Id2"};

     IDynamoDBContext dbContext = new DynamoDBContext(new AmazonDynamoDBClient());

     TransactWrite<Table1> t1 = dbContext.CreateTransactWrite<Table1>(TransactionDynamoDbOperationConfig);
     t1.AddSaveItem(table1);

     TransactWrite<Table2> t2 = dbContext.CreateTransactWrite<Table2>(TransactionDynamoDbOperationConfig);
     t2.AddSaveItem(table2);

     MultiTableTransactWrite transaction = dbContext.CreateMultiTableTransactWrite(t1,t2);
     await transaction.ExecuteAsync();
}

[DynamoDBTable("Table1")]
public class Table1
{
    [DynamoDBHashKey("Id")]
    public string Id { get; set; } = string.Empty;

    [DynamoDBRangeKey("RelatedId")]
    public string RelatedId { get; set; } = string.Empty;
}

[DynamoDBTable("Table2")]
public class Table2
{
    [DynamoDBHashKey("Id")]
    public string Id { get; set; } = string.Empty;

    [DynamoDBRangeKey("RelatedId")]
    public string RelatedId { get; set; } = string.Empty;

    [DynamoDBProperty("NonKey")]
    public string NonKey { get; set; } = string.Empty;
}

## Types of changes

## Checklist

- [ x] I have read the **README** document

## License

- [ x ] I confirm that this pull request can be released under the Apache 2 license

### DIFF
--- a/sdk/src/Services/DynamoDBv2/Custom/DataModel/TransactWrite.cs
+++ b/sdk/src/Services/DynamoDBv2/Custom/DataModel/TransactWrite.cs
@@ -16,6 +16,7 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
+using System.Linq;
 #if AWS_ASYNC_API
 using System.Threading;
 using System.Threading.Tasks;
@@ -134,11 +135,8 @@ namespace Amazon.DynamoDBv2.DataModel
             Expression conditionExpression = CreateConditionExpressionForVersion(storage);
             SetNewVersion(storage);
 
-            DocumentTransaction.AddDocumentToUpdate(storage.Document, new TransactWriteItemOperationConfig
-            {
-                ConditionalExpression = conditionExpression,
-                ReturnValuesOnConditionCheckFailure = DocumentModel.ReturnValuesOnConditionCheckFailure.None
-            });
+            AddDocumentTransaction(storage, conditionExpression);
+            
             var objectItem = new DynamoDBContext.ObjectWithItemStorage
             {
                 OriginalObject = item,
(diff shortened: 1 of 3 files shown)


sdk/src/Services/DynamoDBv2/Custom/DataModel/TransactWrite.cs
```diff
@@ -437,6 +435,43 @@ private Expression CreateConditionExpressionForVersion(ItemStorage storage)
...
             return DynamoDBContext.CreateConditionExpressionForVersion(storage, conversionConfig);
         }
+        
+
+        private void AddDocumentTransaction(ItemStorage storage, Expression conditionExpression)
+        {
+            var hashKeyPropertyNames = storage.Config.HashKeyPropertyNames;
+            var rangeKeyPropertyNames = storage.Config.RangeKeyPropertyNames;
+
+            var keyList = storage.Document.Keys.ToList();
```
normj, line 445: Can you name the variable `attributeNames`? My first scan of the PR got confused thinking `storage.Document.Keys` was returning the DynamoDB key names. I should have known better but maybe seeing the variable as `attributeNames` instead of `keyList` would help the next reader not jump to that collection. Ideally we would have named `Document.Keys` something better but we can't do anything about that now.

sdk/src/Services/DynamoDBv2/Custom/DataModel/TransactWrite.cs
```diff
@@ -437,6 +435,43 @@ private Expression CreateConditionExpressionForVersion(ItemStorage storage)
...
+            {
+                keyList.Remove(keyPropertyName);
+            }
+
+            foreach (var rangeKeyPropertyName in rangeKeyPropertyNames)
+            {
+                keyList.Remove(rangeKeyPropertyName);
+            }
+
+            if (keyList.Any())
```
normj, line 457: Add comment that if there is nothing in the `attributeNames` collection that means the user is attempting to save a POCO that only has DDB keys defined and you can't not do an update call when there are no data attributes.

normj (commented):
We need a test for this change. I would suggest adding a test to `sdk\test\Services\DynamoDBv2\IntegrationTests\DataModelTests.cs`.

Also as you create PRs you need to add a change file so the release system will know to update the version and generate the change log. For this PR you would add a file `generator/.DevConfigs/<random-guid>.json` here with the following content. Feel free to tweak the message.

```
{
  "services": [
    {
      "serviceName": "DynamoDBv2",
      "type": "patch",
      "changeLogMessages": [
        "Fixed issue with TransactWrite using the DataModel not handling when only keys are being saved."
      ]
    }
  ]
}
```

normj:
Changes look good. Running the change through the internal validation system.

normj (approved):
Internal dry run successful.

philasmar (approved):

PR #3619, opened 2025-01-24
Refactor services that depend on LitJson to use STJ

muhammad-othman:
## Description
Refactored services custom code that depends on `LitJSON` to use `System.Text.Json` instead so we can eventually remove `LitJSON` as a dependency.

## Motivation and Context
`DOTNET-7783`

## Testing
`DRY_RUN-a842b009-4ba5-4587-a52e-87a3b24516b1`

## Types of changes

## Checklist

## License


sdk/src/Services/S3/Custom/Util/_bcl/S3PostUploadSignedPolicy.cs
```diff
@@ -26,9 +26,10 @@
 
 using Amazon.Runtime;
 using Amazon.Util;
-using ThirdParty.Json.LitJson;
 using System.Globalization;
 using System.Collections.Generic;
+using System.Text.Json.Nodes;
```
peterrsongg, line 31: I didn't know this namespace existed, but after looking at it... is this safe to use? `System.Text.Json.Nodes` isn't available in `NETCoreApp31` which we still support in v4

https://learn.microsoft.com/en-us/dotnet/api/system.text.json.nodes.jsonnode?view=net-9.0

sdk/src/Services/EC2/Custom/_bcl/Util/EC2Metadata.cs
```diff
@@ -237,7 +236,7 @@ public static IAMInfo IAMInstanceProfileInfo
                 IAMInfo info;
                 try
                 {
-                    info = JsonMapper.ToObject<IAMInfo>(json);
+                    info = JsonSerializer.Deserialize<IAMInfo>(json);
```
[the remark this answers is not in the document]
dscpinheiro: > I believe this project doesn't yet enable AOT compilation

I don't think this is true, all service clients have been marked as trimmable: https://github.com/aws/aws-sdk-net/blob/dce5b75f1461482d198af2ec93d1938b5f1308c5/sdk/src/Services/EC2/AWSSDK.EC2.NetStandard.csproj#L31-L33

peterrsongg (changes requested):

sdk/src/Services/EC2/Custom/_bcl/Util/EC2Metadata.cs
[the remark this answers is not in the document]
muhammad-othman: You’re absolutely correct, @dscpinheiro. 
I’ve updated the implementation to make it AOT safe. Please let me know if there are anything else I've missed.

sdk/src/Services/S3/Custom/Util/_bcl/S3PostUploadSignedPolicy.cs
muhammad-othman: You’re absolutely correct, however [S3PostUploadSignedPolicy.cs](https://github.com/aws/aws-sdk-net/pull/3619/files/dce5b75f1461482d198af2ec93d1938b5f1308c5#diff-3c2fc591e9fd5dc35ba3928ba8093d51547e3d39508e1e9971801dbe4b29707e) is available for `bcl` only where `System.Text.Json.Nodes` is available for the .net framework that we target, hence the dry run succeeded.

I did use it here as it is a bit easier to use it to make json updates in place, which is done multiple times in this class. Still it is possible to use `JsonDocument` for these changes but it will require a bit more parsing and serialization.

peterrsongg: ah I didn't see it was under the _bcl folder. No need to refactor then since it is available in net472 targets👍

peterrsongg (approved):

sdk/src/Services/EC2/Custom/_bcl/Util/EC2Metadata.cs
dscpinheiro, line 1: @peterrsongg This class (`EC2Metadata`) is already deprecated in V3. Can we not remove it entirely in V4?

peterrsongg: i removed this class and re-ran the dry run which passed

dscpinheiro (approved):

PR #3622, opened 2025-01-24
[SRA] Update auth resolver to correctly handle multiple schemes scenario

dscpinheiro:
Updates the `AuthResolverHandler` to correctly handle scenarios where multiple schemes are defined. For example, if a service / operation defines `aws.auth#sigv4a,aws.auth#sigv4` and the CRT dependency is not available, we should still be able to resolve the identity (using SigV4).

## Testing
- Dry-run: `DRY_RUN-7a4951f9-165e-4cdd-92a0-23f2183911d9`
- There's only a few operations on the S3 model that have multiple auth schemes at the moment (and they're only applicable when running on Outposts), so I manually modified the model files, ran the generator, and confirmed that:
  - `"auth": ["aws.auth#sigv4a", "aws.auth#sigv4"]` without the CRT will use SigV4
  - `"auth": ["aws.auth#sigv4a"]` fails as expected without the CRT
  - `"auth": ["aws.auth#sigv4a", "aws.auth#sigv4"]` uses SigV4A when the CRT is available

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs
+++ b/sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs
@@ -17,6 +17,7 @@ using Amazon.Runtime.Credentials.Internal;
 using Amazon.Runtime.Endpoints;
 using Amazon.Runtime.Identity;
 using Amazon.Runtime.Internal.Auth;
+using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
@@ -51,7 +52,7 @@ namespace Amazon.Runtime.Internal
             var authOptions = ResolveAuthOptions(executionContext);
             if (authOptions == null || authOptions.Count == 0)
             {
-                throw new AmazonClientException($"No valid authentication schemes defined for ${executionContext.RequestContext.RequestName}");
+                throw new AmazonClientException($"No valid authentication schemes defined for {executionContext.RequestContext.RequestName}");
             }
 
             var clientConfig = executionContext.RequestContext.ClientConfig;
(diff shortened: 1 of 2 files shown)


sdk/test/UnitTests/Custom/Runtime/BearerTokenSignerTests.cs
```diff
@@ -230,49 +230,16 @@ public async Task SignerShouldOverrideExistingHeaderAsync()
...
-        {
-            var mockConfig = new MockClientConfig
-            {
-                AWSTokenProvider = new AWSTokenProviderChain() // chain is empty
-            };
-
-            var pipeline = CreateMockPipeline();
-            var context = CreateTestContext(new BearerTokenSigner(), null, mockConfig);
-
+            var signer = new BearerTokenSigner();
```
dscpinheiro, line 233: This test validates the signer throws an error when the access token is invalid, but it was invoking the request pipeline to achieve that.

I refactored it to call the signer class directly.

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs
```diff
@@ -65,40 +66,64 @@ protected void PreInvoke(IExecutionContext executionContext)
...
+                    if (executionContext.RequestContext.Identity != null)
+                    {
+                        return;
+                    }
+                }
+                catch (Exception)
+                {
+                    // If there are multiple authentication schemes and we cannot resolve the identity for some reason (e.g. the CRT is
+                    // required for SigV4A signing), we'll attempt the next option before returning.
+                    if (authOptions.Count > 1)
```
normj, line 115: From a theoretical point of view we should be checking if there are more authOptions to evaluate not that there are just more then one. In theory what if there were 2 authOptions and the first one cleanly didn't resolve a token so a `continue` happened and then the second one threw an exception. We would exit this loop with no exception and no identity.

normj: Can we add an SDK debug logging here?

dscpinheiro: Addressed both comments in https://github.com/aws/aws-sdk-net/pull/3622/commits/4e6589d34a4ac75a084ea9a17c412562f084afdb

normj (approved):

PR #3628, opened 2025-01-30
[SRA] Remove default assignment and obsolete attribute from AWSTokenProvider property

dscpinheiro:
## Description
Another change based on our bug bash feedback: `AWSTokenProvider` was marked as deprecated but there isn't a lot of documentation on how to extend the new `IIdentityResolverConfiguration` interface (something we'll need to include in the developer guide).

For now, remove the obsolete attribute and the default assignment (so that the SDK uses the SRA resolvers by default), but still honor the token provider if it's set (which both PowerShell and the Toolkit do).

## Testing
- Dry-run: `DRY_RUN-a1367950-84d3-462d-b9ef-c78aaf6bfd88`

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSTokenIdentityResolver.cs
+++ b/sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSTokenIdentityResolver.cs
@@ -26,10 +26,8 @@ namespace Amazon.Runtime.Credentials
     {
         private readonly IAWSTokenProvider _tokenProvider;
 
-        public DefaultAWSTokenIdentityResolver() : this(null) { }
-
-        public DefaultAWSTokenIdentityResolver(string profileName)
-            => _tokenProvider = new AWSTokenProviderChain(new ProfileTokenProvider(profileName));
+        public DefaultAWSTokenIdentityResolver() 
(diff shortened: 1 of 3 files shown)


sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSTokenIdentityResolver.cs
```diff
@@ -26,10 +26,8 @@ public class DefaultAWSTokenIdentityResolver : IIdentityResolver<AWSToken>
     {
         private readonly IAWSTokenProvider _tokenProvider;
 
-        public DefaultAWSTokenIdentityResolver() : this(null) { }
-
-        public DefaultAWSTokenIdentityResolver(string profileName)
```
dscpinheiro, line 31: This constructor was redundant, the `ProfileTokenProvider` already searches for the profile name in the correct places (config file, environment variable, etc...): https://github.com/aws/aws-sdk-net/blob/63cb4e09f3f21b33a37d2f79cc79744c9a9feef8/sdk/src/Core/Amazon.Runtime/_bcl%2Bnetstandard/Tokens/ProfileTokenProvider.cs#L102-L105

normj (approved):

afroz429 (approved):

Issue #2985, opened 2023-06-28
Saving nested items with class hierarchies


atjohns:
### Describe the bug

This is a similar issue to a very old discussion posted [here ](https://github.com/aws/aws-sdk-net/issues/245) on class hierarchies in Dynamo, but this one is explicitly about saving items. I am trying to use DynamoDB to save a model that mimics the following structure
```

    public class A
    {
        public string Name { get; set; }

        public int MyPropA { get; set; } = 1;
    }

    public class B : A
    {
        public int MyPropB { get; set; } = 2;
    }

    [DynamoDBTable("Test")]
    public class ModelA
    {
        [DynamoDBHashKey]
        public Guid id { get; set; }

        public List<A> MyClasses { get; set; }
    }
```

If we instantiate a new instance of this model like so...
```

            var a1 = new A { Name = "A1" };
            var b1 = new B { Name = "B1" };

            Guid id = Guid.NewGuid();

            var model = new ModelA
            {
                id = id,
                MyClasses = new List<A> { a1, b1 }
            };
```

We end up with a list that contains both Class B & Class A types that are being saved into Dynamo incorrectly.

### Expected Behavior

Expectation would be for both Model A & Model B to save with the properties that are pushed by the code and be able to recognize the Model B inheritance 

### Current Behavior

When this model is saved to Dynamo it only saves the Class A properties. Here's a screenshot showing what the object looks like when the Save call is made vs what it looks like within Dynamo

![image](https://github.com/aws/aws-sdk-net/assets/94479740/112e0fa5-ede2-4c6a-88d0-f8e8a4ad84cb)

### Reproduction Steps

See above

### AWS .NET SDK and/or Package version used

AWSSDK.DynamoDBv2 3.7.104

### Targeted .NET Platform

.NET 7

### Operating System and version

Windows 10

ashishdhingra:
Reproducible using below code:
```C#
using System;
using System.Collections.Generic;
using Amazon.DynamoDBv2;
using Amazon.DynamoDBv2.DataModel;
using Amazon.Runtime;

using (AmazonDynamoDBClient client = new AmazonDynamoDBClient())
{
    DynamoDBContext context = new DynamoDBContext(client);
    var a1 = new A { Name = "A1" };
    var b1 = new B { Name = "B1" };

    Guid id = Guid.NewGuid();

    var model = new ModelA
    {
        id = id,
        MyClasses = new List<A> { a1, b1 }
    };

[...]

PR #3633, opened 2025-02-03
implement DynamoDBDerivedTypeAttribute for polymorphism support

irina-herciu:
implement DynamoDBDerivedTypeAttribute for polymorphism support that should solve Saving nested items with class hierarchies  https://github.com/aws/aws-sdk-net/issues/2985

## Description
New DynamoDB PropertyAttribute added as DynamoDBDerivedTypeAttribute in order to add support for polymorphic types serialization and deserialization 
[DynamoDBDerivedType("B", typeof(B))] 

Type Attribute name can be configurable from DerivedTypeAttributeName on DynamoDBContextConfig and by default uses "$type".

## Motivation and Context
https://github.com/aws/aws-sdk-net/issues/2985

## Testing
Integration tests added

## Screenshots (if appropriate)

image

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project
- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [x] I have read the **README** document
- [x] I have added tests to cover my changes
- [x] All new and existing tests passed

## License

- [x] I confirm that this pull request can be released under the Apache 2 license

[issues]: https://github.com/aws/aws-sdk-net/issues
[license]: http://aws.amazon.com/apache2.0/
[cla]: http://en.wikipedia.org/wiki/Contributor_License_Agreement

PR #3643, opened 2025-02-07
implement polymorphism support for DynamoDB entries

irina-herciu:
- implement DynamoDBPolymorphicTypeAttribute for polymorphism support for table entries and properties polymorphism support that should solve Saving nested items with class hierarchies  https://github.com/aws/aws-sdk-net/issues/2985

## Description
New DynamoDB Attribute added as DynamoDBPolymorphicTypeAttribute in order to add support for polymorphic types serialization and deserialization
`[DynamoDBPolymorphicType("A2", typeof(ModelA2))]`

Type Attribute name can be configurable from DerivedTypeAttributeName on DynamoDBContextConfig and by default uses "$type".

```cs
      //class level use
      [DynamoDBPolymorphicType("dog", typeof(Dog))]
      [DynamoDBPolymorphicType("cat", typeof(Cat))]
      public class Animal { }

       //property level
       public class Zoo
       {
        [DynamoDBPolymorphicType("dog", typeof(Dog))]
        [DynamoDBPolymorphicType("cat", typeof(Cat))]
        public Animal Resident { get; set; }
       }
```

## Motivation and Context
https://github.com/aws/aws-sdk-net/issues/2985

## Testing
Integration tests added

image

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/DynamoDBv2/Custom/DataModel/Attributes.cs
+++ b/sdk/src/Services/DynamoDBv2/Custom/DataModel/Attributes.cs
@@ -72,37 +72,90 @@ namespace Amazon.DynamoDBv2.DataModel
     }
 
     /// <summary>
-    /// DynamoDB attribute that marks a class for polymorphism support.
-    /// 
+    /// DynamoDB attribute that marks a class or property for polymorphism support.
+    /// This allows DynamoDB to store and retrieve instances of derived types 
+    /// while preserving their original types during serialization and deserialization.
     /// </summary>
-    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct, Inherited = true, AllowMultiple = true)]
+    /// <remarks>
+    /// To enable polymorphic serialization, this attribute should be applied multiple times, 
+    /// once for each possible subtype. The <see cref="TypeDiscriminator"/> serves as a unique
+    /// identifier used in the database to distinguish between different derived types.
+    ///
+    /// The name of the stored discriminator attribute in DynamoDB can be configured via
+    /// <see cref="DynamoDBContextConfig.DerivedTypeAttributeName"/>.
+    /// If not explicitly set, the SDK will use a default attribute name for the discriminator.
+    /// </remarks>
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct | AttributeTargets.Field | AttributeTargets.Property, Inherited = true, AllowMultiple = true)]
     public sealed class DynamoDBPolymorphicTypeAttribute : DynamoDBAttribute
     {
         /// <summary>
         /// Unique name discriminator of the derived type.
         /// </summary>
+        /// <remarks>
+        /// The <see cref="TypeDiscriminator"/> is stored in DynamoDB and is used to 
+        /// determine the actual type of the object when deserializing. 
+        /// It should be unique among all declared polymorphic types for a given base type.
+        ///
+        /// The attribute name under which this discriminator is stored in DynamoDB
+        /// is configurable via <see cref="DynamoDBContextConfig.DerivedTypeAttributeName"/>.
+        ///
+        /// Example:
+        /// <code>
+        /// [DynamoDBPolymorphicType("dog", typeof(Dog))]
+        /// [DynamoDBPolymorphicType("cat", typeof(Cat))]
+        /// public class Animal { }
+        /// </code>
+        ///
+        /// When retrieving an item, DynamoDB will use this discriminator value to 
+        /// deserialize into the appropriate derived type.
+        /// </remarks>
         public string TypeDiscriminator { get; }
 
         /// <summary>
-        /// Derived type of the Property type.
-        /// Type must be a subclass of the Property type.
+        /// The specific derived type that corresponds to this polymorphic entry.
         /// </summary>
+        /// <remarks>
+        /// This should be a subclass of the property type where the attribute is applied. 
+        /// When an instance of this type is stored in DynamoDB, the <see cref="TypeDiscriminator"/> 
+        /// will be saved along with it, allowing correct type resolution during deserialization.
+        /// </remarks>
         [DynamicallyAccessedMembers(InternalConstants.DataModelModeledType)]
         public Type DerivedType { get; }
 
         /// <summary>
-        /// Construct an instance of DynamoDBPolymorphicTypeAttribute
+        /// Constructs an instance of <see cref="DynamoDBPolymorphicTypeAttribute"/>.
         /// </summary>
-        /// <param name="typeDiscriminator"></param>
-        /// <param name="derivedType"></param>
+        /// <param name="typeDiscriminator">
+        /// A unique string identifier representing this derived type.
+        /// This value is stored in DynamoDB and used for deserialization.
+        /// </param>
+        /// <param name="derivedType">
+        /// The actual derived type that this attribute represents.
+        /// Must be a subclass of the property type to which it is applied.
+        /// </param>
+        /// <example>
+        /// Usage for a polymorphic property:
+        /// <code>
+        /// public class Zoo
+        /// {
+        ///     [DynamoDBPolymorphicType("dog", typeof(Dog))]
+        ///     [DynamoDBPolymorphicType("cat", typeof(Cat))]
+        ///     public Animal Resident { get; set; }
+        /// }
+        /// </code>
+        /// 
+        /// In this example, when a `Dog` instance is stored, DynamoDB will save `"dog"` as its discriminator
+        /// under the attribute name configured in <see cref="DynamoDBContextConfig.DerivedTypeAttributeName"/>.
+        /// When retrieved, the stored discriminator ensures that the value is deserialized as a `Dog` instance.
+        /// </example>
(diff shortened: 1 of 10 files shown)


dscpinheiro:
Can the other PR (that was targeting V3) - https://github.com/aws/aws-sdk-net/pull/3633 - be closed?

irina-herciu:
@dscpinheiro the issue that the PR has in scope to fix was affecting AWSSDK.DynamoDBv2 3.7.104, but as @normj instructed in https://github.com/aws/aws-sdk-net/pull/3633 the fix is raised for V4 also.

dscpinheiro:
Ok, my read from Norm's comment in the other PR was that we should only make this change for V4. 

But I can double-check with him offline.

normj:
@irina-herciu Can you add a change file like you did in your previous PR. https://github.com/aws/aws-sdk-net/pull/3614/files

normj:
@irina-herciu A concern I have is a user saves the item with .NET but another parts reads the item with a different language have a mechanism that only stores the .NET type won't be portable. The AWS Kotlin SDK has similar support for polymorphism in their high level. It supports a mechanism customers can use to control the identifier for the type and they could potentially use a language neutral identifier. Can you rework this to have some sort of system that allows users to register a type conversation mapping. By default it can rely on .NET type but if a user can opt-in their own mechanism that would unblock polyglot customers.

Validation should be a customer can save a item using polymorphism in our Kotlin SDK and have a way to read it back out using your changes. My understanding in the Kotlin SDK you use the `HeterogenerousItemConverter` https://sdk.amazonaws.com/kotlin/api/latest/dynamodb-mapper/aws.sdk.kotlin.hll.dynamodbmapper.items/-heterogeneous-item-converter/index.html

irina-herciu:
@normj The functionality implemented of this feature is different from the 'HeterogenerousItemConverter', as the Kotlin SDK feature is designed to convert top-level items in a DynamoDB table, meaning it applies polymorphism only when retrieving or saving items at the root level and requires a custom 'AttributeConverter' to handle polymorphic nested items.

irina-herciu:
> @irina-herciu Can you add a change file like you did in your previous PR. https://github.com/aws/aws-sdk-net/pull/3614/files

Change file added

sdk/src/Services/DynamoDBv2/Custom/DataModel/Attributes.cs
```diff
@@ -71,6 +71,38 @@ public DynamoDBTableAttribute(string tableName, bool lowerCamelCaseProperties)
         }
     }
 
+    /// <summary>
+    /// DynamoDB attribute that marks a class for polymorphism support.
```
normj, line 75: Can you add more depth to the docs. Explaining how you add multiple declaration of the attribute for each possible subtype and how the TypeDiscriminator  is used.

sdk/src/Services/DynamoDBv2/Custom/DataModel/Attributes.cs
```diff
@@ -71,6 +71,38 @@ public DynamoDBTableAttribute(string tableName, bool lowerCamelCaseProperties)
...
+    public sealed class DynamoDBPolymorphicTypeAttribute : DynamoDBAttribute
+    {
+        /// <summary>
+        /// Unique name discriminator of the derived type.
+        /// </summary>
+        public string TypeDiscriminator { get; }
+
+        /// <summary>
+        /// Derived type of the Property type.
+        /// Type must be a subclass of the Property type.
```
normj, line 88: What happens if the property type is an interface? Instead of a subclass can it be a type that implements the interface?

sdk/src/Services/DynamoDBv2/Custom/DataModel/Attributes.cs
```diff
@@ -243,6 +275,36 @@ public DynamoDBPropertyAttribute(string attributeName, bool storeAsEpoch)
         public bool StoreAsEpoch { get; set; }
     }
 
+    /// <summary>
+    /// DynamoDB Polymorphic Property Attribute that marks up current member for polymorphism support.
```
normj, line 279: I'm confused on the purpose of separating the attribute that goes on the type vs the property. Seems like they solve same purpose. You are giving them the option of defining the polymorphism on either the type or the property. What happens if the declaration is on both the type and the property that uses the type. Should that be allowed?

sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs
```diff
@@ -140,6 +140,11 @@ public DynamoDBContextConfig()
         /// Service calls made via <see cref="AmazonDynamoDBClient"/> will always return 
         /// <see cref="DateTime"/> attributes in UTC.</remarks>
         public bool? RetrieveDateTimeInUtc { get; set; }
+
+        /// <summary>
+        /// Property indicating the name of the attribute used for derived types conversion.
```
normj, line 145: Expand on this documentation. I assume this is the attribute that stores the metadata used by this library to figure out the type to instantiate.

This attribute is missing listing the default attribute.

sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs
```diff
@@ -550,6 +559,12 @@ public DynamoDBFlatConfig(DynamoDBOperationConfig operationConfig, DynamoDBConte
         // State of the operation using this config
         internal OperationState State { get; private set; }
 
+        /// <summary>
+        /// Property indicating the name of the attribute used for derived types conversion.
+        /// Default value is "$type" if not set in the config.
+        /// </summary>
+        public string DerivedTypeAttributeName { get; set; }
```
normj, line 566: Expand on this documentation. I assume this is the attribute that stores the metadata used by this library to figure out the type to instantiate.

sdk/src/Services/DynamoDBv2/Custom/DataModel/InternalModel.cs
```diff
@@ -57,18 +57,42 @@ internal class SimplePropertyStorage
...
+        // derived type information used for polymorphic serialization
+        public Dictionary<Type, string> DerivedTypesDictionary { get; private set; }
+
+        // derived type information used for polymorphic deserialization
+        public Dictionary<string, Type> DerivedTypeKeysDictionary { get; private set; }
+
         [UnconditionalSuppressMessage("ReflectionAnalysis", "IL2072",
             Justification = "The user's type has been annotated with DynamicallyAccessedMemberTypes.All with the public API into the library. At this point the type will not be trimmed.")]
-        internal SimplePropertyStorage(MemberInfo member)
+        public SimplePropertyStorage(MemberInfo member)
```
normj, line 74: Why is this being changed to `public`?

sdk/src/Services/DynamoDBv2/Custom/DataModel/InternalModel.cs
```diff
@@ -341,10 +373,17 @@ internal StorageConfig([DynamicallyAccessedMembers(InternalConstants.DataModelMo
     /// <summary>
     /// Storage information for a specific class that is associated with a table
     /// </summary>
-    internal class ItemStorageConfig : StorageConfig
+    internal class ItemStorageConfig
```
normj, line 376: Can you explain why the removal of the base class? When you propose this implementation changes please give background in the PR description to help speed up PR reviews.

normj (changes requested):

sdk/src/Services/DynamoDBv2/Custom/DataModel/InternalModel.cs
irina-herciu: `StorageConfig` was removed as the base class of `ItemStorageConfig` and composition used for Storage configuration in order to support multiple Storage configurations for supported derived types.

sdk/src/Services/DynamoDBv2/Custom/DataModel/Attributes.cs
irina-herciu: added extra code comments for docs

sdk/src/Services/DynamoDBv2/Custom/DataModel/Attributes.cs
irina-herciu: both interfaces and and subclasses are supported. Integration tests added cover this scenario.

sdk/src/Services/DynamoDBv2/Custom/DataModel/Attributes.cs
irina-herciu: I reworked the implementation in order to have one single attribute for this scope. If the attribute is used on both the type and property, the property one allows more granular control, overriding or supplementing the class-level behavior.

sdk/src/Services/DynamoDBv2/Custom/DataModel/InternalModel.cs
irina-herciu: reverted to internal

normj:
Running the branch through the internal build system for validation.

normj (approved):

normj:
Internal Dry run DRY_RUN-2b085cb8-c005-4606-9a0f-aafb76c2ad6e was successful

philasmar (approved):

PR #3652, opened 2025-02-13
Fixes XML Map unmarhsalling to handle any key-value pair and xmlnamespaces

peterrsongg:
## Description

It is possible for a service to specify a different key and value name in the model via "locationName". This PR removes the hardcoded reference to "key" and "value" and instead we pass in the modeled key & value name from the model. This also fixes an issue with unmarshalling a a key / value pair that has a namespace attached to it. 

## Motivation and Context

Fixes 8 protocol tests.
## Testing

DRY RUN 5b12205f-6ab4-4c59-81a0-f045ed954fca
## Types of changes

## Checklist

## License


sdk/src/Core/Amazon.Runtime/Internal/Transform/SimpleTypeUnmarshaller.cs
```diff
@@ -844,32 +844,40 @@ public class XmlKeyValueUnmarshaller<K, V, KUnmarshaller, VUnmarshaller> :
...
             int targetDepth = originalDepth + 1;
 
             while (context.Read())
             {
-                if (context.TestExpression("key", targetDepth))
+                if (context.TestExpression(keyName, targetDepth))
                 {
                     key = this.keyUnmarshaller.Unmarshall(context);
                 }
                 else if (context.TestExpression("name", targetDepth))
```
boblodgett, line 876: What is this one for? Is there still a case where it is explicitly called "name" even though this is the key/value unmarshaller?

peterrsongg: I didn't touch that because it was there before and I wasn't sure why. Maybe there is a case where the default is name instead? would you like me try to remove it and see if the dry run blows up?

peterrsongg: new dry run with this removed pass
DRY_RUN-6984cf04-cd0e-453e-86cf-4b7df81c2045

AlexDaines (commented):
LGTM  ✅

- [x] All tests fail as expected without proposed changes
- [x] All tests pass after inclusion of branch features
- [x] No additional tracked files created by local generation
- [x] Prod dry run succeeded @ Feb 18, 2025, 06:49:21.063

sdk/src/Core/Amazon.Runtime/Internal/Transform/SimpleTypeUnmarshaller.cs
boblodgett: I mean maybe it would be that way but it would still be out of spec if there is some strange unmodeled `name` case and I doubt other SDK teams would know to do such a thing especially if there isn't a protocol test for it. Can you double check with the Smithy team? Otherwise this looks good to me.

AlexDaines (approved):

PR #3664, opened 2025-02-22
Remove deprecated IMDSv1 support

normj:
## Description
Remove deprecated IMDS v1 support from the SDK. IMDS v2 has been deployed in all areas so there is no reason to use IMDS v1 and this makes the .NET SDK consistent with other AWS SDKs.

The approach to enforce no V1 behavior is done is the `FetchApiToken` always logs and throws an exception if it was not able to retrieve the token. The `GetItems` call which calls `FetchApiToken` will immediately return null if the `FetchApiToken` throws an exception. Returning null from `GetItems` was needed to maintain behavior in V3 where if you try to access data from the `EC2InstanceMetadata` and you weren't running on EC2 instance you would get a null.

## Testing
Refactored the EC2 instance metadata tests and got them passing.
Manually deployed an application to EC2 and confirm credentials were retrieved
Dry run: DRY_RUN-292cd345-1ae4-4fff-993a-068cd7ed8550 (successful)

### DIFF
--- a/sdk/src/Core/Amazon.Util/EC2InstanceMetadata.cs
+++ b/sdk/src/Core/Amazon.Util/EC2InstanceMetadata.cs
@@ -63,8 +63,6 @@ namespace Amazon.Util
 
         private static Dictionary<string, string> _cache = new Dictionary<string, string>();
 
-        private static bool useNullToken = false;
-
         private static ReaderWriterLockSlim metadataLock = new ReaderWriterLockSlim(); // Lock to control getting metadata across multiple threads.
         private static readonly TimeSpan metadataLockTimeout = TimeSpan.FromMilliseconds(5000);
         private static readonly string _userAgent = InternalSDKUtils.BuildUserAgentString(string.Empty, string.Empty);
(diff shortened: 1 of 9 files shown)


sdk/src/Core/Amazon.Util/EC2InstanceMetadata.cs
```diff
@@ -735,16 +686,20 @@ private static List<string> GetItems(string relativeOrAbsolutePath, int tries, b
...
             {
-                token = FetchApiToken(DEFAULT_RETRIES);
+                try
+                {
+                    token = FetchApiToken(DEFAULT_RETRIES);
+                }
+                catch(InvalidOperationException e)
+                {
+                    Logger.GetLogger(typeof(EC2InstanceMetadata)).InfoFormat("Failed to retrieve IMDS data \"{0}\" because IMDS API token could not be retrieved: {1}", relativeOrAbsolutePath, e.Message);
+                    return null; // If we could not get a token then assume we are not running in an EC2 instance and return null.
```
normj, line 696: Return null here is to maintain consistency with V3 that if the SDK failed to access IMDS for one of the properties of `EC2InstanceMetadata` null is returned to the user.

dscpinheiro (approved):

PR #3667, opened 2025-02-23
Remove special EC2 DryRun method

normj:
## Description
Remove the custom `DryRun` method from the EC2 service client. In place of that the customization that removed adding the`DryRun` property on EC2 requests was removed. To do a dry run in V4 users will use the individual EC2 service APIs and set the `DryRun` property to `true`. This makes the EC2 service client match the service model and is consistent with other SDKs and the EC2 docs.

The custom `DryRun` method is a legacy design choice when the EC2 service client was hand coded. We didn't want to duplicate the same property onto every operation as well as continue maintaining the `DryRun` feature whenever new operations were added. So the idea back in the day was to create a single `DryRun` method to reduce the development effort. The EC2 client is generated these days so those maintenance issues are not a concern anymore.

To keep the PR small I only included a few of the generated files so reviewers can see the generator effect.

## Testing
Dry Run: Pending


normj:
Whoops, didn't realize this was being worked on by another member.

PR #3675, opened 2025-02-26
Manual Sweep of S3 for protocol test work

peterrsongg:
## Description

Since s3 is handcoded, the protocol test changes we made didn't propagate to S3. S3 was mostly in line with the protocol tests. There were some very minor changes such as adding missing xml namespaces and prefixes. 

I also removed the `IXmlUnmarshaller<JsonUnmarshallerContext>` implementation for the structure unmarshallers since the unmarshallers don't need to implement both interfaces anymore. 

## Motivation and Context

Updates S3 to be inline with other services.
## Testing

DRY_RUN-0cbfaf75-64e3-4948-9d69-ec49e21f9d54
## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/Model/Internal/MarshallTransformations/PutBucketACLRequestMarshaller.cs
+++ b/sdk/src/Services/S3/Custom/Model/Internal/MarshallTransformations/PutBucketACLRequestMarshaller.cs
@@ -125,7 +125,7 @@ namespace Amazon.S3.Model.Internal.MarshallTransformations
                             {
                                 if (publicRequestAccessControlPolicyGrantsValue.Grantee != null)
                                 {
-                                    xmlWriter.WriteStartElement( "Grantee");
+                                    xmlWriter.WriteStartElement( "xsi","Grantee", "http://www.w3.org/2001/XMLSchema-instance");
                                     if (publicRequestAccessControlPolicyGrantsValue.Grantee.IsSetType())
                                         xmlWriter.WriteAttributeString("xsi", "type", "http://www.w3.org/2001/XMLSchema-instance", S3Transforms.ToXmlStringValue(publicRequestAccessControlPolicyGrantsValue.Grantee.Type));
                                     if (publicRequestAccessControlPolicyGrantsValue.Grantee.IsSetDisplayName())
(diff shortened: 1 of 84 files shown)


sdk/src/Services/S3/Custom/Model/Internal/MarshallTransformations/PutBucketACLRequestMarshaller.cs
```diff
@@ -125,7 +125,7 @@ public IRequest Marshall(PutBucketAclRequest publicRequest)
                             {
                                 if (publicRequestAccessControlPolicyGrantsValue.Grantee != null)
                                 {
-                                    xmlWriter.WriteStartElement( "Grantee");
+                                    xmlWriter.WriteStartElement( "xsi","Grantee", "http://www.w3.org/2001/XMLSchema-instance");
                                     if (publicRequestAccessControlPolicyGrantsValue.Grantee.IsSetType())
                                         xmlWriter.WriteAttributeString("xsi", "type", "http://www.w3.org/2001/XMLSchema-instance", S3Transforms.ToXmlStringValue(publicRequestAccessControlPolicyGrantsValue.Grantee.Type));
```
dscpinheiro, line 130: Why does this line have `xsi` as well but the other ones below don't? Is it because it's an attribute instead of an element?

sdk/src/Services/S3/Custom/Model/Internal/MarshallTransformations/PutBucketLoggingRequestMarshaller.cs
```diff
@@ -84,7 +84,7 @@ public IRequest Marshall(PutBucketLoggingRequest putBucketLoggingRequest)
                                         var granteeGrantee = loggingEnabledLoggingEnabledtargetGrantsListValue.Grantee;
                                         if (granteeGrantee != null)
                                         {
-                                            xmlWriter.WriteStartElement("Grantee");
+                                            xmlWriter.WriteStartElement("xsi","Grantee", "http://www.w3.org/2001/XMLSchema-instance");
```
dscpinheiro, line 87: We have a lot of integration tests for S3, but I couldn't find one for this specific operation. We need to verify it doesn't break with this change.

sdk/src/Services/S3/Custom/Model/Internal/MarshallTransformations/PutBucketACLRequestMarshaller.cs
peterrsongg: Type specifically has "xsi" because it is modeled that way with the xsi prefix
```
        "Type":{
          "shape":"Type",
          "documentation":"<p>Type of grantee</p>",
          "locationName":"xsi:type",
          "xmlAttribute":true
        },
```
the others don't have that modeled

sdk/src/Services/S3/Custom/Model/Internal/MarshallTransformations/PutBucketLoggingRequestMarshaller.cs
peterrsongg: Okay i will add integ tests for those that don't have coverage

boblodgett (approved):

PR #3676, opened 2025-02-27
Rework how the SDK internal logging is output

normj:
## Description
In V3 the logging is output to either the console or if Log4net if found via reflection it attaches itself to that. The reflection approach was done to avoid the SDK taking a dependency on log4net. This was also done back in the V2 days when the SDK was a single .NET assembly and we hadn't ventured into creating multiple packages yet.

Log4net is likely not the most popular logging framework these days. We know Microsoft.Extensions.Logging is an ask but unlike log4net it doesn't rely on a singleton programming model and we need an actual instance if `ILogFactory` to do logging. But again we don't want AWSSDK.Core taking dependency on all possible logging frameworks.

This redesign removes the reflection for log4net and instead moves it out into a separate package `AWSSDK.Extensions.Logging.Log4NetAdaptor` along with another package `AWSSDK.Extensions.Logging.ILoggerAdaptor` for Microsoft.Extensions.Logger support. Console and System.Diagnostics are still built into AWSSDK.Core since they don't require any dependencies.

If a user wants integrate the SDK logging with `Microsoft.Extensions.Logger` they would add the `AWSSDK.Extensions.Logging.ILoggerAdaptor` package and add the following code to their startup.
```
var builder = WebApplication.CreateBuilder(args);

var app = builder.Build();

app.Services.GetRequiredService<ILoggerFactory>()
    .ConfigureAWSSDKLogging();
```

For log4net since it uses static singleton the extensions pattern can't be done like `Microsoft.Extensions.Logger`. So a user would add the `AWSSDK.Extensions.Logging.ILoggerAdaptor` package and use code similar to the following:
```
[assembly: log4net.Config.XmlConfigurator(ConfigFile = "log4net.config")]

Log4NetAWSExtensions.ConfigureAWSSDKLogging();
var logger = LogManager.GetLogger(typeof(Program));
```

How the SDK internally logs has not changed. I did do some adjustments to SDK log levels due to what I considered too chatting logging messages. Basically downgraded log messages saying there was not value for something to debug level.

The internals of the SDK's internal logger have changed by removing all the reflection for log4net. The internal logger now keeps a collection of registered `IAdaptorLoggerFactory` that are implemented per supported logging framework. The factory is in charge of creating `IAdaptorLogger` that are wrappers around the loggers for the logging framework.

This PR does add 2 new extension NuGet packages and the build system will need to be updated to package them up.

## Motivation and Context
Remove legacy reflection that causes Native AOT issues and might cause unexpected logging by users that just want to turn log4net in their own applications.

## Testing
Dry run: DRY_RUN-8947d6cb-2e55-4539-a792-e52de2753086 successful


sdk/src/Core/Amazon.Runtime/Internal/Util/ILogger.cs
```diff
@@ -23,6 +23,5 @@ public interface ILogger
         void Debug(Exception exception, string messageFormat, params object[] args);
         void DebugFormat(string messageFormat, params object[] args);
         void Error(Exception exception, string messageFormat, params object[] args);
-        void Flush();
```
normj, line 26: Flush is not supported in Microsoft.Extensions.Logger and we only used in 2 places in the entire SDK that I don't think needed to be used there.

extensions/src/AWSSDK.Extensions.Logging.Log4NetAdaptor/AWSSDK.Extensions.Logging.Log4NetAdaptor.4.0.0-preview.8.nupkg
dscpinheiro, line 1: This shouldn't be here, right? Was this supposed to be the `.nuspec` file?

extensions/src/AWSSDK.Extensions.Logging.Log4NetAdaptor/AWSSDK.Extensions.Logging.Log4NetAdaptor.csproj
```diff
@@ -0,0 +1,17 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+	<GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>	  	  
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="log4net" Version="3.0.1" />
```
dscpinheiro, line 10: `log4net` needs to be added to our NOTICE files (`Notice.txt` and `buildtools/NoticeForZips.txt`) as well.

extensions/AWSSDK.Extensions.sln
```diff
@@ -54,10 +54,26 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "AWSSDK.Extensions.CloudFron
...
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "AWSSDK.Extensions.Logging.ILoggerAdaptor", "src\AWSSDK.Extensions.Logging.ILoggerAdaptor\AWSSDK.Extensions.Logging.ILoggerAdaptor.csproj", "{0A257E97-E0E9-43DE-8FEE-67382307725A}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AWSSDK.Extensions.Logging.Log4NetAdaptor", "src\AWSSDK.Extensions.Logging.Log4NetAdaptor\AWSSDK.Extensions.Logging.Log4NetAdaptor.csproj", "{346C35DE-D261-4090-A3A9-4D047B60210A}"
+EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "AWSSDK.Extensions.Bedrock.MEAI.NetFramework", "src\AWSSDK.Extensions.Bedrock.MEAI\AWSSDK.Extensions.Bedrock.MEAI.NetFramework.csproj", "{4A94F623-0C71-47BD-B927-CB6FA28D33A1}"
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "AWSSDK.Extensions.Bedrock.MEAI.NetStandard", "src\AWSSDK.Extensions.Bedrock.MEAI\AWSSDK.Extensions.Bedrock.MEAI.NetStandard.csproj", "{B174860A-0D1B-4B7D-9E46-7DBFC9AA5AAB}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AWSSDK.UnitTestUtilities.NetStandard", "..\sdk\test\UnitTests\Custom\AWSSDK.UnitTestUtilities.NetStandard.csproj", "{57891233-C9B6-45C2-AEAD-56F81143C042}"
```
dscpinheiro, line 65: Are these changes (adding the generator and test utilities projects) intentional? I don't see any tests for the new projects in the PR.

normj: It was intentional but not related to the logging it was to fix the solution building in Visual Studio. For example the `CloudFront.Signers.Tests.NetStandard` project already in the solution had a project reference  to `AWSSDK.UnitTestUtilities.NetStandard` but which has a dependency on `ServiceClientGeneratorLib` and `AWSSDK.CommonTest.NetStandard`. The solution will compile from the command line but in Visual Studio you get errors complaining about the projects not being part of the solution.

extensions/src/AWSSDK.Extensions.Logging.Log4NetAdaptor/AWSSDK.Extensions.Logging.Log4NetAdaptor.csproj
normj: Updated and also added `Microsoft.Extensions.Logging.Abstractions` in the file.

sdk/src/Core/Amazon.Runtime/Logging/ConsoleAdaptorLogger.cs
```diff
@@ -0,0 +1,72 @@
...
+using System.Threading;
+
+namespace Amazon.Runtime.Logging
+{
+    /// <summary>
+    /// Adaptor for outputing the SDK logs to the console.
+    /// </summary>
+    internal class ConsoleAdaptorLogger : IAdaptorLogger
+    {
+        public static long _sequanceId;
```
peterrsongg, line 28: nit: `_sequanceId` => `sequenceId`

sdk/src/Core/Amazon.Runtime/Logging/IAdaptorLogger.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// <summary>
+        /// The name of the factory used to store the factory in the registry.
+        /// </summary>
+        string Name { get; }
+
+        /// <summary>
+        /// Creates a logger.
+        /// </summary>
+        /// <param name="type"></param>
+        /// <returns></returns>
```
peterrsongg, line 55: nit: `<returns></returns>` is empty

peterrsongg (approved):
tested in a web app, approved with some minor non-blocking comments

sdk/src/Core/Amazon.Runtime/Logging/ConsoleAdaptorLogger.cs
normj: Fixed

sdk/src/Core/Amazon.Runtime/Logging/IAdaptorLogger.cs
normj: Done

sdk/src/Core/Amazon.Runtime/Logging/IAdaptorLogger.cs
```diff
@@ -0,0 +1,58 @@
...
+    /// The adaptor logger that wraps the logger from the targeted logging framework.
+    /// </summary>
+    public interface IAdaptorLogger
+    {
+        /// <summary>
+        /// True if the log level is enabled.
+        /// </summary>
+        /// <param name="level"></param>
+        /// <returns></returns>
+        bool IsEnabled(SdkLogLevel level);
```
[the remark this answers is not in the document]
normj: Done

extensions/src/AWSSDK.Extensions.Logging.ILoggerAdaptor/AWSSDK.Extensions.Logging.ILoggerAdaptor.nuspec
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package> 
+  <metadata> 
+    <id>AWSSDK.Extensions.Logging.ILoggerAdaptor</id>
+    <title>AWSSDK - Adaptor for Microsoft.Extensions.Logger</title>
```
dscpinheiro, line 5: I thought the Microsoft package name was `Microsoft.Extensions.Logging`. Is this name correct? (same comment applies to the description on line 8)

normj: Good catch. Fixed in latest commit.

dscpinheiro (approved):

PR #3680, opened 2025-02-28
Remove obsolete endpoint logic code

normj:
## Description
As part of doing a sweep for obsolete code that should be removed for V4 this PR removes all of the obsolete code from Core except for the obsolete on the `ReadWriteTimeout`. I want to do that as a separate PR.

The main chuck of the code that was removed was old endpoint code when we relied on endpoints.json. The main reason the old endpoint logic was being used where an endpoint needed to computed with an alternate endpoint. Generally a presigned url use case copying a resource from another region. I update the endpoints 2.0 code to allow overriding the region with an alternate endpoint. This required a small generator change. In the PR I only included a few of the generated results that I needed to address tests and give an example of what the change will look like.

## Testing
Dry run: DRY_RUN-eee71360-5560-4668-90fb-ddd8e052c3f0 Successful


sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4Signer.cs
```diff
@@ -663,12 +662,6 @@ public static string DetermineSigningRegion(IClientConfig clientConfig,
             // client config properties.
             if (alternateEndpoint != null)
             {
-#pragma warning disable CS0612,CS0618
-                var serviceEndpoint = alternateEndpoint.GetEndpointForService(serviceName, clientConfig.ToGetEndpointForServiceOptions());
-#pragma warning restore CS0612,CS0618
-                if (serviceEndpoint.AuthRegion != null)
```
normj, line 669: It was safe to remove the `serviceEndpoint.AuthRegion != null` check because the endpoint returned from `GetEndpointForService` sets the auth region to null.

sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4Signer.cs
```diff
@@ -694,16 +687,6 @@ public static string DetermineSigningRegion(IClientConfig clientConfig,
             var endpoint = clientConfig.RegionEndpoint;
             if (endpoint != null)
             {
-#pragma warning disable CS0612,CS0618
-                var serviceEndpoint = endpoint.GetEndpointForService(serviceName, clientConfig.ToGetEndpointForServiceOptions());
-                if (!string.IsNullOrEmpty(serviceEndpoint.AuthRegion))
```
normj, line 699: It was safe to remove the `serviceEndpoint.AuthRegion != null` check because the endpoint returned from `GetEndpointForService` sets the auth region to null.

sdk/src/Services/DocDB/Custom/Internal/PreSignedUrlRequestHandler.cs
```diff
@@ -98,7 +97,7 @@ protected virtual void PreInvoke(IExecutionContext executionContext)
                 var iRequest = marshaller.Marshall(preSignedUrlRequest as AmazonWebServiceRequest);
                 iRequest.UseQueryString = true;
                 iRequest.HttpMethod = HTTPGet;
-                iRequest.Endpoint = new UriBuilder(UriSchemeHTTPS, endpoint.GetEndpointForService(config).Hostname).Uri;
+                iRequest.Endpoint = new Uri(config.DetermineServiceOperationEndpoint(new Runtime.Endpoints.ServiceOperationEndpointParameters(executionContext.RequestContext.OriginalRequest, RegionEndpoint.GetBySystemName(preSignedUrlRequest.SourceRegion))).URL);
```
normj, line 100: This is using the new logic I added for the endpoints 2.0 code path allowing to pass in an alternate endpoint. In that case the source region of the resource being copied. This same logic is done in other services that have a copy feature like RDS and Neptune.

generator/ServiceClientGeneratorLib/Generators/Endpoints/EndpointResolver.partial.cs
```diff
@@ -14,7 +14,7 @@ private string GetValueSource(Parameter parameter)
         {
             switch (parameter.builtIn)
             {
-                case "AWS::Region": return "config.RegionEndpoint?.SystemName";
+                case "AWS::Region": return "requestContext.Request.AlternateEndpoint?.SystemName ?? config.RegionEndpoint?.SystemName;";
```
dscpinheiro, line 17: There's an extra `;` here, I ran the generator and the resolvers ended up with:
```csharp
result.Region = requestContext.Request.AlternateEndpoint?.SystemName ?? config.RegionEndpoint?.SystemName;;
```

generator/ServiceClientGeneratorLib/Generators/SourceFiles/ServiceConfig.tt
```diff
@@ -135,6 +135,9 @@ namespace <#=this.Config.Namespace#>
                 ClientConfig = this,
                 OriginalRequest = parameters.Request,
                 Request = new DefaultRequest(parameters.Request, ServiceId)
+                {
```
dscpinheiro, line 138: Not related to this PR, but what's the use case for `AlternateEndpoint`? I did not know about it and even had to double check it wasn't added in the PR.

normj: It is being used in scenarios like RDS and DocDB where they support copying a database from a different region. In this this code path is getting triggered by the changes `PreSignedUrlRequestHandler.cs` classes. We essentially have to create a presigned URL to the resource in the region the database is being copied. So we need use the current client config settings but override the region with the region the database is in.

In V3 the URL for the presigned URL was being generated by the old obsolete method that didn't use EP 2.0.

generator/ServiceClientGeneratorLib/Generators/Endpoints/EndpointResolver.partial.cs
normj: Removed the extra `;`

muhammad-othman (approved):

dscpinheiro (approved):

PR #3686, opened 2025-03-02
Remove S3 code that was obsolete in V3 for V4.

normj:
## Description
Deleted all obsolete code in S3 package for V4. Initially I thought I would leave some in but given that we want to look at auto generating S3 at some point I figured it would be best to remove as much customization logic as possible.

The obsolete code I was most on the fence about deleting
* The `Expires` property in favor of `ExpiresString`.
* The `GetAcl` and `PutAcl` service methods. We have created new gets and puts per object and bucket which map to the S3 model. I debated about keeping this obsolete code to give an easier migration but since this was the SDK moving closer to the model it seemed better for a future auto generation effort.
* CalculateMD5Header I debated about keeping because it had been marked obsolete fairly recently in V3. I figured it was best for simplifying the SDK logic and the migration experience is just delete the code that is ever setting it because the SDK now always computes a checksum.

## Testing
Dry Run: DRY_RUN-bd4e3e24-9cd1-4569-a2b7-eb07c6ae51eb Successful


dscpinheiro:
Thanks for removing the checksum related properties, I should've done that in the last V3 sync. 

But for `GetAcl` and `PutAcl`, they were not deprecated in V3 (Peter added the `Obsolete` attribute in V4: https://github.com/aws/aws-sdk-net/pull/3408/files#diff-b64870a2cb12bfa4f8f57ed10c69c61b95fe5d79bbb8e650bfde31d4f1a17ec5). Is it really OK to remove them now?

normj:
I have reverted the removal of the `GetACL` and `PutACL` methods.

dscpinheiro (approved):

peterrsongg (approved):

PR #3698, opened 2025-03-10
Add telemetry data for Credentials retrieval

muhammad-othman:
## Description
Add telemetry data for Credentials retrieval.

## Motivation and Context
The SDK used to emit telemetry data for Credentials retrieval here https://github.com/aws/aws-sdk-net-staging/blob/main-staging/sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/CredentialsRetriever.cs#L57
The `CredentialsRetriever` handler was removed in V4 as part of the SRA I&A changes, this PRs adds the telemetry data back to the `Signer` handler.

## Testing

`DRY_RUN-990010e0-a8db-4830-8fa6-fa05b2b8a155`

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/Signer.cs
+++ b/sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/Signer.cs
@@ -17,6 +17,7 @@ using Amazon.Runtime.Internal.Auth;
 using Amazon.Runtime.Internal.Util;
 using Amazon.Runtime.Telemetry;
 using Amazon.Runtime.Telemetry.Metrics;
+using Amazon.Runtime.Telemetry.Tracing;
 using Amazon.Util;
 using System;
 using System.IO;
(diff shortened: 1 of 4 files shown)


sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/Signer.cs
```diff
@@ -110,7 +111,14 @@ public static void SignRequest(IRequestContext requestContext)
             using (requestContext.Metrics.StartEvent(Metric.RequestSigningTime))
             using (MetricsUtilities.MeasureDuration(requestContext, TelemetryConstants.AuthSigningDurationMetricName))
             {
-                var immutableCredentials = (requestContext.Identity as AWSCredentials)?.GetCredentials();
+                ImmutableCredentials immutableCredentials = null;
```
dscpinheiro, line 114: Don't you need to add this in the async method as well? The previous pipeline handler had both.

---

Also: I mentioned this offline (and I know this now matches V3), but is tracking the time to retrieve credentials from the context what customers expect? Wouldn't it be more accurate to see how much time it took to resolve credentials instead?

normj: I'm assuming the `GetCredentials()` would be the resolution of reaching out to IMDS or Assume role if the current cache of credentials has expired.

Do you mean resolving what type of credentials to use as part of the SRA work? That to me would be a separate trace but I assume wouldn't the cost of that be essentially nil since we don't make any network calls or anything to resolve that.

dscpinheiro: Yes, we didn't add a trace to the SRA auth resolver and maybe it makes sense.

But I don't follow the "cost would be nil": if there's a role to be assumed for example, we'd call it here right? https://github.com/aws/aws-sdk-net/blob/v4-development/sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs#L105 and https://github.com/aws/aws-sdk-net/blob/v4-development/sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs#L55

normj: I was thinking the logic for determining if we need to use `DefaultAWSCredentialsIdentityResolver`. That would be nil. But you are right that `DefaultAWSCredentialsIdentityResolver` itself will make expensive calls reaching out to IMDS. Running through the resolver should be a separate trace.

normj (approved):

dscpinheiro (approved):

PR #3699, opened 2025-03-11
Better exception messages

opottone:
Improve error messages.

## Description

In some cases when the SDK throws an exception, the error message is "Value cannot be null. (Parameter 'Options property cannot be empty: ClientName')". It should be "Value cannot be null. (Parameter 'ClientName')". The problem is caused by incorrect usage of `ArgumentNullException` constructor.

## Types of changes

## Checklist

## License


dscpinheiro:
Thanks for the contribution, but as you noticed (in #3704) this class has a bigger problem where it doesn't handle expired credentials, and updating the error message doesn't solve the problem.

I'll close this PR and we'll triage that issue instead.

PR #3710, opened 2025-03-19
Fix bug in V4 where copy object wasn't working for some special characters

peterrsongg:
## Description

In V4 we [updated our valid path characters ](https://github.com/aws/aws-sdk-net/blob/v4-development/sdk/src/Core/Amazon.Util/AWSSDKUtils.cs#L122-L128)list to be in accordance with RFC3389. We also updated url encode logic to only encode the characters in that list if the resource path was part of a label substitution. If the character is not part of a label subsitution, we don't encode those characters. CopyObject's marshaller has a special method which was calling its own AmazonS3Util.UrlEncode method which was passing in `true` to UrlEncode, essentially not encoding. In this PR I remove that method altogether and update all occurences to use AWSSDKUtils.UrlEncode. I pass in false so we do url-encode as we should.

## Motivation and Context

Internal ticket
## Testing

DRY_RUN-4a57b081-ad7e-4a0e-a5e8-d0af921d8554
Added integ tests
## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/test/Services/S3/IntegrationTests/CopyObjectTests.cs
+++ b/sdk/test/Services/S3/IntegrationTests/CopyObjectTests.cs
@@ -73,6 +73,7 @@ namespace AWSSDK_DotNet.IntegrationTests.Tests.S3
             BaseClean();
         }
         [DataTestMethod]
+        [DataRow(testKey, testKey)]
         [DataRow("ObjectWithAllSpecialCharacters/'()!*$+,;=&", "DestinationObjectWithAllSpecialCharacters/'()!*$+,;=&")]
         [TestCategory("S3")]
         public void TestCopyObject(string sourceKey, string destinationKey)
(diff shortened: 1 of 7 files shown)


sdk/test/Services/S3/IntegrationTests/CopyObjectTests.cs
```diff
@@ -72,18 +72,39 @@ public void TestCleanup()
             
             BaseClean();
         }
-
-        [TestMethod]
+        [DataTestMethod]
+        [DataRow("ObjectWithAllSpecialCharacters/'()!*$+,;=&", "DestinationObjectWithAllSpecialCharacters/'()!*$+,;=&")]
```
muhammad-othman, line 76: It seems reasonable to keep the original test by adding another data row that uses `testKey` instead of removing it.

muhammad-othman (approved):

sdk/test/Services/S3/IntegrationTests/CopyObjectTests.cs
peterrsongg: Oops, didn't realize I deleted the existing test case. I added it back just in case

afroz429 (approved):

Issue #3362, opened 2024-06-29
V4 Development Tracker


normj:
### Describe the issue

Placeholder

### Links

placeholder

dscpinheiro:
We'll publish the preview packages to NuGet (e.g. `<PackageReference Include="AWSSDK.AccessAnalyzer" Version="4.0.0.0-preview" />`), we're just working on some changes to our internal build system at the moment (we'll update this issue as soon as the packages are available).

normj:
First preview packages have been released. Blog post for the release https://aws.amazon.com/blogs/developer/preview-1-of-aws-sdk-for-net-v4/

martincostello:
Are you able to provide a preview version of `Amazon.AspNetCore.DataProtection.SSM` that depends on `AWSSDK.Extensions.NETCore.Setup`?

Trying to enable native AoT for a project that depends on the SSM package gives a bunch of warnings:

```text
ILC : Trim analysis error IL2026: Microsoft.Extensions.DependencyInjection.ExtensionMethods.PersistKeysToAWSSystemsManager(IDataProtectionBuilder,String,Action`1<PersistOptions>): Using member 'Microsoft.Extensions.DependencyInjection.ServiceCollectionExtensions.TryAddAWSService<IAmazonSimpleSystemsManagement>(IServiceCollection,ServiceLifetime)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. The AWSSDK.Extensions.NETCore.Setup package has not been updated to support Native AOT compilations. 
C:\codebuild\tmp\output\src1982900738\src\aws-sdk-net\sdk\src\Core\Amazon.Runtime\ConstantClass.cs(84): Trim analysis error IL2072: Amazon.Runtime.ConstantClass.Intern(): 'type' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicFields' in call to 'Amazon.Runtime.ConstantClass.LoadFields(Type)'. The return value of method 'System.Object.GetType()' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
  ILC: Method '[AWSSDK.Extensions.NETCore.Setup]Amazon.Extensions.NETCore.Setup.DefaultClientConfig..ctor()' will always throw because: Missing method 'Void Amazon.Runtime.ClientConfig..ctor()'
[...]

PR #3715, opened 2025-03-20
Introduce AddCredentialsFactory() extension methods

chase-miller:
## Description
This PR updates the [AWSSDK.Extensions.NETCore.Setup](https://github.com/aws/aws-sdk-net/tree/main/extensions/src/AWSSDK.Extensions.NETCore.Setup) nuget package with the following:

- Introduces `IServiceCollection.[Try]AddCredentialsFactory` extension methods
- Provides a `DefaultAWSCredentialsFactory` that is effectively a move of the private `ClientFactory.CreateCredentials` method
- Updates the `ClientFactory` to use the injected `IAWSCredentialsFactory`
- Adds a registration for `AWSCredentials` so that it can be injected by consuming code.

## Motivation and Context
There are scenarios where it's necessary to obtain `AWSCredentials`. One scenario is adding sigv4 headers when making RESTful calls using the idiomatic `HttpClient`; as an aside I happen to be using the [AwsSignatureVersion4](https://www.nuget.org/packages/AwsSignatureVersion4) nuget package for this. Today a consumer has to effectively copy the `ClientFactory.CreateCredentials` method to be able to register an `AWSCredentials` object that is created in a way that leverages `AWSOptions`. With these changes a consumer simply needs to call `services.AddCredentialsFactory()` and then inject `AWSOptions` where needed.

These updates also provide a hook for a consumer to customize the way in which `AWSCredentials` are provided to ServiceClients.

## Testing
I added 10 unit tests to the `DependencyInjectionTests`. I'm also using what is effectively a fork with these changes in production code.

## Types of changes

## Checklist
- [?] My code follows the code style of this project

## License

### DIFF
--- a/extensions/src/AWSSDK.Extensions.NETCore.Setup/AWSOptions.cs
+++ b/extensions/src/AWSSDK.Extensions.NETCore.Setup/AWSOptions.cs
@@ -21,6 +21,7 @@ using Amazon;
 using Amazon.Runtime;
 
 using Amazon.Extensions.NETCore.Setup;
+using AWSSDK.Extensions.NETCore.Setup;
 
 namespace Amazon.Extensions.NETCore.Setup
 {
@@ -110,7 +111,7 @@ namespace Amazon.Extensions.NETCore.Setup
         /// <returns>The service client that implements the service interface.</returns>
         public T CreateServiceClient<T>() where T : IAmazonService
         {
-            return (T)ClientFactory.CreateServiceClient(null, typeof(T), this);
+            return (T)ClientFactory.CreateServiceClient(null, typeof(T), this, new DefaultAWSCredentialsFactory(this));
         }
 
         /// <summary>
--- a/extensions/src/AWSSDK.Extensions.NETCore.Setup/ClientFactory.cs
+++ b/extensions/src/AWSSDK.Extensions.NETCore.Setup/ClientFactory.cs
@@ -16,7 +16,7 @@ using System;
 using System.Reflection;
 using Amazon.Runtime;
 using Amazon.Runtime.CredentialManagement;
-
+using AWSSDK.Extensions.NETCore.Setup;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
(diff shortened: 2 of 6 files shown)


dscpinheiro:
Is there an existing issue with more details on the problem you're trying to solve? 

We made significant changes to both the `AWSSDK.Extensions.NETCore.Setup` package and credential resolution in our upcoming V4 version (https://github.com/aws/aws-sdk-net/issues/3362) and I'm afraid a lot of the work you've done cannot be reviewed as is.

chase-miller:
> Is there an existing issue with more details on the problem you're trying to solve?

Nope I haven't created an issue. What additional detail are you looking for?

> We made significant changes to both the `AWSSDK.Extensions.NETCore.Setup` package and credential resolution in our upcoming V4 version (#3362) and I'm afraid a lot of the work you've done cannot be reviewed as is.

Why can't it be reviewed as is?

chase-miller:
P.S. I looked at the `v4-release`, and it appears that these changes could be applied against that version of the lib. Would that be better? I think the problem I'm trying to solve for remains in v4 (`ClientFactory` still uses a private method to resolve credentials), but I only spent a couple minutes looking at the code.

chase-miller:
@dscpinheiro FYI that I just created issue https://github.com/aws/aws-sdk-net/issues/3716 to provide additional context. Let me know if it makes sense to implement these changes against the `v4-development` branch.

chase-miller:
@dscpinheiro bump :-)

normj:
The feature of having an opt-in mechanism on `AWSSDK.Extensions.NETCore.Setup` that adds `AWSCredentials` to the dependency injection container makes sense to me. We do need to retarget this PR to `v4-development`. V4 is close to being GA and at this point any new features should be first done in V4 and then we can debate about back porting to V3.

A question I have is should we injecting `AWSCredentials` instead of a `IAWSCredentialsFactory` since that is what is likely going to to be used on other service's constructor. There could also be registered the `IAWSCredentialsFactory` and then the service register for `AWSCredentials` would grab the `IAWSCredentialsFactory` and use that. And the register method would be `AddAWSCredentials` instead of `AddCredentialsFactory`. I'm just not sure how many users will not what the factory is and that would appear more of an implementation detail to get to what users want which is the credentials.

chase-miller:
@normj I created a series of draft PRs that target `v4-development` and build off one another. Really just trying to throw out ideas...

1. https://github.com/chase-miller/aws-sdk-net/pull/1 
2. https://github.com/chase-miller/aws-sdk-net/pull/2
3. https://github.com/chase-miller/aws-sdk-net/pull/3
4. https://github.com/chase-miller/aws-sdk-net/pull/4
5. https://github.com/chase-miller/aws-sdk-net/pull/6
6. https://github.com/chase-miller/aws-sdk-net/pull/7

I would probably only do 3+. The 6th in that list is a can of worms, but I figured I'd share it as imo it offers a lot of flexibility to the consumer; it might be advantageous to split `ServiceCollectionExtensions` like in 6 regardless since it's getting pretty big.

> A question I have is should we injecting AWSCredentials instead of a IAWSCredentialsFactory since that is what is likely going to to be used on other service's constructor. There could also be registered the IAWSCredentialsFactory and then the service register for AWSCredentials would grab the IAWSCredentialsFactory and use that. And the register method would be AddAWSCredentials instead of AddCredentialsFactory.

I (think I) implemented this in [the 5th PR](https://github.com/chase-miller/aws-sdk-net/pull/6) in the list above. Does that align with what you're thinking? This seems cleaner to me than creating another type but I haven't tested.

> I'm just not sure how many users will not what the factory is and that would appear more of an implementation detail to get to what users want which is the credentials.

I'm with you there. It also gets a little awkward with `AWSOptions` exposing its `Credentials`property. What if I set this property _and_ call `AddAWSCredentials()`...which one wins?

Note that I ran across an issue with registering `AWSCredentials` as anything but a singleton, so we might want to add a warning or something in the code's xml docs. We could lock it down so that `AWSCredentials` must be a singleton, but that seems too restrictive. The issue was that some AWSCredentials implementations implement `IDisposable`, so the DI framework was disposing the object after use. `FallbackCredentialsFactory` was then trying to reuse a cached instance of the object (and we'd get object disposed exceptions).

Thoughts on any of these iterations?

PR #3717, opened 2025-03-21
Add feature IDs for credential providers

dscpinheiro:
## Description
Updates the credential providers to set the feature ID specifying how they were resolved. Still missing from this PR are:
- `CREDENTIALS_AWS_SDK_STORE`
- `SSO_LOGIN_DEVICE`
- `SSO_LOGIN_AUTH`

I followed a similar approach the JS team did, which was to add a property to our `AWSCredentials` class that can be overwritten by each individual implementation: https://github.com/aws/aws-sdk-js-v3/pull/6546/files

## Testing
Manually built the `AWSSDK.NetFramework.sln` solution (full dry-run will be done before merging to `v4-development`)

## Types of changes

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/CredentialManagement/AWSCredentialsFactory.cs
+++ b/sdk/src/Core/Amazon.Runtime/CredentialManagement/AWSCredentialsFactory.cs
@@ -12,17 +12,17 @@
  * express or implied. See the License for the specific language governing
  * permissions and limitations under the License.
  */
-using Amazon.Runtime.Internal;
 using Amazon.Runtime.CredentialManagement.Internal;
+using Amazon.Runtime.Credentials.Internal;
+using Amazon.Runtime.Internal.Settings;
+using Amazon.Runtime.Internal.UserAgent;
 using Amazon.Util;
+using Amazon.Util.Internal;
 using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Linq;
-using Amazon.Runtime.Credentials.Internal;
-using Amazon.Runtime.Internal.Settings;
-using Amazon.Util.Internal;
 
 namespace Amazon.Runtime.CredentialManagement
 {
@@ -205,7 +205,10 @@ namespace Amazon.Runtime.CredentialManagement
                     case CredentialProfileType.BasicWithServices:
                     case CredentialProfileType.BasicWithGlobalEndpoint:
                     case CredentialProfileType.BasicWithServicesAndGlobalEndpoint:
-                        return new BasicAWSCredentials(options.AccessKey, options.SecretKey);
+                        return new BasicAWSCredentials(options.AccessKey, options.SecretKey)
+                        {
+                            FeatureIdSource = UserAgentFeatureId.CREDENTIALS_PROFILE,
+                        };
                     case CredentialProfileType.Session:
                     case CredentialProfileType.SessionWithServices:
                     case CredentialProfileType.SessionWithGlobalEndpoint:
@@ -276,7 +279,11 @@ namespace Amazon.Runtime.CredentialManagement
                             ExternalId = options.ExternalID,
                             MfaSerialNumber = options.MfaSerial
                         };
-                        return new AssumeRoleAWSCredentials(sourceCredentials, options.RoleArn, roleSessionName, assumeRoleOptions);
+                        return new AssumeRoleAWSCredentials(sourceCredentials, options.RoleArn, roleSessionName, assumeRoleOptions)
+                        { 
+                            FeatureIdSource = UserAgentFeatureId.CREDENTIALS_PROFILE_SOURCE_PROFILE,
+                        };
+                    
                     case CredentialProfileType.AssumeRoleCredentialSource:
                     case CredentialProfileType.AssumeRoleCredentialSourceWithGlobalEndpoint:
                     case CredentialProfileType.AssumeRoleCredentialSourceWithServices:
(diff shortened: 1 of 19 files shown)


sdk/src/Core/Amazon.Runtime/CredentialManagement/AWSCredentialsFactory.cs
```diff
@@ -205,7 +205,10 @@ private static AWSCredentials GetAWSCredentialsInternal(
                     case CredentialProfileType.BasicWithServices:
                     case CredentialProfileType.BasicWithGlobalEndpoint:
                     case CredentialProfileType.BasicWithServicesAndGlobalEndpoint:
-                        return new BasicAWSCredentials(options.AccessKey, options.SecretKey);
+                        return new BasicAWSCredentials(options.AccessKey, options.SecretKey)
+                        {
+                            FeatureIdSource = UserAgentFeatureId.CREDENTIALS_PROFILE,
```
dscpinheiro, line 210: The feature IDs are very granular (e.g. `An operation called using credentials resolved from code` vs `An operation called using credentials resolved from a config file(s) profile with static credentials`) so that's why I had to update this class too.

sdk/src/Core/Amazon.Runtime/Internal/UserAgent/UserAgentFeatureId.cs
```diff
@@ -20,26 +20,11 @@ namespace Amazon.Runtime.Internal.UserAgent
     /// </summary>
     public class UserAgentFeatureId : ConstantClass
     {
-        /// <summary>
```
dscpinheiro, line 23: @muhammad-othman I started removing some of the IDs we don't support right now. 

The only one I hesitated a little bit was `CREDENTIALS_STS_ASSUME_ROLE_SAML`. There's a `CREDENTIALS_STS_FEDERATION_TOKEN` that looks similar and our class for SAML is named `FederatedAWSCredentials`...

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryHandler.cs
```diff
@@ -68,7 +69,16 @@ public RetryHandler(RetryPolicy retryPolicy)
         public override void InvokeSync(IExecutionContext executionContext)
         {
             var requestContext = executionContext.RequestContext;
-            var responseContext = executionContext.ResponseContext;
+            switch (requestContext.ClientConfig.RetryMode)
```
dscpinheiro, line 72: @muhammad-othman This was in the marshaller, but I think it makes more sense here. Let me know if you disagree.

sdk/src/Core/Amazon.Runtime/CredentialManagement/AWSCredentialsFactory.cs
muhammad-othman: I like that we can now find out how the credentials were resolved but I would prefer if the we added an independent property that stores the source of the credentials, since it be useful to figure out how the credentials were obtained, still I'm ok with this implementation for now.

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryHandler.cs
muhammad-othman: Yep this is much better than the marshaller.

muhammad-othman (approved):

sdk/src/Core/Amazon.Runtime/Internal/UserAgent/UserAgentFeatureId.cs
muhammad-othman: I'm not sure about this one too, there are some details about the usage here https://code.amazon.com/packages/AwsDrSeps/blobs/main/--/FEATURES.md#:~:text=STS%20assume%20role%20with%20SAML%20docs
Will take a look and see if we should remove it.

sdk/src/Core/Amazon.Runtime/CredentialManagement/AWSCredentialsFactory.cs
```diff
@@ -302,7 +309,11 @@ private static AWSCredentials GetAWSCredentialsInternal(
 
                         roleSessionName = options.RoleSessionName ?? RoleSessionNamePrefix + AWSSDKUtils.CorrectedUtcNow.Ticks;
                         assumeRoleOptions = new AssumeRoleAWSCredentialsOptions();
-                        return new AssumeRoleAWSCredentials(sourceCredentials, options.RoleArn, roleSessionName, assumeRoleOptions);
+                        return new AssumeRoleAWSCredentials(sourceCredentials, options.RoleArn, roleSessionName, assumeRoleOptions)
+                        {
+                            FeatureIdSource = UserAgentFeatureId.CREDENTIALS_PROFILE_NAMED_PROVIDER,
```
normj, line 314: These all seem like they should have the `CREDENTIALS_STS_ASSUME_ROLE` feature id but I can also see that it is named profile so an argument can be made for `CREDENTIALS_PROFILE_NAMED_PROVIDER`. Does `FeatureIdSource` need to be a collection?

dscpinheiro: Just checked the JS implementation and they do add multiple feature IDs. I tested with a source profile and their user agent had both `CREDENTIALS_STS_ASSUME_ROLE` and `CREDENTIALS_PROFILE_SOURCE_PROFILE`:
```
[profile A]
role_arn = arn:aws:iam::000000000000:role/test-assume-role
credential_source = Environment
```

Updated the feature ID source property to be a collection in https://github.com/aws/aws-sdk-net/pull/3717/commits/711713677aaa84fb87f9bcb9a2f92146ba29585f

normj (approved):

PR #3718, opened 2025-03-21
Feat: Support account ID based endpoints

peterrsongg:
## Description

Adds support for Account ID based endpoints. For most assume role operations, the account Id will be taken from the assumed role arn. For SSO we take the accountID from the request. Most static credentials were updated to include accountId as a value in their constructors in a backwards compatible way. 

**NOTE** One other thing I did as part of this is move `Services` and `EndpointUrl` to `CredentialProfile` and away from `CredentialProfileOptions`. This is because these aren't being used to create unique credential types. It was causing the `CredentialProfileType` enum to get exponentially larger, making us have to account for all the combinations of EndpointUrl and Services. This also meant that I removed a bunch of `CredentialProfileType`s. I figured this was safe to do because it is V4 and this is where we can make breaking changes. Without this change, any addition to `CredentialProfileOptions` would make the `CredentialProfileType` enum grow even larger which i thought is unsustainable.

## Testing

DRY_RUN-2958ec6b-8cf7-417d-8454-5283d4caa519
PS dry run passed
## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/CredentialManagement/CredentialProfile.cs
+++ b/sdk/src/Core/Amazon.Runtime/CredentialManagement/CredentialProfile.cs
@@ -161,8 +161,21 @@ namespace Amazon.Runtime.CredentialManagement
         /// </summary>
         public string ClientAppId { get; set; }
 
-        public string AwsAccountId { get; set; }
-
+        /// <summary>
+        /// The name of the section which contains the custom endpoints for a service or services.
+        /// For example: 
+        /// [profile foo]
+        /// services = bar
+        /// [services bar]
+        /// s3 = 
+        ///   endpoint_url = https://custom-endpoint-s3:80
+        /// ec2 = 
+        ///   endpoint_url = https://custome-endpoint_ec2:80
+        /// This will tell the SDK to look for custom endpoints in "bar" for the profile "foo.
+        /// A single Services section can contain configurations for multiple services.
+        /// </summary>
+        public string Services { get; set; }
+        
 
         /// <summary>
         /// Determines the behavior for calculating checksums for request payloads.
--- a/sdk/src/Core/Amazon.Runtime/CredentialManagement/CredentialProfile.cs
+++ b/sdk/src/Core/Amazon.Runtime/CredentialManagement/CredentialProfile.cs
@@ -161,6 +161,8 @@ namespace Amazon.Runtime.CredentialManagement
         /// </summary>
         public string ClientAppId { get; set; }
 
+        public string AwsAccountId { get; set; }
+
 
         /// <summary>
         /// Determines the behavior for calculating checksums for request payloads.
--- a/sdk/src/Core/Amazon.Runtime/Credentials/AssumeRoleImmutableCredentials.cs
+++ b/sdk/src/Core/Amazon.Runtime/Credentials/AssumeRoleImmutableCredentials.cs
@@ -37,16 +37,27 @@ namespace Amazon.Runtime
         /// <param name="token">The security token for the credentials.</param>
         /// <param name="expiration">The expiration time for the credentials.</param>
         public AssumeRoleImmutableCredentials(string awsAccessKeyId, string awsSecretAccessKey, string token, DateTime expiration)
-            : base(awsAccessKeyId, awsSecretAccessKey, token)
+            : this (awsAccessKeyId, awsSecretAccessKey, token, expiration, null)
         {
-            if (string.IsNullOrEmpty(token)) throw new ArgumentNullException("token");
-            Expiration = expiration;
+
             
         }
 
-        public AssumeRoleImmutableCredentials(string awsAccessKeyId, string awsSecretAccessKey, string token, DateTime expiration, string accountId) : this (awsAccessKeyId, awsSecretAccessKey, token, expiration)
+        /// <summary>
+        /// Constructs an instance with supplied keys, token, expiration, and account id. When the account id is set 
+        /// and the service supports account id based endpoints, AWS will send the request using the account-based endpoint rather
+        /// than the regional endpount. Account-based endpoints take the form https://<paramref name="accountId"/>.ddb.region.amazonaws.com
+        /// the request to 
+        /// </summary>
+        /// <param name="awsAccessKeyId">The AccessKey for the credentials</param>
+        /// <param name="awsSecretAccessKey">The SecretKey for the credentials.</param>
+        /// <param name="token">The security token for the credentials.</param>
+        /// <param name="expiration">The expiration time for the credentials.</param>
+        /// <param name="accountId">The account id for the credentials. The account id is your 12 digit account number with no hyphens. For example: 123456789012.</param>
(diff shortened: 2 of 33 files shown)


dscpinheiro (commented):
I know I'm not one of the reviewers, but I started looking earlier today and had some comments I don't want to lose.

Also, I don't quite understand all the code removal in `sdk/src/Core/Amazon.Runtime/CredentialManagement/Internal/CredentialProfileTypeDetector.cs` (even after reading the PR description, it doesn't look that would have to be completed at the same time as account-ID based endpoints)

sdk/src/Core/Amazon.Runtime/CredentialManagement/CredentialProfile.cs
```diff
@@ -161,6 +161,21 @@ internal Dictionary<string, Dictionary<string, string>> NestedProperties
...
+        /// services = bar
+        /// [services bar]
+        /// s3 = 
+        ///   endpoint_url = https://custom-endpoint-s3:80
+        /// ec2 = 
+        ///   endpoint_url = https://custome-endpoint_ec2:80
+        /// This will tell the SDK to look for custom endpoints in "bar" for the profile "foo.
+        /// A single Services section can contain configurations for multiple services.
+        /// </summary>
+        public string Services { get; set; }
```
normj, line 177: How come `Services` was moved but `EndpointUrl` was not moved?

sdk/src/Core/Amazon.Runtime/Credentials/BasicAWSCredentials.cs
```diff
@@ -38,11 +38,21 @@ public class BasicAWSCredentials : AWSCredentials
...
+        public BasicAWSCredentials(string accessKey, string secretKey) : this (accessKey, secretKey, null)
+        {
+        }
+
+        /// <summary>
+        /// Constructs a BasicAWSCredentials object for the specified accessKey, secretKey, and accountId
+        /// </summary>
+        /// <param name="accessKey"></param>
+        /// <param name="secretKey"></param>
+        /// <param name="accountId"></param>
```
normj, line 50: Lets add some descriptions for this parameters. I think in the past access key and secret key were obvious but account id not so much, particular what it is used for. Also if there are format requirements like the account id with no hyphens we should call that out. Once add comments for the parameters update the other constructor with the docs for access key and secret key.

sdk/src/Core/Amazon.Runtime/Credentials/AssumeRoleImmutableCredentials.cs
```diff
@@ -41,15 +41,20 @@ public AssumeRoleImmutableCredentials(string awsAccessKeyId, string awsSecretAcc
         {
             if (string.IsNullOrEmpty(token)) throw new ArgumentNullException("token");
             Expiration = expiration;
+            
         }
 
+        public AssumeRoleImmutableCredentials(string awsAccessKeyId, string awsSecretAccessKey, string token, DateTime expiration, string accountId) : this (awsAccessKeyId, awsSecretAccessKey, token, expiration)
```
normj, line 47: Constructor needs docs. For account be sure to add any formatting requirements like no hyphens.

sdk/src/Core/Amazon.Runtime/Credentials/_bcl+netstandard/SSOImmutableCredentials.cs
```diff
@@ -39,11 +39,24 @@ public class SSOImmutableCredentials : ImmutableCredentials
...
+        /// <param name="awsSecretAccessKey"></param>
+        /// <param name="token"></param>
+        /// <param name="expiration"></param>
+        /// <param name="accountId"></param>
+        /// <exception cref="ArgumentNullException"></exception>
+        public SSOImmutableCredentials(string awsAccessKeyId, string awsSecretAccessKey, string token, DateTime expiration, string accountId) : base(awsAccessKeyId, awsSecretAccessKey, token, accountId)
         {
             if (string.IsNullOrEmpty(token)) throw new ArgumentNullException(nameof(token));
             Expiration = expiration;
+            AccountId = accountId;
```
normj, line 59: You are passing `accountId` to the base constructor so setting this again here does not seem necessary.

sdk/src/Core/Amazon.Runtime/Credentials/ImmutableCredentials.cs
```diff
@@ -50,7 +50,7 @@ public class ImmutableCredentials
         /// <summary>
         /// Gets the AccountId property for the current credentials.
         /// </summary>
-        public string AccountId { get; private set; }
+        public string AccountId { get; protected set; }
```
normj, line 53: I think this was done for `SSOImmutableCredentials` but that change doesn't need seem necessary. Since this type is supposed to be immutable I really don't want this change allowing a potential subclass to mutated the object.

sdk/src/Core/Amazon.Runtime/Credentials/SessionAWSCredentials.cs
```diff
@@ -31,13 +31,16 @@ public class SessionAWSCredentials : AWSCredentials
         /// <param name="awsAccessKeyId"></param>
         /// <param name="awsSecretAccessKey"></param>
         /// <param name="token"></param>
-        public SessionAWSCredentials(string awsAccessKeyId, string awsSecretAccessKey, string token)
+        public SessionAWSCredentials(string awsAccessKeyId, string awsSecretAccessKey, string token) : this(awsAccessKeyId, awsSecretAccessKey, token, null)
+        {
+        }
+
+        public SessionAWSCredentials(string awsAccessKeyId, string awsSecretAccessKey, string token, string accountId)
```
normj, line 38: Constructor needs docs.

normj (changes requested):
With the rework of `CredentialProfileTypeDetector` how confident do you feel with the existing test coverage and your testing? I like simplifying the logic but there is a lot of specific credential profile configuration I want to be sure we are breaking.

peterrsongg:
> With the rework of `CredentialProfileTypeDetector` how confident do you feel with the existing test coverage and your testing? I like simplifying the logic but there is a lot of specific credential profile configuration I want to be sure we are breaking.

That was added as part of my work for service-specific endpoints and we have extensive test coverage that was added when I did that work. The [tests here](https://github.com/aws/aws-sdk-net/blob/dd7f9d01c7ad0829967fb5ac5c4badbbbf5764e6/sdk/test/UnitTests/Custom/Runtime/ServiceSpecificEndpointTests.cs#L19). I've always wanted to clean this class up b/c I realized only after I did that work that I should've added them to `CredentialProfile` and not `CredentialProfileOptions`. I'm confident we won't break any credential profile configuration, specifically the type of credentials we will be resolving, because in AWSCredentialsFactory all of the cases for endpointUrl + services credential profile types returned the same credentials anyways.

sdk/test/UnitTests/Custom/Runtime/Credentials/AccountIdEndpointTests/accountid-source-testcases.json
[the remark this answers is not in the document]
dscpinheiro: That's interesting, you should create the SEP CR as soon you can. 

We should be confident the issues are with the test cases and not our implementation (not to mention other SDKs have already released support for this feature).

normj (approved):

dscpinheiro (approved):

Issue #3630, opened 2025-01-31
AWS SDK requires .aws credential keys to be lowered case (expectation: case insensitive)


shahzadhuq:
### Describe the bug

AWS SDK is unable to read locally AWS profile credentials (in .aws/credentials) if the keys are all upper case. It requires the keys to all lower case.

**Successfully reads credentials**
- aws_access_key_id= ...
- aws_secret_access_key= ...
- aws_session_token= ...

**Fails silently**
- AWS_ACCESS_KEY_ID= ...
- AWS_SECRET_ACCESS_KEY= ...
- AWS_SESSION_TOKEN= ...

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

AWS SDK should be case insensitive to the casing of the credentials keys in the .aws/credentials file.

### Current Behavior

AWS SDK only supports (able to read) lowered case credentials keys in the .aws/credentials file.

### Reproduction Steps

I used this code to figure out the key case sensitivity issue.
Try this code to test with all 1/ lowered case keys and then 2/ upper case keys.
```C#
// Credential validation method
void ValidateAwsCredentials()
{
    try 
    {
        // Attempt to load credentials from default profile
        var chain = new CredentialProfileStoreChain();
        if (chain.TryGetAWSCredentials("default", out var awsCredentials))
        {
            Console.WriteLine("Credentials successfully loaded from profile");

            // Create STS client to validate credentials
            using var stsClient = new AmazonSecurityTokenServiceClient(awsCredentials);
            
            // Get caller identity
            var request = new GetCallerIdentityRequest();
            var response = stsClient.GetCallerIdentityAsync(request).Result;

            Console.WriteLine($"Account: {response.Account}");
            Console.WriteLine($"User ID: {response.UserId}");
            Console.WriteLine($"ARN: {response.Arn}");
        }
        else
        {
            Console.WriteLine("Failed to load AWS credentials from default profile");
        }
    }
    catch (Exception ex)
    {
        Console.WriteLine($"Credential Validation Error: {ex.Message}");
    }
}
```

### Additional Information/Context

Commonly, we configure such credentials using 'aws configure' command which adds all lowered case keys.

I used IAM Role temp credentials. So, from the AWS console, I copied and pasted the generated key/value pairs - generated keys are all upper case. After updating the .aws/credential file with the upper case keys, the AWS logger stopped writing to the AWS CloudWatch. Only after changing keys to all lower case, AWS logger started to work.

[...]

PR #3728, opened 2025-03-26
Fixed an issue where property keys in credentials/config file were incorrectly considered as case-sensitive.

ashishdhingra:
## Description
Fixed an issue where property keys in credentials/config file were incorrectly considered as case-sensitive.

For a Property Key (refer SEP link in `DOTNET-7965` for more details):
- It is `An identifier defined by the SDK as a supported setting. Property keys MUST be considered case-insensitive.`. This is irrespective of whether it is a `credentials` or `config` file.
- **Implementation Guidance:** After parsing, profiles undergo complex merging logic to merge profile from multiple individual files. To reduce the risk of bugs, SDKs MAY lowercase all profile keys during the parsing phase to simplify later logic.
- Property keys are considered case-insensitive. Property keys SHOULD be stored lower-cased after parsing. When keys are retrieved from the profile, they should also be lower-cased.
- **Side-note:** (outside scope of this issue) When a profile has the same property defined in both the configuration and credentials files, the credentials properties must be used.
 
## Motivation and Context
Issue: https://github.com/aws/aws-sdk-net/issues/3630

## Testing
- Added test case.
- 1st dry-run:
  - .NET (V3) dry-run `DRY_RUN-ecd2eb13-5382-49e6-8705-98b308fc170d` completed successfully.
  - PowerShell (V4) dry-run `DRY_RUN-5e9a6012-086a-4a19-8550-9c1282b4d032` completed successfully.
- 2nd dry-run (after changing to use `ToLowerInvariant()`):
  - .NET (V3) dry-run `DRY_RUN-f6147e3c-cb02-40fe-a36a-9846bb65cf1a` completed successfully.
  - PowerShell (V4) dry-run `DRY_RUN-c8c7f70b-5c63-4c43-9eb9-c6b6cd2bf487` completed successfully.

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Internal/Util/IniFile.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Util/IniFile.cs
@@ -595,6 +595,9 @@ namespace Amazon.Runtime.Internal.Util
                 nestedProperty = null;
                 if (TryParseProperty(Lines[lineNumber], out propertyName, out propertyValue))
                 {
+                    // Property keys are considered case-insensitive. Property keys SHOULD be stored lower-cased after parsing.
+                    propertyName = propertyName.ToLowerInvariant();
+
                     //if propertyValue is empty, that means that it is a continuation property
                     if (String.IsNullOrEmpty(propertyValue))
                     {
@@ -641,7 +644,9 @@ namespace Amazon.Runtime.Internal.Util
                 else if (StartsWithWhitespace(currentLine))
                 {
                     var separatorIndex = trimmedLine.IndexOf(keyValueSeparator, StringComparison.Ordinal);
-                    subpropertyName = trimmedLine.Substring(0, separatorIndex).Trim();
+
+                    // Property keys are considered case-insensitive. Property keys SHOULD be stored lower-cased after parsing.
+                    subpropertyName = trimmedLine.Substring(0, separatorIndex).ToLowerInvariant().Trim();
                     subpropertyValue = trimmedLine.Substring(separatorIndex + 1).Trim();
                     nestedProperty.SubpropertyKeys.Add(subpropertyName);
                     nestedProperty.SubpropertyValues.Add(subpropertyValue);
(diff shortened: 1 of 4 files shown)


peterrsongg (approved):

sdk/src/Core/Amazon.Runtime/Internal/Util/IniFile.cs
```diff
@@ -595,6 +595,9 @@ private bool SeekProperty(ref int lineNumber, out string propertyName, out strin
                 nestedProperty = null;
                 if (TryParseProperty(Lines[lineNumber], out propertyName, out propertyValue))
                 {
+                    // Property keys are considered case-insensitive. Property keys SHOULD be stored lower-cased after parsing.
+                    propertyName = propertyName.ToLower();
```
dscpinheiro, line 599: Since this is a file in the customer's environment, shouldn't you use [ToLowerInvariant](https://learn.microsoft.com/en-us/dotnet/api/system.string.tolowerinvariant) here? (I also know we're not 100% consistent within the SDK, I see both methods being used in other places)

sdk/test/UnitTests/Custom/Runtime/Credentials/SharedCredentialsFileTest.cs
```diff
@@ -497,7 +512,31 @@ public void ReadProfileWithSubproperties()
...
+                CredentialProfileOptions expectedProfileOptions = new CredentialProfileOptions
+                {
+                    AccessKey = "basic_aws_ACCESS_key_id",
+                    SecretKey = "basic_Aws_sEcRet_AcceSS_key",
+                };
+                Dictionary<string, string> expectedProperties = new Dictionary<string, string>
+                {
+                    { "output", "json" }
+                };
+                Dictionary<string, Dictionary<string, string>> expectedNestedProperties = new Dictionary<string, Dictionary<string, string>>();
```
dscpinheiro, line 533: Nit, but `Dictionary<string, Dictionary<string, string>> expectedNestedProperties = new Dictionary<string, Dictionary<string, string>>();` is really hard to read. 

I'd use `var expectedNestedProperties = new ...` instead (same comment applies to all variables in this test).

ashishdhingra: Done

dscpinheiro (approved):

PR #3729, opened 2025-03-27
Generate and Send the User Agent String with Business Metrics

muhammad-othman:
## Description

This PR introduces a new approach for handling user agent construction and management by centralizing logic within `UserAgentDetails`:

* Added `UserAgentComponent` and `GenerateUserAgentWithMetrics` in `UserAgentDetails`.
* Added `UserAgentDetails` to `AmazonWebServiceRequest` so requests can have direct access to user agent incase the request context isn't available.
* Replaced existing user agent usages with `UserAgentDetails`.
* Removed `AmazonS3UserAgentHandler` since it is no longer needed after UA 2.1.
* Moved setting the user agent header from the Marshaller to the HTTP handler just before sending the request.

## Motivation and Context

`#DOTNET-7975`
`#DOTNET-7974`

## Testing
`DRY_RUN-aa6c8737-aca2-4a4b-8973-8848c63bba8f`

## Types of changes

## Checklist

## License


sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/Marshaller.cs
```diff
@@ -124,51 +124,19 @@ private static void SetRecursionDetectionHeader(IDictionary<string, string> head
...
-
-            var userAgentAddition = requestContext.OriginalRequest.UserAgentAddition;
-            if (!string.IsNullOrEmpty(userAgentAddition))
-            {
-                sb.Append(' ').Append(InternalSDKUtils.ReplaceInvalidUserAgentCharacters(userAgentAddition));
-            }
-
-            var userAgent = sb.ToString();
-
-            if (requestContext.ClientConfig.UseAlternateUserAgentHeader)
```
dscpinheiro, line 151: I know we made this change on purpose, but since the user agent is now added later in the request pipeline the header doesn't show up if I enable debug logging (`AWSConfigs.LoggingConfig.LogResponses = ResponseLoggingOption.Always`) - only signed headers are included.

@normj had mentioned we should stop signing this header anyway, but not being to see it when debugging seems like a worse customer experience we need to address.

sdk/src/Core/Amazon.Runtime/Internal/UserAgent/UserAgentDetails.cs
```diff
@@ -43,17 +57,57 @@ public void AddFeature(UserAgentFeatureId featureId)
...
+
         /// <summary>
         /// Generates the User-Agent metrics string.
         /// Ensures the final string does not exceed 1024 bytes.
         /// </summary>
         /// <returns>The formatted User-Agent metrics string.</returns>
-#pragma warning disable CA1822 // Mark members as static
-        public string GenerateMetricsUserAgent()
-#pragma warning restore CA1822 // Mark members as static
+        private string GenerateMetricsUserAgent()
```
dscpinheiro, line 79: Have you checked what the PS user agent looks like? It's only called in one place (https://github.com/aws/aws-tools-for-powershell/blob/fd6b7a8e5049ca55a1bd17b01889d578d6b71b16/modules/AWSPowerShell/Utils/Utils.cs#L59-L62), but we should confirm it doesn't break with our changes.

generator/.DevConfigs/a069861d-cd92-4d43-b45b-3aa11112a39b.json
```diff
@@ -0,0 +1,9 @@
+{
+  "core": {
+    "changeLogMessages": [
+      "Generate and Send the User Agent String with Business Metrics"
```
dscpinheiro, line 4:
```suggestion
      "Update the AWS SDK for .NET to include encoded metrics in the `User-Agent` header to track which features were used for a given request (for example, which retry behavior and how credentials were resolved)"
```

dscpinheiro:
@muhammad-othman I just noticed I left a bug in the auth resolver handler 🤦: Since this line has a return it skips updating the request context feature IDs (https://github.com/aws/aws-sdk-net/blob/Business-Metrics-Tracking-UA-2.1/sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs#L79)

I think it needs to be a `break` instead. Could you fix it in this PR?

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/Marshaller.cs
dscpinheiro: Discussed offline: We can add a debug log here with the final user-agent header (the ideal solution would be to add full request logging, but that's a much larger effort).

sdk/src/Core/Amazon.Runtime/Internal/UserAgent/UserAgentDetails.cs
muhammad-othman: This one place doesn't cause any issues, however this one does 
https://github.com/aws/aws-tools-for-powershell/blob/fd6b7a8e5049ca55a1bd17b01889d578d6b71b16/modules/AWSPowerShell/Common/BaseCmdlets.cs#L489

I updated it locally to use `UserAgentDetails.AddUserAgentComponent`  and I verified that PowerShell specific user agent additions are added (`ft/S3Key#unmodified`)
image

Will create a PR for powershell v5 to use `UserAgentDetails.AddUserAgentComponent` after we merge this feature.

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/Marshaller.cs
muhammad-othman: Added the user agent to the logs

sdk/src/Services/S3/Custom/Transfer/Internal/_async/MultipartUploadCommand.async.cs
```diff
@@ -194,9 +194,7 @@ private void AbortMultipartUpload(string uploadId)
                 WebServiceRequestEventArgs wsArgs = args as WebServiceRequestEventArgs;
                 if (wsArgs != null)
                 {
-                    string currentUserAgent = wsArgs.Headers[AWSSDKUtils.UserAgentHeader];
-                    wsArgs.Headers[AWSSDKUtils.UserAgentHeader] =
-                        currentUserAgent + " ft/s3-transfer md/UploadNonSeekableStream";
+                    ((Runtime.Internal.IAmazonWebServiceRequest)wsArgs.Request).UserAgentDetails.AddUserAgentComponent("ft/s3-transfer md/UploadNonSeekableStream");
```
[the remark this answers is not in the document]
muhammad-othman: Yep, updated.

dscpinheiro (approved):

peterrsongg:
I found one other place that uses the old way of adding to the user agent header. Can you update that as well?
[UserAgentPostFix](https://github.com/aws/aws-sdk-net/blob/v4-development/sdk/src/Services/Glacier/Custom/Transfer/ArchiveTransferManager.cs#L170-L174)

The annoying part is that this is part of an internal method which is then called in a bunch of places, but you should just be able to update this code and the other places shoulld be fixed

peterrsongg (approved):

PR #3732, opened 2025-03-28
Log warning when ServiceUrl and RegionEndpoint are set at the same time

peterrsongg:
## Description

Log a warning whenever serviceurl and regionendpoint are set at the same time
## Motivation and Context

internal app sec ticket
## Testing

dry run in progress
## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/ClientConfig.cs
+++ b/sdk/src/Core/Amazon.Runtime/ClientConfig.cs
@@ -271,6 +271,9 @@ namespace Amazon.Runtime
             }
             set
             {
+                if (!string.IsNullOrEmpty(this.serviceURL))
+                    Logger.GetLogger(GetType()).InfoFormat($"RegionEndpoint and ServiceURL are mutually exclusive. Since " +
+                        $"RegionEndpoint was set last, RegionEndpoint: {value} will be used to make the request.");
                 this.defaultConfigurationBackingField = null;
                 this.serviceURL = null;
                 this.regionEndpoint = value;
@@ -373,6 +376,9 @@ namespace Amazon.Runtime
             }
             set
             {
+                if (regionEndpoint != null)
+                    Logger.GetLogger(GetType()).InfoFormat($"RegionEndpoint and ServiceURL are mutually exclusive. Since " +
+                        $"ServiceUrl was set last, ServiceUrl: {value} will be used to make the request.");
                 this.regionEndpoint = null;
                 this.probeForRegionEndpoint = false;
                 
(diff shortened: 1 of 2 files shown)


sdk/src/Core/Amazon.Runtime/ClientConfig.cs
```diff
@@ -373,6 +376,9 @@ public string ServiceURL
             }
             set
             {
+                if (regionEndpoint != null)
```
GarrettBeatty, line 379: nit: here you use != null but above you use !string.IsNullOrEmpty. can you use string.IsNullOrEmpty here too?

peterrsongg: RegionEndpoint is not a string so i can't use that here

GarrettBeatty: oh i didnt realize that

GarrettBeatty (approved):

sdk/src/Core/Amazon.Runtime/ClientConfig.cs
```diff
@@ -271,6 +271,9 @@ public RegionEndpoint RegionEndpoint
             }
             set
             {
+                if (!string.IsNullOrEmpty(this.serviceURL))
+                    Logger.GetLogger(GetType()).InfoFormat($"RegionEndpoint and ServiceURL are mutually exclusive. Since " +
+                        $"RegionEndpoint was set last, RegionEndpoint: {value} will be used to make the request.");
```
normj, line 276: Can you make it clear in the message that the property `ServiceUrl` has been reset to `null`? I want users to know that not only did we ignore the `ServiceUrl` we cleared it out. That might matter to some people's mocking.

sdk/src/Core/Amazon.Runtime/ClientConfig.cs
```diff
@@ -373,6 +376,9 @@ public string ServiceURL
             }
             set
             {
+                if (regionEndpoint != null)
+                    Logger.GetLogger(GetType()).InfoFormat($"RegionEndpoint and ServiceURL are mutually exclusive. Since " +
```
normj, line 380: Same comment as above but in reverse direction.

sdk/src/Core/Amazon.Runtime/ClientConfig.cs
peterrsongg: done

normj (approved):

muhammad-othman (approved):

PR #3733, opened 2025-03-31
Populate User Agent Details

muhammad-othman:
## Description
Added feature IDs to the user agent:
* `CREDENTIALS_AWS_SDK_STORE`
* `RESOLVED_ACCOUNT_ID`
* `ENDPOINT_OVERRIDE`
* `ACCOUNT_ID_MODE_DISABLED`
* `ACCOUNT_ID_MODE_PREFERRED`
* `ACCOUNT_ID_MODE_REQUIRED`
* `SIGV4A_SIGNING`
* `S3_EXPRESS_BUCKET`

Still missing the following features that were added later to the SEP:
- `DDB_MAPPER`
- `SSO_LOGIN_DEVICE`
- `SSO_LOGIN_AUTH`

## Motivation and Context
`DOTNET-7974`

## Testing
`DRY_RUN-bf6c38e1-0972-4fd9-814d-d4a2ad98e183`

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/CredentialManagement/AWSCredentialsFactory.cs
+++ b/sdk/src/Core/Amazon.Runtime/CredentialManagement/AWSCredentialsFactory.cs
@@ -199,14 +199,16 @@ namespace Amazon.Runtime.CredentialManagement
         {
             if (profileType.HasValue)
             {
+                AWSCredentials awsCredentials = null;
                 switch (profileType)
                 {
                     case CredentialProfileType.Basic:
-                        var basicCredentials = new BasicAWSCredentials(options.AccessKey, options.SecretKey, options.AwsAccountId);
-                        basicCredentials.FeatureIdSources.Add(UserAgentFeatureId.CREDENTIALS_PROFILE);
-                        return basicCredentials;
+                        awsCredentials = new BasicAWSCredentials(options.AccessKey, options.SecretKey, options.AwsAccountId);
+                        awsCredentials.FeatureIdSources.Add(UserAgentFeatureId.CREDENTIALS_PROFILE);
+                        break;
                     case CredentialProfileType.Session:
-                        return new SessionAWSCredentials(options.AccessKey, options.SecretKey, options.Token, options.AwsAccountId);
+                        awsCredentials = new SessionAWSCredentials(options.AccessKey, options.SecretKey, options.Token, options.AwsAccountId);
+                        break;
                     case CredentialProfileType.AssumeRole:
                     case CredentialProfileType.AssumeRoleExternal:
                     case CredentialProfileType.AssumeRoleMFA:
@@ -252,10 +254,10 @@ namespace Amazon.Runtime.CredentialManagement
                             MfaSerialNumber = options.MfaSerial
                         };
 
-                        var assumeRoleCredentials = new AssumeRoleAWSCredentials(sourceCredentials, options.RoleArn, roleSessionName, assumeRoleOptions);
-                        assumeRoleCredentials.FeatureIdSources.Add(UserAgentFeatureId.CREDENTIALS_PROFILE_SOURCE_PROFILE);
-                        return assumeRoleCredentials;
-                    
+                        awsCredentials = new AssumeRoleAWSCredentials(sourceCredentials, options.RoleArn, roleSessionName, assumeRoleOptions);
+                        awsCredentials.FeatureIdSources.Add(UserAgentFeatureId.CREDENTIALS_PROFILE_SOURCE_PROFILE);
+                        break;
+
                     case CredentialProfileType.AssumeRoleCredentialSource:
                     case CredentialProfileType.AssumeRoleCredentialSourceSessionName:
                         // get credentials specified by credentialSource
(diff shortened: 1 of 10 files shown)


sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/HttpHandler.cs
```diff
@@ -489,6 +490,18 @@ private static System.IO.Stream GetInputStream(IRequestContext requestContext, S
 
         private void SetUserAgentHeader(IRequestContext requestContext)
         {
+            var accountIdMode = requestContext.ClientConfig.AccountIdEndpointMode;
```
dscpinheiro, line 493: Does it make sense to move this to the Marshaller? Since it's where we already have logic to update the user agent with properties from the `Config` object: https://github.com/aws/aws-sdk-net/blob/17ed221d3b036a7bd57c8b5729c43589d7740f5d/sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/Marshaller.cs#L127-L138

Then this handler would only care about updating the header in the actual request.

sdk/src/Core/Amazon.Runtime/Credentials/BasicAWSCredentials.cs
```diff
@@ -53,6 +53,7 @@ public BasicAWSCredentials(string accessKey, string secretKey) : this (accessKey
         /// <param name="accountId">The account id for the credentials. The account id is your 12 digit account number with no hyphens. For example: 123456789012</param>
         public BasicAWSCredentials(string accessKey, string secretKey, string accountId)
         {
+            if (string.IsNullOrEmpty(accountId)) FeatureIdSources.Add(UserAgentFeatureId.RESOLVED_ACCOUNT_ID);
```
dscpinheiro, line 56: Is there a better place we can check an account ID was resolved? Maybe in the signer? (where we have access to the immutable credentials object and would not need to update multiple credential providers)

This if is also inverted, right? `accountId` should _not_ be empty for us to track the feature ID.

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseEndpointResolver.cs
```diff
@@ -73,6 +74,10 @@ public virtual void ProcessRequestHandlers(IExecutionContext executionContext)
                 };
                 requestContext.Request.Endpoint = uriBuilder.Uri;
             }
+            if (!string.IsNullOrEmpty(requestContext.ClientConfig.ServiceURL))
+            {
+                executionContext.RequestContext.UserAgentDetails.AddFeature(UserAgentFeatureId.ENDPOINT_OVERRIDE);
```
dscpinheiro, line 79: Nit: `requestContext` is already defined, you don't need to use `executionContext.RequestContext` here.

dscpinheiro (changes requested):

sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/HttpHandler.cs
muhammad-othman: Can move the `AccountIdEndpointMode`, but the `SignatureVersion` has to be done after the signer, since this is where it gets updated.

dscpinheiro (approved):

sdk/src/Core/Amazon.Runtime/CredentialManagement/AWSCredentialsFactory.cs
```diff
@@ -199,14 +199,16 @@ private static AWSCredentials GetAWSCredentialsInternal(
...
                     case CredentialProfileType.Basic:
-                        var basicCredentials = new BasicAWSCredentials(options.AccessKey, options.SecretKey, options.AwsAccountId);
-                        basicCredentials.FeatureIdSources.Add(UserAgentFeatureId.CREDENTIALS_PROFILE);
-                        return basicCredentials;
+                        awsCredentials = new BasicAWSCredentials(options.AccessKey, options.SecretKey, options.AwsAccountId);
+                        awsCredentials.FeatureIdSources.Add(UserAgentFeatureId.CREDENTIALS_PROFILE);
+                        break;
                     case CredentialProfileType.Session:
-                        return new SessionAWSCredentials(options.AccessKey, options.SecretKey, options.Token, options.AwsAccountId);
+                        awsCredentials = new SessionAWSCredentials(options.AccessKey, options.SecretKey, options.Token, options.AwsAccountId);
```
normj, line 210: Why doesn't this `case` block add anything to the `FeatureIdSources`?

sdk/src/Core/Amazon.Runtime/CredentialManagement/AWSCredentialsFactory.cs
```diff
@@ -312,17 +310,18 @@ private static AWSCredentials GetAWSCredentialsInternal(
                                 UserIdentity = options.UserIdentity,
                                 ProfileName = profileName
                             };
-                            return new FederatedAWSCredentials(new SAMLEndpointManager().GetEndpoint(options.EndpointName),
+                            awsCredentials = new FederatedAWSCredentials(new SAMLEndpointManager().GetEndpoint(options.EndpointName),
```
normj, line 313: Do we need a `FeatureIdSources` here?

dscpinheiro: As of now, the SEP doesn't have a `_PROFILE` feature ID for this credential type, so it's only set in that class constructor: https://github.com/aws/aws-sdk-net/blob/Business-Metrics-Tracking-UA-2.1/sdk/src/Core/Amazon.Runtime/Credentials/FederatedAWSCredentials.cs#L80

(Same comment applies to `SessionAWSCredentials`: https://github.com/aws/aws-sdk-net/blob/Business-Metrics-Tracking-UA-2.1/sdk/src/Core/Amazon.Runtime/Credentials/SessionAWSCredentials.cs#L56)

We can follow up to see if it makes sense to track those as well.

Issue #2576, opened 2023-03-31
Allow to set DynamoDBEntryConversion per table


Dreamescaper:
### Describe the feature

Currently DynamoDBEntryConversion is set for the whole DynamoDbContext. It would be great to allow to set it per table via attribute instead.

### Use Case

We have some tables with V1 conversion, and we don't want to change them.
On the other hand, we'd prefer to use V2 for newly created tables.

### Proposed Solution

1. Make ConversionSchema enum public.
2. Add ConversionSchema property to DynamoDBTableAttribute.

Usage:
`[DynamoDBTable("payment-events", ConversionSchema = ConversionSchema.V2)]`

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS .NET SDK and/or Package version used

AWSSDK.DynamoDBv2 3.7.3.13

### Targeted .NET Platform

.NET 6

### Operating System and version

AmazonLinux

dscpinheiro:
This was released in the latest preview: https://github.com/aws/aws-sdk-net/releases/tag/4.0.0.0-preview.13

PR #3735, opened 2025-04-01
Issues/2576 dynamo db entry conversion

irina-herciu:
Allow to set DynamoDBEntryConversion per table

## Description
Update `DynamoDBTableAttribute` in order to allow `ConversionSchema` to be set at table level.

```Csharp
[DynamoDBTable("HashTable", false, ConversionSchema.V2)]
```
The ConversionSchema apply as follows:
- OperationConfig Conversion Schema
- Table Conversion Schema
- Context Conversion Schema 

## Motivation and Context
https://github.com/aws/aws-sdk-net/issues/2576

## Testing
Integration test added

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/DynamoDBv2/Custom/Conversion/DynamoDBEntryConversion.cs
+++ b/sdk/src/Services/DynamoDBv2/Custom/Conversion/DynamoDBEntryConversion.cs
@@ -30,11 +30,17 @@ using Amazon.Runtime.Internal.Util;
 
 namespace Amazon.DynamoDBv2
 {
+
     /// <summary>
     /// Available conversion schemas.
     /// </summary>
-    internal enum ConversionSchema
+    public enum ConversionSchema
     {
+        /// <summary>
+        /// No schema set.
+        /// </summary>
+        Unset = -1,
+
         /// <summary>
         /// Default schema before 2014 L, M, BOOL, NULL support
         /// 
--- a/sdk/src/Services/DynamoDBv2/Custom/Conversion/DynamoDBEntryConversion.cs
+++ b/sdk/src/Services/DynamoDBv2/Custom/Conversion/DynamoDBEntryConversion.cs
@@ -47,15 +47,17 @@ namespace Amazon.DynamoDBv2
         ///
         /// This schema pre-dates support for native DynamoDB types such as L (list), M (map), BOOL, and NULL.
         /// Common .NET type mappings:
-        /// - Number types (byte, int, float, decimal, etc.) → DynamoDB N (number)
-        /// - string, char → S (string)
-        /// - bool → N ("0" for false, "1" for true)
-        /// - DateTime, Guid → S (string)
-        /// - MemoryStream, byte[] → B (binary)
-        /// - List, HashSet, array of numeric types → NS (number set)
-        /// - List, HashSet, array of string types → SS (string set)
-        /// - List, HashSet, array of binary types → BS (binary set)
-        /// - Dictionary{string, object} → M (map)
+        /// <ul>
+        /// <li><para>Number types (byte, int, float, decimal, etc.) → DynamoDB N (number)</para></li>
+        /// <li><para>string, char → S (string)</para></li>
+        /// <li><para>bool → N ("0" for false, "1" for true)</para></li>
+        /// <li><para>DateTime, Guid → S (string)</para></li>
+        /// <li><para>MemoryStream, byte[] → B (binary)</para></li>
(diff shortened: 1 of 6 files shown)


sdk/src/Services/DynamoDBv2/Custom/DataModel/Attributes.cs
```diff
@@ -52,22 +52,41 @@ public sealed class DynamoDBTableAttribute : DynamoDBAttribute
         /// </summary>
         public bool LowerCamelCaseProperties { get; set; }
 
+        /// <summary>
+        /// Gets and sets the Conversion Schema property.
```
normj, line 56: This property should explain the precedence of setting this property on the attribute and setting it through the operation config and context config.

sdk/src/Services/DynamoDBv2/Custom/Conversion/DynamoDBEntryConversion.cs
```diff
@@ -30,11 +30,17 @@
...
-    internal enum ConversionSchema
+    public enum ConversionSchema
     {
+        /// <summary>
+        /// No schema set.
+        /// </summary>
+        Unset = -1,
+
         /// <summary>
         /// Default schema before 2014 L, M, BOOL, NULL support
```
normj, line 45: Since you are making this enum public we need to make sure the public docs are good. Saying "Default schema before 2014" is a confusing way to say this because it is actually still the default for the context and table. This property and the V2 property need their docs updated to be more clear for public docs.

sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs
```diff
@@ -552,10 +553,29 @@ public DynamoDBFlatConfig(DynamoDBOperationConfig operationConfig, DynamoDBConte
         public List<ScanCondition> QueryFilter { get; set; }
 
         /// <summary>
-        /// Conversion specification which controls how conversion between
+        /// Entity Conversion specification which controls how conversion between
```
normj, line 556: Can you add more information what you mean by `Entity`? I assume this is the possible conversation coming from the `DataTable` attribute.

sdk/src/Services/DynamoDBv2/Custom/DataModel/Attributes.cs
irina-herciu: Property summary updated

sdk/src/Services/DynamoDBv2/Custom/Conversion/DynamoDBEntryConversion.cs
irina-herciu: Docs updated

normj (approved):
Dry run DRY_RUN-98864961-5f50-49fe-91a4-d9b735b91842 was successful

sdk/src/Services/DynamoDBv2/Custom/Conversion/DynamoDBEntryConversion.cs
```diff
@@ -30,42 +30,51 @@
...
-        /// Dictionary{string,object} are converted to M
+        /// Indicates that no schema has been explicitly set.
+        /// </summary>
+        Unset = -1,
+
+        /// <summary>
+        /// Legacy conversion schema (and current default for context-level configurations).
+        ///
+        /// This schema pre-dates support for native DynamoDB types such as L (list), M (map), BOOL, and NULL.
+        /// Common .NET type mappings:
```
dscpinheiro, line 49: Changes look good, but here's what the intellisense is looking like in VS for me:

ddb-intellisense-before

Would it be possible to change to something like this?
```csharp
/// <ul>
/// <li><para>Number types (byte, int, float, decimal, etc.) → DynamoDB N (number)</para></li>
/// <li><para>string, char → S (string)</para></li>
/// <li><para>bool → N ("0" for false, "1" for true)</para></li>
/// <li><para>DateTime, Guid → S (string)</para></li>
/// <li><para>etc...</para></li>
/// </ul>
```

Then the intellisense would look better (and yes, the `<para>` inside the `<li>` are weird but unfortunately this is how our API reference doc generator expects lists today...):

ddb-intellisense-after

dscpinheiro (approved):

sdk/src/Services/DynamoDBv2/Custom/Conversion/DynamoDBEntryConversion.cs
irina-herciu: Updated as requested.

normj:
This PR was released as part of V4 preview 13.

Issue #3629, opened 2025-01-31
S3 Breaking Change - 3.7.412.0+ doesn't respect seekable streams and incorrectly sets Content-Length Header


niemyjski:
### Describe the bug

I've tracked it down to this was working in 3.7.411.7 but broken in any future release. Due to how you tag releases and different client versions it's pretty hard to track down...

```
System.Net.Http.HttpRequestException
Sent 298 request content bytes, but Content-Length promised 305.
   at System.Net.Http.HttpConnection.SendRequestContentAsync(HttpRequestMessage request, HttpContentWriteStream stream, Boolean async, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnection.SendRequestContentWithExpect100ContinueAsync(HttpRequestMessage request, Task`1 allowExpect100ToContinueTask, HttpContentWriteStream stream, Timer expect100Timer, Boolean async, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnection.SendAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.SendWithVersionDetectionAndRetryAsync(HttpRequestMessage request, Boolean async, Boolean doRequestAuth, CancellationToken cancellationToken)
   at System.Net.Http.HttpClient.<SendAsync>g__Core|83_0(HttpRequestMessage request, HttpCompletionOption completionOption, CancellationTokenSource cts, Boolean disposeCts, CancellationTokenSource pendingRequestsCts, CancellationToken originalCancellationToken)
   at Amazon.Runtime.HttpWebRequestMessage.GetResponseAsync(CancellationToken cancellationToken)
   at Amazon.Runtime.Internal.HttpHandler`1.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.Runtime.Internal.RedirectHandler.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.Runtime.Internal.Unmarshaller.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.S3.Internal.AmazonS3ResponseHandler.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.Runtime.Internal.ErrorHandler.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.Runtime.Internal.ErrorHandler.InvokeAsync[T](IExecutionContext executionContext)
[...]

PR #3768, opened 2025-04-22
Update HttpRequestMessageFactory to correctly set the Content-Length header

muhammad-othman:
## Description

This PR updates the .NET Standard implementation of the `HttpRequestMessageFactory` to correctly set the Content-Length header, particularly for non-seekable streams, by taking the current stream position into account.

The issue being addressed was originally fixed in #3631. However, due to the test coverage only verifying behavior when `UseChunkEncoding = false`, the regression was reintroduced when the position adjustment was moved behind a check that only applies to seekable streams.

This update ensures that the stream position is correctly accounted for when setting Content-Length, regardless of the stream's seekability.

## Motivation and Context
#3629

## Testing
 * `DRY_RUN-05ce4689-5754-4c4b-afd9-b0119f943459`.
 * Updated `TestResetStreamPosition` test to run with and without `ChunkEncoding`.

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_netstandard/HttpRequestMessageFactory.cs
+++ b/sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_netstandard/HttpRequestMessageFactory.cs
@@ -625,13 +625,12 @@ namespace Amazon.Runtime
             if ((isChunkedUploadWrapperStreamWithLength || isTrailingHeadersWrapperStreamWithLength || isCompressionWrapperStreamWithLength)
                 || (chunkedUploadWrapperStream == null && trailingHeadersWrapperStream == null && compressionWrapperStream == null))
             {
-                long position = 0;
+                long position;
                 try
                 {
-                    if (contentStream.CanSeek)
-                    {
-                        position = contentStream.Position;
-                    }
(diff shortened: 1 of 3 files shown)


niemyjski (approved):
Looks good, thanks for tracking this down.

sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_netstandard/HttpRequestMessageFactory.cs
```diff
@@ -625,13 +625,10 @@ public void WriteToRequestBody(HttpContent requestContent, Stream contentStream,
...
             {
-                long position = 0;
+                long position;
                 try
                 {
-                    if (contentStream.CanSeek)
-                    {
-                        position = contentStream.Position;
-                    }
+                    position = contentStream.Position;
```
normj, line 631: nit: Not relying on the `CanSeek` is due to special circumstances of our `ChunkedUploadWrapperStream`. You might want to add a comment here so somebody in the future, like myself, don't suggest adding back the CanSeek check to avoid the control flow by exception.

normj (approved):

normj (approved):

dscpinheiro (approved):

PR #3788, opened 2025-05-03
Update Microsoft.Extensions.AI version and improve chat client for AWS updates

stephentoub:
## Description

- Bumped version number to 9.4.3-preview.1.25230.7
- Renamed some methods and arguments to match other implementations
- Added support for TextReasoningContent
- Ensured underlying objects propagated out as RawRepresentation
- Added message and response IDs, along with setting additional metadata
- Propagate cached token counts to UsageDetails

## Motivation and Context

The Microsoft.Extensions.AI library gained some new features, e.g. TextReasoningContent, and Bedrock nows supports corresponding features, so update accordingly. Also keep up-to-date with the latest previews of M.E.AI, and update the some naming to match what other implementations are doing (e.g. AsChatClient => AsIChatClient).

## Testing

Unit tests plus some manual testing.

## Types of changes

## Checklist

## License

### DIFF
--- a/extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockEmbeddingGenerator.cs
+++ b/extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockEmbeddingGenerator.cs
@@ -39,17 +39,17 @@ internal sealed partial class BedrockEmbeddingGenerator : IEmbeddingGenerator<st
     /// Initializes a new instance of the <see cref="BedrockEmbeddingGenerator"/> class.
     /// </summary>
     /// <param name="runtime">The <see cref="IAmazonBedrockRuntime"/> instance to wrap.</param>
-    /// <param name="modelId">Model ID to use as the default when no model ID is specified in a request.</param>
-    /// <param name="dimensions">Number of dimensions to use when no number of dimensions is specified in a request.</param>
-    public BedrockEmbeddingGenerator(IAmazonBedrockRuntime runtime, string? modelId, int? dimensions)
+    /// <param name="defaultModelId">Model ID to use as the default when no model ID is specified in a request.</param>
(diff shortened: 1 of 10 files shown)


extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockEmbeddingGenerator.cs
```diff
@@ -116,7 +116,11 @@ public async Task<GeneratedEmbeddings<Embedding<float>>> GenerateAsync(
 
         if (totaltokens is not null)
         {
-            embeddings.Usage = new() { InputTokenCount = totaltokens.Value };
+            embeddings.Usage = new() 
+            {
+                InputTokenCount = totaltokens.Value,
```
dscpinheiro, line 121: I'm running an internal build, but just had a quick question: Is using the same value for `InputTokenCount` and `TotalTokenCount` intentional?

stephentoub: Thanks, @dscpinheiro.

> Is using the same value for InputTokenCount and TotalTokenCount intentional?

It is. Typically in embedding systems, there's no separate output token count or other tokens to be included in an overall total, so the total token count is just equal to the input token count. Is that not the case with these in bedrock? Happy to change it if my assumption is incorrect, I just couldn't find any other token counts in the output that would suggest otherwise.

dscpinheiro (approved):
Internal build succeeded as well (`DRY_RUN-7f9fb226-419a-44d1-8b4f-e05cb368eb49`). 

The commit I pushed was to bump the extensions version and fix an assert that was messing the `.\extensions\AWSSDK.Extensions.sln` locally only.

dscpinheiro:
This was released in https://www.nuget.org/packages/AWSSDK.Extensions.Bedrock.MEAI/4.0.0-preview.16

stephentoub:
Nice, thanks

PR #3811, opened 2025-05-13
Add tool for generating the DevConfig file

normj:
## Description
I got tired of manually creating DevConfig files or explaining to contributors how to create a DevConfig file so I thought I add a console tool build tools for helping generating the DevConfig file would be helpful. At least it would give a head start on the file and users could later further modify the generated file.

DevConfigGenerator


peterrsongg (approved):

buildtools/DevConfigGenerator/DevConfigGenerator/DevConfigFile.cs
```diff
@@ -0,0 +1,64 @@
...
+namespace DevConfigGenerator;
+
+public class DevConfigFile
+{
+    public CoreChange? Core { get; set; }
+
+    public List<Service> Services { get; } = new List<Service>();
+}
+
+public class CoreChange
```
boblodgett, line 12: It would be useful to have `backwardIncompatibilitiesToIgnore`  but can be in a future version.

buildtools/DevConfigGenerator/DevConfigGenerator/DevConfigWriter.cs
```diff
@@ -0,0 +1,59 @@
...
+        using Utf8JsonWriter writer = new Utf8JsonWriter(stream, new JsonWriterOptions { Indented = true });
+
+        writer.WriteStartObject();
+
+        if (config.Core != null)
+        {
+            writer.WriteStartObject("core");
+
+            writer.WriteBoolean("updateMinimum", config.Core.UpdateServiceMinimum);
+            writer.WriteString("type", config.Core.Type);
```
boblodgett, line 24: Should validate Type being either `minor` or `patch` in the future. Any other value isn't technically supported right now for basic DevConfigs.

buildtools/DevConfigGenerator/DevConfigGenerator/Program.cs
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    var updateServiceMinimum = AnsiConsole.Prompt(textPrompt);
+
+    devConfig.Core.UpdateServiceMinimum = updateServiceMinimum;
+}
+
+void GeneratorChange()
+{
+    AnsiConsole.Markup("Generator changes are saved as [green]Core changes[/] with update service minimum set to [green]true[/].\n");
```
boblodgett, line 119: A generator change doesn't always means a core update is needed. Usually it requires a change to a specific service unless it impacts many of them.

buildtools/DevConfigGenerator/DevConfigGenerator/Program.cs
```diff
@@ -0,0 +1,209 @@
...
+            new SelectionPrompt<string>()
+                .PageSize(10)
+                .AddChoices(topMatches));
+    }
+
+
+    var service = devConfig.Services.FirstOrDefault(x => x.Name == serviceName);
+    if (service == null)
+    {
+        service = new Service { Name = serviceName };
```
boblodgett, line 165: When would this case be valid? It seems like it would either be a mistake name entered that has no match or trying to specify a new service which would have a folder name. Looks like this could lead to invalid devconfigs.

boblodgett (approved):

buildtools/DevConfigGenerator/DevConfigGenerator/DevConfigFile.cs
normj: Agreed useful but I'll leave for a future version as we see how useful this tool is.

buildtools/DevConfigGenerator/DevConfigGenerator/Program.cs
normj: Agreed but in those cases a user doesn't have to choose the generator option and could just choose the affect service. I had this option because it felt like the safe way to handle generator changes and avoid people making generator changes that did affect services they didn't realize and not get released.

buildtools/DevConfigGenerator/DevConfigGenerator/Program.cs
normj: This is a check to see if this is the first time the service has been added to the dev config file. If I add a second change log entry for a service I already add I want the same service add to the list of change services.  The first time a change log entry is added for a service this will be null and added to the list of changed services.

This isn't adding a new service to our list of supported service or anything. I added a comment to make it more clear.

PR #3819, opened 2025-05-16
[V4] Add the ConnectTimeout property for .NET 8 target

normj:
## Description
In the .NET 8 target we have access to the [ConnectTimeout](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.socketshttphandler.connecttimeout). This PR exposes the ability to set the `ConnectTimeout` on the service client.

## Motivation and Context
This will help in scenario the SDK is trying to connect to a bad host and gets stuck trying to create the connection for a long period. If the user sets the connect timeout the `SendAsync` will timeout after the connect timeout has expired and since it is an IO exception the SDK will retry possibly to a different host.

## Testing
Running dry run
Manually confirmed by setting a low connect timeout that the SDK does do a retry when the connect timeout happens.

### DIFF
--- a/generator/.DevConfigs/A78BAE5D-9864-4E0F-853F-BBA0ACD3F60D.json
+++ b/generator/.DevConfigs/A78BAE5D-9864-4E0F-853F-BBA0ACD3F60D.json
@@ -6,7 +6,8 @@
             "Add the ConnectTimeout property on the service client config for the .NET 8 target of the SDK."
         ],
 		"backwardIncompatibilitiesToIgnore": [
-			"Amazon.Runtime.ClientConfig/MethodAdded"
+			"Amazon.Runtime.ClientConfig/MethodAdded",
+			"Amazon.Runtime.IClientConfig/MethodAbstractMethodAdded"
 		]
     }
 }
--- a/generator/.DevConfigs/A78BAE5D-9864-4E0F-853F-BBA0ACD3F60D.json
+++ b/generator/.DevConfigs/A78BAE5D-9864-4E0F-853F-BBA0ACD3F60D.json
@@ -1,0 +1,9 @@
+{
+    "core": {
+        "updateMinimum": true,
+        "type": "Patch",
+        "changeLogMessages": [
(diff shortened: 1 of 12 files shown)


sdk/src/Core/Amazon.Runtime/ClientConfig.cs
```diff
@@ -742,7 +742,31 @@ public TimeSpan? Timeout
...
+        /// Gets and sets the connection timeout that will be set on the HttpClient used by the service client to make requests.
+        /// The connection timeout is used control the wait time for the connection to be established to the service. The default
+        /// connection timeout for the HttpClient is infinite waiting period.
+        /// </summary>
+        public TimeSpan? ConnectTimeout
+        {
+            get
+            {
+                if (!this._connectTimeout.HasValue)
+                    return null;
```
normj, line 758: Ideally I would like to default this to `DefaultConfiguration.ConnectTimeout` instead of null but changing the default timeout would be a breaking change. Maybe when we do a 4.1 we can reconsider the default.

generator/.DevConfigs/A78BAE5D-9864-4E0F-853F-BBA0ACD3F60D.json
```diff
@@ -0,0 +1,12 @@
+{
+    "core": {
+        "updateMinimum": true,
+        "type": "Patch",
+        "changeLogMessages": [
+            "Add the ConnectTimeout property on the service client config for the .NET 8 target of the SDK."
+        ],
+		"backwardIncompatibilitiesToIgnore": [
```
GarrettBeatty, line 8: why is this a breaking change for this pr but not in v3?

ashishdhingra (approved):

generator/.DevConfigs/A78BAE5D-9864-4E0F-853F-BBA0ACD3F60D.json
normj: It is in V3 as well but we don't run the assembly comparer in V3 build system anymore

GarrettBeatty (approved):

PR #3820, opened 2025-05-16
[V3] Add the ConnectTimeout property on the service client config for the …

normj:
## Description
In the .NET 8 target we have access to the [ConnectTimeout](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.socketshttphandler.connecttimeout). This PR exposes the ability to set the `ConnectTimeout` on the service client.

In the [V4](https://github.com/aws/aws-sdk-net/pull/3819) version of this AWSSDK.Extensions.NETCore.Setup was also updated but in this V3 PR it was not. That is because in V3 extends from the `ClientConfig` so it automatically inherits the properties. In V4 we removed the inheritance so I need to copy the `ConnectTimeout` to AWSSDK.Extensions.NETCore.Setup and add the logic to copy the setting the service client config.

## Motivation and Context
This will help in scenario the SDK is trying to connect to a bad host and gets stuck trying to create the connection for a long period. If the user sets the connect timeout the `SendAsync` will timeout after the connect timeout has expired and since it is an IO exception the SDK will retry possibly to a different host.

## Testing
Dry run successful: DRY_RUN-394333cd-113a-4bbe-a5f0-a13ba692b64c
Manually confirmed by setting a low connect timeout that the SDK does do a retry when the connect timeout happens.

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/ClientConfig.cs
+++ b/sdk/src/Core/Amazon.Runtime/ClientConfig.cs
@@ -832,6 +832,31 @@ namespace Amazon.Runtime
             }
         }
 
+#if NET8_0_OR_GREATER
+        TimeSpan? _connectTimeout;
+
+        /// <summary>
+        /// Gets and sets the connection timeout that will be set on the HttpClient used by the service client to make requests.
+        /// The connection timeout is used control the wait time for the connection to be established to the service. The default
+        /// connection timeout for the HttpClient is infinite waiting period.
+        /// </summary>
+        public TimeSpan? ConnectTimeout
+        {
+            get
+            {
+                if (!this._connectTimeout.HasValue)
+                    return null;
+
+                return this._connectTimeout.Value;
+            }
+            set
+            {
+                ValidateTimeout(value);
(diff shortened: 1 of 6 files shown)


sdk/src/Core/Amazon.Runtime/ClientConfig.cs
```diff
@@ -832,6 +832,31 @@ public TimeSpan? Timeout
...
+        /// Gets and sets the connection timeout that will be set on the HttpClient used by the service client to make requests.
+        /// The connection timeout is used control the wait time for the connection to be established to the service. The default
+        /// connection timeout for the HttpClient is infinite waiting period.
+        /// </summary>
+        public TimeSpan? ConnectTimeout
+        {
+            get
+            {
+                if (!this._connectTimeout.HasValue)
+                    return null;
```
normj, line 848: Ideally I would like to default this to `DefaultConfiguration.ConnectTimeout` instead of null but changing the default timeout would be a breaking change. Maybe when we do a 4.1 we can reconsider the default.

GarrettBeatty (approved):

Issue #3807, opened 2025-05-12
IAmazonS3.EnsureBucketExistsAsync throws an exception


confect1on:
### Describe the bug

Hello. Sorry if it is not a bug, but this behavior is strange at least.
If the bucket with the specified name already exists, the method throws Amazon.S3.Model.BucketAlreadyOwnedByYouException
This was unexpected for me.
Documentation says: "Check to see if the bucket exists and if it doesn't create the bucket."
The implementation just puts the bucket : https://github.com/aws/aws-sdk-net/blob/8bf9e5d3d716683979fa373a6c13bf2c107a726b/sdk/src/Services/S3/Custom/_async/AmazonS3Client.Extensions.cs#L148-L151

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

The method does not throw an exception if the bucket already exists.

### Current Behavior

The method throws an exception if the bucket already exists.

### Reproduction Steps
```C#
using Amazon;
using Amazon.S3;

var testBucketName = "your-unique-bucket-name";

var configsS3 = new AmazonS3Config {
    ServiceURL = "http://localhost:9000",
    Profile = new Profile("local-minio-profile"),
    ForcePathStyle = true
};
using IAmazonS3 s3Client = new AmazonS3Client(configsS3);
await s3Client.EnsureBucketExistsAsync(testBucketName);
await s3Client.EnsureBucketExistsAsync(testBucketName);
Console.WriteLine("Bucket Exists");
```
### Possible Solution

I've searched a little and found a sample from aws extensions for dotnet CLI. This implementation looks more expected because it catches exceptions.
https://github.com/aws/aws-extensions-for-dotnet-cli/blob/master/src/Amazon.Common.DotNetCli.Tools/Utilities.cs

### Additional Information/Context

I'm using MinIO, but I'm not sure that MinIO can cause this because the code just calls the "PutBucketAsync" method.

### AWS .NET SDK and/or Package version used

AWSSDK.S3 4.0.0.3

### Targeted .NET Platform

.NET Core 9

### Operating System and version

Windows 10

ashishdhingra:
[...]

PR #3826, opened 2025-05-19
V3: Fixed an issue where IAmazonS3.EnsureBucketExists(Async) was throwing an exception if S3 bucket already exists in the executing account.

ashishdhingra:
## Description
Fixed an issue where `IAmazonS3.EnsureBucketExists`(`Async`) was throwing an exception if S3 bucket already exists in the executing account.

Kindly note below for 1st version:
- Refer [Using the handle method to filter inner exceptions](https://learn.microsoft.com/en-us/dotnet/standard/parallel-programming/exception-handling-task-parallel-library#using-the-handle-method-to-filter-inner-exceptions) on how to handle exceptions when using async `Task` based version.
- After handling exception for async version, we return `Task.FromResult(0)`. This is because:
  - If we return the original `Task` object, it re-throws exception in the calling awaited thread.
  - `Task.CompletedTask` doesn't exist in `NET45` target.

## Motivation and Context
Issue #3807 

## Testing
- Added integration test.
- Dry-run `DRY_RUN-32d10e46-52d0-421f-9bc7-943cdd6348cd` completed successfully.

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/_async/AmazonS3Client.Extensions.cs
+++ b/sdk/src/Services/S3/Custom/_async/AmazonS3Client.Extensions.cs
@@ -143,9 +143,15 @@ namespace Amazon.S3
             return this.PutACLAsync(request);
         }
 
-        Task ICoreAmazonS3.EnsureBucketExistsAsync(string bucketName)
+        async Task ICoreAmazonS3.EnsureBucketExistsAsync(string bucketName)
         {
-            return this.PutBucketAsync(bucketName);
+            try
+            {
+                await this.PutBucketAsync(bucketName).ConfigureAwait(false);
+            }
+            catch (BucketAlreadyOwnedByYouException)
+            {
+            }
         }
 
         [Obsolete("This method is deprecated: its behavior is inconsistent and always uses HTTP. Please use Amazon.S3.Util.AmazonS3Util.DoesS3BucketExistV2Async instead.")]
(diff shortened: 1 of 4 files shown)


sdk/src/Services/S3/Custom/_async/AmazonS3Client.Extensions.cs
```diff
@@ -145,7 +145,21 @@ Task ICoreAmazonS3.MakeObjectPublicAsync(string bucket, string objectKey, bool e
 
         Task ICoreAmazonS3.EnsureBucketExistsAsync(string bucketName)
         {
-            return this.PutBucketAsync(bucketName);
+            Task task = this.PutBucketAsync(bucketName);
```
normj, line 148: You should just use the async/await something like.

```
async Task ICoreAmazonS3.EnsureBucketExistsAsync(string bucketName)
{
   try
   {
        await this.PutBucketAsync(bucketName);
   }
   catch(BucketAlreadyOwnedByYouException)
   {
   }
}

normj (changes requested):

sdk/src/Services/S3/Custom/_async/AmazonS3Client.Extensions.cs
ashishdhingra: Fixed. It's working. Starting dry-run.

normj (approved):

afroz429 (approved):

Issue #3828, opened 2025-05-20
Unexpected behavior on DeleteObjects request


grsw92:
### Describe the bug

Hi,

I recently noticed rather inconsistent behavior while using `DeleteObjects` command. 

Example scenario: Sending `DeleteObjects` request where one of the files has invalid key (too long in my case). The client throws `AmazonS3Exception` where I believe the `DeleteObjectsException` should be thrown instead. Moreover the valid object keys listed in the request are not deleted. 

I haven't seen any remarks for this particular case in the [documentation](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html) to this method. 

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

When one of the keys is invalid the `DeleteObjectsException` should be thrown and the objects under valid keys should be deleted. 

### Current Behavior

When one of the keys is invalid the `AmazonS3Exception` is thrown and none of the objects is deleted.

### Reproduction Steps

See this little sample which reproduces the problem:
```csharp
// upload valid object
var content = Guid.NewGuid().ToString();
await s3Client.PutObjectAsync(new PutObjectRequest
{
    BucketName = bucket,
    Key = key,
    InputStream = new MemoryStream(Encoding.UTF8.GetBytes(content)),
    AutoCloseStream = true
}, cancellationToken);

// generate invalid key which should result KeyTooLongError
var invalidObjectKey = string.Join(string.Empty, Enumerable.Repeat("a", 1025));

// delete objects
var deleteObjectsRequest = new DeleteObjectsRequest { BucketName = bucket };

deleteObjectsRequest.AddKey(key);
deleteObjectsRequest.AddKey(invalidObjectKey);

try
{
    await s3Client.DeleteObjectsAsync(deleteObjectsRequest, cancellationToken);
}
catch (DeleteObjectsException e)
{
    Console.WriteLine("This is expected! {0}", e);
} 
catch (AmazonS3Exception e)
{
    Console.WriteLine("This is not expected! {0}", e);
}

// the valid file should be deleted
var response = await s3Client.ListObjectsAsync(bucket, key, cancellationToken);
Console.WriteLine(response.S3Objects.Any() ? "Object found. This is not expected" : "Object deleted. This is expected");
```

### AWS .NET SDK and/or Package version used

AWSSDK.S3 4.0.0.5
AWSSDK.S3 3.7.416.29

### Targeted .NET Platform

.NET 8

[...]

PR #3832, opened 2025-05-20
S3 DeleteObjects Key Validation Fix

AlexDaines:
## Description

Modified the S3 client's `DeleteObjects` operation to properly handle invalid keys in batch delete requests. Instead of failing the entire operation when any single key is invalid, the client now skips invalid keys and proceeds with deleting valid objects. If all keys in a request are invalid, a clear exception is thrown with an appropriate message.
## Motivation and Context

Addresses open issue #3828
---
This change is required because the current implementation prevents valid keys from being deleted when any key in the batch is invalid (e.g., too long). The S3 DeleteObjects API is designed to support partial successes, but the client-side validation was preventing this behavior by throwing an `AmazonS3Exception` before the request reached the service.

This fix allows the DeleteObjects operation to behave as expected, processing all valid keys even when some invalid keys are present in the same batch.
## Testing

The changes were tested with the following scenarios:
- Request with only valid keys (verified all objects deleted)
- Request with mix of valid and invalid keys (verified valid objects deleted)
- Request with all invalid keys (verified appropriate exception thrown)
## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/Model/DeleteObjectsRequest.Extensions.cs
+++ b/sdk/src/Services/S3/Custom/Model/DeleteObjectsRequest.Extensions.cs
@@ -1,4 +1,4 @@
-﻿/*
+/*
  * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
  * 
  * Licensed under the Apache License, Version 2.0 (the "License").
@@ -20,6 +20,7 @@ using System.IO;
 
 using Amazon.Runtime;
 using Amazon.Runtime.Internal;
+using Amazon.Runtime.Internal.Util;
 
 namespace Amazon.S3.Model
 {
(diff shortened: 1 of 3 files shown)


sdk/src/Services/S3/Custom/Model/DeleteObjectsRequest.Extensions.cs
```diff
@@ -30,13 +31,69 @@ namespace Amazon.S3.Model
...
+                });
+                return;
+            }
+            if (key.Length > 1024)
+            {
+                string message = string.Format("Skipping key in DeleteObjects request: key length {0} exceeds maximum of 1024 characters", key.Length);
+                Logger.InfoFormat(message);
+                
+                ValidationErrors.Add(new DeleteError {
+                    Key = key.Substring(0, Math.Min(key.Length, 128)) + (key.Length > 128 ? "..." : ""),
```
plachor, line 90: Min here is not truly required, also does it make sense to trim it it 128 when 1024 length is legit it can be issue when you have common prefix of 128 chars length

sdk/src/Services/S3/Custom/Model/DeleteObjectsRequest.Extensions.cs
```diff
@@ -30,13 +31,69 @@ namespace Amazon.S3.Model
...
+                Logger.InfoFormat(message);
+                
+                ValidationErrors.Add(new DeleteError {
+                    Key = key ?? "(null)",
+                    Code = "InvalidKey",
+                    Message = "Key cannot be null or empty"
+                });
+                return;
+            }
+            if (key.Length > 1024)
```
grsw92, line 84: @AlexDaines I understand this change just skips invalid keys and just sends valid ones with the request to the server? I'm wondering shouldn't this behavior be handled on the server side and proper `DeleteObjectsResponse` be returned from the server in that case?

AlexDaines:
Closing - 

After investigation, the real issue is that `DeleteObjectsResponseUnmarshaller` is missing the `UnmarshallException` override, causing `AmazonS3Exception` instead of `DeleteObjectsException`.

New PR (#3839) created with the proper exception handling fix instead of client-side validation.

PR #3833, opened 2025-05-21
Implement DynamoDbFlatten annotation

irina-herciu:
add new attribute and update write item flow

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/DynamoDBv2/Custom/DataModel/Attributes.cs
+++ b/sdk/src/Services/DynamoDBv2/Custom/DataModel/Attributes.cs
@@ -182,6 +182,35 @@ namespace Amazon.DynamoDBv2.DataModel
     }
 
     /// <summary>
+    /// Indicates that the properties of the decorated field or property type should be "flattened"
+    /// into the top-level attributes of the DynamoDB item. When applied, all public properties
+    /// of the referenced type are serialized as individual top-level attributes in the DynamoDB record,
+    /// rather than as a nested object.
+    /// <para>
+    /// Example:
+    /// <code>
+    /// public class Address
+    /// {
+    ///     public string Street { get; set; }
+    ///     public string City { get; set; }
+    /// }
+    ///
+    /// public class Person
+    /// {
+    ///     public string Name { get; set; }
+    ///     [DynamoDBFlatten]
+    ///     public Address Address { get; set; }
+    /// }
+    /// </code>
+    /// In this example, the <c>Person</c> table will have top-level attributes for <c>Name</c>, <c>Street</c>, and <c>City</c>.
+    /// </para>
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property, Inherited = true, AllowMultiple = false)]
+    public sealed class DynamoDBFlattenAttribute : DynamoDBAttribute
+    {
+    }
+
+    /// <summary>
     /// DynamoDB attribute that directs the specified attribute not to
     /// be included when saving or loading objects.
     /// </summary>
--- a/sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextInternal.cs
+++ b/sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextInternal.cs
@@ -390,29 +390,53 @@ namespace Amazon.DynamoDBv2.DataModel
 
             using (flatConfig.State.Track(document))
             {
-                foreach (PropertyStorage propertyStorage in storageConfig.AllPropertyStorage)
+                foreach (PropertyStorage propertyStorage in storageConfig.Properties)
                 {
-                    string propertyName = propertyStorage.PropertyName;
                     string attributeName = propertyStorage.AttributeName;
-
-                    DynamoDBEntry entry;
-                    if (document.TryGetValue(attributeName, out entry))
+                    if (propertyStorage.FlattenProperty)
                     {
-                        if (ShouldSave(entry, true))
+                        //create instance of the flatten property
+                        var targetType = propertyStorage.MemberType;
+                        object flattenedPropertyInstance = Utils.InstantiateConverter(targetType, this);
+
+                        //populate the flatten properties
+                        foreach (var propertyStorageFlattenProperty in propertyStorage.FlattenProperties)
                         {
-                            object value = FromDynamoDBEntry(propertyStorage, entry, flatConfig);
+                            string flattenedAttributeName = propertyStorageFlattenProperty.AttributeName;
 
-                            if (!TrySetValue(instance, propertyStorage.Member, value))
-                            {
-                                throw new InvalidOperationException("Unable to retrieve value from " + attributeName);
-                            }
+                            PopulateProperty(storage, flatConfig, document, flattenedAttributeName, propertyStorageFlattenProperty, flattenedPropertyInstance);
                         }
-
-                        if (propertyStorage.IsVersion)
-                            storage.CurrentVersion = entry as Primitive;
+                        if (!TrySetValue(instance, propertyStorage.Member, flattenedPropertyInstance))
+                        {
+                            throw new InvalidOperationException("Unable to retrieve value from " + attributeName);
+                        }
+                    }
+                    else
+                    {
+                        PopulateProperty(storage, flatConfig, document, attributeName, propertyStorage, instance);
                     }
                 }
             }
+        }
+
+        private void PopulateProperty(ItemStorage storage, DynamoDBFlatConfig flatConfig, Document document,
+            string attributeName, PropertyStorage propertyStorageFlattenProperty, object instance)
+        {
+            DynamoDBEntry entry;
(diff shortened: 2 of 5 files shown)


sdk/src/Services/DynamoDBv2/Custom/DataModel/Attributes.cs
```diff
@@ -181,6 +181,35 @@ public DynamoDBPolymorphicTypeAttribute(string typeDiscriminator,
         }
     }
 
+    /// <summary>
+    /// Indicates that the properties of the decorated field or property type should be "flattened"
+    /// into the top-level attributes of the DynamoDB item. When applied, all public properties
```
normj, line 186: Are they always flattened to "top-level attributes" or in the case of multi-nested structure the attributes are flattened to the parent structure.

sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextInternal.cs
```diff
@@ -390,29 +390,53 @@ private void PopulateInstance(ItemStorage storage, object instance, DynamoDBFlat
 
             using (flatConfig.State.Track(document))
             {
-                foreach (PropertyStorage propertyStorage in storageConfig.AllPropertyStorage)
+                foreach (PropertyStorage propertyStorage in storageConfig.Properties)
```
normj, line 393: Why the switch from `AllPropertyStorage` to `Properties`?

sdk/src/Services/DynamoDBv2/Custom/DataModel/Attributes.cs
```diff
@@ -181,6 +181,35 @@ public DynamoDBPolymorphicTypeAttribute(string typeDiscriminator,
...
+    ///     public string Name { get; set; }
+    ///     [DynamoDBFlatten]
+    ///     public Address Address { get; set; }
+    /// }
+    /// </code>
+    /// In this example, the <c>Person</c> table will have top-level attributes for <c>Name</c>, <c>Street</c>, and <c>City</c>.
+    /// </para>
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property, Inherited = true, AllowMultiple = false)]
+    public sealed class DynamoDBFlattenAttribute : DynamoDBAttribute
```
normj, line 209: What happens if a POCO has this attribute and the POCO is used in a collection property?

normj:
Is there an ask for this feature or is this parity feature with the Java SDK?

irina-herciu:
> Is there an ask for this feature or is this parity feature with the Java SDK?

this is for feature parity with Java SDK

sdk/src/Services/DynamoDBv2/Custom/DataModel/Attributes.cs
irina-herciu: It should only applies to single nested objects – not to collections, and this behavior is consistent with Java SDK. The only collection that will support flatten are Map<String, String>, but this is a new annotation @DynamoDbFlatten that is proposed to be added in the Java SDK.

sdk/src/Services/DynamoDBv2/Custom/DataModel/Attributes.cs
irina-herciu: The attributes are flattened to the parent structure

sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextInternal.cs
normj: Can you answer this question? I'm worried about the side effects of this change.

sdk/src/Services/DynamoDBv2/Custom/DataModel/InternalModel.cs
```diff
@@ -146,9 +146,17 @@ internal class PropertyStorage : SimplePropertyStorage
         // whether to store Type Discriminator for polymorphic serialization
         public bool PolymorphicProperty { get; set; }
 
+        // whether to store child properties at the same level as the parent property
+        public bool FlattenProperty { get; set; }
```
normj: The names of the properties `FlattenProperty` and `IsFlattened` sound to similar and not descriptive enough to tell the difference. What about `FlattenProperty` -> `ShouldFlattenChildProperties`?

I'm not sure why we didn't use `///` in this class before but maybe adding the `///` for these properties to help the team see the purpose in the intellisense might help.

sdk/src/Services/DynamoDBv2/Custom/DataModel/InternalModel.cs
```diff
@@ -941,6 +952,8 @@ private static void PopulateConfigFromType(ItemStorageConfig config, [Dynamicall
             }
         }
 
+        [UnconditionalSuppressMessage("ReflectionAnalysis", "IL2072",
+            Justification = "The user's type has been annotated with DynamicallyAccessedMemberTypes.All with the public API into the library. At this point the type will not be trimmed.")]
         private static PropertyStorage MemberInfoToPropertyStorage(ItemStorageConfig config, MemberInfo member)
```
normj, line 986: I think we need validate that the flatten child property names do not overlap with any names on the parent objects.

sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextInternal.cs
```diff
@@ -390,29 +390,53 @@ private void PopulateInstance(ItemStorage storage, object instance, DynamoDBFlat
...
+                    else
+                    {
+                        PopulateProperty(storage, flatConfig, document, attributeName, propertyStorage, instance);
+                    }
+                }
+            }
+        }
+
+        private void PopulateProperty(ItemStorage storage, DynamoDBFlatConfig flatConfig, Document document,
+            string attributeName, PropertyStorage propertyStorageFlattenProperty, object instance)
```
normj: Calling the parameter `propertyStorageFlattenProperty` doesn't seem right because in the `PopulateInstance` method if we go into the else block when the property is not flatten we are passing in the regular property storage and not a flatten property storage.

normj:
With your other PR's merged there are now conflicts with this PR that need to be addressed.

sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextInternal.cs
irina-herciu: the switch from AllPropertyStorage to Properties was done due to the fact that at Denormalize flattened child properties were added at parent level AllPropertyStorage while the Properties collection preserves the entity properties. 
While implementing the validation for flatten child property names to not overlap with any names on the parent objects I identified that AllPropertyStorage is not correctly populated and additional work is needed and I will reevaluate this change.

sdk/src/Services/DynamoDBv2/Custom/DataModel/InternalModel.cs
irina-herciu: renamed from FlattenProperty to ShouldFlattenChildProperties

sdk/src/Services/DynamoDBv2/Custom/DataModel/InternalModel.cs
irina-herciu: additional work needed to allow property names to overlap but with different attribute names.

sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextInternal.cs
irina-herciu: renamed

sdk/src/Services/DynamoDBv2/Custom/DataModel/InternalModel.cs
irina-herciu: I think AttributeNames should not overlap but the property names overlapping should work. Also at this moment there can be 2 or more properties with the same attributeName on same level and the value is overwritten inside PopulateItemStorage.

irina-herciu: @normj validation added for both flatten and non flatten properties for attribute names overlap the aligns with Java SDK behavior. [JavaTableSchema](https://github.com/aws/aws-sdk-java-v2/blob/master/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/StaticImmutableTableSchema.java#L166)

sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextInternal.cs
irina-herciu: reverted to AllPropertyStorage

normj:
PR looks good. I kicked off the internal validation build.

normj:
Dry run DRY_RUN-5611e6aa-4bfc-484e-a9d5-0e0c53946515 successful

GarrettBeatty (approved):

PR #3844, opened 2025-05-27
Remove Bouncy Castle dependency when targeting .NET 8 et al.

teo-tsirpanis:
## Description
This PR updates the CloudFront and EC2 extension libraries to import private keys from PEM with the relatively new `ImportFromPem` method when targeting .NET 8. This lets us remove the dependency to Bouncy Castle in this case.

## Motivation and Context
Removing third-party dependencies is always a good thing to do, especially when the dependency is a cryptographic library, and the replacement is trivial.

## Testing
Unit tests succeed locally.

## Types of changes

## Checklist

## License

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@
 
 **/*project.lock.json
 **/project.lock.json
+**/*.nuspec
 
 sdk/packages/**
 sdk/test/PCLTests/AndroidTests/packages/**


peterrsongg:
Thanks for the PR and for cleaning up our extension packages! Ran the tests myself, and it looks good. Do you mind doing a version bump to 4.0.0.1 for this?

You would have to update [this file](https://github.com/aws/aws-sdk-net/blob/main/extensions/src/AWSSDK.Extensions.EC2.DecryptPassword/AWSSDK.Extensions.EC2.DecryptPassword.nuspec#L6) for EC2 and [this file](https://github.com/aws/aws-sdk-net/blob/main/extensions/src/AWSSDK.Extensions.CloudFront.Signers/AWSSDK.Extensions.CloudFront.Signers.nuspec#L6) for cloudfront.

teo-tsirpanis:
@peterrsongg done.

.gitignore
```diff
@@ -18,7 +18,6 @@
 
 **/*project.lock.json
 **/project.lock.json
-**/*.nuspec
```
dscpinheiro: This doesn't look correct. The `nuspec` files are supposed to be ignored.

peterrsongg: ^^ yup this should still be kept here. We put that here b/c we want to ignore the generated nuspec files for all the service packages. For extensions, we update the nuspec and commit it manually.

teo-tsirpanis: Reverted.

peterrsongg:
Running it through our build system. if everything passes I'll hit approve

peterrsongg (approved):
dry run passed. approved

dscpinheiro (approved):

dscpinheiro:
Thanks for the contribution, this will be included in tomorrow's release.

PR #3847, opened 2025-05-30
[Merge 2nd]Generate AbortMultipartUpload, CreateBucketMetadataTableConfiguration, DeleteBucket from the model

peterrsongg:
## Description

This is the first PR of a larger effort to generate s3 fully in a backwards compatible manner. This PR includes updates to the generator so that the generator can fully run without erroring out for S3 and necessary adjustments to generate the first three operations: `AbortMultipartUpload`, `CreateBucketMetadataTableConfiguration`, and `DeleteBucket`. There is a list of operations that will be included in "Phase 1", which contains many commented out operations. As more PRs are created, the operations will be uncommented. 

I've added a folder in our generator `.tt` files which will most likely grow in size as we add more operations. Some of the marshallers have custom logic and since we cannot split a method across multiple files I had to add custom `.tt` files for specific operations which only trigger for that operation. (See DeleteBucketRequestCustomMarshallLogic.tt).

## Testing

DRY_RUN-f27a8b08-6008-4bbc-9756-3c984c5d34a0

I also checked using beyond compare that there were no differences (other than order of generated code). For those where there are differences I will leave a comment & explanation.

## Types of changes

## Checklist

## License


generator/ServiceClientGeneratorLib/GeneratorDriver.cs
```diff
@@ -207,10 +205,16 @@ public void Execute()
                 string.Format("ServiceEnumerations.{0}.cs", Configuration.ClassName) : "ServiceEnumerations.cs";
 
             // Any enumerations for the service
-            this.ExecuteGenerator(new ServiceEnumerations(), enumFileName);
+            // skip s3 until we're at the end of s3 client generation
+            if (this.Configuration.ServiceId != "S3")
```
peterrsongg, line 209: As the comment says, I'm skipping enumerations for now. None of these operations touch the enumerations and there is some shape renaming that must happen that we won't deal with in phase 1.

generator/ServiceClientGeneratorLib/GeneratorDriver.cs
```diff
@@ -207,10 +205,16 @@ public void Execute()
...
+            {
+                this.ExecuteGenerator(new ServiceEnumerations(), enumFileName);
+            }
+
           
             // Any paginators for the service
-            if (Configuration.ServiceModel.HasPaginators)
+            // skip paginators for s3 until we're at the end of s3 client generation
+            if (Configuration.ServiceModel.HasPaginators && Configuration.ServiceId != "S3")
```
peterrsongg, line 217: as the comment says, paginators will not be included in phase 1

generator/ServiceClientGeneratorLib/Generators/Marshallers/BaseMarshaller.tt
```diff
@@ -130,10 +130,11 @@ using Amazon.Runtime.Internal.Util;
 		{
             bool isGreedy = false;
             var marshallLocationName = operation.GetUriResourcePathTarget(member, out isGreedy);
-
             if (member.model.Customizations.SkipUriPropertyValidations.Contains(member.PropertyName))
             {
-				if(isGreedy)
+            if (this.Config.ServiceId == "S3" && marshallLocationName == "{Bucket}")
+                continue;
+		    if(isGreedy && this.Config.ServiceId != "S3")
```
peterrsongg, line 137: We were calling `TrimStart` on the resource path if it is greedy, however this was causing integ tests to fail. I did a test myself of putting an object with a leading slash and then calling `GetObjectACL` on the object. It failed with `KeyDoesn'tExist`. For now I skip this `TrimStart` logic only for S3, but we should take a closer look at this. With the way we split the resource path now on the `/` and with the way we removed the leading slash trimming logic in v4, we shouldn't be doing this anymore.

sdk/src/Services/S3/Generated/Model/CreateBucketMetadataTableConfigurationRequest.cs
```diff
@@ -0,0 +1,190 @@
...
+using System.IO;
+using System.Net;
+
+using Amazon.Runtime;
+using Amazon.Runtime.Internal;
+
+#pragma warning disable CS0612,CS0618,CS1570
+namespace Amazon.S3.Model
+{
+    /// <summary>
```
GarrettBeatty, line 32: just curious when you used the file diff too? it said there werent any changes even though the whitespace and everything changed for these comments?

generator/ServiceClientGeneratorLib/Shape.cs
```diff
@@ -764,10 +764,10 @@ public string DeprecationMessage
             {
                 string message = this.model.Customizations.GetShapeModifier(this._name)?.DeprecationMessage ??
                                  data[DeprecatedMessageKey].CastToString();
-                if (message == null)
+                if (this.model.ServiceId.Equals("S3") && message == null)
```
GarrettBeatty, line 767: why is this needed?

sdk/src/Services/S3/Generated/Model/CreateBucketMetadataTableConfigurationRequest.cs
peterrsongg: hmm not sure what you mean, are you saying that when you used the file diff tool it told you there weren't any changes?

generator/ServiceClientGeneratorLib/Generators/Marshallers/BaseMarshaller.tt
```diff
@@ -130,10 +130,11 @@ using Amazon.Runtime.Internal.Util;
 		{
             bool isGreedy = false;
             var marshallLocationName = operation.GetUriResourcePathTarget(member, out isGreedy);
-
             if (member.model.Customizations.SkipUriPropertyValidations.Contains(member.PropertyName))
             {
-				if(isGreedy)
+            if (this.Config.ServiceId == "S3" && marshallLocationName == "{Bucket}")
+                continue;
```
GarrettBeatty, line 136: can you fix the indentation on this, or use brackets instead of the one line if statement?  the reason being is i wasnt sure why we needed the second check for != sa3 on line 137, but thats a separate if statement and not nested. so maybe adding brackets/fix indentation will help

peterrsongg:
going to close this for now until the 1st one is approved and merged.

PR #3850, opened 2025-05-30
[MERGE 1st] Generate ListBuckets and ListDirectoryBuckets

peterrsongg:
## Description

This is part of phase 1 to generate S3. ListBuckets and ListDirectoryBuckets are being generated together because they both depend on the same shapes. 

This pr branched off of https://github.com/aws/aws-sdk-net/pull/3847 which was created first, which is why it contains some logic such as the deletebucket custom tt file. 

## Testing

**DRY_RUN-6de8c346-77da-432c-b8bf-93e40922ad56 (updated 06/12/2025)**

The assembly comparer output is below.

All of these are expected changes, the only one you could argue is a breaking change is the one where we return a nullable int instead of int for `MaxBuckets`. However the private variable is of type `int` and we just missed changing this from an `int` to an `int?` in the public property since s3 is handcoded. 

changing the parameter name to publicRequets in the marshaller is not a breaking change b/c it is in the internal namespace. The BucketUnmarshaller being renamed to S3Bucketnmarshaller is not a breaking change b/c it is in the internal namespace.

In terms of property renames being missed there are none.
```
C:\Dev\Repos\AwsSdkDotNetBuildUtilities\AssemblyComparer\AssemblyComparer\bin\Debug\net8.0 [mainline ≡]> ./AssemblyComparer.exe assembly-comparer --old-assembly "C:\AssemblyComparisons\AWSSDK.S3.Old.dll" --new-assembly "C:\AssemblyComparisons\AWSSDK.S3.New.dll"
AssemblyComparer AWSSDK.S3.New.dll: Error Amazon.S3.Model.Internal.MarshallTransformations.ListBucketsRequestMarshaller/ParameterNameChanged: Parameter name changed for method Amazon.Runtime.Internal.IRequest Marshall(Amazon.S3.Model.ListBucketsRequest) in Amazon.S3.Model.Internal.MarshallTransformations.ListBucketsRequestMarshaller from listBucketsRequest to publicRequest
AssemblyComparer AWSSDK.S3.New.dll: Message Amazon.S3.Model.Internal.MarshallTransformations.ListBucketsResponseUnmarshaller/MethodAdded: New method Amazon.Runtime.AmazonServiceException UnmarshallException(Amazon.Runtime.Internal.Transform.XmlUnmarshallerContext, System.Exception, System.Net.HttpStatusCode) in Amazon.S3.Model.Internal.MarshallTransformations.ListBucketsResponseUnmarshaller
AssemblyComparer AWSSDK.S3.New.dll: Error Amazon.S3.Model.ListBucketsRequest/MethodRemoved: Missing method System.Int32 get_MaxBuckets() in Amazon.S3.Model.ListBucketsRequest
AssemblyComparer AWSSDK.S3.New.dll: Message Amazon.S3.Model.ListBucketsRequest/MethodAdded: New method System.Nullable<System.Int32> get_MaxBuckets() in Amazon.S3.Model.ListBucketsRequest
AssemblyComparer AWSSDK.S3.New.dll: Error Amazon.S3.Model.ListBucketsRequest/MethodRemoved: Missing method System.Void set_MaxBuckets(System.Int32) in Amazon.S3.Model.ListBucketsRequest
AssemblyComparer AWSSDK.S3.New.dll: Message Amazon.S3.Model.ListBucketsRequest/MethodAdded: New method System.Void set_MaxBuckets(System.Nullable<System.Int32>) in Amazon.S3.Model.ListBucketsRequest
AssemblyComparer AWSSDK.S3.New.dll: Error Amazon.S3.Model.Internal.MarshallTransformations.BucketUnmarshaller/TypeRemoved: Missing type Amazon.S3.Model.Internal.MarshallTransformations.BucketUnmarshaller
AssemblyComparer AWSSDK.S3.New.dll: Message Amazon.S3.Model.Internal.MarshallTransformations.S3BucketUnmarshaller/TypeAdded: New type Amazon.S3.Model.Internal.MarshallTransformations.S3BucketUnmarshaller
AssemblyComparer AWSSDK.S3.New.dll: Message Amazon.S3.AmazonS3Request/TypeAdded: New type Amazon.S3.AmazonS3Request
```

## Types of changes

## Checklist

## License

### DIFF
--- a/generator/ServiceClientGeneratorLib/Generators/Marshallers/RestXmlRequestMarshaller.tt
+++ b/generator/ServiceClientGeneratorLib/Generators/Marshallers/RestXmlRequestMarshaller.tt
@@ -65,26 +65,9 @@ namespace <#=this.Config.Namespace #>.Model.Internal.MarshallTransformations
 	ProcessHeaderMembers("publicRequest", this.Operation.RequestHeaderMembers);		
 	ProcessUriMembers("publicRequest", this.Operation);	
 	ProcessQueryStringMembers("publicRequest", this.Operation);
-	if (this.Config.ServiceId == "S3")
-	{
-		// even though RequestUri contains the bucket name in s3. In reality, the bucket Name
-		// is part of the endpoint so we remove it.
-		string resourcePath = string.Empty;
-		if (this.Operation.RequestUri.StartsWith("/{Bucket}"))
-			resourcePath = this.Operation.RequestUri.Substring(9).Length == 0 ? "/" :  this.Operation.RequestUri.Substring(9);
-		else
-			resourcePath = this.Operation.RequestUri;
-#>
-			request.ResourcePath = "<#=resourcePath#>";
-<#
-	}
-	else
-	{
 #>
 			request.ResourcePath = "<#=this.Operation.RequestUri #>";
-
 <#   
-	}
 	// Process any members which are marshalled as part of the request body
 	if(this.Operation.RequestHasBodyMembers || shouldMarshallPayload)
 	{
--- a/generator/ServiceClientGeneratorLib/Operation.cs
+++ b/generator/ServiceClientGeneratorLib/Operation.cs
@@ -302,30 +302,9 @@ namespace ServiceClientGenerator
                 if (this.RequestStructure != null)
                 {
                     var payload = this.RequestStructure.PayloadMemberName;
-                    // check to see if the payload member has been overridden by another member defined in the service customizations
-                    // file. For example, the payload member could've been included in the "exclude" array and another member
-                    // could've been injected. 
-                    if (this.model.Customizations.ShapeModifiers.ContainsKey(this.RequestStructure.Name))
-                    {
-                        var customization = this.model.Customizations.ShapeModifiers[this.RequestStructure.Name];
-                        if (customization.IsExcludedProperty(payload))
-                        {
-                            foreach (var injectedProperty in customization.InjectedPropertyNames)
-                            {
-                                var propertyInjector = customization.InjectedPropertyData(injectedProperty);
-                                if (string.Equals(propertyInjector.Data[CustomizationsModel.OriginalMemberKey].ToString(), payload, StringComparison.OrdinalIgnoreCase))
-                                {
-                                    payload = injectedProperty;
-                                    break;
-                                }
-                                    
-                            }
-                        }
-                        
-                    }
                     if (!string.IsNullOrWhiteSpace(payload))
                     {
-                        return this.RequestStructure.Members.Single(m => 
+                        return this.RequestStructure.Members.Single(m =>
                             (m.HasModifier && m.MarshallName.Equals(payload, StringComparison.InvariantCultureIgnoreCase))
                             || (!m.HasModifier && m.ModeledName.Equals(payload, StringComparison.InvariantCultureIgnoreCase)));
                     }
--- a/generator/ServiceClientGeneratorLib/Generators/Marshallers/RestXmlRequestMarshaller.cs
+++ b/generator/ServiceClientGeneratorLib/Generators/Marshallers/RestXmlRequestMarshaller.cs
@@ -148,49 +148,21 @@ namespace ServiceClientGenerator.Generators.Marshallers
 	ProcessHeaderMembers("publicRequest", this.Operation.RequestHeaderMembers);		
 	ProcessUriMembers("publicRequest", this.Operation);	
 	ProcessQueryStringMembers("publicRequest", this.Operation);
-	if (this.Config.ServiceId == "S3")
-	{
-		// even though RequestUri contains the bucket name in s3. In reality, the bucket Name
-		// is part of the endpoint so we remove it.
-		string resourcePath = string.Empty;
-		if (this.Operation.RequestUri.StartsWith("/{Bucket}"))
-			resourcePath = this.Operation.RequestUri.Substring(9).Length == 0 ? "/" :  this.Operation.RequestUri.Substring(9);
-		else
-			resourcePath = this.Operation.RequestUri;
 
             
             #line default
             #line hidden
             this.Write("\t\t\trequest.ResourcePath = \"");
             
-            #line 78 "C:\Dev\Repos\aws-sdk-net-staging\generator\ServiceClientGeneratorLib\Generators\Marshallers\RestXmlRequestMarshaller.tt"
-            this.Write(this.ToStringHelper.ToStringWithCulture(resourcePath));
+            #line 69 "C:\Dev\Repos\aws-sdk-net-staging\generator\ServiceClientGeneratorLib\Generators\Marshallers\RestXmlRequestMarshaller.tt"
+            this.Write(this.ToStringHelper.ToStringWithCulture(this.Operation.RequestUri));
             
             #line default
             #line hidden
             this.Write("\";\r\n");
             
-            #line 79 "C:\Dev\Repos\aws-sdk-net-staging\generator\ServiceClientGeneratorLib\Generators\Marshallers\RestXmlRequestMarshaller.tt"
-
-	}
-	else
-	{
-
-            
-            #line default
-            #line hidden
-            this.Write("\t\t\trequest.ResourcePath = \"");
-            
-            #line 84 "C:\Dev\Repos\aws-sdk-net-staging\generator\ServiceClientGeneratorLib\Generators\Marshallers\RestXmlRequestMarshaller.tt"
-            this.Write(this.ToStringHelper.ToStringWithCulture(this.Operation.RequestUri));
-            
-            #line default
-            #line hidden
-            this.Write("\";\r\n\r\n");
-            
-            #line 86 "C:\Dev\Repos\aws-sdk-net-staging\generator\ServiceClientGeneratorLib\Generators\Marshallers\RestXmlRequestMarshaller.tt"
+            #line 70 "C:\Dev\Repos\aws-sdk-net-staging\generator\ServiceClientGeneratorLib\Generators\Marshallers\RestXmlRequestMarshaller.tt"
    
-	}
 	// Process any members which are marshalled as part of the request body
 	if(this.Operation.RequestHasBodyMembers || shouldMarshallPayload)
 	{
@@ -206,7 +178,7 @@ namespace ServiceClientGenerator.Generators.Marshallers
             #line hidden
             this.Write("\r\n");
             
-            #line 99 "C:\Dev\Repos\aws-sdk-net-staging\generator\ServiceClientGeneratorLib\Generators\Marshallers\RestXmlRequestMarshaller.tt"
+            #line 82 "C:\Dev\Repos\aws-sdk-net-staging\generator\ServiceClientGeneratorLib\Generators\Marshallers\RestXmlRequestMarshaller.tt"
 
 	if (this.Operation.UnsignedPayload)
 	{
@@ -216,7 +188,7 @@ namespace ServiceClientGenerator.Generators.Marshallers
             #line hidden
             this.Write("\t\t\trequest.DisablePayloadSigning = true;\r\n");
             
-            #line 104 "C:\Dev\Repos\aws-sdk-net-staging\generator\ServiceClientGeneratorLib\Generators\Marshallers\RestXmlRequestMarshaller.tt"
+            #line 87 "C:\Dev\Repos\aws-sdk-net-staging\generator\ServiceClientGeneratorLib\Generators\Marshallers\RestXmlRequestMarshaller.tt"
 	}
 
 	// If there aren't any members that are marshalled as part of the body or streamed	
@@ -228,7 +200,7 @@ namespace ServiceClientGenerator.Generators.Marshallers
             #line hidden
             this.Write("\t\t\trequest.UseQueryString = true;\r\n");
             
-            #line 111 "C:\Dev\Repos\aws-sdk-net-staging\generator\ServiceClientGeneratorLib\Generators\Marshallers\RestXmlRequestMarshaller.tt"
+            #line 94 "C:\Dev\Repos\aws-sdk-net-staging\generator\ServiceClientGeneratorLib\Generators\Marshallers\RestXmlRequestMarshaller.tt"
 	}	
 
 	// We skip endpoint host prefix handling for S3 and S3 Control as it's implemented by endpoint rules.
@@ -242,7 +214,7 @@ namespace ServiceClientGenerator.Generators.Marshallers
             #line hidden
             this.Write("\r\n\t\t\treturn request;\r\n\t\t}\r\n");
             
-            #line 122 "C:\Dev\Repos\aws-sdk-net-staging\generator\ServiceClientGeneratorLib\Generators\Marshallers\RestXmlRequestMarshaller.tt"
+            #line 105 "C:\Dev\Repos\aws-sdk-net-staging\generator\ServiceClientGeneratorLib\Generators\Marshallers\RestXmlRequestMarshaller.tt"
 				
 	this.AddRequestSingletonMethod();
 
@@ -253,7 +225,7 @@ namespace ServiceClientGenerator.Generators.Marshallers
             return this.GenerationEnvironment.ToString();
         }
         
-        #line 128 "C:\Dev\Repos\aws-sdk-net-staging\generator\ServiceClientGeneratorLib\Generators\Marshallers\RestXmlRequestMarshaller.tt"
+        #line 111 "C:\Dev\Repos\aws-sdk-net-staging\generator\ServiceClientGeneratorLib\Generators\Marshallers\RestXmlRequestMarshaller.tt"
 
 
 	void ProcessRequestBodyMembers(string variableName, Operation operation)
@@ -264,7 +236,7 @@ namespace ServiceClientGenerator.Generators.Marshallers
         #line default
         #line hidden
         
-        #line 133 "C:\Dev\Repos\aws-sdk-net-staging\generator\ServiceClientGeneratorLib\Generators\Marshallers\RestXmlRequestMarshaller.tt"
+        #line 116 "C:\Dev\Repos\aws-sdk-net-staging\generator\ServiceClientGeneratorLib\Generators\Marshallers\RestXmlRequestMarshaller.tt"
 this.Write(@"			var stringWriter = new XMLEncodedStringWriter(CultureInfo.InvariantCulture);
 			using (var xmlWriter = XmlWriter.Create(stringWriter, new XmlWriterSettings() { Encoding = System.Text.Encoding.UTF8, OmitXmlDeclaration = true, NewLineHandling = NewLineHandling.Entitize }))
 			{   
@@ -274,7 +246,7 @@ this.Write(@"			var stringWriter = new XMLEncodedStringWriter(CultureInfo.Invari
         #line default
         #line hidden
         
-        #line 137 "C:\Dev\Repos\aws-sdk-net-staging\generator\ServiceClientGeneratorLib\Generators\Marshallers\RestXmlRequestMarshaller.tt"
+        #line 120 "C:\Dev\Repos\aws-sdk-net-staging\generator\ServiceClientGeneratorLib\Generators\Marshallers\RestXmlRequestMarshaller.tt"
 
 		if(operation.RequestPayloadMember==null)
 		{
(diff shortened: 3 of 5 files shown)


generator/ServiceClientGeneratorLib/Generators/S3/DeleteBucketRequestCustomMarshallLogic.tt
```diff
@@ -0,0 +1,24 @@
+﻿<#@ template language="C#" #>
+<#@ assembly name="System.Core" #>
+<#@ import namespace="System.Linq" #>
+<#@ import namespace="System.Text" #>
+<#@ import namespace="System.Collections.Generic" #>
+            if (publicRequest.BucketRegion != null)
+            {
+                RegionEndpoint regionEndpoint;
+#pragma warning disable CR1004
```
peterrsongg, line 9: as the description says, since i branched from https://github.com/aws/aws-sdk-net/pull/3847 this file is included.

sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/S3BucketUnmarshaller.cs
```diff
@@ -12,26 +12,40 @@
...
 namespace Amazon.S3.Model.Internal.MarshallTransformations
 {
-     /// <summary>
-     ///   Bucket Unmarshaller
-     /// </summary>
-    public class BucketUnmarshaller : IXmlUnmarshaller<S3Bucket, XmlUnmarshallerContext>
+    /// <summary>
+    /// Response Unmarshaller for S3Bucket Object
+    /// </summary>  
+    public class S3BucketUnmarshaller : IXmlUnmarshaller<S3Bucket, XmlUnmarshallerContext>
```
peterrsongg, line 39: Even though we rename the shape from Bucket to S3Bucket, in the custom code the unmarshaller was still called BucketUnmarshaller. Since this is part of the internal namespace, i figured the rename to S3BucketUnmarshaller was safe.

generator/ServiceClientGeneratorLib/ServiceModel.cs
```diff
@@ -460,6 +460,114 @@ public List<Operation> Operations
...
+                        //new Operation(this, "SelectObjectContent", DocumentRoot[OperationsKey]["SelectObjectContent"])
+                    };
+                }
+                return _s3AllowListOperations;
+            }
+        }
+
+        private List<string> _s3AddParametersList;
+        /// <summary>
+        /// Some s3 operations add query string parameters to the Subresources collection instead of the Parameters collection, but we are inconsistent with it across the board.
```
peterrsongg, line 532: this is the change i didn't like making but in some cases we add a query param as a subresource and other times we add as a parameter. Even though they are used in the same way in signing, we don't know how customers might be using the Parameters collection in the wild so I added this so we can keep it the same.

README.md
dscpinheiro, line 1: What happened here? 🤔

peterrsongg: oops :) removed

generator/ServiceClientGeneratorLib/GeneratorDriver.cs
```diff
@@ -227,24 +231,43 @@ public void Execute()
...
             // We use the base exceptions generated for the parent model.
-            if (!this.Configuration.IsChildConfig)
+            if (!this.Configuration.IsChildConfig && this.Configuration.ServiceId != "S3")
             {
                 this.ExecuteGenerator(new BaseServiceException(), "Amazon" + this.Configuration.ClassName + "Exception.cs");
             }
 
-            // Generates the Request, Response, Marshaller, Unmarshaller, and Exception objects for a given client operation
-            foreach (var operation in Configuration.ServiceModel.Operations)
+            if (Configuration.Namespace == "Amazon.S3")
```
normj: Since the only thing different between the `if` and `else` block is what collection we do the foreach can we simplify the logic to something like this:
```csharp
var operations = Configuration.Namespace == "Amazon.S3" ? Configuration.ServiceModel.S3AllowListOperations : Configuration.ServiceModel.Operations;
foreach (var operation in operations)
{
    GenerateRequest(operation);
    GenerateResponse(operation);
    GenerateRequestMarshaller(operation);
    GenerateResponseUnmarshaller(operation);
    GenerateEndpointDiscoveryMarshaller(operation);
    GenerateExceptions(operation);
    GenerateStructures(operation);
    GenerateEventStreamPublisher(operation);
}
```

generator/ServiceClientGeneratorLib/GeneratorDriver.cs
```diff
@@ -700,7 +728,14 @@ private void GenerateUnmarshaller(Shape shape)
 
                 if (this.Configuration.ServiceModel.Customizations.IsSubstitutedShape(nestedStructure.Name))
                     continue;
-
+                if (this.Configuration.ServiceId == "S3")
+                {
+                    if (shape.IsEvent || shape.IsEventStream)
+                    {
+                        this._processedUnmarshallers.Add(nestedStructure.Name);
```
normj: Since we are skipping these unmarshallers why do we need to put them in the `_processedUnmarshallers` collection. The skip checks before aren't doing that.

generator/ServiceClientGeneratorLib/GeneratorDriver.cs
```diff
@@ -1002,6 +1036,14 @@ void GenerateStructures(Operation operation)
                 if (definition.IsEventStream && !Configuration.ServiceModel.Operations.Any(x => string.Equals(x.ResponseEventStreamingMember?.Shape.Name, definition.Name)))
                     continue;
 
+                if (this.Configuration.ServiceId == "S3")
+                {
+                    if (definition.IsEvent || definition.IsEventStream)
+                    {
+                        this._processedStructures.Add(definition.Name);
```
normj: Since we are skipping these unmarshallers why do we need to put them in the _processedStructures collection. The skip checks before aren't doing that.

generator/ServiceClientGeneratorLib/ServiceModel.cs
```diff
@@ -460,6 +460,114 @@ public List<Operation> Operations
...
+            }
+        }
+
+
+        private List<string> _s3RequestMarshallerThrowGenericExceptionList;
+        /// <summary>
+        /// The names of s3 operations where we throw AmazonS3Exception instead of System.ArgumentException
+        /// when required parameters aren't set.
+        /// </summary>
+        public List<string> S3RequestMarshallerThrowGenericExceptionList
```
normj, line 561: This property needs a better name to make it clearer it's purposes. The comment block should also have blurp about why we have this so years from now we know why we have this.

generator/ServiceClientGeneratorLib/Generators/Marshallers/BaseMarshaller.tt
```diff
@@ -149,11 +150,23 @@ using Amazon.Runtime.Internal.Util;
...
+                {
 #>
             if (!<#=variableName#>.IsSet<#=member.PropertyName#>())
                 throw new <#=this.Config.BaseException#>("Request object does not have required field <#=member.PropertyName#> set");
 <#+
-			if(isGreedy)
+                }
+            if (this.Config.ServiceId == "S3" && marshallLocationName == "{Bucket}")
+                continue;
+			if(isGreedy && this.Config.ServiceId != "S3")
```
normj, line 181: Add a comment why S3 doesn't need the same `isGreedy` logic.

generator/ServiceClientGeneratorLib/Generators/Marshallers/BaseMarshaller.tt
```diff
@@ -233,9 +246,18 @@ using Amazon.Runtime.Internal.Util;
 <#+
                 if (member.CustomMarshallerTransformation != null)
                 {
+                    if (this.Config.ServiceId == "S3")
```
normj, line 264: Can you add a comment why S3 has to be different here? I want to make sure readers of TT file in a few years understand why we have these conditions.

generator/ServiceClientGeneratorLib/Generators/Marshallers/BaseMarshaller.tt
```diff
@@ -261,17 +283,35 @@ using Amazon.Runtime.Internal.Util;
                 }
                 else
                 {
+                    if (this.Config.ServiceId != "S3" || this.Config.ServiceModel.S3AddParametersList.Contains(operation.Name))
```
normj: To be consistent with the previous ordering `request.AddSubResource` and `request.Parameters.Add` you can write the `if` condition for it being S3 and then add a comment why S3 has different behavior.

generator/ServiceClientGeneratorLib/Generators/Marshallers/RestXmlRequestMarshaller.tt
```diff
@@ -99,6 +115,20 @@ namespace <#=this.Config.Namespace #>.Model.Internal.MarshallTransformations
 	{
 		ProcessEndpointHostPrefixMembers(3, "publicRequest", this.Operation);
 	}
+#>
+<#
+	if (this.Config.ServiceId == "S3")
```
normj: Since `DeleteBucket` is not being generated you might want to remove this from this PR.

I don't like the pattern of having to add `tt` file for a specific operation special logic. I would rather we add a mechanism where we add a new partial in the `Custom` folder. In this case CustomModel\Internal\MarshallTransformations\DeleteBucketMarshaller.cs` with a method called `CustomizeMarshaller` and then add a flag in the `s3.customization.json` file that tells the generator to add the code to invoke the method. 

That way we keep one off complexity out of the generator. The mechanism is more reusable and we can have a combination of generated code and custom code.

generator/ServiceClientGeneratorLib/Generators/S3/DeleteBucketRequestCustomMarshallLogic.tt
```diff
@@ -0,0 +1,24 @@
+﻿<#@ template language="C#" #>
```
normj: Copied from similar comment above

---

Since `DeleteBucket` is not being generated you might want to remove this from this PR.

I don't like the pattern of having to add `tt` file for a specific operation special logic. I would rather we add a mechanism where we add a new partial in the `Custom` folder. In this case CustomModel\Internal\MarshallTransformations\DeleteBucketMarshaller.cs` with a method called `CustomizeMarshaller` and then add a flag in the `s3.customization.json` file that tells the generator to add the code to invoke the method. 

That way we keep one off complexity out of the generator. The mechanism is more reusable and we can have a combination of generated code and custom code.

generator/ServiceClientGeneratorLib/Member.cs
```diff
@@ -972,10 +992,10 @@ public string DeprecationMessage
             {
                 string message = this.model.Customizations.GetPropertyModifier(this.OwningShape.Name, this._name)?.DeprecationMessage ??
                                  this.data[DeprecatedMessageKey].CastToString();
-                if (message == null)
+                if (!this.model.ServiceId.Equals("S3") && message == null)
```
normj: Why is S3 being excluded from this check. If we are missing deprecated messages would should add them to our customization file.

generator/ServiceClientGeneratorLib/Operation.cs
```diff
@@ -111,7 +111,7 @@ public string DeprecationMessage
             {
                 string message = this.model.Customizations.GetOperationModifiers(this.name)?.DeprecatedMessage ??
                                  data[ServiceModel.DeprecatedMessageKey].CastToString();
-                if (message == null)
+                if (!string.Equals(this.model.ServiceId, "s3", StringComparison.OrdinalIgnoreCase) &&  message == null)
```
normj: Why is S3 being excluded from this check. If we are missing deprecated messages would should add them to our customization file.

generator/ServiceClientGeneratorLib/Operation.cs
```diff
@@ -301,6 +301,25 @@ public Member RequestPayloadMember
                 if (this.RequestStructure != null)
                 {
                     var payload = this.RequestStructure.PayloadMemberName;
+                    // check if PayloadMemberName was removed and something else was injected
```
normj: Add more context to this comment. As it is I don't know what it is telling me till I full understand the following block to get the context, which at that point the comment is meaning less. I assume it should be telling something like:

```
Check to see if the payload member has been overridden by another member defined in the service customization file.
```

generator/ServiceClientGeneratorLib/ServiceModel.cs
normj: Judging by the comment in the SDK codebase for `SubResources` this is used to handle a query string parameter could have a `?` as part of its value like S3 object key. Are you sure this is safe to do at the operation level?

generator/ServiceClientGeneratorLib/Shape.cs
```diff
@@ -764,10 +764,10 @@ public string DeprecationMessage
             {
                 string message = this.model.Customizations.GetShapeModifier(this._name)?.DeprecationMessage ??
                                  data[DeprecatedMessageKey].CastToString();
-                if (message == null)
+                if (this.model.ServiceId.Equals("S3") && message == null)
```
normj: Similar comment as above we shouldn't be excluding S3 from the missing message validation.

normj (changes requested):

generator/ServiceClientGeneratorLib/GeneratorDriver.cs
peterrsongg: I like that much better. done.

generator/ServiceClientGeneratorLib/GeneratorDriver.cs
peterrsongg: I kind of put this here just in case any other methods were accessing the collection, but you're right it's probably not needed.

generator/ServiceClientGeneratorLib/ServiceModel.cs
peterrsongg: Updated the name and comment

generator/ServiceClientGeneratorLib/Generators/Marshallers/BaseMarshaller.tt
peterrsongg: I don't think it is just an S3 thing, I think we could remove the `TrimStart` for all services that utilize the greedy path, but I only tested for S3 so I only did it for S3. I also wanted to keep the scope limited to S3.

I tested with `GetObjectAcl` for example since that uses the `TrimStart` and it couldn't get an object with a leading slash. The reason I don't think we need the `TrimStart` anymore is b/c we correctly split the resource path into segments based on the `/` here: https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Core/Amazon.Util/AWSSDKUtils.cs#L424-L428 if it is greedy for signing, and then join it back together as to not encode the `/`. My guess is the `TrimStart` was added before the path splitting was fixed up?

normj: If you want to limit this to only S3 for now to avoid a potential blast radius that is fine but leave a comment explain the decision to limit the scope of this change for now and a TODO for the future to see we can remove the special case.

generator/ServiceClientGeneratorLib/ServiceModel.cs
peterrsongg: This is how the SDK is today, and I didn't want to change it so I left it as is. We use a mixture of Parameter and SubResources so this list will just act as a way to generate some as adding to the parameters and some as adding to SubResources

generator/ServiceClientGeneratorLib/Generators/Marshallers/BaseMarshaller.tt
peterrsongg: Added the comment and TODO

normj (approved):

generator/ServiceClientGeneratorLib/GeneratorDriver.cs
```diff
@@ -244,6 +249,8 @@ public void Execute()
                 GenerateStructures(operation);
                 GenerateEventStreamPublisher(operation);
             }
+                // Generates the Request, Response, Marshaller, Unmarshaller, and Exception objects for a given client operation
```
dscpinheiro: Was this comment moved on purpose? It feels like it should be before `var operations = ...` like before.

generator/ServiceClientGeneratorLib/GeneratorDriver.cs
```diff
@@ -259,16 +266,19 @@ public void Execute()
                 var fileName = string.Format("{0}EndpointDiscoveryMarshallingTests.cs", Configuration.ClassName);
                 ExecuteTestGenerator(new EndpointDiscoveryMarshallingTests(), fileName, "Marshalling");
             }
-
-            // Test that simple customizations were generated correctly
-            GenerateCustomizationTests();
             ExecuteProjectFileGenerators();
-            if (this.Configuration.ServiceModel.Customizations.HasExamples)
+            // Test that simple customizations were generated correctly
+            if (this.Configuration.ServiceId != "S3")
```
dscpinheiro, line 269: Does this mean we're not generating examples for S3 anymore? There a few of them today: https://github.com/aws/aws-sdk-net/tree/main/docgenerator/AWSSDKDocSamples/S3 (I thought that was the reason you updated the `Example.cs` file not to fail)

generator/ServiceClientGeneratorLib/Generators/Marshallers/RestXmlRequestMarshaller.tt
```diff
@@ -65,10 +65,26 @@ namespace <#=this.Config.Namespace #>.Model.Internal.MarshallTransformations
 	ProcessHeaderMembers("publicRequest", this.Operation.RequestHeaderMembers);		
 	ProcessUriMembers("publicRequest", this.Operation);	
 	ProcessQueryStringMembers("publicRequest", this.Operation);
+	if (this.Config.ServiceId == "S3")
+	{
+		// even though RequestUri contains the bucket name in s3. In reality, the bucket Name
```
dscpinheiro: I don't follow this change... Where did the logic to check for `/{Bucket}` come from? If it's not related to ListBuckets I wonder if it should be left for the other PRs.

generator/ServiceClientGeneratorLib/Member.cs
```diff
@@ -972,10 +992,11 @@ public string DeprecationMessage
             {
                 string message = this.model.Customizations.GetPropertyModifier(this.OwningShape.Name, this._name)?.DeprecationMessage ??
                                  this.data[DeprecatedMessageKey].CastToString();
+                // TODO: Fill in s3 deprecation messages
```
dscpinheiro, line 995: You resolved Norm's comment without a response (https://github.com/aws/aws-sdk-net/pull/3850#discussion_r2127588600) but did we decide not to fill in the deprecation messages for now?

generator/ServiceClientGeneratorLib/Operation.cs
```diff
@@ -301,6 +302,27 @@ public Member RequestPayloadMember
                 if (this.RequestStructure != null)
                 {
                     var payload = this.RequestStructure.PayloadMemberName;
+                    // check to see if the payload member has been overridden by another member defined in the service customizations
```
dscpinheiro: Is this required for ListBuckets? If not, same feedback as the other comment (suggesting leaving for future PRs).

generator/ServiceClientGeneratorLib/Generators/Marshallers/RestXmlRequestMarshaller.tt
peterrsongg: If you look at the request marshallers today, Bucket is not part of the resource path. This is because s3 is unique in that even though the bucket is modeled as part of the resource path, it actually isn't appended to the uri since the bucket is part of the domain name. If I don't include this, we will start sending requests to a different uri.

generator/ServiceClientGeneratorLib/Member.cs
peterrsongg: The 2 operations we are generating right now do not have any deprecated things for me to fill in. I was going to fill them in as those get generated.

generator/ServiceClientGeneratorLib/Operation.cs
peterrsongg: It is not needed for `ListBuckets` but I'd rather leave it in since I know this will be used in future operations. This is something I discovered when trying to run the generator for all operations.

generator/ServiceClientGeneratorLib/GeneratorDriver.cs
peterrsongg: This I wasn't sure on either, technically I think we could start generating examples but I didn't want to blow up this PR. 

However, I think at the moment we aren't generating examples through the `GeneratorDriver` class at least
https://github.com/aws/aws-sdk-net/blob/main/generator/ServiceClientGeneratorLib/GeneratorDriver.cs#L196-L202

we just `return` before we get to that code path.

generator/ServiceClientGeneratorLib/GeneratorDriver.cs
peterrsongg: oops :) i moved it back

generator/ServiceClientGeneratorLib/Operation.cs
dscpinheiro: I'd prefer removing since we're not going to see it in the next PRs (and other reviewers would need more context on why files are changing).

generator/ServiceClientGeneratorLib/GeneratorDriver.cs
dscpinheiro: Got it. But then is the change in the `Example.cs` file even needed?

dscpinheiro:
I ran a console app and the output was the same, but I noticed the `ServiceName` in the request metrics was different for `ListBuckets` (but oddly not for `ListDirectoryBuckets`).

Before:
```json
{
    "properties": {
        "AsyncCall": "True",
        "CanonicalRequest": "...",
        "StringToSign": "...",
        "ServiceName": "AmazonS3",
        "ServiceEndpoint": "https://s3.us-west-2.amazonaws.com/",
        "MethodName": "ListBucketsRequest",
        "StatusCode": "OK"
    }
}
```

After:
```json
{
    "properties": {
        "AsyncCall": "True",
        "CanonicalRequest": "...",
        "StringToSign": "...",
        "ServiceName": "Amazon.S3",
        "ServiceEndpoint": "https://s3.us-west-2.amazonaws.com/",
        "MethodName": "ListBucketsRequest",
        "StatusCode": "OK"
    }
}
```

I assume that's not intentional?

PR #3870, opened 2025-06-10
[Merge 2nd] Generate AbortMultipartUpload, CreateBucketMetadataTableConfiguration and DeleteBucket related operations.

peterrsongg:
## Description

This PR generates more S3 operations, but the main addition here is the CustomizeMarshaller partial method that will be generated on every rest-xml request marshaller. 99% of services won't implement this, but for those operations that have some sort of customization, we will create a partial of the marshaller and implement this method. (Discussed with Norm).

## Testing

**Assembly Comparer was run and no breaking changes in non-internal namespaces**
DRY_RUN-94321c15-4f1e-4c55-aef9-e9f9c8e1e476
## Types of changes

## Checklist

## License


normj (approved):

generator/ServiceClientGeneratorLib/ServiceClientGeneratorLib.csproj
```diff
@@ -736,12 +736,12 @@
     </None>
   </ItemGroup>
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
-	<PackageReference Include="Microsoft.CSharp" Version="4.4.1" />
+	  <PackageReference Include="Microsoft.CSharp" Version="4.4.1" />
```
peterrsongg, line 739: this was already here

https://github.com/aws/aws-sdk-net/blob/d828fcc3e16de5b67939356f59609bbef1df7c37/generator/ServiceClientGeneratorLib/ServiceClientGeneratorLib.csproj#L738-L746

sdk/src/Services/S3/Generated/Model/ListBucketsResponse.cs
```diff
@@ -44,11 +44,6 @@ public partial class ListBucketsResponse : AmazonWebServiceResponse
         /// <para>
         /// The list of buckets owned by the requester.
         /// </para>
-        /// <para />
```
peterrsongg, line 47: this won't actually be deleted. Norm's change is in the structuregenerator.cs file. What happened is it was in the `tt` file but I didnt run the custom tool so when I generated it got deleted. I reran the custom tool to include norm's change so this will remain the same when we actually release.

generator/ServiceClientGeneratorLib/Generators/Marshallers/RestXmlRequestMarshaller.tt
```diff
@@ -99,12 +117,13 @@ namespace <#=this.Config.Namespace #>.Model.Internal.MarshallTransformations
 		ProcessEndpointHostPrefixMembers(3, "publicRequest", this.Operation);
 	}
 #>
-
+			PostMarshallCustomization(request, publicRequest);
```
muhammad-othman, line 120: Since `99% of services won't implement this`, why not add it only for the marshallers that need it instead of affecting all the xml marshallers?

peterrsongg: Norm and I talked and he wanted this mechanism in here instead of another customization entry which would tell us to generate this for a specific subset.

normj: Since the other 99% won't implement the partial `PostMarshallCustomization` the compiler will just compile it away as if it doesn't exists. There is no runtime cost to the other marshallers and I want to avoid as much as possible in the end of this project the number of S3 specific checks we need to do in the generator.

Plus it might be a useful customization point in the future for other services.

PR #3872, opened 2025-06-11
Context expressions - Add support for Scan and Query using LINQ

irina-herciu:
This pull request introduces native support for LINQ expression trees in the IDynamoDBContext API for ScanAsync<T>() and QueryAsync<T>().

## Description
Implement support to automatically translate  LINQ expression trees to DynamoDB Expression 

Example translation:
u => u.Age > 21 && u.Status == "Active"
becomes a ExpressionStatement like: #C0 > :C0 AND #C1 = :C1
with appropriate ExpressionAttributeNames and ExpressionAttributeValues.

- Introduce `ContextExpression` that will be used to define filter LINQ expressions as follows:
```csharp
var contextExpression = new ContextExpression();
contextExpression.SetFilter<Product>(p => p.CompanyInfo.Name == "Acme");
```
Add in LinqDdbExtensions methods to be used inside expression trees to generate DynamoDB Expression functions as follows:
- Between<T>
- AttributeExists
- AttributeNotExists
- AttributeType

Extend `IDynamoDBContext` with a new overloads for Scan API:
```csharp
IEnumerable<T> Scan<T>(ContextExpression filterExpression);
IEnumerable<T> Scan<T>(ContextExpression filterExpression, ScanConfig scanConfig);
```

- Extend `QueryConfig` to allow setting `ExpressionFilter` as `ContextExpression`.
```csharp
var resultExpressionFilter = Context.Query<Employee>("Diane", new QueryConfig
            {
                ExpressionFilter = contextExpression 
            }).ToList();
```

## Testing
unit and integration tests added
benchmarking between filters methods and LINQ expression tree ones
image

## Types of changes

## Checklist

## License


sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs
```diff
@@ -269,6 +269,14 @@ public class DynamoDBOperationConfig
         /// </remarks>
         public List<ScanCondition> QueryFilter { get; set; }
 
+        /// <summary>
+        /// Represents a filter expression that can be used to filter results in DynamoDB operations.
+        /// </summary>
+        /// <remarks>
+        /// Note: Conditions must be against non-key properties.
+        /// </remarks>
+        public ContextExpression Expression { get; set; }
```
normj, line 278: We should be consistent also make the `Expression` available on `ScanConfig`. I see currently it is on `DynamoDBOperationConfig` and `QueryConfig`.

Also since you added the `Scan` overloads that take in the expression seems we should also add the overloads for `Query`.

sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextExpression.cs
```diff
@@ -0,0 +1,242 @@
+﻿using Amazon.DynamoDBv2.DocumentModel;
+using System;
+using System.Collections.Generic;
+using System.Linq.Expressions;
+using ThirdParty.RuntimeBackports;
+using Expression = System.Linq.Expressions.Expression;
```
normj, line 9: Project is failing to compile for .NET 8 because you are missing `using System.Diagnostics.CodeAnalysis;` to include the real version of `DynamicallyAccessedMembers`. The other .NET targets are correctly pulling "fake" version from `ThirdParty.RuntimeBackports`.

sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextExpression.cs
```diff
@@ -0,0 +1,242 @@
...
+            }
+            Filter = filterExpression.Body;
+        }
+    }
+
+    /// <summary>
+    /// Provides extension methods for use in LINQ-to-DynamoDB expression trees.
+    /// These methods are intended for query translation and should not be called directly at runtime.
+    /// </summary>
+    public static class LinqDdbExtensions
```
normj: I'm not comfortable with globally adding extension methods especially since they throw exception if called. For example this code compiles.

```
var o1 = new object();
o1.Between(new object(), new object());
```

normj (changes requested):

sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextExpression.cs
irina-herciu: I removed the extension methods and added the dummy methods under ContextExpression in order to achieve the same functionality.

sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs
irina-herciu: for `Query` there is no overload that allows passing the QueryFilter as 'List<ScanCondition>' as arguments, as it is for `Scan`. 

Also looks like `QueryFilter` from `ScanConfig` is ignored at this moment. Should `Expression` follow the same pattern and be added to config just for consistency? 
image

sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextExpression.cs
```diff
@@ -140,10 +136,45 @@ internal static ConstantExpression GetConstant(Expression expr)
...
+            {
+                throw new InvalidOperationException("MemberExpression does not have an associated expression.");
+            }
+            var constant= GetConstant(memberExpression);
+
+            var value= constant?.Value;
+            if (value != null)
+            {
+                // Use reflection to get the value of the member
+                var memberInfo = value.GetType().GetMember(memberName, BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic).FirstOrDefault();
```
normj: Suspecting this would cause problems for Native AOT I made `GetConstant` public and then create a console application with `PublishAot` set to true.
```
using Amazon.DynamoDBv2.DataModel;
using System.Linq.Expressions;

var constExpr = Expression.Constant(42);
var result = ContextExpressionsUtils.GetConstant(constExpr);
```

When I ran `dotnet publish` I got Native AOT trim warnings on this. 

```
> dotnet publish
Restore complete (1.1s)
  AWSSDK.Core.NetStandard net8.0 succeeded (3.3s) → C:\codebase\v4\aws-sdk-net-v4\sdk\src\Core\bin\Release\net8.0\AWSSDK.Core.dll
  AWSSDK.DynamoDBv2.NetStandard net8.0 succeeded (2.8s) → C:\codebase\v4\aws-sdk-net-v4\sdk\src\Services\DynamoDBv2\bin\Release\net8.0\AWSSDK.DynamoDBv2.dll
  DDBNativeAOTTest succeeded with 1 warning(s) (6.3s) → bin\Release\net8.0\win-x64\publish\
    C:\codebase\v4\aws-sdk-net-v4\sdk\src\Services\DynamoDBv2\Custom\DataModel\ContextExpression.cs(167): Trim analysis warning IL2075: Amazon.DynamoDBv2.DataModel.ContextExpressionsUtils.GetConstantFromMember(MemberExpression): 'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicConstructors', 'DynamicallyAccessedMemberTypes.NonPublicConstructors', 'DynamicallyAccessedMemberTypes.PublicMethods', 'DynamicallyAccessedMemberTypes.NonPublicMethods', 'DynamicallyAccessedMemberTypes.PublicFields', 'DynamicallyAccessedMemberTypes.NonPublicFields', 'DynamicallyAccessedMemberTypes.PublicNestedTypes', 'DynamicallyAccessedMemberTypes.NonPublicNestedTypes', 'DynamicallyAccessedMemberTypes.PublicProperties', 'DynamicallyAccessedMemberTypes.NonPublicProperties', 'DynamicallyAccessedMemberTypes.PublicEvents', 'DynamicallyAccessedMemberTypes.NonPublicEvents' in call to 'System.Type.GetMember(String,BindingFlags)'. The return value of method 'System.Object.GetType()' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
```

This is the tricky thing of Native AOT is the checks done in Roslyn are not as extensive compared to actually doing a `dotnet publish`. Given this PR is doing quite a bit of .NET typesystem/reflection work we need to make sure we do some Native AOT compilation exercising the work.

In your next revision of the PR can you include in the PR description your sample console application that you compiled for Native AOT so we can tell what code paths have been exercised.

sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs
normj: Are you sure `QueryFilter` is being ignored for `Scan`? I see it being copied over in the `ToDynamoDBOperationConfig` operation. If it is not being used that seems like a bug.

I saw we didn't have a list of arguments overload and I'm not sure why. It seems like from a discoverability point of view it makes sense to get the feature more upfront to users as an overload. Besides having yet another overload is there a downside?

normj (changes requested):
We need to get the Native AOT fixed and tested.

sdk/src/Services/DynamoDBv2/Custom/DataModel/Configs.cs
irina-herciu: I've added a integration test for this scenario `TestContext_ScanConfigFilter` in 'DataModelTests', and at this point it fails. To be sure my changes did not affect the existing behavior the same test I executed on main and also fails.

I will push a fix to this issue in this PR.

As for having both an overload and the config, apart form what should happen when both 'Expression's are provided I am not a big fan of having too many arguments in public interfaces and in `Query` method case, for 'IEnumerable<T> Query<T>(object hashKeyValue, QueryOperator op, IEnumerable<object> values, QueryConfig queryConfig);' will look like  'IEnumerable<T> Query<T>(object hashKeyValue, QueryOperator op, IEnumerable<object> values, ContextExpression filterExpression, QueryConfig queryConfig);' and this just just for visibility.

sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextExpression.cs
irina-herciu: I've reimplemented the GetConstant to no longer use reflection and method 'System.Object.GetType()'
console app used to test it is here: [TestNativeConsole](https://github.com/irina-herciu/aws-sdk-net/compare/features/LinqExpressions...irina-herciu:aws-sdk-net:features/LinqExpressionsAOTTEst)

sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextExpression.cs
```diff
@@ -0,0 +1,321 @@
...
+                        expr = null;
+                        break;
+                }
+            }
+
+            return pathNodes;
+        }
+    }
+
+    //internal class ConstantExtractor : ExpressionVisitor
```
normj: nit: We shouldn't have commented out code. If we don't need it then delete it.

normj (approved):
Internal validation `DRY_RUN-027636eb-3411-4bfe-adc6-91a49df023fa` was successful. I'm approving with the assumption you take care of my nit from previous comment.

sdk/test/Services/DynamoDBv2/UnitTests/Custom/ContextExpressionsUtilsTests.cs
```diff
@@ -0,0 +1,378 @@
...
+            var result = ContextExpressionsUtils.GetConstant(memberExpr);
+            Assert.IsNotNull(result);
+            Assert.AreEqual(456, result);
+        }
+
+        [TestMethod]
+        public void GetConstant_ThrowsForUnsupportedNewExpression()
+        {
+            Expression expr = Expression.New(typeof(TestClass));
+            ContextExpressionsUtils.GetConstant(expr);
```
GarrettBeatty: is this test supposed to assert anything? sounds like the test should be asserting an exception is thrown?

sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextExpression.cs
```diff
@@ -0,0 +1,297 @@
...
+        /// Indicates that the value should be compared to see if it falls inclusively between the specified lower and upper bounds.
+        /// Intended for use in LINQ expressions to generate DynamoDB BETWEEN conditions.
+        /// This method is only used inside expression trees and should not be called at runtime.
+        /// </summary>
+        /// <typeparam name="T">The type of the value being compared.</typeparam>
+        /// <param name="value">The value to test.</param>
+        /// <param name="lower">The inclusive lower bound.</param>
+        /// <param name="upper">The inclusive upper bound.</param>
+        /// <returns>This method is intended to be used only within expression definitions (such as LINQ expression trees) and should not be called at runtime.</returns>
+        public static bool Between<[DynamicallyAccessedMembers(InternalConstants.DataModelModeledType)] T>(T value, T lower, T upper) => throw null!;
```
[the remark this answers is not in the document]
GarrettBeatty: is there a reason we do `throw null!` here rather than the suggested notsupportexeception?

sdk/test/Services/DynamoDBv2/UnitTests/Custom/ContextExpressionsUtilsTests.cs
irina-herciu: tests was updated

GarrettBeatty (approved):

GarrettBeatty:
going to rerun the tests and then ill merge this

GarrettBeatty:
started dry run `afba8c16-ce0b-4abc-8704-ae767553950d`.

GarrettBeatty:
the dry run passed

Issue #3869, opened 2025-06-10
If V4 goes for nullable properties - mark them with ? nullable ref type syntax to clearly indicate this


alex-jitbit:
### Describe the bug

For example ListObjectsResponse.S3Objects can be null if no objects are found.

Should be marked with ? - indicating nullable reference type.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

ListObjectsResponse.S3Objects should be declared like this

```
public List<S3Object>? S3Objects //like this
```

### Current Behavior

Currently it's not using nullable reference type. So no warnings from the compiler

### Reproduction Steps

n/a

### AWS .NET SDK and/or Package version used

AWSSDK.S3 4.0.1

### Targeted .NET Platform

.NET 8

### Operating System and version

all

normj:
We had lots of debates on whether to use .NET nullable reference types. The major problems we had with using it was we still need to support .NET Framework which has questionable support for using nullable reference types. The other challenge is with the design of the SDK and every POCO having get and set methods for the properties we would essentially have to mark everything as nullable which would create a swarm of compiler warnings/errors for everybody.

We did add some Roslyn analyzers in V4 to help detect some of the null collection use cases. I think we need to invest more effort into our Roslyn analyzers to help catch potential null exceptions at compile time.

alex-jitbit:
@normj It's OK to not want to that , but you EITHER have to indicate nullable intention at the code level OR mention this in the docs & xml-comments for public members.

For example. Look at the ListObjectResponse docs https://docs.aws.amazon.com/sdkfornet/v4/apidocs/items/S3/TListObjectsResponse.html There is no indication that `S3Objects` property can be `null`. None. Nada. F*cking zero. I had to debug my code that started thrown NREs after migrating to v4. (Even the example code on that page has no null checks)

If the property was nullable (or had proper XML docs) any developer that hovers this property with a mouse - would know what's going on.

normj:
@alex-jitbit You are right we should make this change more obvious in the API docs. I put out this PR that updates our generator to inject the comment into the ndoc which will flow out to the API reference. https://github.com/aws/aws-sdk-net/pull/3874

alex-jitbit:
@normj thanks!! appreciate

normj:
@alex-jitbit PR is out to also update the sample that was in the docs to have the null check: https://github.com/aws/aws-sdk-net/pull/3878

[...]

PR #3874, opened 2025-06-11
Add null documentation comment

normj:
## Description
To help users moving to V4 this PR adds a documentation comment to all generated collection properties about the change in nullability in V4 and how to fallback legacy behavior.

S3 collection properties will not get the comment right now because it is hand coded but we are in the process of moving S3 to the generator and in that move it will get this added doc comment.

## Motivation and Context
https://github.com/aws/aws-sdk-net/issues/3869

## Testing
Dry Run: Pending
Ran generator locally and new doc comment showed where expected
Built the solutions locally.

### DIFF
--- a/generator/ServiceClientGeneratorLib/Generators/BaseGenerator.cs
+++ b/generator/ServiceClientGeneratorLib/Generators/BaseGenerator.cs
@@ -495,7 +495,7 @@ this.Write("    /// </summary>\r\n");
     }
 
     // Adds documentation to the members of a structure
-    public void FormatPropertyDocumentation(Member member, string documentationPreface = "", string evenPublisherDocumentation = "")
+    public void FormatPropertyDocumentation(Member member, string documentationPreface = "", string evenPublisherDocumentation = "", bool isCollectionProperty = false)
     {
         if (documentationPreface != null && documentationPreface != "")
         {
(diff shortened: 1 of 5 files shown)


generator/ServiceClientGeneratorLib/Generators/BaseGenerator.cs
```diff
@@ -526,19 +526,40 @@ public void FormatPropertyDocumentation(Member member, string documentationPrefa
...
+        #line 164 "C:\codebase\v4\aws-sdk-net-v4\generator\ServiceClientGeneratorLib\Generators\BaseGenerator.tt"
+this.Write("        /// <para />\r\n");
+
+        
+        #line default
+        #line hidden
+        
+        #line 166 "C:\codebase\v4\aws-sdk-net-v4\generator\ServiceClientGeneratorLib\Generators\BaseGenerator.tt"
+
+            WriteCommentBlock("        ", "Starting with version 4 of the SDK this property will default to null and if no data for this property is returned");
```
GarrettBeatty: nit: I would make this two sentences. it took me a bit to understand.

"Starting with version 4 of the SDK this property will default to null. If the service returns no data for this property it will be null."

or maybe reword it if you want to keep one sentence? The `and` was throwing my off reading it

ashishdhingra (approved):

ashishdhingra:
Looks good

generator/ServiceClientGeneratorLib/Generators/BaseGenerator.cs
normj: Done

PR #3893, opened 2025-06-23
Generate 8 more s3 operations

peterrsongg:
## Description

Generates the following operations:
```
GetBucketTagging
GetBucketRequestPayment
GetObjectAttributes
GetObjectLegalHold
GetObjectLockConfiguration
GetObjectRetention
GetObjectTagging
GetPublicAccessBlock
```

I added a customization "TreatEnumsAsEnum". For List<Enum> types, enum types are generated as strings, which in S3 isn't always the case. There are cases where List<Enum> is written in custom code. If this customization is used it will treat the enum as an enum instead of as a string.
## Testing

Dry run passes DRY_RUN-e40d50e6-0c92-4258-9f40-5829631b3698
No backwards incompatible changes (re-ran 10:58AM PST 06/24/25)
```
AssemblyComparer AWSSDK.S3.New.dll: Message Amazon.S3.Model.ErrorDocument/TypeAdded: New type Amazon.S3.Model.ErrorDocument
AssemblyComparer AWSSDK.S3.New.dll: Message Amazon.S3.Model.NoSuchKeyException/TypeAdded: New type Amazon.S3.Model.NoSuchKeyException
AssemblyComparer AWSSDK.S3.New.dll: Message Amazon.S3.Model.Condition/TypeAdded: New type Amazon.S3.Model.Condition
AssemblyComparer AWSSDK.S3.New.dll: Message Amazon.S3.Model.GetObjectTaggingResponse/MethodAdded: New method System.String get_VersionId() in Amazon.S3.Model.GetObjectTaggingResponse
AssemblyComparer AWSSDK.S3.New.dll: Message Amazon.S3.Model.GetObjectTaggingResponse/MethodAdded: New method System.Void set_VersionId(System.String) in Amazon.S3.Model.GetObjectTaggingResponse
```

Additional sanity testing:
```
    var getBucketRequestPaymentResponse = await s3Client.GetBucketRequestPaymentAsync(new GetBucketRequestPaymentRequest
    {
        BucketName = bucketName,
        ExpectedBucketOwner = expectedBucketOwner
    }).ConfigureAwait(false);
    Console.WriteLine("The bucket's payer is {0}", getBucketRequestPaymentResponse.Payer);
```
```
The bucket's payer is BucketOwner
```

```
    var getBucketTaggingResponse = await s3Client.GetBucketTaggingAsync(new GetBucketTaggingRequest
    {
        BucketName = bucketName,
        ExpectedBucketOwner = expectedBucketOwner
    }).ConfigureAwait(false);
    getBucketTaggingResponse.TagSet.ForEach(x => Console.WriteLine(x.Key));
```
```
fizz
foo
```
## Types of changes

## Checklist

## License


generator/ServiceModels/s3/s3.customizations.json
```diff
@@ -381,6 +388,13 @@
         },
         "StorageClass":{
             "renameShape": "S3StorageClass"
+        },
+        "Payer":{
+            "renameShape": "string"
         }
+
+    },
+    "treatEnumsAsEnums":{
```
normj: Instead of making up the new term `treatEnumsAsEnums` which by itself sounds weird because if set it false then what does it treat enums if not enums. I suggest reusing the existing term and possible add an `Override` prefix. For example have a customization called `OverrideTreatEnumsAsString` and then add `ObjectAttributesList` with a value `false`.

generator/ServiceClientGeneratorLib/Customizations.cs
```diff
@@ -600,6 +600,22 @@ public List<string> NoArgOverloads
...
+        /// <summary>
+        /// For Lists of enums, the generator automatically treats the enum as a string via passing in true to "treatEnumAsString".
+        /// This customization overrides that so that the enum is treated as an enum. 
+        ///     "treatEnumsAsEnums":{
+        ///        "ObjectAttributesList": true
+        /// }
+        /// 
+        /// </summary>
+        /// <returns></returns>
+        public bool GetTreatEnumAsEnum(string shapeName)
```
normj: Per my comment below rename this to `OverrideTreatEnumsAsString`.

generator/ServiceModels/s3/s3.customizations.json
peterrsongg: thanks for the suggestion, I like that name much better. Changed it and updated customizations.json file as well.

generator/ServiceClientGeneratorLib/Customizations.cs
```diff
@@ -600,6 +600,22 @@ public List<string> NoArgOverloads
             }
         }
 
+        /// <summary>
+        /// For Lists of enums, the generator automatically treats the enum as a string via passing in true to "treatEnumAsString".
+        /// This customization overrides that so that the enum is treated as an enum. 
+        ///     "treatEnumsAsEnums":{
```
normj: You need to update the example to match the rename to `OverrideTreatEnumsAsString`

generator/ServiceModels/s3/s3.customizations.json
```diff
@@ -381,6 +388,13 @@
...
         "StorageClass":{
             "renameShape": "S3StorageClass"
+        },
+        "Payer":{
+            "renameShape": "string"
         }
+
+    },
+    "overrideTreatEnumsAsString":{
+        "ObjectAttributesList": true
```
normj: Shouldn't this be false since we want to turn off treating the enum as a string. I'm viewing this customization is allowing the user manually specify the exact value for the treatEnumAsString parameter on `DetermineType`.

generator/ServiceClientGeneratorLib/Customizations.cs
```diff
@@ -600,6 +600,22 @@ public List<string> NoArgOverloads
...
+        ///        "ObjectAttributesList": true
+        /// }
+        /// 
+        /// </summary>
+        /// <returns></returns>
+        public bool OverrideTreatEnumsAsString(string shapeName)
+        {
+            var data = _documentRoot[OverrideTreatEnumsAsStringKey];
+            if (data == null) return false;
+            return data[shapeName]  == null ? false : true;
```
normj: Since the objects in the list have a value of `true` or `false` whether we should treat enums as strings we should check the value here and not just whether it exists.

generator/ServiceClientGeneratorLib/Member.cs
```diff
@@ -583,7 +583,8 @@ private string DetermineType(JsonData extendedData, bool treatEnumsAsString, boo
                     var valueType = DetermineType(memberShape["value"], true, false);
                     return string.Format("Dictionary<{0}, {1}>", keyType, valueType);
                 case "list":
-                    var listType = DetermineType(memberShape["member"], true, false);
+                    // if customization exists to treat the enum as an enum instead of a string, we pass in false
+                    var listType = this.model.Customizations.OverrideTreatEnumsAsString(this.Extends) ? DetermineType(memberShape["member"], false, false) : DetermineType(memberShape["member"], true, false) ;
```
normj: This `?:` can be reduced down `DetermineType(memberShape["member"], this.model.Customizations.OverrideTreatEnumsAsString(this.Extends), false)`

generator/ServiceModels/s3/s3.customizations.json
peterrsongg: since `DetermineType()` by default passes in false for `treatEnumsAsString` and there is no way to call `DetermineType(JsonData extendedData, bool treatEnumsAsString, bool useNullable = true)` outside the `Member.cs` class I was thinking that this customization would override the default value which is why I passed in `true` here meaning. "Override the default value of `false`".

peterrsongg: I think I just had a different idea of what this boolean would represent. I'll update this

generator/ServiceClientGeneratorLib/Customizations.cs
peterrsongg: since we're now changing what `OverrideTreatEnumsAsString` represents, I think we have to return null if the value doesn't exist, and otherwise just return the value.

generator/ServiceClientGeneratorLib/Member.cs
```diff
@@ -579,11 +579,13 @@ private string DetermineType(JsonData extendedData, bool treatEnumsAsString, boo
                     }
                     return typeName;
                 case "map":
-                    var keyType = DetermineType(memberShape["key"], true, false);
-                    var valueType = DetermineType(memberShape["value"], true, false);
+                    bool overrideMapTreatEnumsAsString = this.model.Customizations.OverrideTreatEnumsAsString(this.Extends) ?? true;
```
dscpinheiro, line 582: Why do we need to use `Extends` here?

peterrsongg: I used `Extends` here b/c I'm just trying to grab the name of the shape that the member targets rather than than the name of the member itself since this is a shape-level modification and not a member-level modification

PR #3895, opened 2025-06-27
Generate 9 more S3 Operations

peterrsongg:
## Description

Generates 9 more operations to conclude phase 1. I had to add a customization to exclude members from property generation but not from the marshallers because of `PartDetail` which inherits from `PartETag`. `PartETag` contains the excluded members. 

I also updated the `RestXmlExceptionUnmarshaller.tt` file to get the `Id2` and `CfId` which is unique to S3, as well as update the exception shapes themselves to add the constructor which takes these values in its constructor.

One change from this is that `PutBucketPolicy`'s payload is set to `request.Content` instead of `request.ContentStream`. I've done a sanity check to make sure it still works. Per the model since the payload is a string, there is no need to wrap it in a memory stream. Also the payload member doesn't target the streaming trait so there is no need to stream it either.

One other minor change is that `Tagging` in `PutBucketTagging` was automatically initialized if it was null. This was a one-off customization we had that I intentionally left out to remove another customization hook. If we feel we want this added back, we can do that, but I will call this out as a breaking change.

## Testing

DRY_RUN-2df7b27d-b938-44c2-aa9d-64c1ce797bf7

```
    var putBucketPolicyResponse = await s3Client.PutBucketPolicyAsync(new PutBucketPolicyRequest
    {
        BucketName = bucketName,
        Policy = 
"""
{
"Version":"2008-10-17",
"Id":"test-put-bucket-policy",
"Statement" : [
    {
        "Effect":"Allow",
        "Sid":"1", 
        "Principal" : {
            "AWS": "arn:aws:iam::<account_id>:***"
        },
        "Action":["s3:*"],
        "Resource":"arn:aws:s3:::<bucket_name>*"
    }
    ]
}
"""
    }).ConfigureAwait(false);
```

executed successfully with flipping request.ContentStream to request.Content
## Types of changes

## Checklist

## License


generator/ServiceClientGeneratorLib/Generators/Marshallers/RestXmlRequestMarshaller.tt
```diff
@@ -729,8 +730,13 @@ WriteXmlAttributeString(level + 1, member, variableName, isPayload: true, operat
 		{
 #>
 <#=new string(' ', level * 4)#>request.Content = Encoding.UTF8.GetBytes(<#=this.Operation.RequestPayloadMember.PrimitiveMarshaller#>(publicRequest.<#=this.Operation.RequestPayloadMember.PropertyName#>));
+<#+
+			if (this.Operation.Name != "PutBucketPolicy")
```
dscpinheiro: Why is this needed if the `sdk/src/Services/S3/Custom/Model/Internal/MarshallTransformations/PutBucketPolicyRequestMarshaller.cs` file is also setting the content-type header?

generator/.DevConfigs/c881e931-b1ff-45ff-9521-9653e179466e.json
```diff
@@ -0,0 +1,13 @@
+{
+    "services": [
+        {
+            "serviceName": "S3",
+            "type": "patch",
+            "changeLogMessages": [
+                "Update ExceptionUnmarshallers and add S3 Exception t4 file. Generate PutBucketPolicy, PutObjectLegalHold, PutObjectLock, PutObjectRetention, PutPublicAccessBlock, HeadBucket, ListParts",
```
dscpinheiro: The PR title and description say 9 operations, but there are only 7 listed here.

generator/ServiceClientGeneratorLib/Operation.cs
```diff
@@ -302,6 +302,27 @@ public Member RequestPayloadMember
                 if (this.RequestStructure != null)
                 {
                     var payload = this.RequestStructure.PayloadMemberName;
+                    // check to see if the payload member has been overridden by another member defined in the service customizations
```
dscpinheiro, line 305: An example here would have helped (these nested if statements are not very easy to read).

sdk/src/Services/S3/Generated/Model/Condition.cs
```diff
@@ -1,97 +0,0 @@
-/*
```
dscpinheiro: Is this removal intentional? It's in the generated folder...

generator/ServiceClientGeneratorLib/Generators/Marshallers/RestXmlRequestMarshaller.tt
peterrsongg: Good catch, at first I thought I needed this since the previous custom code logic was 

```
            if (!defaultRequest.Headers.ContainsKey(HeaderKeys.ContentTypeHeader))
            {
                defaultRequest.Headers.Add(HeaderKeys.ContentTypeHeader, "text/plain");
            }
```

and customers can set the headers via something like:

```
    s3Client.BeforeRequestEvent += (sender, args) =>
    {
        if (args is Amazon.Runtime.WebServiceRequestEventArgs webArgs &&
    webArgs.Request is PutBucketPolicyRequest)
        {
            webArgs.Headers["Content-Type"] = "application/json";
        }
    };
```

but I realize that this would overwrite it after the marshaller logic is invoked, so actually I don't need this extra logic in the generator. I'll remove it.

generator/ServiceClientGeneratorLib/Operation.cs
peterrsongg: added an example and also got rid of one level of nesting by refactoring

generator/.DevConfigs/c881e931-b1ff-45ff-9521-9653e179466e.json
peterrsongg: oops it's actually 8, I updated the devconfig message

sdk/src/Services/S3/Generated/Model/Condition.cs
peterrsongg: This is only used in `GetBucketWebsite` so I'm not sure how it got here...

peterrsongg: it doesn't even exist here.. 
https://github.com/aws/aws-sdk-net/tree/main/sdk/src/Services/S3/Generated/Model

maybe a github UI thing? not sure how that happened

sdk/src/Services/S3/Generated/Model/PartDetail.cs
```diff
@@ -0,0 +1,77 @@
...
+        private DateTime? _lastModified;
+        private long? _size;
+
+        /// <summary>
+        /// Gets and sets the property LastModified. 
+        /// <para>
+        /// Date and time at which the part was uploaded.
+        /// </para>
+        /// </summary>
+        public DateTime? LastModified
```
muhammad-othman, line 46: Looks like `LastModified` to isn't using nullable DateTime in v4, probably we missed it.
Changing it now is a breaking change, are we okay with that?

generator/ServiceClientGeneratorLib/Generators/Marshallers/RestXmlRequestMarshaller.tt
```diff
@@ -729,6 +730,8 @@ WriteXmlAttributeString(level + 1, member, variableName, isPayload: true, operat
 		{
 #>
 <#=new string(' ', level * 4)#>request.Content = Encoding.UTF8.GetBytes(<#=this.Operation.RequestPayloadMember.PrimitiveMarshaller#>(publicRequest.<#=this.Operation.RequestPayloadMember.PropertyName#>));
+<#+
```
dscpinheiro: These lines (733 and 734) can be removed too now, right?

sdk/src/Services/S3/Generated/Model/Condition.cs
dscpinheiro: It's gone from the last diff, so maybe? But yeah need to make sure we don't remove generated code moving forward.

sdk/src/Services/S3/Generated/Model/PartDetail.cs
peterrsongg: Yup, in my previous PRs Norm said as long as we call it out in the DevConfig it is fine, but just for these nullable type changes. So I included it in the dev config

normj: It is not great to make these breaking changes but we are still early in V4 adoption and these fields missed so far are likely not the common fields used. I would prefer to take the hit of taking on the minor breaking change in this early part of V4 then add more complexity in our generator to support some fields not being nullable.

@peterrsongg If you find a property that looks like it would be frequent property being accessed like a bucket name then we should discuss.

peterrsongg: Gotcha, if i feel like the property is very commonly used I will bring it up for discussion

muhammad-othman (approved):

generator/ServiceClientGeneratorLib/Generators/Marshallers/RestXmlRequestMarshaller.tt
dscpinheiro: Can we remove these? There are no other changes in this file.

Edit: Oh, I hadn't seen the `GenerateRequestChecksumHandling` above... But I'd still prefer removing this unnecessary code.

peterrsongg: done

dscpinheiro (approved):

PR #3897, opened 2025-06-30
Add unit tests for existing S3PostUploadSignedPolicy class and fix exception message

GarrettBeatty:
## Description
1. Add unit tests to existing S3PostUploadSignedPolicy  class
2. Fix exception message format parsing. The `string.Format` was missing the parameter for `{0}` in the function call. The unit tests caught this issue.

## Motivation and Context
For implementing https://github.com/aws/aws-sdk-net/issues/1901 we are moving the `S3PostUploadSignedPolicy` class out of the _bcl folder so that other frameworks (net standard/core 3.1, and the existing netframework) can utilize it. In order to ensure that we do not break anything during this move, i am adding unit tests which validate the current functionality. This way once we make changes in the future we can easily determine if we break things. 

Please note that we only run unit tests in netframework but this should at least catch potential future issues for netframework. For other frameworks we can copy/run these unit tests in a console app when testing.

## Testing
1. Ran unit tests locally
2. Dry run `6b777a12-a351-46a6-9b45-89f4f98ad46a` pass

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project
- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [x] I have read the **README** document
- [x] I have added tests to cover my changes
- [x] All new and existing tests passed

## License

- [x] I confirm that this pull request can be released under the Apache 2 license

[issues]: https://github.com/aws/aws-sdk-net/issues
[license]: http://aws.amazon.com/apache2.0/
[cla]: http://en.wikipedia.org/wiki/Contributor_License_Agreement

Issue #1901, opened 2021-08-12
AmazonS3Client is missing `Create Presigned Post`


marns:
`AmazonS3Client` / `IAmazonS3` is missing `Create Presigned Post` functionality found in JS/Python libraries. `GetPreSignedURL` does not appear to support POST. Therefore there seems to be no way to constrain max upload size when providing signed links using aws-sdk-net.

ashishdhingra:
Hi @marns,

Good morning.

Could you please elaborate or point to the JS/Python functionality which appear to use `Create Presigned Post`? Are you referring to https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/modules/_aws_sdk_s3_presigned_post.html (for example)?

Just for reference, [GetPreSignedUrlRequest](https://github.com/aws/aws-sdk-net/blob/96963afe6fef4d9efd8275cf201c2bafe2789210/sdk/src/Services/S3/Custom/Model/GetPreSignedUrlRequest.cs#L38) has a property named [Verb](https://github.com/aws/aws-sdk-net/blob/96963afe6fef4d9efd8275cf201c2bafe2789210/sdk/src/Services/S3/Custom/Model/GetPreSignedUrlRequest.cs#L179) which is of type [Amazon.S3. HttpVerb](https://github.com/aws/aws-sdk-net/blob/96963afe6fef4d9efd8275cf201c2bafe2789210/sdk/src/Services/S3/Custom/S3Enumerations.cs#L573) which only supports GET, PUT, DELETE and HEAD verbs. The example at [Uploading objects using presigned URLs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/PresignedUrlUploadObject.html) makes use of PUT HTTP Verb. Also the Visual Studio toolkit appears to support only GET and PUT HTTP verbs for generating presigned URLs.

This needs discussion with the team and doesn't appear to be bug, rather, the feature request.

Thanks,
Ashish

ashishdhingra:
Just for reference:
- [Authenticating Requests: Browser-Based Uploads Using POST (AWS Signature Version 4)](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-authentication-HTTPPOST.html)
- [ Amazon.S3.HttpVerb](https://github.com/aws/aws-sdk-net/blob/96963afe6fef4d9efd8275cf201c2bafe2789210/sdk/src/Services/S3/Custom/S3Enumerations.cs#L573)

[...]

PR #3900, opened 2025-07-01
Move S3PostUploadSignedPolicy outside of _bcl folder

GarrettBeatty:
## Description
1. Move logic related to generating presigned post urls outside of _bcl folder so the signing logic can be used by netstandard and net core 3.1. 
2. Update json parsing logic to be rely only on System.Text.Json and to be native AOT compatible.
3. For XML related functionality i did not think it was worth making it native aot compatible. i  was having difficulty ( i tried following github.com/dotnet/runtime/issues/106580)

## Motivation and Context
In order to implement https://github.com/aws/aws-sdk-net/issues/1901 , we need to utilize the existing `GetSignedPolicy` function. It is currently in the _bcl folder because it relies on system.text.json.nodes which isnt available in .net core3.1 and the current code wasn't native aot compatible. So in order to fix this, we needed to remove the system.text.json.nodes dependency and move the code out of the _bcl folder.

## Testing
In order to ensure it works on net standard and is native AOT compatible i created a console app

```
Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net8.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
    <PublishAot>true</PublishAot>
    <InvariantGlobalization>true</InvariantGlobalization>
  </PropertyGroup>
  
  <ItemGroup>
    <!-- Reference to the SDK S3 project -->
    <ProjectReference Include="..\sdk\src\Services\S3\AWSSDK.S3.NetStandard.csproj" />
  </ItemGroup>
</Project>

```

```
using Amazon;
using Amazon.Runtime;
using Amazon.S3.Util;
using System;
using System.Text.Json;

namespace S3PostAotDemo
{
    class Program
    {
        static void Main(string[] args)
        {
            // Create a sample S3 POST policy document
            // This policy allows uploads:
            // - To a specific bucket
            // - With a public-read ACL
            // - Keys that start with user/uploads/
            // - With content size between 1KB and 10MB
            string policyJson = @"{
  ""expiration"": ""2030-01-01T00:00:00Z"",
  ""conditions"": [
    {""bucket"": ""example-bucket""},
    {""acl"": ""public-read""},
    [""starts-with"", ""$key"", ""user/uploads/""],
    [""starts-with"", ""$Content-Type"", ""image/""],
    [""content-length-range"", 1024, 10485760]
  ]
}";

            Console.WriteLine("S3 Presigned POST Policy Demo (Native AOT)");
            Console.WriteLine("=========================================");
            Console.WriteLine();
            
            try
            {
                // Set up credentials and region
                // NOTE: In a real application, use environment variables, AWS SDK credential providers,
                // or other secure credential sources instead of hardcoded values
                var credentials = new BasicAWSCredentials("AKIAIOSFODNN7EXAMPLE", "test-secret-key");
                var region = RegionEndpoint.USWest1;

                // Get the signed policy
                Console.WriteLine("Generating presigned policy...");
                var signedPolicy = S3PostUploadSignedPolicy.GetSignedPolicy(policyJson, credentials, region);
                
                // Display results
                Console.WriteLine("\nSigned Policy Details:");
                Console.WriteLine($"Policy (Base64): {signedPolicy.Policy}");
                Console.WriteLine($"Signature: {signedPolicy.Signature}");
                Console.WriteLine($"AccessKeyId: {signedPolicy.AccessKeyId}");
                Console.WriteLine($"Algorithm: {signedPolicy.Algorithm}");
                Console.WriteLine($"Date: {signedPolicy.Date}");
                Console.WriteLine($"Credential: {signedPolicy.Credential}");
                
                // Display human-readable policy
                Console.WriteLine("\nHuman-readable policy:");
                Console.WriteLine(signedPolicy.GetReadablePolicy());
                
                // Show how to use this in an HTML form
                Console.WriteLine("\nSample HTML form fields:");
                Console.WriteLine($"<input type=\"hidden\" name=\"key\" value=\"user/uploads/${{filename}}\">");
                Console.WriteLine($"<input type=\"hidden\" name=\"acl\" value=\"public-read\">");
                Console.WriteLine($"<input type=\"hidden\" name=\"Content-Type\" value=\"image/jpeg\">");
                Console.WriteLine($"<input type=\"hidden\" name=\"bucket\" value=\"example-bucket\">");
                Console.WriteLine($"<input type=\"hidden\" name=\"X-Amz-Algorithm\" value=\"{signedPolicy.Algorithm}\">");
                Console.WriteLine($"<input type=\"hidden\" name=\"X-Amz-Credential\" value=\"{signedPolicy.Credential}\">");
                Console.WriteLine($"<input type=\"hidden\" name=\"X-Amz-Date\" value=\"{signedPolicy.Date}\">");
                Console.WriteLine($"<input type=\"hidden\" name=\"Policy\" value=\"{signedPolicy.Policy}\">");
                Console.WriteLine($"<input type=\"hidden\" name=\"X-Amz-Signature\" value=\"{signedPolicy.Signature}\">");
                Console.WriteLine($"<input type=\"file\" name=\"file\">");
                Console.WriteLine($"<input type=\"submit\" value=\"Upload\">");
                
                // Also demonstrate converting to JSON for use in JavaScript applications
                Console.WriteLine("\nJSON representation (for JavaScript apps):");
                string json = signedPolicy.ToJson();
                Console.WriteLine(json);
            }
            catch (Exception ex)
            {
                Console.WriteLine($"\nERROR: {ex.Message}");
                Console.WriteLine(ex.ToString());
            }
            
            Console.WriteLine("\nPress any key to exit...");
            Console.ReadKey();
        }
    }
}

```

```
PS C:\dev\repos\aws-sdk-net\S3PostAotDemo> dotnet publish -r win-x64 -c Release
Restore complete (1.9s)
  AWSSDK.Core.NetStandard net8.0 succeeded (16.7s) → C:\dev\repos\aws-sdk-net\sdk\src\Core\bin\Release\net8.0\AWSSDK.Core.dll
  AWSSDK.S3.NetStandard net8.0 succeeded (9.9s) → C:\dev\repos\aws-sdk-net\sdk\src\Services\S3\bin\Release\net8.0\AWSSDK.S3.dll
  S3PostAotDemo succeeded (11.6s) → bin\Release\net8.0\win-x64\publish\

Build succeeded in 43.4s
```

```
S3 Presigned POST Policy Demo (Native AOT)
=========================================

Generating presigned policy...

Signed Policy Details:
Policy (Base64): eyJleHBpcmF0aW9uIjoiMjAzMC0wMS0wMVQwMDowMDowMFoiLCJjb25kaXRpb25zIjpbeyJidWNrZXQiOiJleGFtcGxlLWJ1Y2tldCJ9LHsiYWNsIjoicHVibGljLXJlYWQifSxbInN0YXJ0cy13aXRoIiwiJGtleSIsInVzZXIvdXBsb2Fkcy8iXSxbInN0YXJ0cy13aXRoIiwiJENvbnRlbnQtVHlwZSIsImltYWdlLyJdLFsiY29udGVudC1sZW5ndGgtcmFuZ2UiLDEwMjQsMTA0ODU3NjBdLHsieC1hbXotY3JlZGVudGlhbCI6IkFLSUFJT1NGT0ROTjdFWEFNUExFLzIwMjUwNzAxL3VzLXdlc3QtMS9zMy9hd3M0X3JlcXVlc3QvIn0seyJ4LWFtei1hbGdvcml0aG0iOiJBV1M0LUhNQUMtU0hBMjU2In0seyJ4LWFtei1kYXRlIjoiMjAyNTA3MDFUMTcwNzAzWiJ9XX0=
Signature: 8838058f9f8a328616b3fa8bf42ddf6517429f155f1083d5a6f43eef0ac80049
AccessKeyId: AKIAIOSFODNN7EXAMPLE
Algorithm: AWS4-HMAC-SHA256
Date: 20250701T170703Z
Credential: AKIAIOSFODNN7EXAMPLE/20250701/us-west-1/s3/aws4_request/

Human-readable policy:
{"expiration":"2030-01-01T00:00:00Z","conditions":[{"bucket":"example-bucket"},{"acl":"public-read"},["starts-with","$key","user/uploads/"],["starts-with","$Content-Type","image/"],["content-length-range",1024,10485760],{"x-amz-credential":"AKIAIOSFODNN7EXAMPLE/20250701/us-west-1/s3/aws4_request/"},{"x-amz-algorithm":"AWS4-HMAC-SHA256"},{"x-amz-date":"20250701T170703Z"}]}

Sample HTML form fields:
<input type="hidden" name="key" value="user/uploads/${filename}">
<input type="hidden" name="acl" value="public-read">
<input type="hidden" name="Content-Type" value="image/jpeg">
<input type="hidden" name="bucket" value="example-bucket">
<input type="hidden" name="X-Amz-Algorithm" value="AWS4-HMAC-SHA256">
<input type="hidden" name="X-Amz-Credential" value="AKIAIOSFODNN7EXAMPLE/20250701/us-west-1/s3/aws4_request/">
<input type="hidden" name="X-Amz-Date" value="20250701T170703Z">
<input type="hidden" name="Policy" value="eyJleHBpcmF0aW9uIjoiMjAzMC0wMS0wMVQwMDowMDowMFoiLCJjb25kaXRpb25zIjpbeyJidWNrZXQiOiJleGFtcGxlLWJ1Y2tldCJ9LHsiYWNsIjoicHVibGljLXJlYWQifSxbInN0YXJ0cy13aXRoIiwiJGtleSIsInVzZXIvdXBsb2Fkcy8iXSxbInN0YXJ0cy13aXRoIiwiJENvbnRlbnQtVHlwZSIsImltYWdlLyJdLFsiY29udGVudC1sZW5ndGgtcmFuZ2UiLDEwMjQsMTA0ODU3NjBdLHsieC1hbXotY3JlZGVudGlhbCI6IkFLSUFJT1NGT0ROTjdFWEFNUExFLzIwMjUwNzAxL3VzLXdlc3QtMS9zMy9hd3M0X3JlcXVlc3QvIn0seyJ4LWFtei1hbGdvcml0aG0iOiJBV1M0LUhNQUMtU0hBMjU2In0seyJ4LWFtei1kYXRlIjoiMjAyNTA3MDFUMTcwNzAzWiJ9XX0=">
<input type="hidden" name="X-Amz-Signature" value="8838058f9f8a328616b3fa8bf42ddf6517429f155f1083d5a6f43eef0ac80049">
<input type="file" name="file">
<input type="submit" value="Upload">

JSON representation (for JavaScript apps):
{"policy":"eyJleHBpcmF0aW9uIjoiMjAzMC0wMS0wMVQwMDowMDowMFoiLCJjb25kaXRpb25zIjpbeyJidWNrZXQiOiJleGFtcGxlLWJ1Y2tldCJ9LHsiYWNsIjoicHVibGljLXJlYWQifSxbInN0YXJ0cy13aXRoIiwiJGtleSIsInVzZXIvdXBsb2Fkcy8iXSxbInN0YXJ0cy13aXRoIiwiJENvbnRlbnQtVHlwZSIsImltYWdlLyJdLFsiY29udGVudC1sZW5ndGgtcmFuZ2UiLDEwMjQsMTA0ODU3NjBdLHsieC1hbXotY3JlZGVudGlhbCI6IkFLSUFJT1NGT0ROTjdFWEFNUExFLzIwMjUwNzAxL3VzLXdlc3QtMS9zMy9hd3M0X3JlcXVlc3QvIn0seyJ4LWFtei1hbGdvcml0aG0iOiJBV1M0LUhNQUMtU0hBMjU2In0seyJ4LWFtei1kYXRlIjoiMjAyNTA3MDFUMTcwNzAzWiJ9XX0=","signature":"8838058f9f8a328616b3fa8bf42ddf6517429f155f1083d5a6f43eef0ac80049","access_key":"AKIAIOSFODNN7EXAMPLE"}
```

Dry run: `fb816228-3d51-4f02-bff4-3711227d8874` - `SUCCEEDED`.
I have already created unit tests for this functionality in a previous PR https://github.com/aws/aws-sdk-net/pull/3897

## Types of changes

## Checklist

- [] I have added tests to cover my changes

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/Util/S3PostUploadSignedPolicy.cs
+++ b/sdk/src/Services/S3/Custom/Util/S3PostUploadSignedPolicy.cs
@@ -147,17 +147,16 @@ namespace Amazon.S3.Util
         /// <returns>UTF-8 bytes of the modified policy document</returns>
         private static byte[] AddConditionsToPolicy(string policy, Dictionary<string, string> newConditions)
         {
-            // Parse the original policy document
             using (JsonDocument policyDoc = JsonDocument.Parse(policy, options))
             {
                 using (var ms = new MemoryStream())
                 {
                     using (var writer = new Utf8JsonWriter(ms))
                     {
-                        // Start writing the new policy document object
+                        // Step 1: Start writing the new policy document object
                         writer.WriteStartObject();
 
-                        // Step 1: Copy all properties except "conditions" from the original policy
+                        // Step 2: Copy all properties except "conditions" from the original policy
                         // (We'll handle the conditions array separately)
                         foreach (var property in policyDoc.RootElement.EnumerateObject())
                         {
@@ -167,38 +166,39 @@ namespace Amazon.S3.Util
                             }
                         }
 
-                        // Step 2: Start writing the new "conditions" array
+                        // Step 3: Start writing the new "conditions" array
                         writer.WritePropertyName("conditions");
                         writer.WriteStartArray();
 
-                        // Check if the original policy has a conditions array
+                        // Step 4: Check if the original policy has a conditions array
                         bool foundConditions = policyDoc.RootElement.TryGetProperty("conditions", out JsonElement conditionsElement);
 
                         if (foundConditions && conditionsElement.ValueKind == JsonValueKind.Array)
                         {
-                            // Step 3: Process each existing condition in the original policy
+                            // Step 5: Process each existing condition in the original policy
                             foreach (JsonElement condition in conditionsElement.EnumerateArray())
                             {
                                 bool shouldCopyCondition = true;  // Default: copy the original condition as-is
                                 string matchedKey = null;         // Track if we found a key to update
 
-                                // Step 3a: If this condition is an object, check if it contains any keys
+                                // Step 5a: If this condition is an object, check if it contains any keys
                                 // that match our new conditions (meaning we need to update this condition)
                                 if (condition.ValueKind == JsonValueKind.Object)
                                 {
-                                    foreach (var newCond in newConditions.Keys.ToList())
+                                    // Optimized: Directly iterate over dictionary entries to avoid ToList allocation
+                                    foreach (var newCond in newConditions)
                                     {
-                                        if (condition.TryGetProperty(newCond, out _))
+                                        if (condition.TryGetProperty(newCond.Key, out _))
                                         {
                                             // This condition contains a key we want to update
-                                            matchedKey = newCond;
+                                            matchedKey = newCond.Key;
                                             shouldCopyCondition = false;  // Don't copy as-is, we'll update it
                                             break;
                                         }
                                     }
                                 }
 
-                                // Step 3b: Handle the condition based on whether we need to update it
+                                // Step 5b: Handle the condition based on whether we need to update it
                                 if (shouldCopyCondition)
                                 {
                                     // Case 1: This condition doesn't need updating, copy it as-is
@@ -217,7 +217,7 @@ namespace Amazon.S3.Util
                                     {
                                         if (property.Name == matchedKey)
                                         {
-                                            // This is the property we want to update
+                                            // This is the property we want to update with the new value
                                             writer.WriteString(matchedKey, newConditions[matchedKey]);
 
                                             // Remove this key from newConditions since we've now handled it
@@ -227,7 +227,7 @@ namespace Amazon.S3.Util
                                         else
                                         {
                                             // This is a property we want to preserve (not update)
-                                            // Without this, properties other than the one being updated would be lost
+                                            // Without this else clause, properties other than the one being updated would be lost
                                             property.WriteTo(writer);
                                         }
                                     }
@@ -238,7 +238,7 @@ namespace Amazon.S3.Util
                             }
                         }
 
-                        // Step 4: Add any remaining new conditions that weren't updates to existing ones
+                        // Step 6: Add any remaining new conditions that weren't updates to existing ones
                         foreach (var newCond in newConditions)
                         {
                             writer.WriteStartObject();
@@ -246,12 +246,12 @@ namespace Amazon.S3.Util
                             writer.WriteEndObject();
                         }
 
-                        // Step 5: Close the conditions array and the overall policy object
+                        // Step 7: Close the conditions array and the overall policy object
                         writer.WriteEndArray();
                         writer.WriteEndObject();
                         writer.Flush();
 
-                        // Step 6: Return the complete modified policy document as bytes
+                        // Step 8: Return the complete modified policy document as bytes
                         return ms.ToArray();
                     }
                 }


sdk/src/Services/S3/Custom/Util/S3PostUploadSignedPolicy.cs
```diff
@@ -47,14 +47,59 @@ namespace Amazon.S3.Util
...
+        /// </summary>
+        public string Date { get; set; }
+
+        /// <summary>
+        /// In addition to the access key ID, this provides scope information
+        /// used in calculating the signing key for signature calculation. 
+        /// </summary>
+        public string Credential { get; set; }
+
+
```
GarrettBeatty, line 95: this is all existing code, i just moved it to the top of file

sdk/src/Services/S3/Custom/Util/S3PostUploadSignedPolicy.cs
```diff
@@ -91,75 +136,102 @@ public static S3PostUploadSignedPolicy GetSignedPolicy(string policy, AWSCredent
             };
         }
 
-        private static byte[] addConditionsToPolicy(string policy, Dictionary<string, string> newConditions)
+        /// <summary>
+        /// Framework-agnostic implementation to add conditions to a policy document.
+        /// Uses JsonDocument/JsonElement instead of JsonNode.
+        /// </summary>
+        private static byte[] AddConditionsToPolicy(string policy, Dictionary<string, string> newConditions)
```
GarrettBeatty, line 148: This is the only function that i really modified in this PR. the functionality is the same but it uses different logic to parse the json

sdk/src/Services/S3/Custom/Util/S3PostUploadSignedPolicy.cs
```diff
@@ -206,6 +282,41 @@ public string ToXml()
...
+        /// Create an instance of this class from an XML string.
+        /// </summary>
+        /// <param name="policyXml">XML string generated by ToXml()</param>
+        /// <returns>Instance of S3PostUploadSignedPolicy</returns>
+#if NET8_0_OR_GREATER
+        [System.Diagnostics.CodeAnalysis.UnconditionalSuppressMessage("AssemblyLoadTrimming", "IL2026:RequiresUnreferencedCode",
+   Justification = "This suppression is here to ignore the warnings since we have not made this function AOT compatible yet.")]
+
+#endif
+        public static S3PostUploadSignedPolicy GetSignedPolicyFromXml(string policyXml)
```
GarrettBeatty, line 318: this function is the same, i just added the AssemblyLoadTrimming warning

muhammad-othman: I believe we usually add this suppression we we know that the function is AOT compatible but there is a warning that doesn't affect the trimming, but if this isn't AOT compatible then users may get a runtime exception.

GarrettBeatty: hmm so what should i do if its not AOT compatible? not have this suppression at all?

peterrsongg: If it isn't AOT compatible then we should use the `RequiresUnreferencedCode` attribute :)
https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.requiresunreferencedcodeattribute?view=net-9.0

GarrettBeatty: looks like this is the wrong suppression. it should be updated to requires dynamic code

sdk/src/Services/S3/Custom/Util/S3PostUploadSignedPolicy.cs
```diff
@@ -91,75 +136,102 @@ public static S3PostUploadSignedPolicy GetSignedPolicy(string policy, AWSCredent
...
+                                string matchedKey = null;
 
-        /// <summary>
-        /// The signature for the policy.
-        /// </summary>
-        public string Signature { get; set; }
+                                // Check if this condition is an object that contains any of our new condition keys
+                                if (condition.ValueKind == JsonValueKind.Object)
+                                {
+                                    foreach (var newCond in newConditions.Keys.ToList())
```
peterrsongg: this is minor, but instead of calling `ToList` on the `Keys` and allocating another array can we just do something like

```
foreach (var newCond in newConditions)
{
  if (condition.TryGetProperty(newCond.Key, out _ ))
  {
    matchedKey = newCond.Key;
    shouldCopyCondition = false;
    break;
  }
}
```

sdk/src/Services/S3/Custom/Util/S3PostUploadSignedPolicy.cs
```diff
@@ -91,75 +136,102 @@ public static S3PostUploadSignedPolicy GetSignedPolicy(string policy, AWSCredent
...
-        /// The signing algorithm used. Required as a field in the post Amazon
-        /// S3 can re-calculate the signature.
-        /// </summary>
-        public string Algorithm { get; set; }
+                                            // Remove this key from newConditions since we've handled it
+                                            newConditions.Remove(matchedKey);
+                                        }
+                                        else
+                                        {
+                                            property.WriteTo(writer);
```
peterrsongg, line 231: I'm a little bit confused by this `else` statement, maybe I'm not understanding correctly but if we matched the key that means we must update that condition with the new condition. Why are we then writing the property to the writer if it doesn't match? Aren't we already doing that if `shouldCopyCondition` is true?

sdk/src/Services/S3/Custom/Util/S3PostUploadSignedPolicy.cs
GarrettBeatty: fixed @muhammad-othman

sdk/src/Services/S3/Custom/Util/S3PostUploadSignedPolicy.cs
GarrettBeatty: The reason we need this `else` is because __a condition object might contain multiple properties, not just the one we're updating__.

For example, imagine we have a condition object in the policy that looks like this:

```json
{
    "x-amz-meta-uuid": "14365123651274",
    "x-amz-meta-tag": "metadata"
}
```

And let's say we want to update `x-amz-meta-tag` to a new value "updated". The code does:

1. Detects this condition contains a key we want to update (`x-amz-meta-tag`)

2. Sets `shouldCopyCondition = false` and `matchedKey = "x-amz-meta-tag"`

3. Creates a new object with `WriteStartObject()`

4. Iterates through each property in the original object:

   - For `x-amz-meta-uuid`: The name doesn't match our target key, so it goes to the `else` clause and we write the original property unchanged
   - For `x-amz-meta-tag`: The name matches our target key, so we write the new value

Without the `else` statement, any properties in the condition object that aren't being updated would be lost.

sdk/src/Services/S3/Custom/Util/S3PostUploadSignedPolicy.cs
GarrettBeatty: updated

sdk/src/Services/S3/Custom/Util/S3PostUploadSignedPolicy.cs
GarrettBeatty: ive also updated the code to have detailed comments

peterrsongg: ah I see thanks for clarifying that makes sense

GarrettBeatty:
ive retested with my console app and it works on latest commits

peterrsongg (approved):

muhammad-othman (approved):

PR #3905, opened 2025-07-07
CBOR response unmarshallers and related protocol test

muhammad-othman:
## Description
- Implement Cbor response unmarshallers.
- Implement Cbor structure unmarshallers.
- Add shared base classes like `CborUnmarshallerContext` and `CborResponseUnmarshaller` to `AWSSDK.Extensions.CborProtocol` package.
- Generate CBOR response protocol tests.
- Didn't need to `CloudWatch` or `SecretsManager` unmarshallers this time since we have the protocol tests unmarshallers for reviewing the generator output.

## Motivation and Context
- `DOTNET-8161`
- `DOTNET-8162`
- `DOTNET-8165`

## Testing
- Added `smithy-rpc-v2-cbor` to CloudWatch and SecretsManager models `protocols`, generated the marshallers/unmarshallers, and executed some requests using CBOR marshallers and unmarshallers.
- CBOR response protocol tests.
- `DRY_RUN-36078c9c-efcb-46d7-95e3-a6052e91aa33`.

## Types of changes

## License


generator/ProtocolTestsGenerator/smithy-dotnet-codegen/src/main/java/software/amazon/smithy/dotnet/codegen/HttpProtocolTestGenerator.java
```diff
@@ -98,7 +98,7 @@ private void addServiceProtocolSpecificImports() {
             writer.addImport(serviceName, "System.Xml");
             writer.addImport(serviceName, "System.Xml.Linq");
         } else if (this.serviceName.toLowerCase().contains("rpcv2")) {
-            writer.addImport(serviceName, "AWSSDK.Extensions.CborProtocol.Internal");
+            writer.addImport(serviceName, "AWSSDK.Extensions.CborProtocol.Internal.Transform");
```
peterrsongg: This is just a general comment, but I don't see the "QueryCompatbile" protocol tests added here. The CBOR protocol-selection SEP says that a service using the CBOR protocol can make use of the `awsquerycompatible` trait. I'm fine with adding this in a separate PR but I was wondering if that was on your radar

[query-compatbile tests](https://github.com/smithy-lang/smithy/blob/main/smithy-aws-protocol-tests/model/rpcv2Cbor/query-compatible.smithy)

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/Transform/CborResponseUnmarshaller.cs
```diff
@@ -0,0 +1,75 @@
...
+        protected override UnmarshallerContext ConstructUnmarshallerContext(Stream responseStream, bool maintainResponseBody, IWebResponseData response, bool isException)
+        {
+            return new CborUnmarshallerContext(responseStream, maintainResponseBody, response, isException, null);
+        }
+
+        protected override UnmarshallerContext ConstructUnmarshallerContext(Stream responseStream, bool maintainResponseBody, IWebResponseData response, bool isException, IRequestContext requestContext)
+        {
+            return new CborUnmarshallerContext(responseStream, maintainResponseBody, response, isException, requestContext);
+        }
+    }
```
peterrsongg, line 74: Do we need to override
```
ShouldReadEntireResponse(IWebResponseData response, bool readEntireResponse)
```

It is defined in the base class so I'm just wondering if you left it out intentionally

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/Transform/CborUnmarshallerContext.cs
```diff
@@ -0,0 +1,210 @@
...
+                            throw;
+                        }
+                        ArrayPool<byte>.Shared.Return(tempBuffer);
+                        tempBuffer = newBuffer;
+                    }
+                }
+
+                // Create a new byte array to hold only the read data.
+                var actualBytes = new byte[totalRead];
+                Buffer.BlockCopy(tempBuffer, 0, actualBytes, 0, totalRead);
```
peterrsongg: doesn't tempBuffer already contain all the data you need? Why do you need to create another byte array and copy the contents of tempBuffer to actualBytes?

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/Transform/CborResponseUnmarshaller.cs
muhammad-othman: Yeah, didn't need to override it here since we are reading the response anyway for the `CborReader`.

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/Transform/CborUnmarshallerContext.cs
muhammad-othman: When we use `ArrayPool<byte>.Shared.Rent(someSize)` it may give us a buffer of larger size, if we passed this to  `CborReader` it wont know that it should reading data after the size that we specified, so we have to copy only the data that we need to `CborReader`.
I wasn't sure about this whole approach at the beginning since we allocate some additional memory just to pass it to `CborReader`, but I did some benchmarking and even with that it still allocates less memory than `JSON`, and much less than `XML`

generator/ProtocolTestsGenerator/smithy-dotnet-codegen/src/main/java/software/amazon/smithy/dotnet/codegen/HttpProtocolTestGenerator.java
muhammad-othman: Looks like these were added two weeks ago, after I added the protocol tests, will include them in the next PR.

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/Transform/CborUnmarshallerContext.cs
peterrsongg: Ah okay, i remember running into this as well when using array pools. makes sense.

peterrsongg (approved):

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/Transform/CborErrorResponseUnmarshaller.cs
```diff
@@ -0,0 +1,58 @@
...
+using Amazon.Runtime.Internal;
+using Amazon.Runtime.Internal.Transform;
+using Amazon.Runtime.Internal.Util;
+using Amazon.Util;
+using System;
+using System.Collections.Generic;
+using System.Formats.Cbor;
+using System.IO;
+
+namespace AWSSDK.Extensions.CborProtocol.Internal.Transform
```
normj: Part of weird legacy is we never use or should never use `AWSSDK` in the namespace name. For example `AWSSDK.Extensions.NETCore.Setup` uses `Amazon.Extension.NETCore.Setup` as the namespace. I know it is weird but I want to be consistent and not have 1% of code use a different namespace.

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/Transform/CborSimpleTypeUnmarshaller.cs
```diff
@@ -0,0 +1,559 @@
...
+    static class CborSimpleTypeUnmarshaller<T>
+    {
+        public static T Unmarshall(CborUnmarshallerContext context)
+        {
+            var reader = context.Reader;
+            object value;
+
+            if (
+                (
+                    typeof(T) == typeof(int?) ||
```
normj, line 36: nit: Do we need all the type checks? If `reader.PeekState()` is `null` then shouldn't we always call `reader.ReadNull()` and set `value` to `null`?

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/Transform/CborSimpleTypeUnmarshaller.cs
```diff
@@ -0,0 +1,559 @@
...
+                    typeof(T) == typeof(float?) ||
+                    typeof(T) == typeof(string)
+                )
+                && reader.PeekState() == CborReaderState.Null
+            )
+            {
+                reader.ReadNull();
+                value = null;
+            }
+            else if (typeof(T) == typeof(string))
```
normj, line 50: Don't we need DateTime simple unmarsahlers?

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/Transform/CborSimpleTypeUnmarshaller.cs
```diff
@@ -0,0 +1,559 @@
...
+                    typeof(T) == typeof(double?) ||
+                    typeof(T) == typeof(bool?) ||
+                    typeof(T) == typeof(float?) ||
+                    typeof(T) == typeof(string)
+                )
+                && reader.PeekState() == CborReaderState.Null
+            )
+            {
+                reader.ReadNull();
+                value = null;
```
normj: Should `value` be set to `default(T)` so types like `int` would be `0`?

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/Transform/CborUnmarshallerContext.cs
normj: Instead of copying the whole value can't we just wrap the byte buffer in a properly sized ReadOnlyMemory struct? That would avoid the copying?

```
var actualBytes = new ReadOnlyMemory<byte>(tempBuffer, 0, totalRead);
```

When we do that we would not return the `tempBuffer` to the array pool since passed the buffer to another object.

normj (changes requested):

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/Transform/CborUnmarshallerContext.cs
muhammad-othman: This could be a good improvement, but I guess we should keep a reference to the `tempBuffer` then return it when we dispose the Unmarshaller, otherwise we may end up with unnecessary allocations.
Will do some testing and make this change.

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/Transform/CborErrorResponseUnmarshaller.cs
muhammad-othman: Updating the namespace.

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/Transform/CborSimpleTypeUnmarshaller.cs
muhammad-othman: I was trying to be extra protective, when this method was get called with another type that we don't supported yet we should get the following exception even if the value is null
```
throw new NotSupportedException(
                    $"CBOR deserialization for type {typeof(T)} is not supported."
                );
```

Also it means something is wrong if we get null value for any of the non nullable value type, if we get `null` for an `int`, it will go to `reader.ReadInt32();` which will throw an exception as it wont be able to read it, but we we removed the `typeof` conditions it will just return `null` (or `0` if we applied the next comment).

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/Transform/CborSimpleTypeUnmarshaller.cs
muhammad-othman: `int` cannot be `null` and `reader.PeekState() == CborReaderState.Null` wont be true, this is related to the above comment and being extra protective.
But if we kept the the `typeof` conditions then returning  `default(T)` will give the same result as we made sure the type is nullable, but I feel it is more clear that way (we read `null` and we return `null`) but since it is the same output I'm okay with changing it.

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/Transform/CborSimpleTypeUnmarshaller.cs
muhammad-othman: DateTime is a `double` that are reading using  `CborDateTimeUnmarshaller.Unmarshall`
image


From the SEP
```
Timestamps are serialized as Major type 6, value tag 1 (epoch-based date/time) to indicate the tagged value is an epoch timestamp. Values are either major type 0 (unsigned integer) or major type 1 (negative integer) for integer values or major type 7, value tags 25 (half-precision float), 26 (single-precision float), or 27 (double-precision float) for floating-point values.
```

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/Transform/CborSimpleTypeUnmarshaller.cs
normj: Ok

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/Transform/CborErrorResponseUnmarshaller.cs
muhammad-othman: Updated the namespaces to `Amazon.Extensions.CborProtocol`.

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/Transform/CborSimpleTypeUnmarshaller.cs
muhammad-othman: Switched to `default(T)`.

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/Transform/CborUnmarshallerContext.cs
muhammad-othman: @normj Updated the final copy to use `ReadOnlyMemory` and I can see some improvement in the memory allocations over the previous approach.

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/Transform/CborUnmarshallerContext.cs
```diff
@@ -0,0 +1,213 @@
...
+        public string PopPathSegment()
+        {
+            return _pathStack.Pop();
+        }
+
+        protected override void Dispose(bool disposing)
+        {
+            if (!this.disposed)
+            {
+                if (disposing && rentedBuffer != null)
```
normj, line 207: The member variable `rentedBuffer` is never set and is always `null` because `CreateCborReaderFromStream` which is static is working with a local `rentedBuffer` variable. Some logic still needs to be fixed up for the `ArrayPool<byte>.Shared.Return` to work here.

muhammad-othman: lol, how did I get some allocations improvements 🤔

muhammad-othman: Aha, I guess we get the benefits of `ReadOnlyMemory` even if we didn't return the buffer

muhammad-othman: Updated it to set the private member, gonna compare it to the previous version to see if returning the memory made any difference.

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/Transform/CborUnmarshallerContext.cs
```diff
@@ -0,0 +1,213 @@
...
+                {
+                    // Expand the buffer size by doubling it
+                    var newBuffer = ArrayPool<byte>.Shared.Rent(rentedBuffer.Length * 2);
+                    try
+                    {
+                        Buffer.BlockCopy(rentedBuffer, 0, newBuffer, 0, totalRead);
+                    }
+                    catch
+                    {
+                        ArrayPool<byte>.Shared.Return(newBuffer);
```
normj: Since we are exiting out due to an exception shouldn't we also `Return` the `rentedBuffer` as well?

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/Transform/CborUnmarshallerContext.cs
```diff
@@ -0,0 +1,213 @@
...
+            );
+        }
+
+        private CborReader CreateCborReaderFromStream(Stream stream, long? streamSize = null)
+        {
+            int totalRead = 0;
+            if (streamSize.HasValue)
+            {
+                // If we know the size, we can read directly into a buffer of exact size
+                var buffer = new byte[streamSize.Value];
```
normj: Can we do the same trick as below and rent a buffer here and return as part of the dispose?

muhammad-othman: Trying this made me think that we will have an issue if the stream size is larger than `int.MaxValue` since arrays size should be an int, in `JSON` and `XML` where we can read chunks directly from the stream so it doesn't matter if we have a stream larger than 2GB, but here the `CborReader` needs the whole buffer since it suppose to improve its performance being in-memory parser.
I'm not sure if a `Cbor` payload can be larger than that, will double check.

normj (approved):

PR #3909, opened 2025-07-10
S3 Gen Phase 2: Generate PutBucket

peterrsongg:
## Description

This generates PutBucket. The major changes are
1. Added InheritAlternateBaseClass customization for PutBucketRequest to inherit from PutWithAclRequest as defined in custom
2. I moved `PostMarshallCustomization` to before the `try` `catch` block where we set the content because we could be altering the content in that method
3. There is some custom logic that exists for `PutBucketRequestMarshaller` where the user could set `BucketRegion` or `BucketRegionName` and not have to create a `PutBucketConfiguration` object and we would create the appropriate xml for the user. I moved that logic to the setters of both `BucketRegion` and `BucketRegionName`
4. I excluded some un-used members since `PutBucketRequest` has logic with setting ACLs

## Motivation and Context

S3 Generation Phase 2.
## Testing

I did some manual checks to make sure that a user could still only specify the bucket region and name and create a bucket without having to create the `PutBucketConfiguration` object themselves.

Assembly comparer output (empty)
```
Already on 'petesong/generate-s3-phase-2-pr1'
  AWSSDK.Core.NetStandard net8.0 succeeded (39.0s) → C:\Dev\Repos\aws-sdk-net-staging\sdk\src\Core\bin\Debug\net8.0\AWSSDK.Core.dll
  AWSSDK.S3.NetStandard net8.0 succeeded (8.8s) → bin\Debug\net8.0\AWSSDK.S3.dll

Build succeeded in 53.0s

Workload updates are available. Run `dotnet workload list` for more information.
Switched to branch 'development'
Your branch is up to date with 'public/development'.
  AWSSDK.Core.NetStandard net8.0 succeeded (5.3s) → C:\Dev\Repos\aws-sdk-net-staging\sdk\src\Core\bin\Debug\net8.0\AWSSDK.Core.dll
  AWSSDK.S3.NetStandard net8.0 succeeded (4.5s) → bin\Debug\net8.0\AWSSDK.S3.dll

Build succeeded in 10.3s
```
Dry_Run passes caeb33aa-9f50-4418-af23-d054dcd47197
## Types of changes

## Checklist

## License


generator/ServiceClientGeneratorLib/Generators/Marshallers/RestXmlRequestMarshaller.tt
```diff
@@ -181,17 +187,41 @@ WriteXmlAttributeString(level, member, variableName, isPayload: false);
 #>
 				if (<#=variableName + ".IsSet" + operation.RequestPayloadMember.PropertyName#>())
 				{
+<#+
+					// S3 doesn't follow the rule where if the request payload member's location name is the same as the payload member name, we marshall with the payload member's shape's marshall name instead
```
peterrsongg, line 191: without this, we would incorrectly write `PutBucketConfiguration` as the xml element name when it should be `CreateBucketConfiguration`. S3 doesn't follow this rule.

sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/PutObjectRetentionRequestMarshaller.cs
```diff
@@ -97,7 +97,7 @@ public IRequest Marshall(PutObjectRetentionRequest publicRequest)
             {   
                 if (publicRequest.IsSetRetention())
                 {
-                    xmlWriter.WriteStartElement("ObjectLockRetention", "http://s3.amazonaws.com/doc/2006-03-01/");
+                    xmlWriter.WriteStartElement("Retention", "http://s3.amazonaws.com/doc/2006-03-01/");
```
normj, line 100: I assume this was a benign bug that we were using the wrong root element name. I see in the docs `Retention` is the correct value.

peterrsongg: yup, i also tested the operation when i did a sweep of all released operations

muhammad-othman (approved):

PR #3922, opened 2025-07-18
Skip unknown properties when mapping EndpointTests V3

muhammad-othman:
## Description
Skip unknown properties when mapping EndpointTests

## Motivation and Context
This PR change allows the .NET SDKs to ignore unknown properties so the endpoints team can make additive changes and add new endpoint properties without requiring support from the .NET SDK.  
This way they can ship the endpoint rules updates and SDKs will benefit from the addition whenever they decide to implement the feature instead of having to wait for all SDKs to schedule and complete the work.

## Testing
* `DRY_RUN-22399aa8-3385-40dd-a5d6-663347c06e65`
* Ran the generator using some models that includes unknown endpoint properties.

## Types of changes

## Checklist

## License

### DIFF
--- a/generator/.DevConfigs/1e81788b-4e71-4457-9348-d81a456886b7.json
+++ b/generator/.DevConfigs/1e81788b-4e71-4457-9348-d81a456886b7.json
@@ -1,0 +1,8 @@
+{
+  "core": {
+    "changeLogMessages": [
+      "Skip unknown properties when mapping EndpointTests"
+    ],
+    "type": "patch"
+  }
+}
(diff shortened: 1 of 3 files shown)


normj (approved):
Approved but the alternative is how @dscpinheiro made this change into the runtime version of this library in V3 and not add the public `skipUnknownProperties` and just always skip. https://github.com/aws/aws-sdk-net/commit/806f777e5012a9b055d426851a456561e4261299#diff-2a221ac8c2d33c868a08b6d34a568a1ca388d40e03de832776cbb4fcc611dc69

muhammad-othman:
@normj I wanted to avoid this incase new properties were added and we don't supported yet.
Just allowed EndpointProperties since the SEP mentioned we should allow unknown properties, but for the other parts of the models we probably should fail if we face something we don't recognize.

ashishdhingra (approved):

PR #3923, opened 2025-07-19
Add managed SigV4a signer.

### LINKED ISSUE #3881: Add managed SigV4a signer.
### Describe the feature

Using the AWS SigV4a signature scheme it in the AWS SDK for .NET currently requires an extra NuGet dependency to the native `aws-c-auth` library. However, all the frameworks supported by V4 have the required elliptic curve cryptography APIs to implement SigV4a entirely in managed code.

### Use Case

Eliminate optional dependency to native code and increase platform compatibility.

### Proposed Solution

* `AWS4aSigner` class gets added.
* `AWS4aSignerCRTWrapper` class and all related APIs in `AWSSDK.Core` and `AWSSDK.Extensions.CrtIntegration` get deprecated.

### Other Information

I have started an implementation on my machine, will open a PR soon.

### Acknowledgements

- [x] I may be able to implement this feature request

AWSSDK.Core V4

### Targeted .NET Platform

All supported frameworks

All supported operating systems

teo-tsirpanis:
## Description
This PR adds a managed implementation of the AWS SigV4a algorithm, contained in the newly added `Amazon.Runtime.Internal.Auth.AWS4aSigner` class. All uses of the old signer that forward to the CRT were replaced. The CRT signer and related APIs were deprecated, but were left otherwise untouched.

## Motivation and Context
Fixes #3881.

## Testing
* Added a small unit test for the signing key derivation process, ported from the existing tests of the CRT signer in `extensions`.
  * There were several more tests that were not moved over at the moment, because of a perceived low value, considering that: they would necessitate writing a managed SigV4a verifier as well, equivalent tests for SigV4 do not exist, and that the code paths that use the SigV4 and SigV4a signers have become more similar, which reduces the room for error.
* Tested by performing a request and generating a pre-signed URL, on an S3 multi-region access point.
* There are already several existing integration tests (which I did not run), that use SigV4a signing.

## Types of changes

## Checklist

## License


sdk/src/Core/Amazon.Runtime/Credentials/ImmutableCredentials.cs
```diff
@@ -55,6 +56,14 @@ public class ImmutableCredentials
...
         /// </summary>
         public string AccountId { get; private set; }
+
+        /// <summary>
+        /// Gets a cached AWS4a signing key for the current credentials.
+        /// </summary>
+        /// <remarks>
+        /// The key must not be disposed, and must not be returned to the user.
+        /// </remarks>
+        internal ECDsa AWS4aSigningKey { get; set; }
```
teo-tsirpanis: I had the idea of caching the signing key in the `ImmutableCredentials` instance, but it will not work in practice, because we are cloning the immutable credentials before passing them to the signer ([example](https://github.com/aws/aws-sdk-net/blob/9b002cd7e308fc79e2ed80954dcb707f859c2bf4/sdk/src/Core/Amazon.Runtime/Credentials/BasicAWSCredentials.cs#L76)), which doesn't make sense because the immutable credentials objects are immutable.

What to do? Should I open another PR to remove the copyings?

sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4aSigner.cs
```diff
@@ -0,0 +1,594 @@
...
+        private static byte[] ConvertToRfc3279DerSequence(byte[] signature)
+        {
+            var writer = new AsnWriter(AsnEncodingRules.DER);
+            writer.PushSequence();
+            writer.WriteIntegerUnsigned(signature.AsSpan(0, signature.Length / 2)); // R value
+            writer.WriteIntegerUnsigned(signature.AsSpan(signature.Length / 2)); // S value
+            writer.PopSequence();
+            return writer.Encode();
+        }
+#endif
```
teo-tsirpanis, line 554: The fact that we format the signature in ASN.1 is [not documented anywhere](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html), and stumped me for some days. I sent feedback to update the documentation.

sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4aSigningResult.cs
```diff
@@ -99,6 +120,7 @@ public string RegionSet
         /// <summary>
         /// Returns the full presigned Uri
         /// </summary>
+        [Obsolete("This property is always empty in objects returned by AWS4aSigner. Use the ForQueryParameters property instead, to get the query parameters for a presigned URL.")]
```
teo-tsirpanis, line 123: There is a difference in how the SigV4 and SigV4a signing result types return presigned URLs. I think that obsoleting this SigV4a-only property and unifying with SigV4 is the better choice; for reference, there are [zero usages](https://github.com/search?q=%22.PresignedUri%22+language%3AC%23+NOT+is%3Afork&type=code) of this property outside of the S3 library, which will need the Core bump from the replacement of `AWS4aSignerCRTWrapper` either way.

sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4aSigner.cs
```diff
@@ -0,0 +1,564 @@
...
+        #endregion
+    }
+
+    /// <summary>
+    /// AWS4a protocol signer for Amazon S3 presigned urls.
+    /// </summary>
+    public static class AWS4aPreSignedUrlSigner
```
teo-tsirpanis, line 561: I had the idea of making this a static class, unlike the SigV4 counterpart, which mostly has static members and an overriden method that always throws.

generator/.DevConfigs/b495f893-31aa-4e17-866c-af78e4da6e8b.json
teo-tsirpanis, line 1: This might need to be updated to account for the fact that the S3 library needs a newer Core version. How to write this?

dscpinheiro:
I know it's been a while but we're finally starting to look at this PR (I've reached out to the AppSec team at AWS asking for them to review it as well). I also tried to run this branch in our build systems and some existing tests (such as the ones in https://github.com/aws/aws-sdk-net/blob/main/sdk/test/Services/EventBridge/UnitTests/Custom/EventBridgeMRAPTests.cs) failed:
```
Message: 
  Test method AWSSDK.UnitTests.EventBridgeMRAPTests.UseOverrideEndpointWhenEndpointIdIsSet threw exception: 
  System.Security.Cryptography.CryptographicException: The specified key parameters are not valid. Q.X and Q.Y are required fields. Q.X, Q.Y must be the same length. If D is specified it must be the same length as Q.X and Q.Y for named curves or the same length as Order for explicit curves.

Stack Trace: 
  ECParameters.Validate()
  ECCng.ParametersToBlob(ECParameters& parameters, Func`3 namedCurveResolver, Func`2 explicitCurveResolver, CngKeyBlobFormat& format, String& curveName)
  ECDsaCng.ImportParameters(ECParameters parameters)
  ECDsa.Create(ECParameters parameters)
  AWS4aSigner.ComputeSigningKey(String awsAccessKey, String awsSecretAccessKey)
  // etc...
```

Our build system tests on Windows, so I believe that may be related.

teo-tsirpanis:
Thanks for the update @dscpinheiro. The unit test I added succeeded on .NET 8, but I reproduced the failure on .NET Framework, and upon closer look it does not support importing EC parameter sets with only $d$ set. Modern .NET added support relatively recently (dotnet/runtime#33874), which means that the PR is blocked until support gets backported to .NET Framework, if it meets the bar. c.c. @vcsjones

This also means that SigV4a signing will fail on .NET Core 3.1 as well, which is long out of support but is still targeted by tests.

Update: opened .NET Framework backport request in https://developercommunity.visualstudio.com/t/Cannot-import-elliptic-curve-parameter-s/11058177

dscpinheiro:
Thanks for opening that feature request, will keep an eye on it (we'd definitely prefer not having to write custom cryptography code in the SDK if we can avoid it). 

On a side note I did another pass of the PR and only had two minor comments:
- In `sdk/src/Core/Amazon.Runtime/Internal/Util/ChunkedUploadWrapperStream.cs`, you're using `CHUNK_STRING_TO_SIGN_PREFIX` and `TRAILING_HEADER_STRING_TO_SIGN_PREFIX` for both SigV4 and SigV4A. We should use the correct values for SigV4A (`AWS4-ECDSA-P256-SHA256-PAYLOAD` and `AWS4-ECDSA-P256-SHA256-TRAILER`) instead.
- We also have quite a few integration tests that use SigV4A (the S3 solution is the best place to look). They also run on Windows so we'll need to confirm they pass once we decide on the path forward.

teo-tsirpanis:
I fixed `ChunkedUploadWrapperStream`. With that, the `MultiRegionAccessPointsTests` succeed locally, after pointing them to my test MRAP and porting them to .NET 8 (some low-hanging fruit from the effort has been pushed).

vcsjones:
Personally I do not know if this is likely to be backported to any version of .NET Framework. The backport bar to .NET Framework is _very_ high.

One thing that might work, specifically in the case of .NET Framework, is to create a [PKCS#8](https://datatracker.ietf.org/doc/html/rfc5208) with an [`ECPrivateKey`](https://datatracker.ietf.org/doc/html/rfc5915) in it. You can then import that with `CngKey.Import` and a key format of `Pkcs8PrivateBlob`. 

In an ECPrivateKey, the private portion, D, is required, but the public Q parameters `publicKey` is optional.

Then pass that key to a new instance of `ECDsaCng` (a subclass of `ECDsa`).

So:

```C#

ECDsa key;

#if NETFRAMEWORK
  byte[] assemblePkcsPrivateKey;
  CngKey cngKey = CngKey.Import(assemblePkcsPrivateKey, CngKeyBlobFormat. Pkcs8PrivateBlob);
  key = new ECDsaCng(cngKey);
#else
  key = ECDsa.Create();
  // use key.ImportParameters with D populated
#endif
```

The Windows CNG key importer _should_ handle the PKCS#8 of the ECPrivateKey omitting the `publicKey`.

It’s not exactly trivial to implement, but it’s “just” assembling some ASN.1. No crypto point multiplication required.

teo-tsirpanis:
Thanks for chiming in @vcsjones. Upon further investigation, we could simply import a `BCRYPT_ECCPRIVATE_BLOB` with the public parameters set to all zeroes, which is [what .NET does under the hood](https://github.com/dotnet/runtime/blob/c38f37a3a1c39ca921fe156d472990d259977f6f/src/libraries/Common/src/System/Security/Cryptography/ECDsaCng.ImportExport.cs#L58-L66). I will give it a try soon.

teo-tsirpanis:
I updated the .NET Framework implementation to construct and import a `BCRYPT_ECCPRIVATE_BLOB`. With this, `MultiRegionAccessPointsTests` succeed on .NET Framework too.

dscpinheiro:
I ran the latest changes through our build system and there were 3 test failures (2 of them are the same but in different methods):
- `DeriveSigningKey` failed in .NET Core 3.1 (I know... But still supported in the SDK) with `System.Security.Cryptography.CryptographicException: The specified key parameters are not valid. Q.X and Q.Y are required fields. Q.X, Q.Y must be the same length. If D is specified it must be the same length as Q.X and Q.Y for named curves or the same length as Order for explicit curves.`
- Both `TestV4aSignedMultipartUpload` and `PutObjectChunked` (in the S3 solution) failed with `System.ArgumentException: The first 9 bits of the integer value all have the same value. Ensure the input is in big-endian byte order and that all redundant leading bytes have been removed.`
```
   at System.Formats.Asn1.AsnWriter.WriteIntegerUnsignedCore(Asn1Tag tag, ReadOnlySpan`1 value)
   at System.Formats.Asn1.AsnWriter.WriteIntegerUnsigned(ReadOnlySpan`1 value, Nullable`1 tag)
   at Amazon.Runtime.Internal.Auth.AWS4aSigner.ConvertToRfc3279DerSequence(Byte[] signature) in /_/sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4aSigner.cs:line 450
   /// etc.
```

teo-tsirpanis:
I fixed the signature format conversion. I copied the .NET sources that do it and added links to them, hoping it will make security review easier.

With regards to .NET Core 3.1, there's not much we can do. I thought of importing the CNG key like on .NET Framework, but I'm not sure how much helpful it would be, since it would still fail on non-Windows platforms. Are there plans to drop .NET Core 3.1 support? The AWS SDK would still be installable and (mostly) usable on .NET Core 3.1 after stopping explicit targeting and testing, through the .NET Standard 2.0 binaries.

dscpinheiro:
> Are there plans to drop .NET Core 3.1 support? 

Yes, this is something we plan to do in the next minor version of the SDK (i.e. `4.1`), but that likely won't happen until later this year (after V3 reaches end of support - https://aws.amazon.com/blogs/developer/aws-sdk-for-net-v3-maintenance-mode-announcement/).

teo-tsirpanis:
Thanks; I guess we can't merge it until then. No problem, I can wait.

dscpinheiro:
Update: We just published https://aws.amazon.com/blogs/developer/annual-net-target-updates-for-the-aws-sdk-for-net/, which includes our plan to drop support for .NET Core 3.1 (the reason the PR was marked as "do not merge").

We can't commit to a specific date (as the post says _We will attempt to align as closely as possible with the .NET November release cycle but may fall back to December depending on demands for other priorities, particularly for [AWS re:Invent](https://aws.amazon.com/events/reinvent/)_), but the managed SigV4A signer is something we plan to include in our V4.1 release.

dscpinheiro:
Sorry for more conflicts, but I just changed the destination branch here to `feature/4.1` (that's where we'll keep the changes for the next minor version).

Sometime next week we'll also create a tracking issue for `4.1` (similar to what we had for the V4 GA tracker).

teo-tsirpanis:
Conflicts resolved. I also did some upkeeping.

sdk/src/Core/Amazon.Runtime/Internal/Util/ChunkedUploadWrapperStream.cs
```diff
@@ -462,24 +449,26 @@ private string ConstructSignedTrailersChunk()
...
-                    HeaderSigningResult.Scope + "\n" +
-                    PreviousChunkSignature + "\n" +
-                    AWSSDKUtils.ToHex(AWS4Signer.ComputeHash(canonicalizedTrailingHeaders), true);
+            var prefix = HeaderSigningResult is AWS4aSigningResult ? V4A_TRAILING_HEADER_STRING_TO_SIGN_PREFIX : V4_TRAILING_HEADER_STRING_TO_SIGN_PREFIX;
+            var chunkStringToSign =
+                prefix + "\n" +
+                HeaderSigningResult.ISO8601DateTime + "\n" +
+                HeaderSigningResult.Scope + "\n" +
+                PreviousChunkSignature + "\n" +
+                AWSSDKUtils.ToHex(AWS4Signer.ComputeHash(canonicalizedTrailingHeaders), true);
```
martincostello, line 461: While not new code, would it be better to do this with a `StringBuilder`?

sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4aSigningResult.cs
```diff
@@ -87,6 +89,25 @@ public override string ForAuthorizationHeader
...
+        {
+            get
+            {
+                var authParams = new StringBuilder()
+                    .AppendFormat("{0}={1}", AWSSDKUtils.UrlEncode(HeaderKeys.XAmzAlgorithm, false), AWSSDKUtils.UrlEncode(AWS4Signer.AWS4aAlgorithmTag, false))
+                    .AppendFormat("&{0}={1}", AWSSDKUtils.UrlEncode(HeaderKeys.XAmzRegionSetHeader, false), AWSSDKUtils.UrlEncode(RegionSet, false))
+                    .AppendFormat("&{0}={1}", AWSSDKUtils.UrlEncode(HeaderKeys.XAmzCredential, false), AWSSDKUtils.UrlEncode(string.Format(CultureInfo.InvariantCulture, "{0}/{1}", AccessKeyId, Scope), false))
+                    .AppendFormat("&{0}={1}", AWSSDKUtils.UrlEncode(HeaderKeys.XAmzDateHeader, false), AWSSDKUtils.UrlEncode(ISO8601DateTime, false))
+                    .AppendFormat("&{0}={1}", AWSSDKUtils.UrlEncode(HeaderKeys.XAmzSignedHeadersHeader, false), AWSSDKUtils.UrlEncode(SignedHeaders, false))
+                    .AppendFormat("&{0}={1}", AWSSDKUtils.UrlEncode(HeaderKeys.XAmzSignature, false), AWSSDKUtils.UrlEncode(Signature, false));
```
martincostello, line 105: Would it be more performant to avoid `AppendFormat` and just append the individual elements directly?

```csharp
.Append(AWSSDKUtils.UrlEncode(HeaderKeys.XAmzAlgorithm, false))
.Append('=')
.Append(AWSSDKUtils.UrlEncode(AWS4Signer.AWS4aAlgorithmTag, false))
.Append('&')
.Append(AWSSDKUtils.UrlEncode(HeaderKeys.XAmzRegionSetHeader, false))
.Append('=')
// etc
```

sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4aSigner.cs
```diff
@@ -0,0 +1,755 @@
...
+                                                         RequestMetrics metrics)
+        {
+            var dateStamp = FormatDateTime(signedAt, AWSSDKUtils.ISO8601BasicDateFormat);
+            var scope = string.Format(CultureInfo.InvariantCulture, "{0}/{1}/{2}", dateStamp, service, Terminator);
+
+            var stringToSignBuilder = new StringBuilder();
+            stringToSignBuilder.AppendFormat(CultureInfo.InvariantCulture, "{0}\n{1}\n{2}\n",
+                                             AWS4aAlgorithmTag,
+                                             FormatDateTime(signedAt, AWSSDKUtils.ISO8601BasicDateTimeFormat),
+                                             scope);
```
martincostello, line 305: Similar here.

sdk/src/Core/Amazon.Runtime/Internal/Util/ChunkedUploadWrapperStream.cs
teo-tsirpanis: It would be better, but I'm not going to do it in this PR to minimize scope creep. Same with your proposed changes to other parts, which match the equivalent code in SigV4.

generator/.DevConfigs/b495f893-31aa-4e17-866c-af78e4da6e8b.json
dscpinheiro: I know this comment is from last year, but we have a config option to make sure the service packages require latest Core (already set in https://github.com/aws/aws-sdk-net/blob/feature/4.1/generator/.DevConfigs/04942b20-252a-4c6a-9136-a4f58b3cfaa0.json so no action needed here).

sdk/test/Services/S3/IntegrationTests/CapacityManagerTests.cs
```diff
@@ -92,18 +92,18 @@ public void S3CapacityManagerIntegrationTest()
...
                     retryFlag = true;
                     requestCount = 0;
-                    FailureRetryRequests(TotalRequests, RetryRequests, ExtraRequests, client);
+                    await FailureRetryRequests(TotalRequests, RetryRequests, ExtraRequests, client);
                 }
             }
         }
 
-        private static void FailureRetryRequests(int totalRequests, int retryRequests, int extraRequests, AmazonS3Client client)
+        private static async Task FailureRetryRequests(int totalRequests, int retryRequests, int extraRequests, AmazonS3Client client)
```
dscpinheiro, line 101: This entire class is `#if NETFRAMEWORK` only so it failed to compile for me (on my Windows machine). 

It needs a `using System.Threading.Tasks;` at the top and marking `S3CapacityManagerIntegrationTest` as async too.

sdk/test/UnitTests/Custom/Runtime/AWS4aSignerTests.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        private const string SigningTestAccessKeyId = "AKIDEXAMPLE";
+        private const string SigningTestSecretAccessKey = "wJalrXUtnFEMI/K7MDENG+bPxRfiCYEXAMPLEKEY";
+
+        /* The public coordinates of the ecc key derived from the above credentials pair */
+        private const string SigningTestEccPubX = "b6618f6a65740a99e650b33b6b4b5bd0d43b176d721a3edfea7e7d2d56d936b1";
+        private const string SigningTestEccPubY = "865ed22a7eadc9c5cb9d2cbaca1b3699139fedc5043dc6661864218330c8e518";
+
+        [TestMethod]
+        public void DeriveSigningKey()
```
dscpinheiro, line 34: This test failed for me (also on Windows but same error for all TFMs):
```
 DeriveSigningKey
   Source: AWS4aSignerTests.cs line 33
   Duration: 93 ms

  Message: 
Assert.AreEqual failed. Expected:<System.Byte[]>. Actual:<System.Byte[]>. 

  Stack Trace: 
AWS4aSignerTests.DeriveSigningKey() line 37
MethodBaseInvoker.InterpretedInvoke_Method(Object obj, IntPtr* args)
MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```

My environment details if it matters:
```
PS> dotnet --info
.NET SDK:
 Version:           11.0.100-preview.7.26381.103
 Commit:            e2c1e00b3d
 Workload version:  11.0.100-manifests.2ce71a7f
 MSBuild version:   18.10.0-1.26381.103+e2c1e00b3

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.26200
 OS Platform: Windows
 RID:         win-x64
 Base Path:   C:\Program Files\dotnet\sdk\11.0.100-preview.7.26381.1034

Host:
  Version:      11.0.0-preview.7.26381.103
  Architecture: x64
  Commit:       e2c1e00b3d

.NET SDKs installed:
  8.0.424 [C:\Program Files\dotnet\sdk]
  9.0.317 [C:\Program Files\dotnet\sdk]
  10.0.400 [C:\Program Files\dotnet\sdk]
  11.0.100-preview.7.26381.103 [C:\Program Files\dotnet\sdk]
```

sdk/src/Core/AWSSDK.Core.NetStandard.csproj
```diff
@@ -82,6 +82,7 @@
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="8.0.0" />
     <PackageReference Include="System.Buffers" Version="4.6.1" />
+    <PackageReference Include="System.Formats.Asn1" Version="6.0.1" />
```
dscpinheiro, line 85: FYI @normj I had not noticed this dependency was also added... Hopefully not an issue since we're adding others in 4.1 anyway.

sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4aSigner.cs
```diff
@@ -0,0 +1,755 @@
...
+        }
+
+        /// <summary>
+        /// Compares two byte arrays in constant time.
+        /// </summary>
+        /// <param name="lhs">The first byte array to compare.</param>
+        /// <param name="rhs"> The second byte array to compare.</param>
+        /// <exception cref="ArgumentException"><paramref name="lhs"/> and <paramref name="rhs"/> have a different length.</exception>
+        /// <remarks>
+        /// Implementation adapted from <see href="https://github.com/awslabs/aws-c-auth/blob/e8360a65e0f3337d4ac827945e00c3b55a641a5f/source/key_derivation.c#L191-L215"/>.
```
dscpinheiro, line 344: Thanks for this link, I verified the implementation against it and found one small difference: `CompareConstantTime` initializes `eq = 0`, but the aws-c-auth original ([key_derivation.c](https://github.com/awslabs/aws-c-auth/blob/main/source/key_derivation.c)) initializes `eq = 1`. With `eq = 0` the `s_be_compare_test_cases` equivalent in C# fails today:

```csharp
[TestMethod]
public void CompareConstantTimeMatchesReference()
{
    // Test vectors ported from aws-c-auth tests/key_derivation_tests.c (be_sequence_compare).
    var lessLhs = new byte[] { 0x00, 0x00, 0x00 };
    var lessRhs = new byte[] { 0x00, 0x00, 0x01 };
    var equalLhs = new byte[] { 0xAB, 0xCD, 0x80, 0xFF, 0x01, 0x0A };
    var equalRhs = new byte[] { 0xAB, 0xCD, 0x80, 0xFF, 0x01, 0x0A };
    var greaterLhs = new byte[] { 0xFF, 0xCD, 0x80, 0xFF, 0x01, 0x0A };
    var greaterRhs = new byte[] { 0xFE, 0xCD, 0x80, 0xFF, 0x01, 0x0A };

    Assert.AreEqual(-1, Math.Sign(AWS4aSigner.CompareConstantTime(lessLhs, lessRhs)));
    Assert.AreEqual(1, Math.Sign(AWS4aSigner.CompareConstantTime(lessRhs, lessLhs)));
    Assert.AreEqual(0, Math.Sign(AWS4aSigner.CompareConstantTime(equalLhs, equalRhs)));
    Assert.AreEqual(0, Math.Sign(AWS4aSigner.CompareConstantTime(equalRhs, equalLhs)));
    Assert.AreEqual(1, Math.Sign(AWS4aSigner.CompareConstantTime(greaterLhs, greaterRhs)));
    Assert.AreEqual(-1, Math.Sign(AWS4aSigner.CompareConstantTime(greaterRhs, greaterLhs)));
}
```

```
 CompareConstantTimeMatchesReference
   Source: AWS4aSignerTests.cs line 43
   Duration: 47 ms

  Message: 
Assert.AreEqual failed. Expected:<1>. Actual:<-1>. 

  Stack Trace: 
AWS4aSignerTests.CompareConstantTimeMatchesReference() line 54
MethodBaseInvoker.InterpretedInvoke_Method(Object obj, IntPtr* args)
MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```

With `eq = 1` all six assertions pass (we can even add `s_be_add_one_test_cases` as a regression check for `AddOneConstantTime`).

sdk/test/Services/S3/IntegrationTests/CapacityManagerTests.cs
teo-tsirpanis: Oops, fixed.

sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4aSigner.cs
teo-tsirpanis: Fixed, and added test cases.

sdk/test/UnitTests/Custom/Runtime/AWS4aSignerTests.cs
teo-tsirpanis: Fixed.

sdk/test/UnitTests/AWSSDK.UnitTests.Core.csproj
```diff
@@ -34,9 +34,6 @@
          sdk/Directory.Build.props (the latter conditioned on the .NETFramework target identifier). -->
     <NoWarn>CS1591,CS0612,CS0618</NoWarn>
   </PropertyGroup>
-  <PropertyGroup>
```
dscpinheiro, line 37: Removed this line since the language version is now defined in: https://github.com/aws/aws-sdk-net/blob/feature/4.1/sdk/Directory.Build.props#L12

dscpinheiro (approved):

generator/.DevConfigs/b495f893-31aa-4e17-866c-af78e4da6e8b.json
```diff
@@ -0,0 +1,10 @@
+{
+  "core": {
+    "updateMinimum": false,
+    "type": "patch",
+    "changeLogMessages": [
+      "Added AWS4aSigner class that implements SigV4a signing in managed code. AWS CRT is no longer used by the SDK for this purpose, and the related APIs were deprecated.",
+      "The AWS4aSigningResult.PresignedUri property was deprecated and will always be empty in objects returned by AWS4aSigner. Use the ForQueryParameters property instead, to get the query parameters for a presigned URL."
```
normj: This is an internal field the users of the SDK wouldn't understand. We don't need to call out internal API changes in the change log.

sdk/src/Core/Amazon.Runtime/Credentials/ImmutableCredentials.cs
```diff
@@ -55,6 +56,15 @@ public class ImmutableCredentials
...
         public string AccountId { get; private set; }
+
+        /// <summary>
+        /// Contains a cached AWS4a signing key for the current credentials. Do
+        /// not access this property directly; use <see cref="Internal.Auth.AWS4aSigner.GetCachedSigningKey"/> instead.
+        /// </summary>
+        /// <remarks>
+        /// The key must not be disposed, and must not be returned to the user.
+        /// </remarks>
+        internal ECDsa AWS4aSigningKey;
```
normj: nit: Can we call this `CachedAWS4aSigningKey` to make it a little more clear this is a lazily set property.

sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4aSigner.cs
```diff
@@ -0,0 +1,755 @@
...
+        }
+
+        /// <summary>
+        /// Adds one to a large unsigned integer represented by a sequence of bytes in constant time.
+        /// </summary>
+        /// <remarks>
+        /// Implementation adapted from <see href="https://github.com/awslabs/aws-c-auth/blob/e8360a65e0f3337d4ac827945e00c3b55a641a5f/source/key_derivation.c#L106-L189"/>.
+        /// </remarks>
+        [MethodImpl(MethodImplOptions.NoInlining | MethodImplOptions.NoOptimization)]
+        private static void AddOneConstantTime(byte[] data)
```
normj: I would feel more comfortable with this method if it was made `internal` and added specific unit tests so we can easily tell its corner cases are covered by unit tests.

sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4aSigner.cs
```diff
@@ -0,0 +1,755 @@
...
+            }
+            return newKey;
+        }
+
+        /// <summary>
+        /// Returns the ECDSA signature for an arbitrary blob using the specified credentials.
+        /// </summary>
+        /// <param name="credentials">The credentials to use.</param>
+        /// <param name="data">The data to sign.</param>
+        public static byte[] SignBlob(ImmutableCredentials credentials, string data)
```
normj, line 480: Can we get some unit tests covering this method? What I'm most concern about is making sure we are purposefully covering the conditional blocks in `WriteKeyParameterInteger`.

normj (changes requested):
Some minor comments but PR is looking good. Excited to get this merged, thanks for your effort.

generator/.DevConfigs/b495f893-31aa-4e17-866c-af78e4da6e8b.json
teo-tsirpanis: Removed.

sdk/src/Core/Amazon.Runtime/Credentials/ImmutableCredentials.cs
teo-tsirpanis: Renamed

sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4aSigner.cs
teo-tsirpanis: Added test cases from aws-c-auth.

sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4aSigner.cs
teo-tsirpanis: Added unit tests that sign a message and verify it. We can't do much more than that since the signatures are non-deterministic. I also had to add the corresponding conditional blocks for signature verification.

normj (approved):

dscpinheiro (approved):

Issue #3932, opened 2025-07-24
No Amazon.Runtime.ClientConfig.HttpClientCacheSize available for .net framework.


LeuisKen:
### Describe the bug

`HttpClientCacheSize` property is not available for .net framework apps.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

The app should work without exception thrown.

### Current Behavior

Unhandled Exception: System.MissingMethodException: Method not found: 'Void Amazon.Runtime.ClientConfig.set_HttpClientCacheSize(Int32)'.
   at Amazon.Extensions.NETCore.Setup.ClientFactory`1.CreateConfig(AWSOptions options)
   at Amazon.Extensions.NETCore.Setup.ClientFactory`1.CreateServiceClient(ILogger logger, AWSOptions options) in C:\codebuild\tmp\output\src2114625843\src\aws-sdk-net\extensions\src\AWSSDK.Extensions.NETCore.Setup\ClientFactory.cs:line 100
   at Amazon.Extensions.NETCore.Setup.AWSOptions.CreateServiceClient[T]() in C:\codebuild\tmp\output\src2114625843\src\aws-sdk-net\extensions\src\AWSSDK.Extensions.NETCore.Setup\AWSOptions.cs:line 104
   at AWSConsoleApp.Program.Main(String[] args) in D:\Projects\AWSConsoleApp\Program.cs:line 14

### Reproduction Steps

I created a repo for reproduction, you can see an error thrown when running the app. https://github.com/LeuisKen/aws-sdk-error-reproduce

The reproduction code is just like this:

```csharp
using Amazon.SecretsManager;
using Microsoft.Extensions.Configuration;
using System;

namespace AWSConsoleApp
{
    internal class Program
    {
        static void Main(string[] args)
        {
            var configBuilder = new ConfigurationBuilder();
            var config = configBuilder.Build();
            var awsOptions = config.GetAWSOptions();
            awsOptions.CreateServiceClient<IAmazonSecretsManager>();
            Console.WriteLine("AWS Options Loaded Successfully");
            Console.WriteLine(awsOptions);
        }
    }
}

```

### Possible Solution

Based on the code here: https://github.com/aws/aws-sdk-net/blob/835b063a95a3b7a79d0bc7bbb3863f0e176f67ea/sdk/src/Core/Amazon.Runtime/ClientConfig.cs#L1093-L1118

[...]

PR #3935, opened 2025-07-25
Avoid setting config properties that don't exist in .NET Framework variant

normj:
## Description
This is the part 2 of PR https://github.com/aws/aws-sdk-net/pull/3934 that makes it so AWSSDK.Extensions.NETCore.Setup does not attempt to set service client config properties that do not exist in the .NET Framework variant of the SDK.

## Motivation and Context
https://github.com/aws/aws-sdk-net/issues/3932

## Testing
Built a .NET Framework console application using `AWSSDK.Extensions.NETCore.Setup` and confirmed I got the reported `MissingMethodException`. Then built a new package version with this change and used that and confirmed the exception went away.

### DIFF
--- a/extensions/src/AWSSDK.Extensions.NETCore.Setup/ClientFactory.cs
+++ b/extensions/src/AWSSDK.Extensions.NETCore.Setup/ClientFactory.cs
@@ -17,6 +17,7 @@ using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.CompilerServices;
 using System.Xml.Linq;
 using Amazon.Runtime;
 using Amazon.Runtime.CredentialManagement;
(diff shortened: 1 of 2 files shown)


boblodgett (approved):

muhammad-othman (approved):

extensions/src/AWSSDK.Extensions.NETCore.Setup/ClientFactory.cs
```diff
@@ -367,9 +364,27 @@ private ClientConfig CreateConfig(AWSOptions options)
...
+            // if the SDK is not the .NET Framework variant.
+            if (!AWSSDKUtils.IsNETFramework())
+            {
+                SetNETStandardAndAboveSettings(defaultConfig, config);
+            }
+
             return config;
         }
 
+        private void SetNETStandardAndAboveSettings(DefaultClientConfig defaultConfig, ClientConfig config)
```
teo-tsirpanis, line 382: To be extra sure that the non-existent properties don't get resolved, this method should be marked with `[MethodImpl(MethodImplOptions.NoInlining)]`.

ashishdhingra (approved):

extensions/src/AWSSDK.Extensions.NETCore.Setup/ClientFactory.cs
normj: Done, thanks for the suggestion.

PR #3939, opened 2025-07-30
Add CborReaderProxy to support process CBOR stream in chunks

muhammad-othman:
## Description
The current code loads the whole response stream in memory before start the unmarshalling process. This works for small payloads but may cause issues for large response payloads.
The `System.Formats.Cbor.CborReader` doesn't accept stream and requires in memory byte array for deserialization.

This PR adds `CborReaderProxy` that wraps `System.Formats.Cbor.CborReader` to support processing the stream in chunks without ever holding the whole payload in memory, feeding the chunks into `CborReader` and process them one by one.

The default `CborReader.ReadEndMap()`/`CborReader.ReadEndArray()` methods assume the full map is already present in the input buffer and cannot continue reading if the buffer ends in the middle of a container.
This means the reader will fail to advance if the next token is a container break (0xFF) but the start wasn't read into the current buffer (it was read in a previous part of the stream).

To support maps/arrays across multiple refills I did the following:
* Implement custom `ReadStartMap`, `ReadEndMap`, `ReadStartArray`, `ReadEndArray`.
* Detect and skip container break markers `0xFF` as needed.
* Added `_nestingStack` to track when reading a nested CBOR structure (e.g., an array of maps), and correctly match calls to `ReadStartMap` with `ReadEndMap`, and `ReadStartArray` with `ReadEndArray`.

## Motivation and Context
`#DOTNET-8252`

## Testing
* Added `CborReaderProxyTests.cs` to cover the new changes.
* Tested various `CloudWatch` and `SecretsManager` operations with CBOR.
* `DRY_RUN-4ff92fab-9e40-46cc-88d0-746023de37cf`.

## Types of changes

## Checklist

## License

### DIFF
--- a/extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
+++ b/extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
@@ -1,0 +1,295 @@
+﻿/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ * 
+ *  http://aws.amazon.com/apache2.0
+ * 
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+using Amazon.Runtime.Internal.Util;
+using System;
+using System.Buffers;
+using System.Collections.Generic;
+using System.Formats.Cbor;
+using System.IO;
+
+namespace Amazon.Extensions.CborProtocol.Internal
+{
+    /// <summary>
+    /// A streaming CBOR reader that processes large CBOR data streams in chunks without loading
+    /// the entire payload into memory at once. This class wraps <see cref="CborReader"/>
+    /// to provide streaming capabilities while maintaining the same reading interface.
+    /// </summary>
+    public class CborStreamReader : IDisposable
+    {
+        /// <summary>
+        /// Enum to track the type of CBOR container (map or array)
+        /// for state management within the CborStreamReader.
+        /// </summary>
+        private enum CborContainerType
+        {
+            Map,
+            Array
+        }
+
+        private static readonly ILogger _logger = Logger.GetLogger(typeof(CborStreamReader));
+        private readonly Stack<CborContainerType> _nestingStack = new Stack<CborContainerType>();
+        private readonly Stream _stream;
+        private byte[] _buffer;
+        private CborReader _internalCborReader;
+        private int _currentChunkSize;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CborStreamReader"/> class that reads CBOR data
+        /// from the specified stream.
+        /// </summary>
+        /// <param name="stream">The input stream containing CBOR-formatted data.
+        /// The stream must be readable and remain open for the lifetime of the reader.</param>
+        /// <exception cref="ArgumentNullException">Thrown when the stream parameter is null.</exception>
+        /// <remarks>
+        /// The reader uses a configurable initial buffer size (from <see cref="AWSConfigs.CborReaderInitialBufferSize"/>)
+        /// and will automatically resize the buffer if needed to handle larger CBOR items.
+        /// </remarks>
+        public CborStreamReader(Stream stream)
+        {
+            _stream = stream ?? throw new ArgumentNullException(nameof(stream));
+            _buffer = ArrayPool<byte>.Shared.Rent(AWSConfigs.CborReaderInitialBufferSize);
+
+            _currentChunkSize = _stream.Read(_buffer, 0, _buffer.Length);
+            var memorySlice = new ReadOnlyMemory<byte>(_buffer, 0, _currentChunkSize);
+
+            // We must allow multiple root values because when refilling the new chunk is just a fragment of the whole stream.
+            _internalCborReader = new CborReader(memorySlice, allowMultipleRootLevelValues: true);
+        }
+
+        /// <summary>
+        /// This method is called when a read operation fails because it needs more
+        /// data than is currently available in the buffer. It handles stitching leftover
+        /// data with a new chunk from the stream and, if necessary, resizing the buffer.
+        /// </summary>
+        /// <param name="bytesToSkip">Number of bytes to skip before reading new data (e.g., 1 to skip CBOR break byte 0xFF)</param>
+        private void RefillBuffer(int bytesToSkip = 0)
+        {
+            int leftoverBytesCount = _internalCborReader.BytesRemaining;
+
+            // Determine where the leftover bytes start
+            int leftoverStartIndex = _currentChunkSize - leftoverBytesCount;
+
+            // If we are skipping bytes, we need to move the start forward
+            leftoverStartIndex += bytesToSkip;
+            leftoverBytesCount = leftoverBytesCount - bytesToSkip;
+
+            // If the leftover data completely fills the buffer, grow it
+            if (leftoverBytesCount >= _buffer.Length)
+            {
+                int newSize = _buffer.Length * 2;
+                var newBuffer = ArrayPool<byte>.Shared.Rent(newSize);
+
+                Buffer.BlockCopy(_buffer, leftoverStartIndex, newBuffer, 0, leftoverBytesCount);
+                ArrayPool<byte>.Shared.Return(_buffer);
+                _buffer = newBuffer;
+            }
+            else if (leftoverBytesCount > 0)  // Shift leftovers (after skipping) to the beginning of the buffer
+            {
+                Buffer.BlockCopy(_buffer, leftoverStartIndex, _buffer, 0, leftoverBytesCount);
+            }
+
+            // Read from stream into buffer after leftovers
+            int bytesReadFromStream = _stream.Read(_buffer, leftoverBytesCount, _buffer.Length - leftoverBytesCount);
+
+            // Update the total size of valid data in our buffer.
+            _currentChunkSize = leftoverBytesCount + bytesReadFromStream;
+
+            // Check for a malformed stream: if we have leftovers but the stream is empty,
+            // it means the CBOR data was truncated.
+            if (bytesReadFromStream == 0 && leftoverBytesCount > 0)
+            {
+                throw new CborContentException("Stream ended unexpectedly with an incomplete CBOR data item.");
+            }
+
+            var newMemorySlice = new ReadOnlyMemory<byte>(_buffer, 0, _currentChunkSize);
+            _internalCborReader.Reset(newMemorySlice);
+
+            _logger.DebugFormat("Buffer refilled: read {0} byte(s), total in buffer now: {1}.", bytesReadFromStream, _currentChunkSize);
+        }
+
+        /// <summary>
+        /// Executes a CBOR read operation, refilling the buffer and retrying if a CborContentException is thrown.
+        /// </summary>
+        /// <typeparam name="T">The return type of the CBOR read operation.</typeparam>
+        /// <param name="readOperation">A delegate representing the read operation to execute.</param>
+        /// <returns>The result of the read operation.</returns>
+        /// <exception cref="CborContentException">
+        /// Thrown if too many retries are attempted or if the stream ends unexpectedly.
+        /// </exception>
+        private T ExecuteRead<T>(Func<CborReader, T> readOperation)
+        {
+            int maxRetries = 64;
+            int retryCount = 0;
+
+            while (true)
+            {
+                try
+                {
+                    return readOperation(_internalCborReader);
+                }
+                catch (CborContentException ex)
+                {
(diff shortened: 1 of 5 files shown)


extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborReaderProxy.cs
```diff
@@ -0,0 +1,274 @@
...
+        }
+
+        private static readonly ILogger _logger = Logger.GetLogger(typeof(CborReaderProxy));
+        private readonly Stack<CborContainerType> _nestingStack = new Stack<CborContainerType>();
+        private readonly Stream _stream;
+        private byte[] _buffer;
+        private CborReader _internalCborReader;
+        private int _currentChunkSize;
+
+        public CborReaderProxy(Stream stream)
```
normj: Prefer the name `CborStreamReader`

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborReaderProxy.cs
```diff
@@ -0,0 +1,274 @@
...
+        {
+            _stream = stream ?? throw new ArgumentNullException(nameof(stream));
+            _buffer = ArrayPool<byte>.Shared.Rent(AWSConfigs.CborReaderInitialBufferSize);
+
+            _currentChunkSize = _stream.Read(_buffer, 0, AWSConfigs.CborReaderInitialBufferSize);
```
normj: Since the `Rent` call might have returned back a buffer larger then `AWSConfigs.CborReaderInitialBufferSize` shouldn't we use `_buffer.Length`?

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborReaderProxy.cs
```diff
@@ -0,0 +1,274 @@
...
+                {
+                    if (_currentChunkSize == 0 && _internalCborReader.BytesRemaining == 0)
+                    {
+                        // Fail fast if we’ve already consumed all input and nothing remains to refill.
+                        throw;
+                    }
+
+                    if (++retryCount > maxRetries)
+                    {
+                        throw new CborContentException("Too many retries during CBOR stream parsing. Possible malformed or infinite data.");
```
normj: Can you add the original `ex` exception as the inner exception?

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborReaderProxy.cs
```diff
@@ -0,0 +1,274 @@
...
+                    throw new CborContentException($"Unexpected end of {expectedType.ToString().ToLowerInvariant()}.");
+
+                var state = CborReaderState.Finished;
+                try
+                {
+                    state = r.PeekState();
+                }
+                catch (CborContentException)
+                {
+                    _logger.InfoFormat($"CborContentException caught during PeekState while expecting end of {expectedType.ToString().ToLowerInvariant()}");
```
normj: Why is okay to ignore this exception.

Also this log message should either be a warning if it is very uncommon or a debug. I don't want to flood the log with lots of info messages that are not actionable by users.

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborReaderProxy.cs
muhammad-othman: This way we can use the large buffer since we rented it anyway, let me test it and see how it goes.

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborReaderProxy.cs
muhammad-othman: This exception will be thrown if we reached the end of the current `CborReader` or the `CborReader` didn't contain the start of the array/map that we are expecting to end, at this case  the state will be `CborReaderState.Finished` and we will refill the buffer and pop an item from the stack.
So it isn't a real error that why I didn't do a debug or a warning log, I wasn't really sure if we should log this error, but it seems like it can be useful to trace how we got to certain situation.
I'm okay with removing it though.

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborReaderProxy.cs
```diff
@@ -0,0 +1,274 @@
...
+            // it means the CBOR data was truncated.
+            if (bytesReadFromStream == 0 && leftoverBytesCount > 0)
+            {
+                throw new CborContentException("Stream ended unexpectedly with an incomplete CBOR data item.");
+            }
+
+            var newMemorySlice = new ReadOnlyMemory<byte>(_buffer, 0, _currentChunkSize);
+            _internalCborReader.Reset(newMemorySlice);
+
+            _logger.InfoFormat("Buffer refilled: read {0} byte(s), total in buffer now: {1}.", bytesReadFromStream, _currentChunkSize);
```
normj: Make this a debug log message.

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborReaderProxy.cs
normj: Add that information as a comment here so we remember the reason for the future.

normj (changes requested):

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
```diff
@@ -0,0 +1,279 @@
...
+                    state = r.PeekState();
+                }
+                catch (CborContentException)
+                {
+                    // This exception is expected in two cases:
+                    // 1. When we've reached the end of the current CBOR buffer chunk and need to refill.
+                    // 2. When the current buffer does not contain the start of the array/map we're trying to end.
+                    // In both cases, the internal CborReader's state will be `Finished` and we will trigger a buffer refill.
+                    // This is not a true error, but logging at Info level can help trace how we arrived at a certain state.
+                    _logger.InfoFormat($"CborContentException caught during PeekState while expecting end of {expectedType.ToString().ToLowerInvariant()}");
```
normj: I Still think this should be `DebugFormat`. Also instead of using string interpolation with the `$` use the format methods parameters. That way we avoid the string allocation when the logging is not enabled.

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
```diff
@@ -0,0 +1,279 @@
...
+                        // Fail fast if we’ve already consumed all input and nothing remains to refill.
+                        throw;
+                    }
+
+                    if (++retryCount > maxRetries)
+                    {
+                        throw new CborContentException("Too many retries during CBOR stream parsing. Possible malformed or infinite data.", ex);
+                    }
+
+                    _logger.Debug(ex, $"CborContentException caught (attempt #{retryCount}), attempting to refill buffer.");
```
normj: Use `DebugFormat` and use the methods parameters instead of doing the string interpolation with the `$`.

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
```diff
@@ -0,0 +1,279 @@
...
+                {
+                    // Translate a Break code to the appropriate container end state
+                    // based on our own nesting stack.
+                    if (_nestingStack.Count > 0)
+                    {
+                        var inferredState = _nestingStack.Peek() == CborContainerType.Map
+                            ? CborReaderState.EndMap
+                            : CborReaderState.EndArray;
+
+                        _logger.Debug(ex, $"CborContentException during PeekState interpreted as {inferredState} due to nesting stack.");
```
normj: Use `DebugFormat` and use the methods parameters instead of doing the string interpolation with the `$`.

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
muhammad-othman: I kept `Debug` to allow passing the exception, but I removed the string interpolation.

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
```diff
@@ -0,0 +1,279 @@
...
+        }
+
+        private static readonly ILogger _logger = Logger.GetLogger(typeof(CborStreamReader));
+        private readonly Stack<CborContainerType> _nestingStack = new Stack<CborContainerType>();
+        private readonly Stream _stream;
+        private byte[] _buffer;
+        private CborReader _internalCborReader;
+        private int _currentChunkSize;
+
+        public CborStreamReader(Stream stream)
```
peterrsongg, line 59: this class is missing documentation

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
```diff
@@ -0,0 +1,279 @@
...
+using Amazon.Runtime.Internal.Util;
+using System;
+using System.Buffers;
+using System.Collections.Generic;
+using System.Formats.Cbor;
+using System.IO;
+
+namespace Amazon.Extensions.CborProtocol.Internal
+{
+    public class CborStreamReader : IDisposable
```
peterrsongg, line 29: class is missing documentation

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
```diff
@@ -0,0 +1,279 @@
...
+        private byte[] _buffer;
+        private CborReader _internalCborReader;
+        private int _currentChunkSize;
+
+        public CborStreamReader(Stream stream)
+        {
+            _stream = stream ?? throw new ArgumentNullException(nameof(stream));
+            _buffer = ArrayPool<byte>.Shared.Rent(AWSConfigs.CborReaderInitialBufferSize);
+
+            _currentChunkSize = _stream.Read(_buffer, 0, _buffer.Length);
```
peterrsongg, line 64: If the intention is to fill the whole buffer or fill until the stream has no data left, this isn't guaranteed to happen with one `Stream.read` call, since the docs state that it could be less than the bytes you requested if the stream doesn't have that data available yet.

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
```diff
@@ -0,0 +1,279 @@
...
+
+            while (true)
+            {
+                try
+                {
+                    return readOperation(_internalCborReader);
+                }
+                catch (CborContentException ex)
+                {
+                    if (_currentChunkSize == 0 && _internalCborReader.BytesRemaining == 0)
```
peterrsongg, line 144: I'm curious how we would be in a `CborContentException` is the _currentChunkSize == 0 and the BytesRemaining == 0? Wouldn't this indicate that we successfully read everything?

normj (approved):

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
```diff
@@ -0,0 +1,279 @@
...
+
+            // Determine where the leftover bytes start
+            int leftoverStartIndex = _currentChunkSize - leftoverBytesCount;
+
+            // If we are skipping bytes, we need to move the start forward
+            leftoverStartIndex += bytesToSkip;
+            leftoverBytesCount = leftoverBytesCount - bytesToSkip;
+
+            // If the leftover data completely fills the buffer, grow it
+            if (leftoverBytesCount >= _buffer.Length)
```
peterrsongg, line 89: just wondering.. is there a reason you used >= here? Isn't it impossible for the leftoverBytesCount to be greather than the buffer's length, since the reader will be instantiated with at most the data in the buffer? or am I missing something? Have you seen this case happen before?

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
muhammad-othman: This is why we are using `_currentChunkSize` to know how much did we read and we create the `ReadOnlyMemory` from the start of the buffer to `_currentChunkSize` not to the buffer size. 

From `Read` documentation:
image

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
muhammad-othman: It didn't happen while testing, but just wanted to be safe incase we had an off by one error or some, at this case I would prefer expanding the buffer rather than go to the next case and lose data when copying to the start of the buffer.

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
muhammad-othman: Yes, and after reading everything the code that uses the `CborStreamReader` might try to read more data which means there is an issue with that code, at this case `CborReader` will throw this exception and we will throw it back for the other code to handle or fail.

peterrsongg (approved):

PR #3940, opened 2025-07-31
Generate Get&PutLifecycleConfiguration. Generate Get and PutBucketNotification

peterrsongg:
…tNotification

## Description

Generates 4 more operations. I've added a lot more customization hooks to get this to work. This is because these 4 operations follow the pattern:
```
        "PutBucketLifecycle": {
            "exclude": true
        },
        "PutBucketLifecycleConfiguration": {
            "name": "PutLifecycleConfiguration"
        },
```

where one is excluded and another is renamed, and in some cases there will be a shape that is shared between two operations even though it isn't defined like that in the model. This is why I had to add 
```
                        "shape": "LifecycleConfiguration",
                        "originalMemberIsOutsideContainingShape":true,
                        "outsideOriginalMember":"PutBucketLifecycleConfigurationRequest,LifecycleConfiguration"
```

## Testing

For Get and PutLifecycleConfiguration I've added unit tests just to make sure we generated the right marshaller/unmarshaller
since we are using extensive customization hooks.

```
Already on 'petesong/phase-2-pr2'
  AWSSDK.Core.NetStandard net8.0 succeeded (32.6s) → C:\Dev\Repos\aws-sdk-net-staging\sdk\src\Core\bin\Debug\net8.0\AWSSDK.Core.dll
  AWSSDK.S3.NetStandard net8.0 succeeded (13.2s) → bin\Debug\net8.0\AWSSDK.S3.dll

Build succeeded in 59.0s

Workload updates are available. Run `dotnet workload list` for more information.
error: pathspec 'petesong/s3-phase2' did not match any file(s) known to git
  AWSSDK.Core.NetStandard net8.0 succeeded (0.4s) → C:\Dev\Repos\aws-sdk-net-staging\sdk\src\Core\bin\Debug\net8.0\AWSSDK.Core.dll
  AWSSDK.S3.NetStandard net8.0 succeeded (0.4s) → bin\Debug\net8.0\AWSSDK.S3.dll

Build succeeded in 1.2s
```
## Types of changes

## Checklist

## License


generator/ServiceClientGeneratorLib/Generators/Marshallers/BaseMarshaller.cs
```diff
@@ -3392,68 +3392,81 @@ protected void GenerateRequestChecksumHandling(Operation operation, string reque
...
             {
-                // For some S3 operations, in addition to the checksum specific headers (such as "x-amz-checksum-crc32"), there's also another
-                // header which specifies the algorithm used: "x-amz-sdk-checksum-algorithm"
-                // Even though S3 doesn't validate the latter, it's modeled and could be enforced in the future.
-                var member = operation.RequestStructure.Members.FirstOrDefault(m => string.Equals(m.PropertyName, operation.ChecksumConfiguration.RequestAlgorithmMember));
-                var headerName = member.MarshallLocation == MarshallLocation.Header ? member.MarshallLocationName : string.Empty;
+                // if a rename happened for the operation, look for the original operation in the model and see if a request algorithm member exists
+                string headerName = null;
+                if (this.Config.ServiceModel.DocumentRoot["operations"][operation.Name] != null && this.Config.ServiceModel.DocumentRoot["operations"][operation.Name][ServiceModel.HttpChecksumKey] != null)
```
peterrsongg, line 3397: Basically, this is for the case where an operation was renamed, but the renamed operation doesn't have `httpChecksumRequired`. ex: PutBucketNotificationConfiguration doesn't have `httpChecksumRequired = true` but `PutBucketNotification` (the original method) does. 

https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Services/S3/Custom/Model/Internal/MarshallTransformations/PutBucketNotificationRequestMarshaller.cs#L146-L158

^^ clearly it does checksum here.

generator/ServiceClientGeneratorLib/GeneratorDriver.cs
```diff
@@ -1675,5 +1698,26 @@ public static void GenerateS3Enumerations(GeneratorOptions options)
             var text = generator.TransformText();
             WriteFile(generatedFileRoot, null, fileName, text);
         }
+
+        // in some cases we have deviated too far from the S3 model. For example for "Filter" shapes where we implement
+        // the visitor pattern for the different predicate types. This method will check to see if additional members
+        // are being added to these shapes so that we know to make the appropriate customization
+        private void S3NeedsCustomUpdate(Shape shape)
```
peterrsongg, line 1704: Norm and I talked about this. Since we have strayed so far from the model with the LifecycleFilter class (with all the LifecycleFilterPredicate polymorphism), we won't generate those classes and unmarshallers. So instead we take a snapshot of what the model looks like at this point in time and if it changes, we throw an exception to make sure we make the appropriate update.

peterrsongg:
dry run 40ca0bcc-605f-43fa-85e6-db18fd8cf3b5 succceded

sdk/test/Services/S3/UnitTests/Custom/LifecycleConfigurationTests.cs
```diff
@@ -0,0 +1,221 @@
...
+            #endregion
+            XDocument expectedDoc = XDocument.Parse(expectedBody);
+            ComparisonUtils.AreDocumentsEqual(expectedDoc, actualDoc);
```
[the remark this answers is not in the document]
GarrettBeatty: should this be updated

sdk/test/Services/S3/UnitTests/Custom/LifecycleFilterTests.cs
```diff
@@ -26,8 +26,8 @@ namespace AWSSDK.UnitTests
     [TestClass]
     public class LifecycleFilterTests
     {
-        private readonly string ExpectedRequestFormat = "<LifecycleConfiguration xmlns=\"http://s3.amazonaws.com/doc/2006-03-01/\"><Rule><Transition><Days>1</Days><StorageClass>GLACIER" +
-            "</StorageClass></Transition>{0}<Status>Disabled</Status></Rule></LifecycleConfiguration>";
+        private readonly string ExpectedRequestFormat = "<LifecycleConfiguration xmlns=\"http://s3.amazonaws.com/doc/2006-03-01/\"><Rule>{0}<Status>Disabled</Status><Transition><Days>1</Days><StorageClass>GLACIER" +
```
GarrettBeatty, line 29: why is this being changed?

sdk/src/Services/S3/Custom/Model/PutBucketNotificationRequest.cs
```diff
@@ -115,20 +115,6 @@ namespace Amazon.S3.Model
     public partial class PutBucketNotificationRequest : AmazonWebServiceRequest
```
GarrettBeatty, line 115: general question: why is this class still left over if a new class was made in generated? why couldnt checksum algorithm also moved?

sdk/test/Services/S3/UnitTests/Custom/LifecycleFilterTests.cs
peterrsongg: The order of the Xml document doesn't matter, I changed this because it was saying the xml document wasn't the same due to ordering, so I put the filter element where it is being written with the newly generated marshallers

sdk/src/Services/S3/Custom/Model/PutBucketNotificationRequest.cs
peterrsongg: This is because "CheckSum Algorithm" isn't modeled on the original request shape
https://github.com/aws/aws-sdk-net/blob/32933d593014977245226d19bdc8d89c95899f3e/generator/ServiceModels/s3/s3-2006-03-01.normal.json#L9535-L9565

This is because this operation is really a combination of two operations (PutBucketNotificationRequest and PutBucketNotificationConfigurationRequest)


If you see https://github.com/aws/aws-sdk-net/blob/32933d593014977245226d19bdc8d89c95899f3e/generator/ServiceModels/s3/s3-2006-03-01.normal.json#L9589 it is modeled there. But since we are generating from `PutBucketNotificationConfiguration` I had to leave this custom partial. I could've maybe used the customization I added
```
                        "originalMemberIsOutsideContainingShape":true,
                        "outsideOriginalMember":"PutBucketNotificationRequest, ChecksumAlgorithm"
```

but that didn't feel right either because we weren't excluding a member and injecting something else in its place.

GarrettBeatty (approved):

generator/.DevConfigs/58c91090-8610-40e3-970b-5cb6e77ea5e9.json
```diff
@@ -0,0 +1,9 @@
+{
+  "core": {
+    "updateMinimum": true,
```
muhammad-othman: I don't think we need this `devConfig`, we don't change anything a the core package and with `updateMinimum` set to `true` we will do an unnecessary release.

generator/ServiceClientGeneratorLib/GeneratorDriver.cs
```diff
@@ -447,6 +447,9 @@ var propertyModifier
                     operation.ResponseStructure == null ||
                     this.Configuration.ServiceModel.Customizations.ResultGenerationSuppressions.Contains(operation.Name);
 
+                if (this.Configuration.ServiceModel.Customizations.ExcludeShapes().Contains(operation.Name + "Response"))
+                    return;
```
muhammad-othman: Instead of skipping the response and request response separately, can you skip them here https://github.com/aws/aws-sdk-net/blob/cbor-protocol/generator/ServiceClientGeneratorLib/GeneratorDriver.cs#L240 ?

generator/ServiceClientGeneratorLib/GeneratorDriver.cs
muhammad-othman: Is it possible that one of the members change? Can we compare the json directly to account of all the changes not just the number of members?

generator/.DevConfigs/03962e88-1f87-4fb6-b63a-188eff9588f5.json
```diff
@@ -0,0 +1,11 @@
+{
+  "services": [
+    {
+      "serviceName": "S3",
+      "type": "patch",
+      "changeLogMessages": [
```
muhammad-othman, line 6: nit: These two devConfigs can be merged.

generator/.DevConfigs/58c91090-8610-40e3-970b-5cb6e77ea5e9.json
peterrsongg: This is for all the rest-xml protocols that will change due to the `XmlStructureUnmarshallCustomization` method that will be added. Do you mean add a devconfig for each rest-xml service instead?

peterrsongg:
>Is it possible that one of the members change? Can we compare the json directly to account of all the changes not just the number of members?

@muhammad-othman I'm not sure why I can't quote the reply directly so replying here😅. If one of the members changes (in a backwards compatible way of course), it wouldn't matter since we generate the member's unmarshallers (if it is a structure). For example, if a new member is added to `Tag`, we would still generate the new member on the `Tag` class and unmarshall it on its unmarshallers, with no need to update the `Filter` class. So really, the only time we need to update the `Filter` class is if they add new members, in which case we will need to add  new predicates or operands depending on the type of member.

generator/ServiceClientGeneratorLib/GeneratorDriver.cs
peterrsongg: done

muhammad-othman (approved):

PR #3956, opened 2025-08-13
Improve CborStreamReader Logic to Handle Nesting and End-of-Buffer Corner Cases

muhammad-othman:
## Description
This PR updates the `CborStreamReader` streaming logic to correctly handle scenarios where the buffer ends just after an item was read, particularly when working with nested containers.

Previously, the streaming CBOR reader could incorrectly signal completion when the current buffer segment ended but nested containers were still open. This caused issues with large or segmented CBOR payloads where container boundaries did not align with buffer boundaries.

* Avoids prematurely treating `PeekState() == Finished` as a final read state when there are still open containers in the nesting stack.
* Ensures containers are only closed when the corresponding `EndArray` or `EndMap` is explicitly read when end-of-current-buffer-part scenarios.

## Motivation and Context
`DOTNET-8252`

## Testing
- `DRY_RUN-08c395ab-328f-4d34-8928-282cecedfd90`
- Added unit tests to cover when the buffer ends close to the end of containers.

## Types of changes

## Checklist

## License

### DIFF
--- a/extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
+++ b/extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
@@ -134,6 +134,14 @@ namespace Amazon.Extensions.CborProtocol.Internal
                 }
                 catch (Exception ex) when (ex is InvalidOperationException || ex is CborContentException)
                 {
+                    // Both InvalidOperationException and CborContentException can occur when the reader
+                    // runs out of buffered data mid-item or encounters an incomplete CBOR structure.
+                    // - CborContentException: typically thrown when encountering invalid CBOR content,
+                    //   such as a premature break or truncated container.
+                    // - InvalidOperationException: can happen when the reader attempts to interpret data
+                    //   as a different type due to hitting a buffer boundary before the full item is available.
+                    // We catch both, trigger a buffer refill, and retry before deciding if it is a genuine error.
+
                     if (_currentChunkSize == 0 && _internalCborReader.BytesRemaining == 0)
                     {
                         // Fail fast if we’ve already consumed all input and nothing remains to refill.
@@ -193,11 +201,30 @@ namespace Amazon.Extensions.CborProtocol.Internal
 
                 if (state == CborReaderState.Finished)
                 {
+                    // We got CborReaderState.Finished which means the reader has exhausted the bytes currently
+                    // given to it and the next token may live in the next chunk that we haven't read yet.
+                    //
+                    // For indefinite-length containers the end of the container is a break marker byte (0xFF).
+                    // If that break byte is the next byte in the stream, we must consume it, otherwise the reader
+                    // will become desynchronized (it will still think we're inside a container while removed the
+                    // container from the _nestingStack).
+                    // This byte can exist in the next chunk, so when we hit Finished we first attempt to refill
+                    // the buffer (no skip) so the reader can see the next byte(s). Only after refilling can we safely
+                    // determine whether the next byte is a break byte that terminates the current container.
+
                     // Try to refill first, maybe the break marker is in the next chunk.
                     RefillBuffer(0);
 
                     if (IsNextByteEndOfContainer())
                     {
+                        // If the next raw byte after refill is the CBOR "break" (0xFF), that indicates an
+                        // indefinite-length container terminator. We must explicitly consume that break byte
+                        // so we can remove the item from the _nestingStack and parsing can continue correctly.
+                        //
+                        // We call RefillBuffer(1) to skip the break byte and  rebuild the internal reader.
+                        // It consumes the break byte from the leftover buffer so subsequent calls to the CborReader
+                        // see the next item.
+
                         RefillBuffer(1); // Skip the break marker (0xFF)
                     }
                     _nestingStack.Pop();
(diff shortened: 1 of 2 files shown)


extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
```diff
@@ -106,13 +106,6 @@ private void RefillBuffer(int bytesToSkip = 0)
             // Update the total size of valid data in our buffer.
             _currentChunkSize = leftoverBytesCount + bytesReadFromStream;
 
-            // Check for a malformed stream: if we have leftovers but the stream is empty,
-            // it means the CBOR data was truncated.
-            if (bytesReadFromStream == 0 && leftoverBytesCount > 0)
```
muhammad-othman, line 111: This check isn't valid anymore since the current logic does call `RefillBuffer` multiple times to make sure we are really at the end of the stream not just the current buffer, so we will fall back to the outer `ExecuteRead` to catch when a stream is ending with an incomplete CBOR data item.

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
```diff
@@ -139,7 +132,7 @@ private T ExecuteRead<T>(Func<CborReader, T> readOperation)
                 {
                     return readOperation(_internalCborReader);
                 }
-                catch (CborContentException ex)
+                catch (Exception ex) when (ex is InvalidOperationException || ex is CborContentException)
```
muhammad-othman, line 135: Catching `InvalidOperationException` to account when CBOR reader assumes that we are reading a different type because of how the current buffer ends.

normj: Can you add this comment in the code?

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
```diff
@@ -247,26 +239,42 @@ public CborReaderState PeekState()
...
+                        if (state == CborReaderState.Finished && _nestingStack.Count > 0)
+                        {
+                            _logger.DebugFormat("PeekState returned Finished, but nesting stack is not empty. Attempting refill.");
+                            RefillBuffer(0);
+                            continue;
+                        }
+
+                        return state;
+                    }
+                    catch (CborContentException ex)
```
normj, line 284: Confirming you don't need to also catch the `InvalidOperationException` exception like you did above?

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
```diff
@@ -200,14 +193,13 @@ private void ReadEndContainer(CborContainerType expectedType, CborReaderState ex
 
                 if (state == CborReaderState.Finished)
                 {
+                    // Try to refill first, maybe the break marker is in the next chunk.
+                    RefillBuffer(0);
+
                     if (IsNextByteEndOfContainer())
```
normj, line 218: Add a comment why it matters that the next byte is end of container.

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
```diff
@@ -200,14 +193,13 @@ private void ReadEndContainer(CborContainerType expectedType, CborReaderState ex
...
                     {
                         RefillBuffer(1); // Skip the break marker (0xFF)
```
normj, line 228: Add a comment about calling `RefillBuffer` a second time on a already filled buffer to trigger an expansion of the size of the buffer for the reasons describe in the comment you will add for line 199.

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
muhammad-othman: Yes, this is intentional. The `CborContentException` is thrown when invalid CBOR content is encountered, which can occur due to a premature break caused by chunking, as might happen here. 
On the other hand, `InvalidOperationException` occurs when the reader tries to interpret data as a different type. This happens when the previous item consumed the last part of the current buffer, causing the `CborReader` to believe the next item is `CborReaderState.Finished`, unaware that more data exists in the stream. 
For instance, if we try calling `ReadTextString()`, we’ll get an `InvalidOperationException` because we are reading a different type. This is why both exceptions are caught when reading data. However, `.PeekState()` will just return `CborReaderState.Finished` without throwing an exception, as it doesn’t interpret the type when called. 

Getting `CborReaderState.Finished` isn’t accurate if we are still inside a container, which is why the condition `if (state == CborReaderState.Finished && _nestingStack.Count > 0)` is used above.

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
muhammad-othman: Added a comment explaining catching both exceptions.

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
muhammad-othman: Added a comment explaining this situation.

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
muhammad-othman: Added a comment explaining why we are calling it twice, this wont trigger an expansion though, since we will remove a byte from the buffer, which means this condition wont be true `if (leftoverBytesCount >= _buffer.Length)`

Issue #3950, opened 2025-08-08
AmazonS3Client's constructor overwrites credentials in caller provided AmazonS3Config object


ryagin:
### Describe the bug

Certain overloads of AmazonS3Client constructor accept both login/password (awsAccessKeyId/awsSecretAccessKey) and an instance of AmazonS3Config object.

Down the call chain, in AmazonServiceClient, the config object is modified by overwriting DefaultAWSCredentials:
`
            if(credentials != null)
                config.DefaultAWSCredentials = credentials;
`

If several instances of AmazonS3Client are created with same config instance but different credentials, previously set credentials will be overwritten by latest call.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

Common expectation is that config object created by the user code should not be modified by the library code.

### Current Behavior

Config object created by the user code is modified by the library code, causing confusion.

### Reproduction Steps

`
var config = new AmazonS3Config();

var client1 = new AmazonS3Client(username1, password1, config);

var client2 = new AmazonS3Client(username2, password2, config);

// client1 now inadvertently uses username2/password2
`

### Possible Solution

Save login/password independently from configuration or create private copy of configuration object.

### AWS .NET SDK and/or Package version used

AWSSDK.S3 4.0.4.2, but current code in main is still affected

### Targeted .NET Platform

All

### Operating System and version

All

afroz429:
Thanks for reporting the issue. The team will review tomorrow and provide an update.

dscpinheiro:
I just merged a PR fixing this issue (https://github.com/aws/aws-sdk-net/pull/3958) and it'll be included in today's release.

I'll update this issue when the release is complete.

dscpinheiro:
Today's release just completed (https://github.com/aws/aws-sdk-net/releases/tag/4.0.71.0), if you update the S3 package to the latest version this issue should be resolved.

PR #3958, opened 2025-08-14
Update SDK not to mutate credentials in config object

dscpinheiro:
Fixes https://github.com/aws/aws-sdk-net/issues/3950

## Description
In V4, we moved the `DefaultAWSCredentials` property from the service client to the client config (the reason was that other credentials related properties - such as configuring the SSO token provider - were already there). However, this created an issue when a) credentials are explicitly specified when creating a service client, and b) the same config object is shared across multiple clients (see original issue for an example).

This PR re-introduces the property so that we don't mutate the customer's config object. If set the explicit credentials will take preference, but customers can still use `s3Config.DefaultAWSCredentials = new ...` to share the same set of credentials across multiple clients.

## Testing
- Dry-run (in progress): `DRY_RUN-7a35fc89-b550-4a38-9a19-85832da65598`

## Types of changes

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/AmazonServiceClient.cs
+++ b/sdk/src/Core/Amazon.Runtime/AmazonServiceClient.cs
@@ -167,11 +167,7 @@ namespace Amazon.Runtime
             config.Validate();
             _config = config;
 
-            if (credentials != null)
-            {
-                ExplicitAWSCredentials = credentials;
-            }
-
+            ExplicitAWSCredentials = credentials;
             EndpointDiscoveryResolver = new EndpointDiscoveryResolver(config, _logger);
             Initialize();
             UpdateSecurityProtocol();
--- a/sdk/src/Core/Amazon.Runtime/AmazonServiceClient.cs
+++ b/sdk/src/Core/Amazon.Runtime/AmazonServiceClient.cs
@@ -45,6 +45,12 @@ namespace Amazon.Runtime
         protected RuntimePipeline RuntimePipeline { get; set; }
         public IClientConfig Config => _config;
         private readonly ClientConfig _config;
+
+        /// <summary>
+        /// Credentials explicitly specified when constructing the client.
+        /// </summary>
+        protected internal AWSCredentials DefaultAWSCredentials { get; private set; }
+
         protected virtual IServiceMetadata ServiceMetadata { get; } = new ServiceMetadata();
         protected virtual bool SupportResponseLogging
         {
@@ -159,11 +165,13 @@ namespace Amazon.Runtime
                 _logger = Logger.GetLogger(this.GetType());
 
             config.Validate();
-
-            if(credentials != null)
-                config.DefaultAWSCredentials = credentials;
-
             _config = config;
+
+            if (credentials != null)
+            {
+                DefaultAWSCredentials = credentials;
+            }
+
             EndpointDiscoveryResolver = new EndpointDiscoveryResolver(config, _logger);
             Initialize();
             UpdateSecurityProtocol();
@@ -205,6 +213,7 @@ namespace Amazon.Runtime
                 new RequestContext(this.Config.LogMetrics)
                 {
                     ClientConfig = this.Config,
+                    DefaultAWSCredentials = this.DefaultAWSCredentials,
                     Marshaller = options.RequestMarshaller,
                     OriginalRequest = request,
                     Unmarshaller = options.ResponseUnmarshaller,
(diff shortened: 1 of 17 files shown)


sdk/src/Services/DocDB/Custom/Internal/PreSignedUrlRequestHandler.cs
```diff
@@ -41,7 +42,7 @@ public class PreSignedUrlRequestHandler : PipelineHandler
         /// <param name="credentials"></param>
         public PreSignedUrlRequestHandler(AWSCredentials credentials)
         {
-            _credentials = credentials;
```
dscpinheiro, line 44: I don't think this code (and the other handlers below) were working properly... Credentials would only be not-null if they were explicitly specified when creating the service client (different than V3, where they'd be populated because of where the `FallbackCredentialsFactory` ran).

I also tried to modify these handlers to read the `Identity` from the context, but they run _before_ the auth resolver.

sdk/src/Core/Amazon.Runtime/IClientConfig.cs
```diff
@@ -65,15 +65,15 @@ public partial interface IClientConfig
         /// <summary> 
         /// <para> 
         /// The AWS credentials used for authenticating calls to AWS for services using AWS signature version 4 (SigV4). 
-        /// SigV4 is the most common authentication mechanism used for AWS service calls. If AWSCredentials are used as a 
-        /// parameter to the service client's constructor the value will be set on this property. 
+        /// SigV4 is the most common authentication mechanism used for AWS service calls.
```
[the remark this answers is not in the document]
dscpinheiro: The removed section is not true anymore, we don't want to modify the provided config object.

sdk/src/Core/Amazon.Runtime/AmazonServiceClient.cs
```diff
@@ -45,6 +45,12 @@ public abstract class AmazonServiceClient : IDisposable
         protected RuntimePipeline RuntimePipeline { get; set; }
         public IClientConfig Config => _config;
         private readonly ClientConfig _config;
+
+        /// <summary>
+        /// Credentials explicitly specified when constructing the client.
+        /// </summary>
+        protected internal AWSCredentials DefaultAWSCredentials { get; private set; }
```
normj: To avoid having 2 properties called `DefaultAWSCredentials` that mean slightly different things what do you think of `ExplicitAWSCredentials`? The meaning that this is set when a user explicitly passes in credentials to the service client.

sdk/src/Core/Amazon.Runtime/Pipeline/Contexts.cs
```diff
@@ -34,6 +34,7 @@ public interface IRequestContext
         RequestMetrics Metrics { get; }
         ISigner Signer { get; set; }
         BaseIdentity Identity { get; set; }
+        AWSCredentials DefaultAWSCredentials { get; }
```
normj: Same comment as the property name for service clients to use the name `ExplicitAWSCredentials` instead of reusing `DefaultAWSCredentials`.

ashishdhingra (approved):
Looks good. So we are falling back to `Config.DefaultAWSCredentials` if these are not explicitly set at service client level.

normj (changes requested):

philasmar (approved):

sdk/src/Core/Amazon.Runtime/AmazonServiceClient.cs
dscpinheiro: Updated in https://github.com/aws/aws-sdk-net/pull/3958/commits/d3c3771378f6dca00ecaee25c18708817c1e07a1

sdk/src/Core/Amazon.Runtime/Pipeline/Contexts.cs
dscpinheiro: Updated in https://github.com/aws/aws-sdk-net/pull/3958/commits/d3c3771378f6dca00ecaee25c18708817c1e07a1

sdk/src/Core/Amazon.Runtime/AmazonServiceClient.cs
```diff
@@ -159,11 +165,13 @@ protected AmazonServiceClient(AWSCredentials credentials, ClientConfig config)
                 _logger = Logger.GetLogger(this.GetType());
 
             config.Validate();
+            _config = config;
 
-            if(credentials != null)
-                config.DefaultAWSCredentials = credentials;
+            if (credentials != null)
```
normj: nit: Not sure why we need the `null` check. Can't you just always use `ExplicitAWSCredentials = credentials;` and if `credentials` is `null` then it just resets `ExplicitAWSCredentials` to null?

dscpinheiro: Updated in https://github.com/aws/aws-sdk-net/pull/3958/commits/bbb9a209641214c86f39e6c41cc438dffa626852

PR #3961, opened 2025-08-15
Fix issue #3959. Do not send empty lists over headers for S3

### LINKED ISSUE #3959: AWSSDK.S3 - AmazonS3Client.ListObjectsV2Async is broken in version 4.0.6.5 of this C# package
### Describe the bug

```csharp
                var request = new ListObjectsV2Request
                {
                    BucketName = ...,
                    Prefix = ...
                };
                response = await _s3Client.ListObjectsV2Async(request);
```

works in 4.0.6.4 and 100s of previous versions, but breaks in 4.0.6.5 with a BadRequest exception.  The error message is "Invalid attribute name specified.".  The exception details don't say which attribute and what's wrong with it.

I have reverted to 4.0.6.4 and all is well.

### Regression Issue

- [x] Select this option if this issue appears to be a regression.

### Expected Behavior

Objects are listed.

### Current Behavior

An exception is thrown.

### Reproduction Steps

```csharp
                var request = new ListObjectsV2Request
                {
                    BucketName = ...,
                    Prefix = ...
                };
                response = await _s3Client.ListObjectsV2Async(request);
```

### Possible Solution

Don't install version 4.0.6.5

_No response_

AWSSDK.S3 4.0.6.5

### Targeted .NET Platform

.NET 8

Windows 11

peterrsongg:
## Description

Fixes #3959 
## Motivation and Context

No longer sends empty header lists.

## Testing

I tested that the user's scenario works
## Types of changes

## Checklist

## License

### DIFF
--- a/generator/.DevConfigs/9c4edfd8-0eba-4193-9752-efb29cbc4de6.json
+++ b/generator/.DevConfigs/9c4edfd8-0eba-4193-9752-efb29cbc4de6.json
@@ -4,7 +4,7 @@
       "serviceName": "S3",
       "type": "patch",
       "changeLogMessages": [
-        "Fixes issue #3959. Do not send empty lists for headers"
+        "Fixes issue #3959. Do not send empty lists for headers. ListObjectsV2 calls were failing if AWSConfigs.InitializeCollections was set to true. This was causing the SDK to send the `x-amz-optional-attributes` as an empty string which is not a valid value."
       ]
     }
   ]
--- a/generator/.DevConfigs/9c4edfd8-0eba-4193-9752-efb29cbc4de6.json
+++ b/generator/.DevConfigs/9c4edfd8-0eba-4193-9752-efb29cbc4de6.json
@@ -1,0 +1,11 @@
+{
+  "services": [
+    {
+      "serviceName": "S3",
+      "type": "patch",
+      "changeLogMessages": [
+        "Fixes issue #3959. Do not send empty lists for headers"
+      ]
+    }
+  ]
+}
--- a/generator/ServiceClientGeneratorLib/Generators/SourceFiles/StructureGenerator.tt
+++ b/generator/ServiceClientGeneratorLib/Generators/SourceFiles/StructureGenerator.tt
@@ -385,7 +385,8 @@ namespace <#=this.Config.Namespace#>.Model
                     // restxml and restjson is the only protocol where we send empty lists in headers. See following protocol tests.
                     // https://github.com/smithy-lang/smithy/blob/de486fa42c7bce4afc802bef95990795eeeed25a/smithy-aws-protocol-tests/model/restJson1/http-headers.smithy#L427
                     // https://github.com/smithy-lang/smithy/blob/de486fa42c7bce4afc802bef95990795eeeed25a/smithy-aws-protocol-tests/model/restXml/http-headers.smithy#L370
-                        if (this.Config.ServiceModel.Type == ServiceType.Rest_Xml || this.Config.ServiceModel.Type == ServiceType.Rest_Json)
+                    // S3 is the exception to this rule. We do not want to send empty header lists for S3.
+                        if (this.Config.ServiceId != "S3" && this.Config.ServiceModel.Type == ServiceType.Rest_Xml || this.Config.ServiceModel.Type == ServiceType.Rest_Json)
                         {
 
 #>
@@ -395,7 +396,7 @@ namespace <#=this.Config.Namespace#>.Model
                         else
                         {
 #>
-            return this.<#=member.VariableName#> != null && this.<#=member.VariableName#>.Count > 0;
+            return this.<#=member.VariableName#> != null && (this.<#=member.VariableName#>.Count > 0 || !AWSConfigs.InitializeCollections);
 <#
                         }
                     }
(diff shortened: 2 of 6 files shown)


generator/ServiceClientGeneratorLib/Generators/SourceFiles/StructureGenerator.tt
```diff
@@ -395,7 +396,7 @@ namespace <#=this.Config.Namespace#>.Model
                         else
                         {
 #>
-            return this.<#=member.VariableName#> != null && this.<#=member.VariableName#>.Count > 0;
+            return this.<#=member.VariableName#> != null && (this.<#=member.VariableName#>.Count > 0 || !AWSConfigs.InitializeCollections);
```
boblodgett, line 399: Does this work when AWSConfigs.InitializeCollections = true but there isn't anything in the list?  Seems in that case you still won't send the empty header list.

peterrsongg: if AWSConfigs.InitializeCollections = true, but there isn't anything in the list, it won't send the empty header list which is the behavior that we want

normj: My understanding is we don't want to send the empty list in that scenario. S3 doesn't handle it and in the case of an operation that used to have a value and an update is being called we would accidently clear out the existing value when the user doesn't modify the property at all.

peterrsongg: the problem was we were sending empty lists over headers and i guess for S3 we shouldn't be doing that... even though

https://github.com/smithy-lang/smithy/blob/de486fa42c7bce4afc802bef95990795eeeed25a/smithy-aws-protocol-tests/model/restXml/http-headers.smithy#L370

generator/ServiceClientGeneratorLib/Generators/SourceFiles/StructureGenerator.tt
```diff
@@ -385,7 +385,8 @@ namespace <#=this.Config.Namespace#>.Model
                     // restxml and restjson is the only protocol where we send empty lists in headers. See following protocol tests.
                     // https://github.com/smithy-lang/smithy/blob/de486fa42c7bce4afc802bef95990795eeeed25a/smithy-aws-protocol-tests/model/restJson1/http-headers.smithy#L427
                     // https://github.com/smithy-lang/smithy/blob/de486fa42c7bce4afc802bef95990795eeeed25a/smithy-aws-protocol-tests/model/restXml/http-headers.smithy#L370
-                        if (this.Config.ServiceModel.Type == ServiceType.Rest_Xml || this.Config.ServiceModel.Type == ServiceType.Rest_Json)
+                    // S3 is the exception to this rule. We do not want to send empty header lists for S3.
+                        if (this.Config.ServiceId != "S3" && this.Config.ServiceModel.Type == ServiceType.Rest_Xml || this.Config.ServiceModel.Type == ServiceType.Rest_Json)
                         {
 
 #>
```
normj, line 392: Why wouldn't this block also need the same check for `AWSConfigs.InitializeCollections`?

generator/.DevConfigs/9c4edfd8-0eba-4193-9752-efb29cbc4de6.json
```diff
@@ -0,0 +1,11 @@
+{
+  "services": [
+    {
+      "serviceName": "S3",
+      "type": "patch",
+      "changeLogMessages": [
+        "Fixes issue #3959. Do not send empty lists for headers"
```
normj: It would be good to put more in the description on how the user's is affected by this? What was broken from the user's point of view.

generator/ServiceClientGeneratorLib/Generators/SourceFiles/StructureGenerator.tt
peterrsongg: hmm I guess if the protocol test is right, we should be doing something like

```
return this.<#=member.VariableName#> != null || !AWSConfigs.InitializeCollections
```

however, that would change a bunch of services and I'd rather keep this change limited to S3. If I make this change now, then I'll have to do a full dry run. Whereas if I limit the change to just S3 then we know it only changes 3 files and we can release the fix quickly

generator/.DevConfigs/9c4edfd8-0eba-4193-9752-efb29cbc4de6.json
peterrsongg: fair enough i'll update this

generator/ServiceClientGeneratorLib/Generators/SourceFiles/StructureGenerator.tt
boblodgett: So if the user has initialized the collection but it is empty and AWSConfigs.InitializeCollections = false we will still send an empty header list. Is that correct?

peterrsongg: If the user initialized the collection, but put nothing in there it will fail. But that has always been the case, even with the previous code.

boblodgett: Are you sure because there was a length check before so it wouldn't send it.

generator/ServiceClientGeneratorLib/Generators/SourceFiles/StructureGenerator.tt
normj: Okay, with shipping this fix for S3 now but we should revisit this condition check before resolving this issue.

boblodgett (approved):

PR #3986, opened 2025-08-28
Add specific inject xml code customization hooks for getters & setters. Remove unused custom partials.

peterrsongg:
## Description

In this PR i had more customization hooks to allow injecting code in specific places such as the getters, setters and private member declarations.
## Motivation and Context

The reason I wanted to add more customization hooks is because, whenever possible we want to generate members of shapes. If we decide to exclude a member via the `excludeMembers` customization hook I added, then the docs for that shape won't be generated and we could easily miss a doc update for that member if the docs were to be updated. This makes it so that the member will still be generated since instead of moving that member to a custom partial of a class, we just inject the customized code.

## Testing

Dry run DRY_RUN-32ff75b0-08bd-4a88-ae5f-18dc16d8e114

## Types of changes

## Checklist

## License


sdk/src/Services/S3/Custom/Model/GetObjectMetadataRequest.cs
```diff
@@ -1,83 +0,0 @@
...
-        /// <summary>
-        /// Returns the object only if it has been modified since the specified time, 
-        /// otherwise returns a PreconditionFailed.
-        /// </summary>
-        public DateTime? ModifiedSinceDate
-        {
-            get { return this.modifiedSinceDate ?? DateTime.SpecifyKind(default, DateTimeKind.Utc); }
-            set
-            {
-                if (value == null)
```
peterrsongg, line 40: I wasn't sure why this `if` `else` was here. Isn't this just the same thing as doing `this.modifiedSinceDate=value`? If not, i can inject this as well. but it seemed redundant

sdk/src/Services/S3/Generated/Model/GetObjectMetadataRequest.cs
```diff
@@ -423,6 +425,47 @@ internal bool IsSetKey()
...
+        /// <para>
+        /// Then Amazon S3 returns the <c>304 Not Modified</c> response code.
+        /// </para>
+        ///  
+        /// <para>
+        /// For more information about conditional requests, see <a href="https://tools.ietf.org/html/rfc7232">RFC
+        /// 7232</a>.
+        /// </para>
+        /// </summary>
+        public DateTime? ModifiedSinceDate
```
peterrsongg, line 457: docs are generated :)

generator/ServiceModels/s3/s3.customizations.json
```diff
@@ -479,6 +486,30 @@
...
+                            "   return null;",
+                            "}"
+
+                        ],
+                        "skipSetter" : true,
+                        "injectXmlIsSet":[
+                            "return Type != null;"
+                        ],
+                        "injectXmlPrivateMemberAssignment":[
+                            ""
```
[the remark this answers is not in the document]
peterrsongg: this is my way of telling the generator to skip private member assignment. I'd rather do this than create another hook to only skip private member assignemnts

dscpinheiro (approved):

sdk/src/Services/S3/Custom/Model/GetObjectMetadataRequest.cs
boblodgett: The `if` isn't needed.

PR #3987, opened 2025-08-28
Update transfer utility class hierarchy

GarrettBeatty:
## Description
TransferDownloadDirectoryRequest currently has fields which are already located in [BaseDownloadRequest](https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Services/S3/Custom/Transfer/BaseDownloadRequest.cs). This change makes TransferDownloadDirectoryRequest extend BaseDownloadRequest so that we dont have to duplicate fields. 

In the next PR, we will be adding more common fields to BaseDownloadRequest (in order to comply with having all of the required s3 fields in our high level Transfer Utility Request). To make it easier to review I am separating the changes into multiple PRs (this PR and future PRs)

Please note there is one small difference which is we are changing `ModifiedSinceDate` and `UnmodifiedSinceDate` default behavior.

TransferUtilityDownloadDirectoryRequest
```
 public DateTime ModifiedSinceDate
        {
            get { return this.modifiedSinceDate.GetValueOrDefault(); }
            set { this.modifiedSinceDate = value; }
        }
```

to be 

BaseDownloadRequest
```
   public DateTime ModifiedSinceDate
        {
            get { return this.modifiedSinceDate ?? DateTime.SpecifyKind(default, DateTimeKind.Utc); }
            set
            {
                this.modifiedSinceDate = value;
            }
        }
```

From my testing (see Peter's comment in PR) - the only difference between the two is that the `DateTimeKind` was not set in the first approach. And in the new code it will be set to UTC.

## Motivation and Context
DOTNET-8277

## Testing
dry run - 52990dbb-1153-40cc-aad0-04c8b6c39b70 pass

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/Transfer/TransferUtilityDownloadDirectoryRequest.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/TransferUtilityDownloadDirectoryRequest.cs
@@ -36,7 +36,7 @@ namespace Amazon.S3.Transfer
     /// <summary>
     /// Request object for downloading a directory with the TransferUtility.
     /// </summary>
-    public class TransferUtilityDownloadDirectoryRequest : BaseDownloadRequest
+    public class TransferUtilityDownloadDirectoryRequest : BaseTransferRequest
     {
         private string s3Directory;
         private string localDirectory;
--- a/sdk/src/Services/S3/Custom/Transfer/TransferUtilityDownloadDirectoryRequest.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/TransferUtilityDownloadDirectoryRequest.cs
@@ -36,46 +36,12 @@ namespace Amazon.S3.Transfer
     /// <summary>
     /// Request object for downloading a directory with the TransferUtility.
     /// </summary>
-    public class TransferUtilityDownloadDirectoryRequest
+    public class TransferUtilityDownloadDirectoryRequest : BaseDownloadRequest
     {
-        private string bucketName;
         private string s3Directory;
         private string localDirectory;
         private bool downloadFilesConcurrently = false;
-        private DateTime? modifiedSinceDate;
-        private DateTime? unmodifiedSinceDate;
         private bool disableSlashCorrection = false;
-
-        private ServerSideEncryptionCustomerMethod serverSideCustomerEncryption;
-        private string serverSideEncryptionCustomerProvidedKey;
-        private string serverSideEncryptionCustomerProvidedKeyMD5;
-
-        private RequestPayer requestPayer;
-
-        /// <summary>
-        /// 	Gets or sets the name of the bucket.
-        /// </summary>
-        /// <value>
-        /// 	The name of the bucket.
-        /// </value>
-        public string BucketName
-        {
-            get { return this.bucketName; }
-            set { this.bucketName = value; }
-        }
-
-        /// <summary>
-        /// Gets whether or not the bucket name is set.
-        /// </summary>
-        /// <returns>
-        /// 	A value of <c>true</c> if the bucket name is set.
-        ///    Otherwise, returns <c>false</c>.
-        /// </returns>
-        internal bool IsSetBucketName()
-        {
-            return !System.String.IsNullOrEmpty(this.bucketName);
-        }
-
 
         /// <summary>
         /// 	Gets or sets the local directory where objects from Amazon S3 will be downloaded.  
@@ -132,51 +98,6 @@ namespace Amazon.S3.Transfer
             return !System.String.IsNullOrEmpty(this.s3Directory);
         }
 
-        /// <summary>
-        /// 	Gets or sets the <c>ModifiedSinceDate</c> property.  
-        /// 	Only objects that have been modified since this date will be
-        /// 	downloaded.
-        /// </summary>
-        /// <value>
-        /// 	The <c>ModifiedSinceDate</c> property. 
-        /// </value>
-        public DateTime ModifiedSinceDate
-        {
-            get { return this.modifiedSinceDate.GetValueOrDefault(); }
-            set { this.modifiedSinceDate = value; }
-        }
-
-        /// <summary>
-        /// Checks if ModifiedSinceDate property is set.
-        /// </summary>
-        /// <returns>A value of <c>true</c> if ModifiedSinceDate property is set.
-        /// 	Otherwise, returns <c>false</c>.</returns>
-        internal bool IsSetModifiedSinceDate()
-        {
-            return modifiedSinceDate.HasValue;
-        }
-
-        /// <summary>
-        /// 	Gets or sets the <c>UnmodifiedSinceDate</c> property.  
-        /// 	Only objects that have not been modified since this date will be downloaded.
-        /// </summary>
-        /// <value>
-        /// 	The <c>UnmodifiedSinceDate</c> property.
-        /// </value>
-        public DateTime UnmodifiedSinceDate
-        {
-            get { return this.unmodifiedSinceDate.GetValueOrDefault(); }
-            set { this.unmodifiedSinceDate = value; }
-        }
-
-        /// <summary>
-        /// Checks if UnmodifiedSinceDate property is set.
-        /// </summary>
-        /// <returns>true if UnmodifiedSinceDate property is set.</returns>
-        internal bool IsSetUnmodifiedSinceDate()
-        {
-            return unmodifiedSinceDate.HasValue;
-        }
 
         /// <summary>
         /// Gets or sets the DownloadFilesConcurrently property.
@@ -205,56 +126,6 @@ namespace Amazon.S3.Transfer
             set { this.disableSlashCorrection = value; }
         }
 
-        /// <summary>
-        /// The Server-side encryption algorithm to be used with the customer provided key.
-        /// </summary>
-        public ServerSideEncryptionCustomerMethod ServerSideEncryptionCustomerMethod
-        {
-            get { return this.serverSideCustomerEncryption; }
-            set { this.serverSideCustomerEncryption = value; }
-        }
-
-        /// <summary>
-        /// The base64-encoded encryption key for Amazon S3 to use to decrypt the object
-        /// <para>
-        /// Using the encryption key you provide as part of your request Amazon S3 manages both the encryption, as it writes 
-        /// to disks, and decryption, when you access your objects. Therefore, you don't need to maintain any data encryption code. The only 
-        /// thing you do is manage the encryption keys you provide.
-        /// </para>
-        /// <para>
-        /// When you retrieve an object, you must provide the same encryption key as part of your request. Amazon S3 first verifies 
-        /// the encryption key you provided matches, and then decrypts the object before returning the object data to you.
-        /// </para>
-        /// <para>
-        /// Important: Amazon S3 does not store the encryption key you provide.
-        /// </para>
-        /// </summary>
-        [AWSProperty(Sensitive = true)]
-        public string ServerSideEncryptionCustomerProvidedKey
-        {
-            get { return this.serverSideEncryptionCustomerProvidedKey; }
-            set { this.serverSideEncryptionCustomerProvidedKey = value; }
-        }
-
-        /// <summary>
-        /// The MD5 of the customer encryption key specified in the ServerSideEncryptionCustomerProvidedKey property. The MD5 is
-        /// base 64 encoded. This field is optional, the SDK will calculate the MD5 if this is not set.
-        /// </summary>
-        public string ServerSideEncryptionCustomerProvidedKeyMD5
-        {
-            get { return this.serverSideEncryptionCustomerProvidedKeyMD5; }
-            set { this.serverSideEncryptionCustomerProvidedKeyMD5 = value; }
-        }
-
-        /// <summary>
-        /// Confirms that the requester knows that they will be charged for the request. 
-        /// Bucket owners need not specify this parameter in their requests.
-        /// </summary>
-        public RequestPayer RequestPayer
-        {
-            get { return this.requestPayer; }
-            set { this.requestPayer = value; }
-        }
 
         /// <summary>
         /// The event for DownloadedDirectoryProgressEvent notifications. All
--- a/generator/.DevConfigs/d31339e8-845c-4988-8657-8840683fe4b2.json
+++ b/generator/.DevConfigs/d31339e8-845c-4988-8657-8840683fe4b2.json
@@ -1,0 +1,11 @@
+{
+  "services": [
+    {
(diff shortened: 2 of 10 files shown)


sdk/src/Services/S3/Custom/Transfer/TransferUtilityDownloadDirectoryRequest.cs
```diff
@@ -132,51 +98,6 @@ internal bool IsSetS3Directory()
...
-
-        /// <summary>
-        /// 	Gets or sets the <c>UnmodifiedSinceDate</c> property.  
-        /// 	Only objects that have not been modified since this date will be downloaded.
-        /// </summary>
-        /// <value>
-        /// 	The <c>UnmodifiedSinceDate</c> property.
-        /// </value>
-        public DateTime UnmodifiedSinceDate
-        {
```
peterrsongg, line 167: Can you check that the behavior isn't different? In the `BaseDownloadRequest` I see
https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Services/S3/Custom/Transfer/BaseDownloadRequest.cs#L145-L152

there is some extra logic. Same goes for `ModifiedSinceDate`

peterrsongg (changes requested):

sdk/src/Services/S3/Custom/Transfer/TransferUtilityDownloadDirectoryRequest.cs
GarrettBeatty: good catch ill have to double check this

GarrettBeatty: so i made a test program and validated that the two approaches are basically identical. The only thing that is different is the `Kind` -  `Unspecified vs Utc`.

In both cases the date will be `1/1/0001 12:00:00 AM` when its not set.

```
using System;

namespace DateTimeComparisonApp
{
    class Program
    {
        static void Main(string[] args)
        {
            Console.WriteLine("=== ModifiedSinceDate Implementation Comparison ===");
            Console.WriteLine();

            // Test both implementations when values are not set (null)
            TestUnsetValues();
            Console.WriteLine();

            // Test both implementations when values are explicitly set
            TestSetValues();
            Console.WriteLine();

            // Test edge cases
            TestEdgeCases();
            Console.WriteLine();

            Console.WriteLine("Press any key to exit...");
            Console.ReadKey();
        }

        static void TestUnsetValues()
        {
            Console.WriteLine("📋 TESTING UNSET VALUES (when DateTime? is null):");
            Console.WriteLine("".PadRight(50, '='));

            // Simulate the two approaches
            DateTime? modifiedSinceDate = null;  // Simulating unset property
            DateTime? unmodifiedSinceDate = null; // Simulating unset property

            // Implementation 1: GetValueOrDefault() approach (current TransferUtilityDownloadDirectoryRequest?)
            var result1 = modifiedSinceDate.GetValueOrDefault();

            // Implementation 2: Null coalescing with SpecifyKind approach (BaseDownloadRequest)
            var result2 = unmodifiedSinceDate ?? DateTime.SpecifyKind(default, DateTimeKind.Utc);

            Console.WriteLine($"Method 1 (GetValueOrDefault): {result1} (Kind: {result1.Kind})");
            Console.WriteLine($"Method 2 (Null coalescing + UTC): {result2} (Kind: {result2.Kind})");
            Console.WriteLine();

            // Compare the results
            Console.WriteLine("🔍 COMPARISON RESULTS:");
            Console.WriteLine($"✓ Same Ticks: {result1.Ticks == result2.Ticks}");
            Console.WriteLine($"✓ Same Date: {result1.Date == result2.Date}");
            Console.WriteLine($"✓ Same TimeOfDay: {result1.TimeOfDay == result2.TimeOfDay}");
            Console.WriteLine($"✓ Same DateTime Value: {result1 == result2}");
            Console.WriteLine($"❗ Different Kind: {result1.Kind} vs {result2.Kind}");

            // Check if they're equivalent when converted to UTC
            var result1Utc = result1.ToUniversalTime();
            var result2Utc = result2.ToUniversalTime();
            Console.WriteLine($"✓ UTC Conversion Identical: {result1Utc == result2Utc}");

            Console.WriteLine();
            Console.WriteLine("🎯 CONCLUSION: Implementations are functionally equivalent!");
            Console.WriteLine("   Only DateTimeKind differs (Unspecified vs UTC)");
            Console.WriteLine("   UTC is more appropriate for S3 operations.");
        }

        static void TestSetValues()
        {
            Console.WriteLine("📋 TESTING SET VALUES (when DateTime? has a value):");
            Console.WriteLine("".PadRight(50, '='));

            var testDateTime = new DateTime(2023, 12, 25, 14, 30, 0, DateTimeKind.Local);
            
            // Simulate both approaches with set values
            DateTime? modifiedSinceDate = testDateTime;
            DateTime? unmodifiedSinceDate = testDateTime;

            var result1 = modifiedSinceDate.GetValueOrDefault();
            var result2 = unmodifiedSinceDate ?? DateTime.SpecifyKind(default, DateTimeKind.Utc);

            Console.WriteLine($"Test Value: {testDateTime} (Kind: {testDateTime.Kind})");
            Console.WriteLine($"Method 1 Result: {result1} (Kind: {result1.Kind})");
            Console.WriteLine($"Method 2 Result: {result2} (Kind: {result2.Kind})");
            Console.WriteLine();

            Console.WriteLine("🔍 COMPARISON RESULTS:");
            Console.WriteLine($"✓ Same Ticks: {result1.Ticks == result2.Ticks}");
            Console.WriteLine($"✓ Same DateTime Value: {result1 == result2}");
            Console.WriteLine($"✓ Same Kind: {result1.Kind == result2.Kind}");

            Console.WriteLine();
            Console.WriteLine("🎯 CONCLUSION: When values are set, both implementations behave identically!");
        }

        static void TestEdgeCases()
        {
            Console.WriteLine("📋 TESTING EDGE CASES:");
            Console.WriteLine("".PadRight(50, '='));

            // Test with different DateTimeKind values
            var utcDateTime = new DateTime(2023, 1, 1, 0, 0, 0, DateTimeKind.Utc);
            var localDateTime = new DateTime(2023, 1, 1, 0, 0, 0, DateTimeKind.Local);
            var unspecifiedDateTime = new DateTime(2023, 1, 1, 0, 0, 0, DateTimeKind.Unspecified);

            Console.WriteLine("Testing with different DateTimeKind values:");
            
            TestEdgeCase("UTC DateTime", utcDateTime);
            TestEdgeCase("Local DateTime", localDateTime);
            TestEdgeCase("Unspecified DateTime", unspecifiedDateTime);
        }

        static void TestEdgeCase(string description, DateTime testValue)
        {
            DateTime? modifiedSinceDate = testValue;
            DateTime? unmodifiedSinceDate = testValue;

            var result1 = modifiedSinceDate.GetValueOrDefault();
            var result2 = unmodifiedSinceDate ?? DateTime.SpecifyKind(default, DateTimeKind.Utc);

            Console.WriteLine($"  {description}:");
            Console.WriteLine($"    Input: {testValue} (Kind: {testValue.Kind})");
            Console.WriteLine($"    Method 1: {result1} (Kind: {result1.Kind})");
            Console.WriteLine($"    Method 2: {result2} (Kind: {result2.Kind})");
            Console.WriteLine($"    Identical: {result1 == result2 && result1.Kind == result2.Kind}");
            Console.WriteLine();
        }
    }
}

```

```
 TESTING UNSET VALUES (when DateTime? is null):
==================================================
Method 1 (GetValueOrDefault): 1/1/0001 12:00:00 AM (Kind: Unspecified)
Method 2 (Null coalescing + UTC): 1/1/0001 12:00:00 AM (Kind: Utc)

🔍 COMPARISON RESULTS:
✓ Same Ticks: True
✓ Same Date: True
✓ Same TimeOfDay: True
✓ Same DateTime Value: True
❗ Different Kind: Unspecified vs Utc
✓ UTC Conversion Identical: False

🎯 CONCLUSION: Implementations are functionally equivalent!
   Only DateTimeKind differs (Unspecified vs UTC)
   UTC is more appropriate for S3 operations.

📋 TESTING SET VALUES (when DateTime? has a value):
==================================================
Test Value: 12/25/2023 2:30:00 PM (Kind: Local)
Method 1 Result: 12/25/2023 2:30:00 PM (Kind: Local)
Method 2 Result: 12/25/2023 2:30:00 PM (Kind: Local)

🔍 COMPARISON RESULTS:
✓ Same Ticks: True
✓ Same DateTime Value: True
✓ Same Kind: True

🎯 CONCLUSION: When values are set, both implementations behave identically!

📋 TESTING EDGE CASES:
==================================================
Testing with different DateTimeKind values:
  UTC DateTime:
    Input: 1/1/2023 12:00:00 AM (Kind: Utc)
    Method 1: 1/1/2023 12:00:00 AM (Kind: Utc)
    Method 2: 1/1/2023 12:00:00 AM (Kind: Utc)
    Identical: True

  Local DateTime:
    Input: 1/1/2023 12:00:00 AM (Kind: Local)
    Method 1: 1/1/2023 12:00:00 AM (Kind: Local)
    Method 2: 1/1/2023 12:00:00 AM (Kind: Local)
    Identical: True

  Unspecified DateTime:
    Input: 1/1/2023 12:00:00 AM (Kind: Unspecified)
    Method 1: 1/1/2023 12:00:00 AM (Kind: Unspecified)
    Method 2: 1/1/2023 12:00:00 AM (Kind: Unspecified)
    Identical: True
```

GarrettBeatty: my initial thoughts are that the `Kind` should not matter in this case in reality since the default value would be `1/1/0001` and there would be no real functional difference between UTC and unspecified kind in this case because the date is so old

normj (commented):
The inherited `Key` and `VersionId` from the base class would make no sense for the download directory request and could cause confusion.

If this was needed for some consolidation of logic reading the properties I suggest adding an internal interface that has just the properties you need.

GarrettBeatty:
> The inherited `Key` and `VersionId` from the base class would make no sense for the download directory request and could cause confusion.
> 
> If this was needed for some consolidation of logic reading the properties I suggest adding an internal interface that has just the properties you need but don't make this inheritance change.

good point i didnt notice those extra properties

GarrettBeatty:
@normj im wondering if we can do something like this (i updated this pr with a sample implementation of below)

Current
```
BaseDownloadRequest (abstract)
├── BucketName
├── Key
├── VersionId
├── ModifiedSinceDate, UnmodifiedSinceDate
├── ServerSideEncryptionCustomerMethod
├── ServerSideEncryptionCustomerProvidedKey
├── ServerSideEncryptionCustomerProvidedKeyMD5
├── ChecksumMode
├── RequestPayer
└── (All shared S3 single-object download properties)
    │
    ├── TransferUtilityDownloadRequest : BaseDownloadRequest
    │   ├── FilePath
    │   └── WriteObjectProgressEvent
    │
    └── TransferUtilityOpenStreamRequest : BaseDownloadRequest
        └── (No new fields; used for streaming)
        
TransferUtilityDownloadDirectoryRequest (standalone class)
├── BucketName
├── S3Directory
├── LocalDirectory
├── ModifiedSinceDate, UnmodifiedSinceDate
├── ServerSideEncryptionCustomerMethod
├── ServerSideEncryptionCustomerProvidedKey
├── ServerSideEncryptionCustomerProvidedKeyMD5
├── RequestPayer
├── etc
```

Proposed


```
BaseTransferRequest (NEW)
├── BucketName
├── ModifiedSinceDate, UnmodifiedSinceDate
├── Server-side encryption properties
├── ChecksumMode, RequestPayer
└── (All shared S3 transfer properties)
    │
    ├── BaseDownloadRequest : BaseTransferRequest (UPDATED)
    │   ├── Key (object-specific)
    │   └── VersionId (object-specific)
    │       │
    │       └── TransferUtilityDownloadRequest : BaseDownloadRequest
    │           └── FilePath, additional download-specific properties
    │
    └── TransferUtilityDownloadDirectoryRequest : BaseTransferRequest (UPDATED)
        ├── S3Directory, LocalDirectory  
        ├── SearchPattern, SearchOption
        └── (Directory-specific properties - NO Key/VersionId)

```

we would need to change some of the hierarchy in the public classes. This would reduce the code duplication but then still let each class have its specific properties when needed

normj:
> @normj im wondering if we can do something like this (i updated this pr with a sample implementation of below)
> 
> Current
> 
> ```
> BaseDownloadRequest (abstract)
> ├── BucketName
> ├── Key
> ├── VersionId
> ├── ModifiedSinceDate, UnmodifiedSinceDate
> ├── ServerSideEncryptionCustomerMethod
> ├── ServerSideEncryptionCustomerProvidedKey
> ├── ServerSideEncryptionCustomerProvidedKeyMD5
> ├── ChecksumMode
> ├── RequestPayer
> └── (All shared S3 single-object download properties)
>     │
>     ├── TransferUtilityDownloadRequest : BaseDownloadRequest
>     │   ├── FilePath
>     │   └── WriteObjectProgressEvent
>     │
>     └── TransferUtilityOpenStreamRequest : BaseDownloadRequest
>         └── (No new fields; used for streaming)
>         
> TransferUtilityDownloadDirectoryRequest (standalone class)
> ├── BucketName
> ├── S3Directory
> ├── LocalDirectory
> ├── ModifiedSinceDate, UnmodifiedSinceDate
> ├── ServerSideEncryptionCustomerMethod
> ├── ServerSideEncryptionCustomerProvidedKey
> ├── ServerSideEncryptionCustomerProvidedKeyMD5
> ├── RequestPayer
> ├── etc
> ```
> 
> Proposed
> 
> ```
> BaseTransferRequest (NEW)
> ├── BucketName
> ├── ModifiedSinceDate, UnmodifiedSinceDate
> ├── Server-side encryption properties
> ├── ChecksumMode, RequestPayer
> └── (All shared S3 transfer properties)
>     │
>     ├── BaseDownloadRequest : BaseTransferRequest (UPDATED)
>     │   ├── Key (object-specific)
>     │   └── VersionId (object-specific)
>     │       │
>     │       └── TransferUtilityDownloadRequest : BaseDownloadRequest
>     │           └── FilePath, additional download-specific properties
>     │
>     └── TransferUtilityDownloadDirectoryRequest : BaseTransferRequest (UPDATED)
>         ├── S3Directory, LocalDirectory  
>         ├── SearchPattern, SearchOption
>         └── (Directory-specific properties - NO Key/VersionId)
> ```
> 
> we would need to change some of the hierarchy in the public classes. This would reduce the code duplication but then still let each class have its specific properties when needed

I'm 99% sure this causes a binary breaking change and users would have to recompile which might not be possible depending on if the SDK is being brought in as a transitive dependency.

PR #3995, opened 2025-09-03
Fix Cbor edge cases when unmarshalling deeply nested maps/arrays and optimized numbers

muhammad-othman:
## Description
This PR fixes edge cases in the CBOR protocol reader/unmarshaller:

Added support in `CborSimpleTypeUnmarshaller` for reading CBOR’s smallest numeric representation.

For the nested maps/arrays issue, CBOR has two ways to end arrays/maps:
* Indefinite-length: there’s an explicit break byte 0xFF.
* Definite-length: there is no terminator byte; the container is “over” once you’ve read the declared number of items (for maps, pairs).

Our streaming reader relied on `PeekState()` or break bytes (`0xFF`) to detect container ends. This worked when containers ended at the stream boundary or stayed within a single chunk, but failed for deeply nested, definite-length containers that spanned multiple chunks.
After reading the last item, `PeekState()` would return the next item’s state instead of EndArray/EndMap, since there was no on-wire terminator. This caused desync issues when calling ReadEndArray/Map.

We now track containers explicitly:
* Push a frame on ReadStartArray/Map with RemainingItems (arrays = count, maps = count × 2).
* Decrement RemainingItems after each item read/skip.
* When RemainingItems == 0, we infer EndArray/EndMap without consuming bytes.
* For indefinite-length containers, we still check for and skip the 0xFF break marker.
* After closing a nested container, the parent’s frame is decremented once.

By maintaining a per-container item counter (arrays: n, maps: 2n) and decrementing on each value, we can know container end precisely, keep `PeekState()` correct, and close containers properly even when the boundary is split across chunks.

## Motivation and Context
`DOTNET-8252`

## Testing
- `DRY_RUN-c81ddfaa-9196-4bb2-8d89-f35fe9e9b6bf`
- Added unit tests to cover these fixes.

## Types of changes

## Checklist

## License

### DIFF
--- a/extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
+++ b/extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
@@ -28,11 +28,30 @@ namespace Amazon.Extensions.CborProtocol.Internal
     /// </summary>
     public class CborStreamReader : IDisposable
     {
+        /// <summary>
+        /// Represents a single container frame (map or array) within the nesting stack
+        /// of the <see cref="CborStreamReader"/>.
+        /// </summary>
         private class ContainerFrame
         {
+            /// <summary>
+            /// Gets the type of CBOR container (map or array).
+            /// </summary>
             public CborContainerType Type { get; }
+
+            /// <summary>
+            /// Gets the number of items left to be read in this container.
+            /// For indefinite-length containers, this value is <c>null</c>.
+            /// </summary>
             public int? RemainingItems { get; private set; }
 
+            /// <summary>
+            /// Initializes a new instance of the <see cref="ContainerFrame"/> class.
+            /// </summary>
+            /// <param name="type">The container type (map or array).</param>
+            /// <param name="remainingItems">
+            /// The number of items in the container, or <c>null</c> for indefinite-length containers.
+            /// </param>
             public ContainerFrame(CborContainerType type, int? remainingItems)
             {
                 Type = type;
@@ -44,6 +63,10 @@ namespace Amazon.Extensions.CborProtocol.Internal
                 }
             }
 
+            /// <summary>
+            /// Decrements the <see cref="RemainingItems"/> count when an item is consumed.
+            /// For indefinite-length containers, this method does nothing.
+            /// </summary>
             public void ConsumeItem()
             {
                 RemainingItems = RemainingItems.HasValue ? RemainingItems.Value - 1 : (int?)null;
@@ -58,8 +81,12 @@ namespace Amazon.Extensions.CborProtocol.Internal
                 }
             }
 
+            /// <summary>
+            /// Gets a value indicating whether this container has been fully consumed.
+            /// </summary>
             public bool IsComplete => RemainingItems == 0;
         }
+
         /// <summary>
         /// Enum to track the type of CBOR container (map or array)
         /// for state management within the CborStreamReader.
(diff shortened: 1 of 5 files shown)


extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
```diff
@@ -28,6 +28,38 @@ namespace Amazon.Extensions.CborProtocol.Internal
     /// </summary>
     public class CborStreamReader : IDisposable
     {
+        private class ContainerFrame
+        {
+            public CborContainerType Type { get; }
+            public int? RemainingItems { get; private set; }
+
+            public ContainerFrame(CborContainerType type, int? remainingItems)
```
peterrsongg, line 55: can you add docs for this? Like what is a `ContainerFrame` and what does it represent? I know it is in the PR description but we should comment here so we know

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
```diff
@@ -28,6 +28,38 @@ namespace Amazon.Extensions.CborProtocol.Internal
...
+            {
+                RemainingItems = RemainingItems.HasValue ? RemainingItems.Value - 1 : (int?)null;
+
+                if (RemainingItems < 0)
+                {
+                    // This should never happen if container boundaries are respected.
+                    // It indicates that more items were read than the container declared.
+                    throw new CborContentException(
+                        $"Too many items read from {Type} container. Expected {RemainingItems + 1} more item(s)."
+                    );
```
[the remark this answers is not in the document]
muhammad-othman: We don't have the original count, and I don't think we want to keep track of the original count just for the exception message especially that this isn't an expected exception anyway and shouldn't happen in normal circumstances.

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
```diff
@@ -305,17 +381,38 @@ public CborReaderState PeekState()
...
+        public int ReadInt32() => ExecuteValueRead(r => r.ReadInt32());
+        public long ReadInt64() => ExecuteValueRead(r => r.ReadInt64());
+        public ulong ReadUInt64() => ExecuteValueRead(r => r.ReadUInt64());
+        public decimal ReadDecimal() => ExecuteValueRead(r => r.ReadDecimal());
+        public double ReadDouble() => ExecuteValueRead(r => r.ReadDouble());
+        public bool ReadBoolean() => ExecuteValueRead(r => r.ReadBoolean());
+        public float ReadSingle() => ExecuteValueRead(r => r.ReadSingle());
+        public byte[] ReadByteString() => ExecuteValueRead(r => r.ReadByteString());
+        public void ReadNull() => ExecuteValueRead(r => { r.ReadNull(); return true; });
+        public void SkipValue() => ExecuteValueRead(r => { r.SkipValue(); return true; });
```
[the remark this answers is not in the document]
muhammad-othman: I prefer returning a dummy true value over creating an additional method.

extensions/test/CborProtocol.Tests/CborStreamReaderTests.cs
```diff
@@ -618,5 +618,232 @@ public void Unmarshall_MultipleNestedIndefiniteMapsEndingAtStreamEnd_ShouldSucce
         reader.ReadEndMap(); // should correctly handle 0xFF at end
     }
 
+    [Fact]
+    public void PeekState_DefiniteArrayAcrossBuffer_ReturnsEndArrayBeforeNextValue()
```
peterrsongg, line 622: have you tried testing something like this  (also across buffers)


Nested array (same goes for a nested map which has a mixture of definite and indefinite length)
```
0x9f018202039f0405ffff
9F        -- Start indefinite-length array
   01     -- 1
   82     -- Array of length 2
      02  -- 2
      03  -- 3
   9F     -- Start indefinite-length array
      04  -- 4
      05  -- 5
      FF  -- "break" (inner array)
   FF     -- "break" (outer array)
```

muhammad-othman: Add unit tests to cover these cases.

extensions/src/AWSSDK.Extensions.CborProtocol/Internal/CborStreamReader.cs
muhammad-othman: Added.

PR #3999, opened 2025-09-04
Add Support for Multiple Auth Schemes and SigV4a

AlexDaines:
## Description
Implements multi-auth scheme support. This enables clients to configure and prioritize the available authentication schemes (SigV4, SigV4a, Bearer, NoAuth) through multiple configuration sources. 

Key changes made:
- Added `AuthScheme`, `AuthSchemePreference`, and resolver infrastructure
- Implemented configuration hierarchy: client > env vars > config file > global
- Added SigV4a region set configuration support
- Preserved backwards compatibility via extension methods pattern (no breaking interface changes)
- Added comprehensive test coverage

## Motivation and Context
Required for upcoming SigV4a adoption and services that need flexible auth scheme selection. Several teams have been waiting on this capability for cross-region signing and failover scenarios.

## Types of changes

## Checklist

## License


sdk/src/Core/Amazon.Runtime/AuthScheme.cs
dscpinheiro, line 1: I added @muhammad-othman and I as reviewers, but even before we take a look you shouldn't have to create these classes. V4 already has auth schemes defined here: https://github.com/aws/aws-sdk-net/tree/b6b4057137cbb6bf39bc9f22687a3441921b1d33/sdk/src/Core/Amazon.Runtime/Credentials/Internal

AlexDaines: very good call out. refactored to remove the duplicated classes

sdk/src/Core/Amazon.Runtime/ClientConfigExtensions.cs
```diff
@@ -0,0 +1,103 @@
...
+    public static class ClientConfigExtensions
+    {
+        /// <summary>
+        /// Gets a value indicating whether the SignatureMethod property was explicitly set by the user.
+        /// This is used for backwards compatibility to determine when legacy SignatureMethod configuration
+        /// should take precedence over auth scheme preferences.
+        /// </summary>
+        /// <param name="clientConfig">The client configuration.</param>
+        /// <returns>True if SignatureMethod was explicitly set; otherwise, false.</returns>
+        public static bool GetIsSignatureMethodExplicitlySet(this IClientConfig clientConfig)
```
dscpinheiro: Let's follow-up offline, but I don't understand why this is needed. Wouldn't it be simpler to update the `IClientConfig` interface to include the properties so we don't have to use reflection here?

(Also I wonder if this would work with an AOT application)

sdk/src/Core/Amazon.Runtime/Internal/InternalConfiguration.cs
```diff
@@ -285,6 +299,18 @@ private string GetClientAppIdEnvironmentVariable()
 
             return rawValue;
         }
+
+        /// <summary>
+        /// Gets a simple string value from an environment variable.
+        /// </summary>
+        private string GetStringEnvironmentVariable(string name)
```
dscpinheiro: Does the SEP say anything about trimming the value? There's already a `GetEnvironmentVariable<T>`  method that could be re-used (`T` for the new variable would be `string`).

generator/.DevConfigs/3aa6313d-9526-40ba-b09c-e046e0d4ef2f.json
```diff
@@ -0,0 +1,13 @@
+{
+  "core": {
+    "changeLogMessages": [
+      "Implement multi-auth SEP with authentication scheme preference and SigV4a region set configuration",
```
dscpinheiro: For the final changelog, we will want to refactor these (SEP doesn't mean anything to customers - and even internal terms like auth resolver are not relevant). 

We should focus on the actual impact - e.g. ability to set SigV4A explicitly.

sdk/src/Core/Amazon.Runtime/Credentials/Internal/AuthSchemeOption.cs
```diff
@@ -23,11 +23,31 @@ public class AuthSchemeOption : IAuthSchemeOption
...
+        /// </summary>
+        /// <param name="schemeId">The fully qualified scheme ID (e.g., "aws.auth#sigv4").</param>
+        /// <returns>The short name (e.g., "sigv4") or the original schemeId if no '#' is present.</returns>
+        public static string GetNameFromSchemeId(string schemeId)
+        {
+            if (string.IsNullOrEmpty(schemeId))
+                return schemeId;
+
+            var parts = schemeId.Split('#');
+            return parts.Length > 1 ? parts[1] : schemeId;
```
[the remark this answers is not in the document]
dscpinheiro: This seems like a good suggestion to accept.

sdk/src/Core/Amazon.Runtime/ClientConfig.cs
```diff
@@ -221,7 +222,20 @@ public abstract string ServiceVersion
...
+                this.isSignatureMethodExplicitlySet = true;
+            }
+        }
+
+        /// <summary>
+        /// Gets a value indicating whether the SignatureMethod property was explicitly set by the user.
+        /// This is used for backwards compatibility to determine when legacy SignatureMethod configuration
+        /// should take precedence over auth scheme preferences.
+        /// </summary>
+        public bool IsSignatureMethodExplicitlySet
```
dscpinheiro: I don't follow why this is needed... The naming is confusing but signature method is not going to be SigV4 / SigV4A, it's this enum instead: https://github.com/aws/aws-sdk-net/blob/0980c7ca63599f992adf1f7560a5caac8119906d/sdk/src/Core/Amazon.Runtime/Enumerations.cs#L24-L28

sdk/src/Core/Amazon.Util/Internal/RootConfig.cs
```diff
@@ -70,6 +75,8 @@ public RootConfig()
             CborReaderInitialBufferSize = AWSConfigs._cborReaderInitialBufferSize;
             CorrectForClockSkew = true;
             DisableLegacyPersistenceStore = AWSConfigs._disableLegacyPersistenceStore;
+            
+            // Authentication configuration is resolved on-demand from environment and config files
```
dscpinheiro: Out-of-place comment?

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs
```diff
@@ -283,6 +283,133 @@ protected static List<IAuthSchemeOption> RetrieveSchemesFromEndpoint(Endpoint en
...
+
+        /// <summary>
+        /// Resolves authentication scheme preference using configuration precedence hierarchy.
+        /// </summary>
+        private string ResolveAuthSchemePreference(IClientConfig clientConfig)
+        {
+            // Check for legacy SignatureMethod configuration (backwards compatibility for S3)
+            if (clientConfig?.SignatureMethod != null && clientConfig.IsSignatureMethodExplicitlySet)
+            {
+                bool isS3Service = clientConfig.ServiceId?.Equals("S3", StringComparison.OrdinalIgnoreCase) == true ||
```
dscpinheiro: (I left the other comment first...) I see the SEP mentions S3 (and Control) but I don't think this logic is needed here (the SEP was referring how the service model - which the auth resolvers are generated from - would keep older properties with the same values for backwards compatibility).

sdk/src/Core/Amazon.Runtime/ClientConfig.cs
```diff
@@ -1185,6 +1199,28 @@ public TelemetryProvider TelemetryProvider
...
+        /// <summary>
+        /// Gets or sets the SigV4a signing region set for this client.
+        /// <para>
+        /// This property allows you to specify a comma-separated list of regions (e.g., "us-east-1,us-west-2")
+        /// that will be used for SigV4a signing. The region set determines which regions the signed request is valid for.
+        /// If not set, the client will use environment variables, configuration files,
+        /// endpoints metadata, or fall back to the client's configured region.
+        /// </para>
+        /// </summary>
+        public string SigV4aSigningRegionSet { get; set; }
```
dscpinheiro: Unless I'm missing something obvious... Where is this used? I don't see it referenced in the auth resolver handler (and also for legacy reasons, we have SigV4A logic in the endpoint resolver too...)

dscpinheiro (changes requested):

sdk/src/Core/Amazon.Runtime/Internal/InternalConfiguration.cs
```diff
@@ -266,6 +282,20 @@ private string GetEC2MetadataEndpointEnvironmentVariable()
             return rawValue;
         }       
 
+        /// <summary>
+        /// Loads a string value from an environment variable.
+        /// </summary>
+        /// <returns>The environment variable value or null if not set</returns>
+        private string GetStringEnvironmentVariable(string environmentVariableName)
```
muhammad-othman, line 288: I think we can just use `GetEnvironmentVariable<string>`, not need to add an additional version of it.

sdk/src/Core/Amazon.Runtime/ClientConfig.cs
```diff
@@ -444,6 +446,46 @@ public string AuthenticationServiceName
...
+            get 
+            { 
+                if (!string.IsNullOrEmpty(this.authSchemePreference))
+                    return this.authSchemePreference;
+                
+                // Fallback to environment variable or config file:
+                // 1. Environment variable: AWS_AUTH_SCHEME_PREFERENCE
+                // 2. Config file: auth_scheme_preference
+                return FallbackInternalConfigurationFactory.AuthSchemePreference;
+            }
```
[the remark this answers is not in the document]
AlexDaines: This is intentional behavior. empty strings trigger fallback behavior. Tested in TestPreferenceList_EmptyString_NoReordering and TestSigV4aRegionSet_EmptyConfiguration.

sdk/src/Core/Amazon.Runtime/ClientConfig.cs
```diff
@@ -444,6 +446,46 @@ public string AuthenticationServiceName
...
+            get 
+            { 
+                if (!string.IsNullOrEmpty(this.sigV4aSigningRegionSet))
+                    return this.sigV4aSigningRegionSet;
+                
+                // Fallback to environment variable or config file:
+                // 1. Environment variable: AWS_SIGV4A_SIGNING_REGION_SET
+                // 2. Config file: sigv4a_signing_region_set
+                return FallbackInternalConfigurationFactory.SigV4aSigningRegionSet;
+            }
```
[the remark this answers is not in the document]
AlexDaines: See above

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs
```diff
@@ -283,6 +289,94 @@ protected static List<IAuthSchemeOption> RetrieveSchemesFromEndpoint(Endpoint en
...
+        private static string GetFullSchemeId(string schemeName)
+        {
+            // Handle common auth scheme names (case-sensitive as per specification)
+            if (schemeName == "sigv4")
+                return "aws.auth#sigv4";
+            if (schemeName == "sigv4a")
+                return "aws.auth#sigv4a";
+            if (schemeName == "httpBearerAuth" || schemeName == "bearer")
+                return "smithy.api#httpBearerAuth";
+            if (schemeName == "noAuth")
```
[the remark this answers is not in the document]
AlexDaines: Design spec explicitly requires case-sensitive matching

extensions/src/AWSSDK.Extensions.CrtIntegration/CrtAWS4aSigner.cs
```diff
@@ -150,6 +159,7 @@ public AWS4aSigningResult SignRequest(IRequest request,
             var signedCrtRequest = signingResult.Get().SignedRequest;
 
             CrtHttpRequestConverter.CopyHeadersFromCrtRequest(request, signedCrtRequest);
+            request.Headers[HeaderKeys.XAmzRegionSetHeader] = regionSet;
```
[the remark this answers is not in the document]
muhammad-othman: I don't think we should set this header, we should just update the region that the signer uses and it should set whatever header it needs.

extensions/src/AWSSDK.Extensions.CrtIntegration/CrtAWS4aSigner.cs
```diff
@@ -194,14 +204,18 @@ public AWS4aSigningResult Presign4a(IRequest request,
...
+                // S3 requests require special URI encoding handling for compatibility
                 request.UseDoubleEncoding = false;
             }
 
             var signedAt = AWS4Signer.InitializeHeaders(request.Headers, request.Endpoint);
             request.SignedAt = CorrectClockSkew.GetCorrectedUtcNowForEndpoint(request.Endpoint.ToString());
-            var regionSet = overrideSigningRegion ?? AWS4Signer.DetermineSigningRegion(clientConfig, clientConfig.RegionEndpointServiceName, request.AlternateEndpoint, request);
+            
+            // Use explicit override, then SigV4aSigningRegionSet, then fall back to single region
+            var regionSet = overrideSigningRegion
```
muhammad-othman: nit: multiple ternary operator is is bit hard to read for me, I think simple if conditions is more readable.

extensions/src/AWSSDK.Extensions.CrtIntegration/CrtAWS4aSigner.cs
```diff
@@ -194,14 +204,18 @@ public AWS4aSigningResult Presign4a(IRequest request,
         {
             if (serviceSigningName == "s3")
             {
-                // Older versions of the S3 package can be used with newer versions of Core, this guarantees no double encoding will be used.
-                // The new behavior uses endpoint resolution rules, which are not present prior to 3.7.100
+                // S3 requests require special URI encoding handling for compatibility
```
muhammad-othman: Multiple comments are updated in this file, is this intentional?

extensions/src/AWSSDK.Extensions.CrtIntegration/CrtAWS4aSigner.cs
```diff
@@ -119,12 +119,21 @@ public AWS4aSigningResult SignRequest(IRequest request,
...
+            string regionSet;
+            if (!string.IsNullOrEmpty(request.SigV4aSigningRegionSet))
+            {
+                regionSet = request.SigV4aSigningRegionSet;
+            }
+            else
+            {
+                regionSet = AWS4Signer.DetermineSigningRegion(clientConfig, clientConfig.RegionEndpointServiceName, request.AlternateEndpoint, request);
+            }
             request.DeterminedSigningRegion = regionSet;
```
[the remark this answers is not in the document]
muhammad-othman: This is an interesting take, I don't think we have to add another property, but we should test some requests with multiple regions and update the documentation of `DeterminedSigningRegion` to mention that it can be multiple regions.

extensions/test/CrtIntegrationTests/V4aSignerTests.cs
```diff
@@ -136,6 +136,36 @@ internal static IRequest BuildHeaderRequestToSign(string resourcePath, Dictionar
...
+        /// <summary>
+        /// Dummy request class needed for DefaultRequest constructor
+        /// </summary>
+        private class DummyRequest : AmazonWebServiceRequest { }
+
+        /// <summary>
+        /// Creates a test request with mutable collections to verify header modifications
+        /// during signing.
+        /// </summary>
+        internal static IRequest CreateDefaultRequest(string resourcePath)
```
muhammad-othman: Is there a reason to create this method instead of using `BuildHeaderRequestToSign`?

extensions/test/CrtIntegrationTests/V4aSignerTests.cs
```diff
@@ -440,5 +470,167 @@ public void TestChunkedRequestWithTrailingHeaders()
...
+        /// <summary>
+        /// Tests multi-region SigV4a signing by comparing signed requests with different region configurations.
+        ///
+        /// The CRT library handles signature calculation but does not automatically add the x-amz-region-set
+        /// header to the HTTP request. The SDK must add this header after signing so AWS services can
+        /// validate which regions the signature covers.
+        ///
+        /// </summary>
+        [Fact]
+        public void TestMultiRegionSigV4a_DifferentialVerification()
```
muhammad-othman: I think we should use `InlineData` for the three test cases that are tested here, similar to `SignRequestViaHeadersWithSigv4a`

extensions/src/AWSSDK.Extensions.CrtIntegration/CrtAWS4aSigner.cs
AlexDaines: I actually just swapped it to be ternary because the branching looked quite messy 😅

sdk/src/Core/Amazon.Runtime/Internal/InternalConfiguration.cs
AlexDaines: Can't use the generic method as it has a `where T : struct` constraint. The string handling code in there seems to be dead code that never runs 🤔

extensions/test/CrtIntegrationTests/V4aSignerTests.cs
AlexDaines: We need the real DefaultRequest because CRT modifies the headers directly during signing. The mock version returns an immutable dictionary so the CRT can't add headers to it. The tests check that those headers actually got added to the request

muhammad-othman: image
The dictionary looks like a regular dictionary that the CRT can add to it.

dscpinheiro: I'm still not sure we actually need to change any tests in the extensions folder.

sdk/src/Core/Amazon.Runtime/Internal/DefaultRequest.cs
```diff
@@ -475,6 +475,13 @@ public string CanonicalResourcePrefix
         /// </summary>
         public string AuthenticationRegion { get; set; }
 
+        /// <summary>
+        /// The signing region set to use for SigV4a requests.
+        /// Contains a comma-separated list of regions for multi-region signing.
+        /// Set from Config.SigV4aSigningRegionSet or endpoints metadata.
+        /// </summary>
+        public string SigV4aSigningRegionSet { get; set; }
```
dscpinheiro: I don't think this is needed. We should be able to leverage the existing logic for SigV4A without adding new properties here.

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs
```diff
@@ -65,7 +68,7 @@ protected void PreInvoke(IExecutionContext executionContext)
                 if (scheme == null)
                 {
                     // Current auth scheme option is not enabled / supported, continue iterating.
-                    Logger.DebugFormat($"{authOptions[i].SchemeId} scheme is not supported for {executionContext.RequestContext.RequestName}");
+                    Logger?.DebugFormat($"{authOptions[i].SchemeId} scheme is not supported for {executionContext.RequestContext.RequestName}");
```
dscpinheiro, line 71: Why this change? `Logger` won't be null.

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs
```diff
@@ -283,6 +289,94 @@ protected static List<IAuthSchemeOption> RetrieveSchemesFromEndpoint(Endpoint en
...
+            // Then add any remaining options that weren't in the preference list
+            foreach (var option in authOptions)
+            {
+                if (!reorderedOptions.Contains(option))
+                {
+                    reorderedOptions.Add(option);
+                }
+            }
+
+            // CRITICAL: Ensure NoAuth/Anonymous is always last in the list.
```
dscpinheiro, line 333: I thought we discussed this, but if we didn't: This is not necessary.

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs
```diff
@@ -283,6 +289,94 @@ protected static List<IAuthSchemeOption> RetrieveSchemesFromEndpoint(Endpoint en
...
+                return "smithy.api#httpBearerAuth";
+            if (schemeName == "noAuth")
+                return "smithy.api#noAuth";
+            
+            // If it already looks like a full ID (contains #), return as-is
+            if (schemeName.Contains("#"))
+                return schemeName;
+                
+            // Otherwise, assume it's an AWS auth scheme
+            return $"aws.auth#{schemeName}";
```
dscpinheiro, line 368: We should not be assuming anything here. I think the logic could be a lot simpler by comparing the scheme name from preferences with the supported value from the model.

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseEndpointResolver.cs
```diff
@@ -88,6 +88,25 @@ public virtual void ProcessRequestHandlers(IExecutionContext executionContext)
             {
                 requestContext.Request.AuthenticationRegion = config.AuthenticationRegion;
             }
+            
+            // Set SigV4a region set if configured (either from endpoint metadata or explicit config)
+            if (requestContext.Request.SignatureVersion == SignatureVersion.SigV4a)
```
dscpinheiro: Similar to my other comment: This should use the existing signing region set logic.

sdk/src/Core/Amazon.Runtime/Pipeline/Contexts.cs
```diff
@@ -175,6 +180,11 @@ public IDictionary<string, object> ContextAttributes
         }
 
         public IHttpRequestStreamHandle RequestStreamHandle { get; set; }
+        
+        /// <summary>
+        /// The region set for SigV4a signing.
+        /// </summary>
+        public string SigV4aSigningRegionSet { get; set; }
```
dscpinheiro: Can probably be removed too.

sdk/src/Core/Amazon.Runtime/ClientConfig.cs
```diff
@@ -444,6 +446,46 @@ public string AuthenticationServiceName
...
+        /// </summary>
+        public string AuthSchemePreference
+        {
+            get 
+            { 
+                if (!string.IsNullOrEmpty(this.authSchemePreference))
+                    return this.authSchemePreference;
+                
+                // Fallback to environment variable or config file:
+                // 1. Environment variable: AWS_AUTH_SCHEME_PREFERENCE
```
dscpinheiro: These comments are redundant. I don't expect the variable name to ever change but these comments don't add any value.

sdk/test/NetStandard/UnitTests/ClientConfigTests.cs
```diff
@@ -72,7 +72,9 @@ public class ClientConfigTests
             "RequestChecksumCalculation",
             "ResponseChecksumValidation",
             "IdentityResolverConfiguration",
-            "DefaultAWSCredentials"
+            "DefaultAWSCredentials",
```
dscpinheiro, line 75: I should've noticed this before: When this test fails, it mentions other places that should change too (e.g. `AWSSDK.Extensions.NETCore.Setup` - see the comment on line 97)

sdk/test/UnitTests/Custom/Runtime/AlternativeAuthResolutionTests.cs
```diff
@@ -0,0 +1,364 @@
...
+            };
+
+            var responseContext = new ResponseContext();
+
+            return new Amazon.Runtime.Internal.ExecutionContext(requestContext, responseContext);
+        }
+
+        /// <summary>
+        /// Test implementation of BaseAuthResolverHandler that provides test auth options.
+        /// This simulates how endpoints 2.0 auth schemes override service defaults.
```
dscpinheiro: I'm being annoying again, but EP2.0 does not matter here. All of these tests seem more redundant than anything.

Issue #4003, opened 2025-09-07
`NullReferenceException` when calling `CreatePresignedPostAsync`


akilin:
### Describe the bug

Hi, I was tinkering with `AWSSDK.S3` and a local `Minio` instance and encountered a `NullReferenceException` when attempting to create a presigned post URL.

The issue might be related to the fact that I am using the `ServiceURL` property (its description states that using it will set Region to null).

Once I call `CreatePresignedPostAsync` - it goes down the stack until it reaches this method:

https://github.com/aws/aws-sdk-net/blob/97bce892d2066a69ad725938efc479989a4968a8/sdk/src/Services/S3/Custom/Util/S3PostUploadSignedPolicy.cs#L116-L123

which, as far as I can tell, throws `NullReferenceException` when trying to get `region.SystemName` because `region` is `null`

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

Url is created and no error is throw

### Current Behavior

Error is thrown

### Reproduction Steps

```csharp
        var endpoint = "http://localhost:9000";
        var accessKey = "admin";
        var secretKey = "admin";

        var bucket = "demo-bucket";

        var config = new AmazonS3Config
        {
            ServiceURL = endpoint,
            ForcePathStyle = true
        };
        var credentials = new BasicAWSCredentials(accessKey, secretKey);
        IAmazonS3 s3 = new AmazonS3Client(credentials, config);

        await s3.EnsureBucketExistsAsync(bucket);

        var req = new CreatePresignedPostRequest
        {
            Key = "potato",
            BucketName = bucket,
            Expires = DateTime.UtcNow.AddMinutes(60)
        };
        var res = await s3.CreatePresignedPostAsync(req);
```

### AWS .NET SDK and/or Package version used

`<PackageReference Include="AWSSDK.S3" Version="4.0.6.13" />`

### Targeted .NET Platform

net9.0

### Operating System and version

win11

akilin:
cc @GarrettBeatty (you asked to let you know if there were any issues with `CreatePresignedPostAsync` as part of the discussion in here https://github.com/aws/aws-sdk-net/issues/1901)

ashishdhingra:
`NullReferenceException` is also thrown by using S3 URL explicitly:
```C#
var endpoint = "https://s3.us-east-2.amazonaws.com";
var bucket = "testbucket-presignedurlpost-issue4003";

var config = new AmazonS3Config
{
[...]

PR #4006, opened 2025-09-11
Fix NullReferenceException creating a presigned URL with a service url

normj:
## Description
If a service url is set for S3 service client the region property on the service config will be null. The presigned url code was only checking for region on the service config and getting a NullReferenceException. I added the code to check to see if there is a ServiceUrl and get the region from there. I change the signature of some internal APIs to take in `string` instead of `RegionEndpoint` because the utility methods we have for getting the region from the service url return back a string.

## Motivation and Context
https://github.com/aws/aws-sdk-net/issues/4003

## Testing
Add new test to reproduce the issue.
Dry Run: Success (DRY_RUN-c72ee41b-9de4-4a31-9cb7-005ac958d17d)

### DIFF
--- a/sdk/src/Services/S3/Custom/AmazonS3Client.Extensions.cs
+++ b/sdk/src/Services/S3/Custom/AmazonS3Client.Extensions.cs
@@ -677,7 +677,11 @@ namespace Amazon.S3
             var policyDocument = BuildPolicyDocument(request);
 
             string region;
-            if (!string.IsNullOrEmpty(this.Config.ServiceURL))
+            if (!string.IsNullOrEmpty(this.Config.AuthenticationRegion))
+            {
+                region = this.Config.AuthenticationRegion;
+            }
+            else if (!string.IsNullOrEmpty(this.Config.ServiceURL))
             {
                 region = AWSSDKUtils.DetermineRegion(this.Config.ServiceURL);
             }
--- a/sdk/test/Services/S3/UnitTests/Custom/CreatePresignedPostRequestTests.cs
+++ b/sdk/test/Services/S3/UnitTests/Custom/CreatePresignedPostRequestTests.cs
@@ -13,15 +13,15 @@
  * permissions and limitations under the License.
  */
 
-using System;
-using System.Threading.Tasks;
-using System.Linq;
-using Microsoft.VisualStudio.TestTools.UnitTesting;
-
+using Amazon.Runtime;
 using Amazon.S3;
 using Amazon.S3.Model;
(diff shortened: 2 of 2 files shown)


sdk/src/Services/S3/Custom/AmazonS3Client.Extensions.cs
```diff
@@ -676,8 +676,18 @@ private CreatePresignedPostResponse BuildPresignedPostResponse(CreatePresignedPo
             // Build the policy document
             var policyDocument = BuildPolicyDocument(request);
 
+            string region;
+            if (!string.IsNullOrEmpty(this.Config.ServiceURL))
```
ashishdhingra: Should it not also try using `AuthenthicationRegion` as we do in `GetPreseignedUrl`? In customer's scenario, they are using Minio.

normj: Good catch. Updated the code to check for `AuthenthicationRegion`.

sdk/src/Services/S3/Custom/AmazonS3Client.Extensions.cs
```diff
@@ -676,8 +676,22 @@ private CreatePresignedPostResponse BuildPresignedPostResponse(CreatePresignedPo
...
             var policyDocument = BuildPolicyDocument(request);
 
+            string region;
+            if (!string.IsNullOrEmpty(this.Config.AuthenticationRegion))
+            {
+                region = this.Config.AuthenticationRegion;
+            }
+            else if (!string.IsNullOrEmpty(this.Config.ServiceURL))
+            {
+                region = AWSSDKUtils.DetermineRegion(this.Config.ServiceURL);
```
GarrettBeatty, line 686: i dont know if it matters but i see in https://github.com/aws/aws-sdk-net/blob/a2b02603307114026d69940e6d3894600fb5d88b/sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4Signer.cs#L665 we do `ToLowerInvariant` too. I am assuming it doesnt matter but might be good to confirm

normj: The call to `ToLowerInvariant` in the signer seems unnecessary. There is no example of AWS regions ever being in anything besides lowercase. I assume that was done because else were in that code is getting the region from the URL and the URL could come from somewhere that used mix cased. That is a guess but I don't think we need to continue adding the unnecessary `ToLowerInvariant` call.

PR #4009, opened 2025-09-13
Migrate the self service model option of generator to V4

shruti0085:
The [VS IDE extensions telemetry generator ](https://github.com/aws/aws-toolkit-common/blob/8c88537fae2ac7e6524fb2b29ae336c606850eeb/telemetry/csharp/TelemetryClient.proj#L72)uses the AWS SDK generator to produce its own service client. It is now being migrated to use sdk v4. However the self service option of the generator still points to v3.x of dependencies such as Core. This change updates it to v4.

## License

### DIFF
--- a/generator/.DevConfigs/57988500-f010-4dca-80ba-ddc4ab2b2271.json
+++ b/generator/.DevConfigs/57988500-f010-4dca-80ba-ddc4ab2b2271.json
@@ -1,9 +1,0 @@
-{
-  "core": {
-    "updateMinimum": true,
-    "type": "minor",
-    "changeLogMessages": [
-      "Migrates the self service model option of the generator to use V4"
-    ]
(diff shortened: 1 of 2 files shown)


generator/.DevConfigs/57988500-f010-4dca-80ba-ddc4ab2b2271.json
```diff
@@ -0,0 +1,9 @@
+{
```
normj: You don't need a dev config for this generator change because it doesn't affect any code shipped as part of the release. Only the self service mode of the generator for internal users like yourself.

shruti0085: Updated

Issue #4012, opened 2025-09-19
S3 PutObject fails with zero-byte InputStream and DisablePayloadSigning


wjrogers:
### Describe the bug

Sending a `PutObjectRequest` with a zero-byte `InputStream` and `DisablePayloadSigning = true` throws an exception:

> The request contained trailing data that was not well-formed or did not conform to our published schema.

### Regression Issue

- [x] Select this option if this issue appears to be a regression.

### Expected Behavior

I expect the operation to succeed even when the input is zero bytes long.

### Current Behavior

```txt
Amazon.S3.AmazonS3Exception: The request contained trailing data that was not well-formed or did not conform to our published schema.
 ---> Amazon.Runtime.Internal.HttpErrorResponseException: Exception of type 'Amazon.Runtime.Internal.HttpErrorResponseException' was thrown.
   at Amazon.Runtime.HttpWebRequestMessage.ProcessHttpResponseMessage(HttpResponseMessage responseMessage)
   at Amazon.Runtime.HttpWebRequestMessage.GetResponseAsync(CancellationToken cancellationToken)
   at Amazon.Runtime.Internal.HttpHandler`1.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.Runtime.Internal.RedirectHandler.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.Runtime.Internal.Unmarshaller.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.S3.Internal.AmazonS3ResponseHandler.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.Runtime.Internal.ErrorHandler.InvokeAsync[T](IExecutionContext executionContext)
   --- End of inner exception stack trace ---
   at Amazon.Runtime.Internal.HttpErrorResponseExceptionHandler.HandleExceptionStream(IRequestContext requestContext, IWebResponseData httpErrorResponse, HttpErrorResponseException exception, Stream responseStream)
   at Amazon.Runtime.Internal.HttpErrorResponseExceptionHandler.HandleExceptionAsync(IExecutionContext executionContext, HttpErrorResponseException exception)
   at Amazon.Runtime.Internal.ExceptionHandler`1.HandleAsync(IExecutionContext executionContext, Exception exception)
   at Amazon.Runtime.Internal.ErrorHandler.ProcessExceptionAsync(IExecutionContext executionContext, Exception exception)
[...]

PR #4013, opened 2025-09-20
Skip adding extra empty chunk for empty streams in TrailingHeadersWrapperStream

muhammad-othman:
## Description
Fixes an issue where sending a `PutObjectRequest` with a zero-byte InputStream and `DisablePayloadSigning = true` throws:
```
The request contained trailing data that was not well-formed or did not conform to our published schema.
```

This happens because we are adding add an extra empty chunk (`0`) at the end to mark the end of the stream before the trailing headers
```
4
x  
0
x-amz-checksum-crc32:iQSVhA==
```

This is correct for requests with some data like the above one, but for zero byte stream we send the following 
```
0

0
x-amz-checksum-crc32:AAAAAA==
```
And sending 2x `0` isn't acceptable by S3 and causing `The request contained trailing data that was not well-formed or did not conform to our published schema.` exception.

This PR fixes this issue by only adding the additional empty chunk when there the stream isn't empty.

## Motivation and Context
https://github.com/aws/aws-sdk-net/issues/4012

## Testing
* Tested the reproduction steps in https://github.com/aws/aws-sdk-net/issues/4012.
* Added unit tests that uploads empty stream with `DisablePayloadSigning`  and without it.
* `DRY_RUN-d4ddc2eb-3a0f-4ddc-a28e-4d4663bfe9a8`

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Internal/Util/TrailingHeadersWrapperStream.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Util/TrailingHeadersWrapperStream.cs
@@ -216,11 +216,16 @@ namespace Amazon.Runtime.Internal.Util
         {
             var trailer = new StringBuilder();
 
-            // End the data chunk
-            trailer.Append(STREAM_NEWLINE);
+            // Avoid adding an empty chunk for an empty stream. Adding the empty chunk for an empty stream will trigger
+            // service error for malformed trailing headers.
+            if (_baseStream.Length > 0)
+            {
+                // End the data chunk
+                trailer.Append(STREAM_NEWLINE);
 
-            // Append a chunk of size 0
-            trailer.Append(EMPTY_CHUNK);
+                // Append a chunk of size 0
+                trailer.Append(EMPTY_CHUNK);
+            }
 
             // Append trailing headers, including special handling for the checksum.
             // The order here must match the order of keys sent already in the X-Amz-Trailer header.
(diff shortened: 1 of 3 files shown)


sdk/src/Core/Amazon.Runtime/Internal/Util/TrailingHeadersWrapperStream.cs
```diff
@@ -216,11 +216,15 @@ private string GenerateTrailingHeaderChunk()
         {
             var trailer = new StringBuilder();
 
-            // End the data chunk
-            trailer.Append(STREAM_NEWLINE);
+            // If this is an empty stream then there is no need to add an additional empty chunk
```
normj: nit: Suggest comment rewrite because the first time I read this my brain assumed the comment here was explaining the logic if it went into the if block. Sure it was just my brain being off so if you prefer your comment I'm okay.
```
// Avoid adding an empty chunk for an empty stream. Adding the empty chunk for an empty stream will trigger
// service error for malformed trailing headers.
```

muhammad-othman: Nope, it isn't you, I'm very bad with comments :D. Updated to the suggested comment.

sdk/test/Services/S3/IntegrationTests/PutObjectTests.cs
```diff
@@ -333,6 +333,57 @@ public void PutObject()
...
+            {
+                BucketName = bucketName,
+                Key = "inputStreamPut" + random.Next(),
+                DisablePayloadSigning = true,
+                DisableDefaultChecksumValidation = true,
+                InputStream = new MemoryStream(),
+            };
+            PutObjectResponse response = Client.PutObject(request);
+
+            Console.WriteLine("S3 generated ETag: {0}", response.ETag);
```
dscpinheiro: Can we remove the `Console.WriteLine` from the tests?

dscpinheiro (approved):
We need a V3 PR as well, right? Customer was using V4 but the same problem happens on latest V3.

sdk/test/Services/S3/IntegrationTests/PutObjectTests.cs
dscpinheiro: Was also thinking: We could probably consolidate these tests into one, and have the `DisablePayloadSigning` and `DisableDefaultChecksumValidation` options via `DataRows`.

What do you think?

muhammad-othman: Done.

muhammad-othman: Created https://github.com/aws/aws-sdk-net/pull/4016 for V3

PR #4019, opened 2025-09-23
Update devconfig generator for extension support.

normj:
Now that the `DevConfig` files support add entries for the extensions this PR updates the dev config generator to support adding extensions to the DevConfig.

image

### DIFF
--- a/buildtools/DevConfigGenerator/DevConfigGenerator/DevConfigWriter.cs
+++ b/buildtools/DevConfigGenerator/DevConfigGenerator/DevConfigWriter.cs
@@ -60,15 +60,15 @@ public class DevConfigWriter
         {
             writer.WriteStartArray("extensions");
 
-            foreach (var service in config.Extensions)
+            foreach (var extension in config.Extensions)
             {
                 writer.WriteStartObject();
-                writer.WriteString("extensionName", service.Name);
-                ValidateType(service.Type);
-                writer.WriteString("type", service.Type);
+                writer.WriteString("extensionName", extension.Name);
+                ValidateType(extension.Type);
+                writer.WriteString("type", extension.Type);
 
                 writer.WriteStartArray("changeLogMessages");
-                foreach (var message in service.Messages)
+                foreach (var message in extension.Messages)
                 {
                     writer.WriteStringValue(message);
                 }
--- a/buildtools/DevConfigGenerator/DevConfigGenerator/Program.cs
+++ b/buildtools/DevConfigGenerator/DevConfigGenerator/Program.cs
@@ -188,6 +188,10 @@ void ExtensionChange()
     var extensionsNames = new List<string>();
     foreach (var extensionDirectory in Directory.GetDirectories(Path.Combine(FindRepoRoot(), "extensions", "src")))
     {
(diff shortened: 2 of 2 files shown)


buildtools/DevConfigGenerator/DevConfigGenerator/DevConfigWriter.cs
```diff
@@ -56,6 +56,30 @@ public static void Write(DevConfigFile config, string filePath)
             writer.WriteEndArray();
         }
 
+        if (config.Extensions.Any())
+        {
+            writer.WriteStartArray("extensions");
+
+            foreach (var service in config.Extensions)
```
boblodgett: `var extension` would be more accurate

buildtools/DevConfigGenerator/DevConfigGenerator/Program.cs
```diff
@@ -175,6 +183,35 @@ void ServiceChange()
     service.Type = AskChangeType(service.Type);
 }
 
+void ExtensionChange()
+{
+    var extensionsNames = new List<string>();
+    foreach (var extensionDirectory in Directory.GetDirectories(Path.Combine(FindRepoRoot(), "extensions", "src")))
```
boblodgett, line 189: You have to filter out `AWSSDK.Extensions.CognitoAuthentication`. A better way to do this an the way the build system does it is it only counts the extension if there is a .nuspec file inside with the name of the extension.

Example: https://github.com/aws/aws-sdk-net/commit/d20f35033812878f6bf157450295244d3d94347e#diff-6dd3371737d8c5c7059d214cd8baf5cdb50de8e84036e297159b2221c5cefb53R244

buildtools/DevConfigGenerator/DevConfigGenerator/Program.cs
```diff
@@ -175,6 +183,35 @@ void ServiceChange()
...
+    {
+        extensionsNames.Add(new DirectoryInfo(extensionDirectory).Name.Replace("AWSSDK.", string.Empty));
+    }
+
+    var extensionName = AnsiConsole.Prompt(
```
boblodgett, line 198: It looks like you are picking the extension from a list. If so that is fine. I did add a schema for DevConfig file since the extensions must start with `Extension.` but this code will not be able to access the schema.

buildtools/DevConfigGenerator/DevConfigGenerator/DevConfigWriter.cs
normj: Fixed

buildtools/DevConfigGenerator/DevConfigGenerator/Program.cs
normj: Fixed. Although it has been 7 years since we moved `AWSSDK.Extensions.CognitoAuthentication` out of the SDK repo. I imagine we could get rid of that placeholder folder now.

buildtools/DevConfigGenerator/DevConfigGenerator/Program.cs
normj: Yes the strings in `extensionsNames` are presented to the user a selectable list for what extension to update.

boblodgett (approved):

PR #4031, opened 2025-10-09
[V4] Update credentials resolver to handle empty profile names in config

### LINKED ISSUE #4028: Service constructors with significantly neutered production behavior should have been removed and replaced
### Describe the bug

In the transition from SDK 3.x, a (possibly unintentional) malicious change was made.  

Previously, the service object constructors that took a profile name fell back to using the IAM role when a null profile was passed.  In 4.x, the constructor still exists, still accepts null, and doesn't throw an exception, but subsequently all calls fail due to the lack of credentials.  This is the worst possible way this breaking change could have been handled.  With normal usage patterns, developers use local credentials (classically API keys) on their local machines that used the named profile, and passed in null for the profile in production.  The way this change was made caused everything to work for devs on their local machines and for unit tests, but fail in production, not even at construction time, but later when the code attempted to use the service with various somewhat confusing access denied errors.  Even after finding out there were problems, the only way to figure out what was going wrong was extensive logging or running a debugger *inside* AWS.  

### Regression Issue

### Expected Behavior

When I got my code upgraded from using SDK3.x to SDK4.x compiling and the unit tests running on my local machine, I expected that code to work when deployed to AWS.  I expected that *whenever possible*, breaking changes break in the compilation step.

### Current Behavior

When upgrading from SDK 3.x, code compiles without errors and works outside of AWS when debugging and running unit tests, but then when you deploy to AWS, EVERYTHING fails with some kind of access error.  Debugging and fixing the code is extremely difficult because it MUST be done on a machine that relies on IAM--that's the only case that broke.

### Reproduction Steps

Write code using SDK3.x that uses service constructors that take a profile name, passing a profile name in dev mode and null in production mode (like just about everyone was doing).  Next, upgrade to SDK4.x and attempt to get your code to work in production.  Good luck!

### Possible Solution

When the decision was made to change the behavior of a profile specified as null, ideally those constructors would have been removed completely or marked with proper non-nullable attributes so that compiler errors would have indicated to devs that they needed to change something.  Second best would have been to have the constructor throw an exception immediately when null is passed for the profile.  I understand if you don't want to make a breaking change for this by removing the constructor, but YOU ALREADY BROKE EVERYONE with this change--they just won't know it until they try to deploy into AWS.  Breaking at compile time is MUCH PREFERRED to breaking at production runtime.  At a minimum you should throw an exception in the constructor in this case--at least the exception will give developers a clue as to the horrific break-only-in-production change you've made.

_No response_

ANY SDK 3.x and ANY SDK 4.x

### Targeted .NET Platform

.NET 9.0

Windows 11, Amazon Linux 2023

dscpinheiro:
Fixes #4028 

## Motivation and Context
Fixes a regression in V4 where the credentials resolver was not handling the scenario where: The `Profile` property in the `ClientConfig` object was set, but without a value in non-local environments (causing an `AmazonClientException` to be thrown when running on an EC2 instance).

## Testing
Dry-run: `DRY_RUN-8d688ddb-21af-49aa-b2d4-5c3888e18cc4` (in progress)

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/test/UnitTests/Custom/Runtime/Credentials/DefaultCredentialsResolverTests.cs
+++ b/sdk/test/UnitTests/Custom/Runtime/Credentials/DefaultCredentialsResolverTests.cs
@@ -13,9 +13,11 @@
  * permissions and limitations under the License.
  */
 
+using Amazon;
 using Amazon.Runtime;
 using Amazon.Runtime.CredentialManagement;
 using Amazon.Runtime.Credentials;
+using Amazon.S3;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using System;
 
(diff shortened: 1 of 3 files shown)


sdk/test/UnitTests/Custom/Runtime/Credentials/DefaultCredentialsResolverTests.cs
```diff
@@ -88,5 +90,31 @@ public void CredentialsAreReevaluatedWhenProfileChanges()
...
+        /// <para />
+        /// If the name property in the "Profile" property is not set (which can happen when its value
+        /// is only set in a non-production environment), the resolver should not throw an exception and 
+        /// move to the next provider in the chain (which in this test will be the access / secret key pair
+        /// set in the environment variables).
+        /// </summary>
+        [DataTestMethod]
+        [DataRow("")]
+        [DataRow(null)]
+        public void CredentialsResolver_HandlesEmptyProfileNameInConfig(string profileName)
```
muhammad-othman, line 105: nit: for completeness we can add another case what the profile name is provided, which means we will need to add another parameter for the identity type since it will be different for this case.

dscpinheiro: Discussed offline, but will skip this for now since I'd actually need to write the "existing" profile like we do in https://github.com/aws/aws-sdk-net/blob/185a9bcf1be18e665cf80791e10626a56b30eafe/sdk/test/UnitTests/Custom/Runtime/Credentials/FallbackFactoryTest.cs (and right now I don't want to introduce the `FallbackFactoryTestFixture` class here...)

PR #4040, opened 2025-10-13
Add ServerSideEncryptionCustomerProvidedKeyMD5 and ContentLanguage to…

GarrettBeatty:
Add ServerSideEncryptionCustomerProvidedKeyMD5  to GetObjectResponse. I noticed  these were missing this property https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html#API_GetObject_ResponseElements while working on DOTNET-8275.

I also noticed we were missing ContentLanguage on  GetObjectResponse 

## Description
1. Add missing fields

## Motivation and Context
To be in compliance with s3 api docs & model

## Testing
1. Integration Tests
2. Dry Run 
```
 "build_id": "c308f474-14ef-4b09-8bb7-ecee1efe46b7",
    "build_type": "DRY_RUN",
    "build_status": "SUCCEEDED",
```

## Types of changes

## Checklist

## License

### DIFF
--- a/generator/.DevConfigs/a1b2c3d4-e5f6-7890-abcd-ef1234567890.json
+++ b/generator/.DevConfigs/a1b2c3d4-e5f6-7890-abcd-ef1234567890.json
@@ -1,0 +1,12 @@
+{
+  "services": [
+    {
+      "serviceName": "S3",
+      "type": "patch",
+      "changeLogMessages": [
+        "Added ServerSideEncryptionCustomerProvidedKeyMD5 property to GetObjectResponse",
+        "Added ContentLanguage support to GetObjectResponse"
+      ]
+    }
+  ]
+}


sdk/test/Services/S3/IntegrationTests/GetObjectTests.cs
```diff
@@ -110,5 +110,187 @@ public void TestObjectAttributesLastModified()
...
+                    using (response)
+                    {
+                        // Verify the ServerSideEncryptionCustomerProvidedKeyMD5 property is populated
+                        Assert.IsNotNull(response.ServerSideEncryptionCustomerProvidedKeyMD5, 
+                            "ServerSideEncryptionCustomerProvidedKeyMD5 should not be null");
+                        Assert.AreEqual(keyMD5, response.ServerSideEncryptionCustomerProvidedKeyMD5,
+                            "ServerSideEncryptionCustomerProvidedKeyMD5 should match the provided MD5");
+                    }
+                }
+                finally
```
dscpinheiro: The cleanup method in this class already takes care of deleting all objects in the test bucket: https://github.com/aws/aws-sdk-net/blob/a4a102c05263c08ae54304a9e0b52d17c6f76175/sdk/test/Services/S3/IntegrationTests/GetObjectTests.cs#L43-L48

So you shouldn't need the `try / finally` in any of the tests you added.

GarrettBeatty: fixed in latest revision

sdk/src/Services/S3/Custom/Model/Internal/MarshallTransformations/GetObjectResponseUnmarshaller.cs
```diff
@@ -85,6 +85,8 @@ private static void UnmarshallResult(XmlUnmarshallerContext context,GetObjectRes
                 response.Headers.ContentDisposition = S3Transforms.ToString(responseData.GetHeaderValue("Content-Disposition"));
             if (responseData.IsHeaderPresent("Content-Encoding"))
                 response.Headers.ContentEncoding = S3Transforms.ToString(responseData.GetHeaderValue("Content-Encoding"));
+            if (responseData.IsHeaderPresent("Content-Language"))
+                response.Headers.ContentLanguage = S3Transforms.ToString(responseData.GetHeaderValue("Content-Language"));
```
peterrsongg: Instead of adding `ContentLanguage` to the `HeadersCollection` and then adding it to `Headers` can we just add `ContentLanguage` as a property on `GetObjectResponse` and unmarshall it to that property? Later when we generate this operation, that is what would be generated. 

Unless there's a reason you need it in the headers collection?

peterrsongg: Another benefit of that is that we will generate the documentation too for Content-Language instead of the custom comment that you wrote in the `HeadersCollection.cs` file.

GarrettBeatty: i only added there to be consistent with the current implementation (Even though in the model it gets unmarshalled to the response as a property). i can update it to be a property though because thats the "correct" way

GarrettBeatty: ive updated it to just unmarshall it to the object

sdk/src/Services/S3/Custom/Model/Internal/MarshallTransformations/GetObjectResponseUnmarshaller.cs
```diff
@@ -85,6 +85,8 @@ private static void UnmarshallResult(XmlUnmarshallerContext context,GetObjectRes
                 response.Headers.ContentDisposition = S3Transforms.ToString(responseData.GetHeaderValue("Content-Disposition"));
             if (responseData.IsHeaderPresent("Content-Encoding"))
                 response.Headers.ContentEncoding = S3Transforms.ToString(responseData.GetHeaderValue("Content-Encoding"));
+            if (responseData.IsHeaderPresent("Content-Language"))
+                response.ContentLanguage = S3Transforms.ToString(responseData.GetHeaderValue("Content-Language"));
```
GarrettBeatty, line 89: this is intentionally not setting in the Headers field because we want to match the model

PR #4045, opened 2025-10-14
Added PutObjectResponse to TransferUtilityUploadResponse mapping

GarrettBeatty:
## Description

This change creates a `TransferUtilityUploadResponse` class. This class contains the common fields between `PutObjectResponse` and `CompleteMultipartUploadResponse`. The reason it contains the common fields between `PutObjectResponse` and `CompleteMultipartUploadResponse` is because this same object should be utilized in both regular upload and multi part upload (as per the SEP). 

This class will be used in https://github.com/aws/aws-sdk-net/pull/4059 and https://github.com/aws/aws-sdk-net/pull/4061 Where for progress events (e.g. UploadCompleted), we return a `TransferUtilityUploadResponse` object as part of it. This class will also be used when we create new apis as part of SEP compliance (`UploadWithResponseAsync`). it will be the return value of this function (instead of `void` like the existing `UploadAsync`).

```
TranserUtilityUploadRequest req = ... 
TransferUtilityUploadResponse response = transferUtility.UploadWithResponse(req)

```

In order to create this `TransferUtilityUploadResponse` class, I have created a conversion function `MapPutObjectResponse` which maps the `PutObjectResponse` to the `TransferUtilityUploadResponse`. Eventually, this class will also contain a `MapCompletemultipartResponse` function as well  https://github.com/aws/aws-sdk-net/pull/4060

In order to comply with the SEP and ensure that all fields are mapped, I have added unit tests which read the `mapping.json` file from the SEP and validate that each of them get mapped in my conversion function between the two classes. 

## Motivation and Context
1. DOTNET-8275

## Testing
1. Unit tests
3. 4728749f-fc1d-45b4-a2a3-3605bf5dcd60 dry run pass

## Types of changes

## Checklist

## License


generator/ServiceClientGeneratorLib/Generators/SourceFiles/AssemblyInfo.tt
```diff
@@ -26,6 +26,10 @@ using System.Runtime.CompilerServices;
 [assembly: InternalsVisibleTo("AWSSDK.UnitTests.DynamoDBv2.NetFramework, PublicKey=0024000004800000940000000602000000240000525341310004000001000100db5f59f098d27276c7833875a6263a3cc74ab17ba9a9df0b52aedbe7252745db7274d5271fd79c1f08f668ecfa8eaab5626fa76adc811d3c8fc55859b0d09d3bc0a84eecd0ba891f2b8a2fc55141cdcc37c2053d53491e650a479967c3622762977900eddbf1252ed08a2413f00a28f3a0752a81203f03ccb7f684db373518b4")]
 [assembly: InternalsVisibleTo("AWSSDK.UnitTests.NetFramework, PublicKey=0024000004800000940000000602000000240000525341310004000001000100db5f59f098d27276c7833875a6263a3cc74ab17ba9a9df0b52aedbe7252745db7274d5271fd79c1f08f668ecfa8eaab5626fa76adc811d3c8fc55859b0d09d3bc0a84eecd0ba891f2b8a2fc55141cdcc37c2053d53491e650a479967c3622762977900eddbf1252ed08a2413f00a28f3a0752a81203f03ccb7f684db373518b4")]
 <# } #>
+<# if (this.Config.AssemblyTitle=="AWSSDK.S3") { #>
```
GarrettBeatty, line 29: this is added so we can test the internal ResponseMapper class i made

sdk/src/Services/S3/Custom/Transfer/TransferUtilityUploadResponse.cs
```diff
@@ -0,0 +1,470 @@
...
+using Amazon.Runtime.Internal;
+
+namespace Amazon.S3.Transfer
+{
+    /// <summary>
+    /// Response object for Transfer Utility upload operations.
+    /// Contains unified response fields from both simple uploads (PutObjectResponse) 
+    /// and multipart uploads (CompleteMultipartUploadResponse).
+    /// </summary>
+    public class TransferUtilityUploadResponse : AmazonWebServiceResponse
```
GarrettBeatty, line 35: right now if we release this change to development, the customer can access this class even though it wont really be used for anything until the next PRs. which i think should go in the feature branch. wondering if we should target feature/transfermanager for this change even though its not a runtime change?

sdk/test/Services/S3/UnitTests/Custom/EmbeddedResource/property-aliases.json
```diff
@@ -0,0 +1,10 @@
+{
+  "PropertyAliases": {
```
GarrettBeatty, line 2: .net's property values are slightly different than whats in the mapping so i made another mapping to the .net values. the reason i didnt modify the original mapping.json is so we can copy and paste it from the SEP code directly if we ever need to update it

sdk/test/Services/S3/UnitTests/Custom/ResponseMapperTests.cs
```diff
@@ -0,0 +1,491 @@
...
+        }
+
+
+        [TestMethod]
+        [TestCategory("S3")]
+        public void ValidateGetObjectResponseDefinitionCompleteness()
+        {
+            ValidateResponseDefinitionCompleteness<GetObjectResponse>(
+                new[] { "Definition", "DownloadResponse", "GetObjectResponse" },
+                "GetObjectResponse",
```
GarrettBeatty: wondering if we should be updating GetObjectResponse soon/now instead of doing this to match the model? e.g. ContentType is in the HeaderCollection property currently. but the model https://github.com/aws/aws-sdk-net/blob/main/generator/ServiceModels/s3/s3-2006-03-01.api.json#L4575 i think maps it to its own attribute @philasmar @peterrsongg ?

GarrettBeatty: also this test case isnt really needed as part of this work. i added it just for completeness sake. its validating that getobjectresponse has all of the expected fields in the mapping.json

peterrsongg: @GarrettBeatty yeah, if it is mapped to a property on the output then we should follow the model, especially if it isn't there right now.

peterrsongg:
Approved, based on assumption that the actual mapping will happen in another PR (unless my read of this is mistaken)

GarrettBeatty:
> Change looks good, but I have a question.
> 
> You're not actually calling `MapPutObjectResponse` anywhere in this PR. So where does the mapping actually happen? Is that going to be part of another PR?

yes - description of the PR has links to the ones that will call it

sdk/test/Services/S3/UnitTests/Custom/ResponseMapperTests.cs
GarrettBeatty: i think i might just remove this test case because we dont need to copy content type to transferutilityuploadresponse anyway, so this test case really isnt validating anything important

GarrettBeatty: ended up removing this since if we update the getobjectresponse model anyway it will be backward compatible so the fields we have now are fine and we dont need a test for it

Issue #4010, opened 2025-09-15
Support nonseekable streams for UploadPartRequest.InputStream


StephenCleary:
### Describe the feature

In the scenario where `DisablePayloadSigning` is `true`, the `UploadPartRequest.InputStream` should support nonseekable streams.

### Use Case

Currently, the `InputStream` is always wrapped in a `PartialWrapperStream` just in case the user passes in a stream that's too big. This validation requires the underlying stream to be seekable.

Please provide an option that avoids the `PartialWrapperStream` for users that provide non-seekable streams guaranteed to be equal to the part size (or smaller for the last part).

### Proposed Solution

Add a flag/option to `UploadPartRequest` to skip [creating the `PartialWrapperStream`](https://github.com/aws/aws-sdk-net/blob/ff7ac2940e1d85b94caffa263e47a129541ff62c/sdk/src/Services/S3/Custom/Internal/AmazonS3PostMarshallHandler.cs#L128).

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS .NET SDK and/or Package version used

Latest

### Targeted .NET Platform

.NET 8

### Operating System and version

Windows 11

normj:
Hi @StephenCleary. S3 always requires content length as well as the SDK does compute checksums as headers so the put object and upload part APIs do not support non seekable streams.

In the SDK we have the `Amazon.S3.Transfer.TransferUtility` class and it's `UploadAsync` methods will simulate the upload of a non-seekable stream by buffering up to a part size, default 5MB, in memory then uploading that buffer using the upload part API.

StephenCleary:
> the SDK does compute checksums as headers

Not if `DisablePayloadSigning` is `true` and the S3 additional checksums are already calculated by my code.

normj:
@StephenCleary I found some time to do some more investigation. There are some discrepancies with what is possible with `PutObject` and `UploadPart` when working  nonseekable streams. I'm attempting to normalize the code which I think will unblock you.

dscpinheiro:
Today's release (https://github.com/aws/aws-sdk-net/releases/tag/4.0.259.0) includes an update where an unseekable stream can be passed to `UploadPart` without any extra config options (and as you pointed out, as long as the payload signing is disabled). 

An example from our tests: https://github.com/aws/aws-sdk-net/blob/main/sdk/test/Services/S3/IntegrationTests/UploadPartUnseekableStreamTests.cs

PR #4048, opened 2025-10-14
Fixed issue calling UploadPart with an unseekable stream and disablin…

normj:
PR in draft mode till full dry run is successul

## Description
The PutObject operation does work with an unseekable stream if you set both `DisablePayloadSigning` is set to `true` but the UploadPart which also fail. That is because it is always wrapping the input stream in a `PartialWrapperStream` which requires the stream to be seekable. The PR switches the code to use the `GetStreamWithLength` method like PutObject which returns back a `PartialReadOnlyWrapperStream` instance which does not require the wrapped stream to be seekable.

The content-length will be determine by either the stream's length if available and then fallback to `PartSize property of the `UploadPartRequest`.

## Motivation and Context
https://github.com/aws/aws-sdk-net/issues/4010

## Testing
Added new integ tests
Dry run: pending

### DIFF
--- a/generator/.DevConfigs/32a12d7c-afc6-4bcf-a2d8-9c49b335b935.json
+++ b/generator/.DevConfigs/32a12d7c-afc6-4bcf-a2d8-9c49b335b935.json
@@ -1,0 +1,11 @@
+{
+  "services": [
+    {
+      "serviceName": "S3",
+      "type": "patch",
+      "changeLogMessages": [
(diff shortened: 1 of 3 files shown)


normj:
The PR is turning into a bit of a whack a mole experience. We'll keep plugging away but there are currently some tests failing making this change.

dscpinheiro:
I addressed this in a different PR that will be included in today's release.

PR #4059, opened 2025-10-18
Add Progress listeners for initiated, complete, and failed for simple upload

GarrettBeatty:
Stacked PRs:
 * #4064
 * #4063
 * #4062
 * #4061
 * #4060
 * __->__#4059

--- --- ---

## Description

This change adds progress listeners for "initiated", "complete", and "failed" lifecycle events for the SimpleUploadCommand (non multipart). Consumers can subscribe to these callbacks to receive notifications when a simple upload starts, finishes successfully, or fails. We will also make this change for multi part upload command here https://github.com/aws/aws-sdk-net/pull/4061

## Example usage
```
// Create a TransferUtilityUploadRequest
TransferUtilityUploadRequest request = new TransferUtilityUploadRequest();

// Subscribe to the progress events
request.UploadInitiatedEvent += Request_UploadInitiatedEvent;
request.UploadCompletedEvent += Request_UploadCompletedEvent;
request.UploadFailedEvent += Request_UploadFailedEvent;

// Use the Transfer Utility to upload the file
using (var transferUtility = new TransferUtility(s3Client))
{
    transferUtility.UploadAsync(request);
}

private void Request_UploadInitiatedEvent(object sender, UploadInitiatedEventArgs e)
{
    Console.WriteLine($"Upload initiated for file: {e.FilePath}");
    Console.WriteLine($"Total bytes to transfer: {e.TotalBytes}");
}

private void Request_UploadCompletedEvent(object sender, UploadCompletedEventArgs e)
{
    Console.WriteLine($"Upload completed for file: {e.FilePath}");
    Console.WriteLine($"Transferred bytes: {e.TransferredBytes}");
    Console.WriteLine($"ETag: {e.Response.ETag}");
    Console.WriteLine($"Version ID: {e.Response.VersionId}");
}

private void Request_UploadFailedEvent(object sender, UploadFailedEventArgs e)
{
    Console.WriteLine($"Upload failed for file: {e.FilePath}");
    Console.WriteLine($"Transferred bytes: {e.TransferredBytes} / {e.TotalBytes}");
}


```

## Motivation and Context
1. To adhere to the SEP

## Testing
1. 46a4bf17-39fd-41e9-b625-1fba1b49f5a4 - pass
2. Integration tests

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/Transfer/Internal/SimpleUploadCommand.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/Internal/SimpleUploadCommand.cs
@@ -41,12 +41,18 @@ namespace Amazon.S3.Transfer.Internal
         IAmazonS3 _s3Client;
         TransferUtilityConfig _config;
         TransferUtilityUploadRequest _fileTransporterRequest;
+        long _totalTransferredBytes;
+        private readonly long _contentLength;
 
         internal SimpleUploadCommand(IAmazonS3 s3Client, TransferUtilityConfig config, TransferUtilityUploadRequest fileTransporterRequest)
         {
             this._s3Client = s3Client;
             this._config = config;
             this._fileTransporterRequest = fileTransporterRequest;
+            
+            // Cache content length immediately while stream is accessible to avoid ObjectDisposedException in failure scenarios
+            this._contentLength = this._fileTransporterRequest.ContentLength;
+            
             var fileName = fileTransporterRequest.FilePath;
         }
 
@@ -108,9 +114,48 @@ namespace Amazon.S3.Transfer.Internal
 
         private void PutObjectProgressEventCallback(object sender, UploadProgressArgs e)
         {
-            var progressArgs = new UploadProgressArgs(e.IncrementTransferred, e.TransferredBytes, e.TotalBytes, 
-                e.CompensationForRetry, _fileTransporterRequest.FilePath);
+            // Keep track of the total transferred bytes so that we can also return this value in case of failure
+            long transferredBytes = Interlocked.Add(ref _totalTransferredBytes, e.IncrementTransferred - e.CompensationForRetry);
+            
+            var progressArgs = new UploadProgressArgs(e.IncrementTransferred, transferredBytes, _contentLength, 
+                e.CompensationForRetry, _fileTransporterRequest.FilePath, _fileTransporterRequest);
             this._fileTransporterRequest.OnRaiseProgressEvent(progressArgs);
+        }
+
+        private void FireTransferInitiatedEvent()
+        {
+            var initiatedArgs = new UploadInitiatedEventArgs(
+                request: _fileTransporterRequest,
+                filePath: _fileTransporterRequest.FilePath,
+                totalBytes: _contentLength
+            );
+            
+            _fileTransporterRequest.OnRaiseTransferInitiatedEvent(initiatedArgs);
+        }
+
+        private void FireTransferCompletedEvent(TransferUtilityUploadResponse response)
+        {
+            var completedArgs = new UploadCompletedEventArgs(
+                request: _fileTransporterRequest,
+                response: response,
+                filePath: _fileTransporterRequest.FilePath,
+                transferredBytes: Interlocked.Read(ref _totalTransferredBytes),
+                totalBytes: _contentLength
+            );
+            
+            _fileTransporterRequest.OnRaiseTransferCompletedEvent(completedArgs);
+        }
+
+        private void FireTransferFailedEvent()
+        {
+            var failedArgs = new UploadFailedEventArgs(
+                request: _fileTransporterRequest,
+                filePath: _fileTransporterRequest.FilePath,
+                transferredBytes: Interlocked.Read(ref _totalTransferredBytes),
+                totalBytes: _contentLength
+            );
+            
+            _fileTransporterRequest.OnRaiseTransferFailedEvent(failedArgs);
         }
     }
 }
--- a/sdk/src/Services/S3/Custom/Transfer/TransferUtilityUploadRequest.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/TransferUtilityUploadRequest.cs
@@ -25,6 +25,7 @@ using System.Collections.Specialized;
 using System.IO;
 using System.Text;
 
+using Amazon.Runtime;
 using Amazon.Runtime.Internal;
 using Amazon.S3.Model;
 using Amazon.Util;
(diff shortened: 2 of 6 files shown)


sdk/src/Services/S3/Custom/Transfer/TransferUtilityUploadRequest.cs
```diff
@@ -835,11 +962,164 @@ internal UploadProgressArgs(long incrementTransferred, long transferred, long to
...
+        /// <summary>
+        /// The original TransferUtilityUploadRequest created by the user.
+        /// Contains all the upload parameters and configuration.
+        /// </summary>
+        public TransferUtilityUploadRequest Request { get; private set; }
+
+        /// <summary>
+        /// Gets the file path being uploaded.
+        /// </summary>
+        public string FilePath { get; private set; }
```
GarrettBeatty, line 689: i put this here for consistency with the UploadProgress. i think technically its part of the Request (Request.FilePath). im open to removing this though

sdk/src/Services/S3/Custom/Transfer/Internal/SimpleUploadCommand.cs
```diff
@@ -103,9 +109,48 @@ private PutObjectRequest ConstructRequest()
...
+            var progressArgs = new UploadProgressArgs(e.IncrementTransferred, transferredBytes, _contentLength, 
+                e.CompensationForRetry, _fileTransporterRequest.FilePath, _fileTransporterRequest);
             this._fileTransporterRequest.OnRaiseProgressEvent(progressArgs);
         }
+
+        private void FireTransferInitiatedEvent()
+        {
+            var initiatedArgs = new UploadInitiatedEventArgs(
+                request: _fileTransporterRequest,
+                filePath: _fileTransporterRequest.FilePath,
```
GarrettBeatty, line 129: im including filepath here to be consistent with progress args but technically this can come from request.filepath. so not sure if we want it?

sdk/src/Services/S3/Custom/Transfer/Internal/SimpleUploadCommand.cs
```diff
@@ -103,9 +109,48 @@ private PutObjectRequest ConstructRequest()
 
         private void PutObjectProgressEventCallback(object sender, UploadProgressArgs e)
         {
-            var progressArgs = new UploadProgressArgs(e.IncrementTransferred, e.TransferredBytes, e.TotalBytes, 
-                e.CompensationForRetry, _fileTransporterRequest.FilePath);
+            // Keep track of the total transferred bytes so that we can also return this value in case of failure
+            long transferredBytes = Interlocked.Add(ref _totalTransferredBytes, e.IncrementTransferred - e.CompensationForRetry);
+            
+            var progressArgs = new UploadProgressArgs(e.IncrementTransferred, transferredBytes, _contentLength,
```
peterrsongg, line 120: why use _contentLength over e.totalBytes here?

sdk/src/Services/S3/Custom/Transfer/TransferUtilityUploadRequest.cs
```diff
@@ -835,11 +962,164 @@ internal UploadProgressArgs(long incrementTransferred, long transferred, long to
...
+    /// </summary>
+    public class UploadCompletedEventArgs : EventArgs
+    {
+        /// <summary>
+        /// Initializes a new instance of the UploadCompletedEventArgs class.
+        /// </summary>
+        /// <param name="request">The original TransferUtilityUploadRequest created by the user</param>
+        /// <param name="response">The unified response from Transfer Utility</param>
+        /// <param name="filePath">The file that was uploaded</param>
+        /// <param name="transferredBytes">The total number of bytes transferred</param>
```
peterrsongg, line 709: just for my understanding. `TransferredBytes` and `totalBytes` should always be the same when the  `UploadCompleted` event fires right? I was just looking at the parameter descriptions and they were the same, so i was a bit confused😅

peterrsongg: ah okay, i just read the properties below.. maybe include that also in this param name xml comment?

```
        /// <param name="transferredBytes">The total number of bytes transferred</param>
        /// <param name="totalBytes">The total number of bytes that were transferred (should equal transferredBytes for successful uploads)./param>
```

sdk/test/Services/S3/IntegrationTests/TransferUtilityTests.cs
```diff
@@ -1393,6 +1543,87 @@ public void OnProgressEvent(object sender, T progress)
...
+                    throw;
+                }
+            }
+
+            public void AssertEventFired()
+            {
+                if (EventException != null)
+                    throw EventException;
+
+                // Add some time for the background thread to finish before checking
```
peterrsongg: can we make this comment a bit more explicit? took me a bit to understand but I get it now
```
Since  AWSSDKUtils.InvokeInBackground fires the event in the background it is possible that we check too early that the event has fired. In this case, we sleep and check again.
```

sdk/src/Services/S3/Custom/Transfer/Internal/SimpleUploadCommand.cs
GarrettBeatty: https://github.com/aws/aws-sdk-net/blob/b55e7807a0010c4345710a46618df5cddcbc33aa/sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_netstandard/HttpRequestMessageFactory.cs#L781

https://github.com/aws/aws-sdk-net/blob/b55e7807a0010c4345710a46618df5cddcbc33aa/sdk/src/Core/Amazon.Runtime/Internal/StreamReadTracker.cs#L57

looks like e.TotalBytes is the length of the stream but at a lower level (closer to the http level) where we wrap it in another stream before sending off to AWS. 

this is what Cline says potential differences are


## Potential Differences

__In most cases, they should be identical__, but differences could occur if:

1. __Stream Position Changes__: If the stream position changed between when `_contentLength` was cached and when the HTTP request was made
2. __Stream Length Changes__: If the underlying stream length changed (rare, but theoretically possible)
3. __Different Stream References__: If somehow different stream objects were involved (very unlikely in normal scenarios)
4. __Wrapper Stream Effects__: If there are wrapper streams that calculate length differently

GarrettBeatty: my understanding is it doesnt actually matter for SimpleUploadCommand. The values should theoretically be the same. 
 
 https://github.com/aws/aws-sdk-net/blob/b55e7807a0010c4345710a46618df5cddcbc33aa/sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_netstandard/HttpRequestMessageFactory.cs#L781 is where e.totalbytes is set

GarrettBeatty: multi part upload uses content length here https://github.com/aws/aws-sdk-net/blob/b55e7807a0010c4345710a46618df5cddcbc33aa/sdk/src/Services/S3/Custom/Transfer/Internal/MultipartUploadCommand.cs#L374.


(it uses e.totalbytes https://github.com/aws/aws-sdk-net/blob/b55e7807a0010c4345710a46618df5cddcbc33aa/sdk/src/Services/S3/Custom/Transfer/Internal/MultipartUploadCommand.cs#L470 here though because thats the individual part tracking not the overall file). 

so in our case for simple upload since its one file and one put request, the values should be equivalent

sdk/test/Services/S3/IntegrationTests/TransferUtilityTests.cs
GarrettBeatty: fixed

sdk/src/Services/S3/Custom/Transfer/TransferUtilityUploadRequest.cs
GarrettBeatty: fixed

peterrsongg (approved):

sdk/src/Services/S3/Custom/Transfer/Internal/SimpleUploadCommand.cs
normj: Which events are you trying to be consistent with? Since the event already contains the request it does seem redundant to have file path at the top level.

sdk/src/Services/S3/Custom/Transfer/Internal/SimpleUploadCommand.cs
normj: My memory is the content length on the `TransferUtilityUploadRequest` is optional and if doesn't exist we eventually use the length of the stream as the content length. So I don't think you can rely on the content length on `TransferUtilityUploadRequest`.

sdk/src/Services/S3/Custom/Transfer/Internal/SimpleUploadCommand.cs
GarrettBeatty: trying to be consistent with the InProgress event which has FilePath exposed https://github.com/aws/aws-sdk-net/blob/f8f03e34d23e4f360bfb9edd5441455f4674216f/sdk/src/Services/S3/Custom/Transfer/TransferUtilityUploadRequest.cs#L834

in this change we are adding the original TransferUtilityUploadRequest to the new progress events (initiated, completed, failed) and we are also adding TransferUtilityUploadRequest  to the existing UploadProgressArgs

sdk/src/Services/S3/Custom/Transfer/Internal/SimpleUploadCommand.cs
GarrettBeatty: https://github.com/aws/aws-sdk-net/blob/f8f03e34d23e4f360bfb9edd5441455f4674216f/sdk/src/Services/S3/Custom/Transfer/TransferUtilityUploadRequest.cs#L436 the getter calculates it here. not sure why this wouldnt always work though?

sdk/test/Services/S3/UnitTests/Custom/ResponseMapperTests.cs
```diff
@@ -165,7 +165,11 @@ public void MapUploadRequest_PutObjectRequest_AllMappedProperties_WorkCorrectly(
                     var simpleUploadCommand = new SimpleUploadCommand(null, null, sourceRequest);
                     return simpleUploadCommand.ConstructRequest();
                 },
-                usesHeadersCollection: false);
+                usesHeadersCollection: false,
+                (sourceRequest) =>
+                {
+                    sourceRequest.InputStream = new MemoryStream(1024);
```
GarrettBeatty, line 171: was getting null pointer on content length without this change. in reality, input stream or file path will always be set so users shouldnt get null pointer

sdk/src/Services/S3/Custom/Transfer/Internal/SimpleUploadCommand.cs
GarrettBeatty: removed filepath from here since it can be derived from request

sdk/src/Services/S3/Custom/Transfer/Internal/SimpleUploadCommand.cs
GarrettBeatty: my reasoning for switching e.totalbytes to content length here was to be consisent because in the new UploadInitiated, UploadFailed and UploadCompleted i use content length. so i want to use this value in all places. I think this should be fine since multi part upload actually uses content length for the total size of the file too.

normj: Got it I see `ContentLength` from the request is not a user provided value but the computed value from the input stream. I assume that if `ContentLength` returns back `-1` then the check will happen before here and the execution will got into the unknown content length code path.

sdk/src/Services/S3/Custom/Transfer/Internal/SimpleUploadCommand.cs
normj: Makes sense. In the `UploadProgressArgs` there is no request object so file path made more sense. 

Just double checking, instead of removing file path in the new event args should we have added the request `TransferUtilityUploadRequest` to the `UploadProgressArgs` for consistency?

sdk/src/Services/S3/Custom/Transfer/Internal/SimpleUploadCommand.cs
GarrettBeatty: Right. If the stream is unseekable steam  (where content length is -1) there is branching logic before this where it calls Multipartuoloadcommand.cs (where the unseekable stream logic is located). Simpleuploadcommand.cs only hands seekable streams anyway.

sdk/src/Services/S3/Custom/Transfer/Internal/SimpleUploadCommand.cs
GarrettBeatty: >instead of removing file path in the new event args should we have added the request TransferUtilityUploadRequest to the UploadProgressArgs for consistency?

so now it looks like

UploadProgressArgs  - contains both TransferUtilityUploadRequest (new)  and Filepath (for backward compatibility)
UploadCompleted, UploadInitiated, UploadFailed - contain TransferUtilityUploadRequest  and not FilePath (edited) 

------

or were you thinking that it should be
UploadProgressArgs  - contains both TransferUtilityUploadRequest (new)  and Filepath (for backward compatibility)
UploadCompleted, UploadInitiated, UploadFailed - contain TransferUtilityUploadRequest  and Filepath  (to be consistent)
(this is what i had originally done. )

GarrettBeatty: i just updated it now to be 

UploadProgressArgs - contains both TransferUtilityUploadRequest (new) and Filepath (for backward compatibility)
UploadCompleted, UploadInitiated, UploadFailed - contain TransferUtilityUploadRequest and Filepath (to be consistent)


and we can always come back to this before merging into development

PR #4060, opened 2025-10-18
Add mapping of CompletemultipartUploadResponse to TransferUtilityUploadResponse

GarrettBeatty:
Stacked PRs:
 * #4064
 * #4063
 * #4062
 * #4061
 * __->__#4060

--- --- ---

## Description
This change adds a function to map the CompletemultipartUploadResponse to TransferUtilityUploadResponse. This is needed because we eventually need to return `TransferUtilityUploadResponse` as part of the progress listener for multi part upload here https://github.com/aws/aws-sdk-net/pull/4061

## Motivation and Context
1. DOTNET-8276

## Testing
1. Unit tests
2. 8338c01e-36d5-4617-b5af-b4a255e7c141 - pass

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/Transfer/Internal/ResponseMapper.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/Internal/ResponseMapper.cs
@@ -99,6 +99,67 @@ namespace Amazon.S3.Transfer.Internal
 
             return response;
         }
+
+        /// <summary>
+        /// Maps a CompleteMultipartUploadResponse to TransferUtilityUploadResponse.
+        /// Uses the field mappings defined in mapping.json "Conversion" -> "CompleteMultipartResponse" -> "UploadResponse".
+        /// </summary>
+        /// <param name="source">The CompleteMultipartUploadResponse to map from</param>
+        /// <returns>A new TransferUtilityUploadResponse with mapped fields</returns>
+        internal static TransferUtilityUploadResponse MapCompleteMultipartUploadResponse(CompleteMultipartUploadResponse source)
+        {
+            if (source == null)
+                return null;
+
+            var response = new TransferUtilityUploadResponse();
+
+            // Map all fields as defined in mapping.json "Conversion" -> "CompleteMultipartResponse" -> "UploadResponse"
+            if (source.IsSetBucketKeyEnabled())
+                response.BucketKeyEnabled = source.BucketKeyEnabled.GetValueOrDefault();
+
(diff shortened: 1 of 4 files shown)


sdk/src/Services/S3/Custom/Transfer/Internal/ResponseMapper.cs
```diff
@@ -99,6 +99,67 @@ internal static TransferUtilityUploadResponse MapPutObjectResponse(PutObjectResp
...
+        /// <returns>A new TransferUtilityUploadResponse with mapped fields</returns>
+        internal static TransferUtilityUploadResponse MapCompleteMultipartUploadResponse(CompleteMultipartUploadResponse source)
+        {
+            if (source == null)
+                return null;
+
+            var response = new TransferUtilityUploadResponse();
+
+            // Map all fields as defined in mapping.json "Conversion" -> "CompleteMultipartResponse" -> "UploadResponse"
+            if (source.IsSetBucketKeyEnabled())
```
peterrsongg, line 117: so anytime a new property is added to `CompleteMultipartUploadResponse` we have to remember to update this mapping?

sdk/test/Services/S3/UnitTests/Custom/EmbeddedResource/property-aliases.json
```diff
@@ -112,6 +112,10 @@
     },
     "AbortMultipartUploadRequest": {
       "Bucket": "BucketName"
+    },
+    "CompleteMultipartUploadResponse": {
```
GarrettBeatty, line 116: in mapping.json (existing file) we are looking for ServerSideEncryption and SSEKMSKeyId which have slightly different property names in .net compared to other SDKs so we need to add that mapping here

sdk/src/Services/S3/Custom/Transfer/Internal/ResponseMapper.cs
GarrettBeatty:
```
you don't really have to validate mapping.json against s3 model per se. If s3 adds a new field, the new TM request/response objects will simply not have those new fields, which means customers will not be able to use them unless we manually update mappings.json and ask SDKs to update their TM implementation to add those new fields
```

this is the expectation from zoe right now. that its just a one time check we are currently doing. if we get asked to update the mapping.json file then we would need to make updates

philasmar (approved):

PR #4073, opened 2025-10-23
fix silently failing test

GarrettBeatty:
Fix silently failing test. the logic in OnEventFired was wrong. it was throwing exception but not failing the test. there was 2 test cases not working related to unseekable streams. the issue was i was expecting transferred bytes to set for unseekable streams but in reality we dont set it (https://github.com/aws/aws-sdk-net/blob/3179a71b8f242c7a6177387c9f93752f71821a93/sdk/src/Services/S3/Custom/Transfer/Internal/MultipartUploadCommand.cs#L283). so this updates the test case to expect 0 bytes for these cases for transferredbytes.

Probably we can come back to this logic in multipartupload in the future, im not sure why transferredbytes wouldnt be able to be calculated?

## Motivation and Context
fix integ test

## Testing
ran test locally

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/test/Services/S3/IntegrationTests/TransferUtilityTests.cs
+++ b/sdk/test/Services/S3/IntegrationTests/TransferUtilityTests.cs
@@ -909,7 +909,7 @@ namespace AWSSDK_DotNet.IntegrationTests.Tests.S3
                     Assert.IsNotNull(args.Request);
                     Assert.IsNotNull(args.Response);
                     Assert.AreEqual(-1, args.TotalBytes); // Unseekable streams have unknown length
-                    Assert.AreEqual(20 * MEG_SIZE, args.TransferredBytes); // since we know the actual length via testing it, we can check the transferredbytes size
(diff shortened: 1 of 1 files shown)


sdk/test/Services/S3/IntegrationTests/TransferUtilityTests.cs
```diff
@@ -1758,15 +1758,16 @@ public void OnEventFired(object sender, T eventArgs)
             {
                 try
                 {
-                    EventFired = true;
                     Console.WriteLine("Lifecycle Event Fired: {0}", typeof(T).Name);
                     Validate?.Invoke(eventArgs);
+                    EventFired = true;  // Only set if validation passes
```
GarrettBeatty, line 1763: with this change, i reran the tests and they actually fail now if there is some assertion error. and then once i made the above updates to the test cases they now pass

philasmar (approved):

PR #4076, opened 2025-10-23
Add GetObjectResponse to TransferUtilityOpenStreamResponse mapping.

GarrettBeatty:
Stacked PRs:
 * #4079
 * #4105
 * __->__#4076

--- --- ---

Add GetObjectResponse to TransferUtilityOpenStreamResponse Mapping

## Motivation and Context
To copy all fields to the Transfer Utility Response so we ensure completeness.

## Testing
1. ran test locally
2. `46ee0201-7a68-49ad-8f1d-54a60adc7513` - pass

## Types of changes

- [] Bug fix (non-breaking change which fixes an issue)

## Checklist

## License


sdk/src/Services/S3/Custom/Transfer/Internal/ResponseMapper.cs
```diff
@@ -163,125 +163,156 @@ internal static TransferUtilityUploadResponse MapCompleteMultipartUploadResponse
         }
 
         /// <summary>
-        /// Maps a GetObjectResponse to TransferUtilityDownloadResponse.
-        /// Uses the field mappings defined in mapping.json "Conversion" -> "GetObjectResponse" -> "DownloadResponse".
+        /// Private helper method to populate the common properties from GetObjectResponse to the base response class.
```
GarrettBeatty, line 107: just making a shared helper function for stream and regular download since they have same response object. the only difference is the stream one has an extra responsestream field.

sdk/src/Services/S3/Custom/Transfer/TransferUtilityGetObjectResponseBase.cs
```diff
@@ -0,0 +1,293 @@
+/*******************************************************************************
```
GarrettBeatty, line 1: made this base class because stream and file based downloads have the same properties. the only difference is openstream command will have extra response stream field

sdk/src/Services/S3/Custom/Transfer/Internal/ResponseMapper.cs
```diff
@@ -163,123 +163,154 @@ internal static TransferUtilityUploadResponse MapCompleteMultipartUploadResponse
...
 
             if (source.ExpiresString != null)
-                response.ExpiresString = source.ExpiresString;
+                target.ExpiresString = source.ExpiresString;
 
             if (source.IsSetLastModified())
-                response.LastModified = source.LastModified;
+                target.LastModified = source.LastModified;
 
             if (source.Metadata != null)
```
peterrsongg: nit: source.Metadata won't be null because the getter always creates one if it doesn't exist

sdk/src/Services/S3/Custom/Transfer/Internal/ResponseMapper.cs
```diff
@@ -163,123 +163,154 @@ internal static TransferUtilityUploadResponse MapCompleteMultipartUploadResponse
...
 
             if (source.IsSetETag())
-                response.ETag = source.ETag;
+                target.ETag = source.ETag;
 
             if (source.Expiration != null)
-                response.Expiration = source.Expiration;
+                target.Expiration = source.Expiration;
 
             if (source.ExpiresString != null)
```
peterrsongg: `!IsNullOrEmpty`?

sdk/src/Services/S3/Custom/Transfer/Internal/ResponseMapper.cs
```diff
@@ -163,123 +163,154 @@ internal static TransferUtilityUploadResponse MapCompleteMultipartUploadResponse
...
+        /// <summary>
+        /// Maps a GetObjectResponse to TransferUtilityDownloadResponse.
+        /// Uses the field mappings defined in mapping.json "Conversion" -> "GetObjectResponse" -> "DownloadResponse".
+        /// </summary>
+        /// <param name="source">The GetObjectResponse to map from</param>
+        /// <returns>A new TransferUtilityDownloadResponse with mapped fields</returns>
+        internal static TransferUtilityDownloadResponse MapGetObjectResponse(GetObjectResponse source)
+        {
+            if (source == null)
+                return null;
```
peterrsongg: Should be throw an exception instead of returning null? usually we'd want to avoid returning null from a method

sdk/src/Services/S3/Custom/Transfer/Internal/ResponseMapper.cs
```diff
@@ -163,123 +163,154 @@ internal static TransferUtilityUploadResponse MapCompleteMultipartUploadResponse
...
+        /// <summary>
+        /// Maps a GetObjectResponse to TransferUtilityOpenStreamResponse.
+        /// Uses the same field mappings as DownloadResponse plus the ResponseStream property.
+        /// </summary>
+        /// <param name="source">The GetObjectResponse to map from</param>
+        /// <returns>A new TransferUtilityOpenStreamResponse with mapped fields</returns>
+        internal static TransferUtilityOpenStreamResponse MapGetObjectResponseToOpenStream(GetObjectResponse source)
+        {
+            if (source == null)
+                return null;
```
peterrsongg: same comment about returning null

sdk/src/Services/S3/Custom/Transfer/Internal/ResponseMapper.cs
GarrettBeatty: yea you are right

GarrettBeatty: fixed

sdk/src/Services/S3/Custom/Transfer/Internal/ResponseMapper.cs
GarrettBeatty: i updated everywhere to just do a pass-through and not even need the null checks.

sdk/src/Services/S3/Custom/Transfer/TransferUtilityOpenStreamResponse.cs
```diff
@@ -0,0 +1,42 @@
...
+using System.IO;
+using Amazon.Runtime;
+
+namespace Amazon.S3.Transfer
+{
+    /// <summary>
+    /// Response object for Transfer Utility open stream operations.
+    /// Contains the stream and response metadata from open stream operations.
+    /// </summary>
+    public class TransferUtilityOpenStreamResponse : TransferUtilityGetObjectResponseBase
```
peterrsongg: Do we have to implement `IDisposable` here like we do for  `GetObjectResponse` which inherits `StreamResponse`? This would allow someone to wrap this class in a `using` statement

peterrsongg: https://github.com/aws/aws-sdk-net/blob/6147c3f66e9c6741291dbfb9fe98d37dec6fa6de/sdk/src/Services/S3/Custom/Model/StreamResponse.cs#L28

GarrettBeatty: updated to use idisposable and added test

peterrsongg (approved):

PR #4083, opened 2025-10-28
Add additional validation to UploadPartRequests in Transfer Utility

philasmar:
DOTNET-8274

## Description
Add additional validation to UploadPartRequests in Transfer Utility

## Motivation and Context

Add additional validation to UploadPartRequests in Transfer Utility

## Testing

Ran tests locally
Ran successful dry run
## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/Transfer/Internal/MultipartUploadCommand.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/Internal/MultipartUploadCommand.cs
@@ -222,7 +222,10 @@ namespace Amazon.S3.Transfer.Internal
         {
             UploadPartRequest uploadPartRequest = ConstructGenericUploadPartRequest(initiateResponse);
 
+            // Calculating how many bytes are remaining to be uploaded from the current part.
+            // This is mainly used for the last part scenario.
             var remainingBytes = this._contentLength - filePosition;
+            // We then check based on the remaining bytes and the content length if this is the last part.
             var isLastPart = calculateIsLastPart(remainingBytes);
             uploadPartRequest.PartNumber = partNumber;
             uploadPartRequest.PartSize = isLastPart ? remainingBytes : this._partSize;
--- a/sdk/src/Services/S3/Custom/Transfer/Internal/_async/MultipartUploadCommand.async.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/Internal/_async/MultipartUploadCommand.async.cs
@@ -73,8 +73,13 @@ namespace Amazon.S3.Transfer.Internal
                         var uploadRequest = ConstructUploadPartRequest(i, filePosition, initResponse);
 
                         var expectedFileOffset = (i - 1) * this._partSize;
+                        // Calculating how many bytes are remaining to be uploaded from the current part.
+                        // This is mainly used for the last part scenario.
                         var remainingBytes = this._contentLength - expectedFileOffset;
+                        // We then check based on the remaining bytes and the content length if this is the last part.
                         var isLastPart = calculateIsLastPart(remainingBytes);
+                        // To maintain the same behavior as the ConstructUploadPartRequest.
+                        // We are setting the remainingBytes/partSize when using the IAmazonS3Encryption client to 0.
                         if (isLastPart
                             && _s3Client is Amazon.S3.Internal.IAmazonS3Encryption)
                         {
@@ -190,7 +195,7 @@ namespace Amazon.S3.Transfer.Internal
 
                         if (expectedUploadPart.IsLastPart)
                         {
-                            if (actualContentLength <= 0 ||
+                            if (actualContentLength < 0 ||
                                 actualContentLength > expectedUploadPart.ExpectedContentLength)
                             {
                                 throw new InvalidOperationException($"Cannot complete multipart upload request. The last part " +
--- a/sdk/test/Services/S3/UnitTests/Custom/MultipartUploadValidationTests.cs
+++ b/sdk/test/Services/S3/UnitTests/Custom/MultipartUploadValidationTests.cs
@@ -13,7 +13,7 @@ using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 
-namespace AWSSDK.UnitTests.S3.NetFramework.Custom
+namespace AWSSDK.UnitTests
 {
     [TestClass]
     public class MultipartUploadValidationTests


sdk/src/Services/S3/Custom/Transfer/Internal/MultipartUploadCommand.cs
```diff
@@ -513,4 +519,12 @@ public void OnTransferProgress(object sender, StreamTransferProgressArgs e)
             _lastProgressArgs = e;
         }
     }
+
+    internal class ExpectedUploadPart
```
GarrettBeatty: can this be private class? or is it needed for unit test to be internal?

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartUploadCommand.cs
```diff
@@ -215,8 +216,13 @@ internal UploadPartRequest ConstructUploadPartRequest(int partNumber, long fileP
         {
             UploadPartRequest uploadPartRequest = ConstructGenericUploadPartRequest(initiateResponse);
 
+            var remainingBytes = this._contentLength - filePosition;
```
GarrettBeatty, line 227: super nit: (just a suggestion) is it worth making a function for calculateIsLastPart. and that can be used here and then also in below if statement instead of `filePosition + this._partSize >= this._contentLengt`

sdk/test/Services/S3/UnitTests/Custom/MultipartUploadValidationTests.cs
```diff
@@ -0,0 +1,133 @@
...
+                Assert.AreEqual(i, constructUploadPartRequest.PartNumber);
+                Assert.AreEqual(isLastPart, constructUploadPartRequest.IsLastPart);
+                Assert.AreEqual(
+                    isLastPart ? remainingBytes : partSize, 
+                    constructUploadPartRequest.PartSize);
+                Assert.AreEqual(expectedFileOffset, constructUploadPartRequest.FilePosition);
+
+                filePosition += partSize;
+            }
+
```
GarrettBeatty, line 130: is there any easy way to right a unit test/integration tests which would reach the code path that throws exception? somehow mock the ConstructUploadPartRequest call to put in incorrect values and then when the response is sent and validation is reached we assert validation error?

sdk/src/Services/S3/Custom/Transfer/Internal/_async/MultipartUploadCommand.async.cs
```diff
@@ -150,8 +165,50 @@ private async Task<UploadPartResponse> UploadPartAsync(UploadPartRequest uploadR
         {
             try
             {
-                return await _s3Client.UploadPartAsync(uploadRequest, internalCts.Token)
+                var response = await _s3Client.UploadPartAsync(uploadRequest, internalCts.Token)
                     .ConfigureAwait(continueOnCapturedContext: false);
+
```
GarrettBeatty, line 180: i think all of the validation looks good. just a general question. it looks like we do the other validation for total number of parts = expected here https://github.com/aws/aws-sdk-net/blob/6147c3f66e9c6741291dbfb9fe98d37dec6fa6de/sdk/src/Services/S3/Custom/Transfer/Internal/MultipartUploadCommand.cs#L164 (right before we do complete multi part upload). 

i dont think it matters to me where we do it too much. what do you think?

GarrettBeatty: also will having the logic here fail/work for unseekable streams? i think the reason the other code is in the other place is because for unseekable streams they skip validation. wondering if this code path gets executed for unseekable stream or we need to skip it for it?

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartUploadCommand.cs
GarrettBeatty, line 237: is this going to fail validation if we are using amazonencryption client? do we need to skip it in that case?

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartUploadCommand.cs
philasmar: Updated

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartUploadCommand.cs
philasmar: Added

sdk/test/Services/S3/UnitTests/Custom/MultipartUploadValidationTests.cs
philasmar: Not without restructuring a lot of existing code. Most of the stuff is created within the functions and not really mockable. Let me know if you have thoughts on how to do it.

sdk/src/Services/S3/Custom/Transfer/Internal/_async/MultipartUploadCommand.async.cs
philasmar: The validation `if (this._uploadResponses.Count != this._totalNumberOfParts) ` is being skipped for unseekable streams. The new validation logic we added will and should be skipped. The code path of the UploadPart function where i added the new validation is different from the unseekable stream code path.

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartUploadCommand.cs
philasmar: I don't know why the Part Size gets set to 0, but i updated our validation to match the same behavior so that we don't cause unexpected issues.

sdk/test/Services/S3/UnitTests/Custom/MultipartUploadValidationTests.cs
```diff
@@ -0,0 +1,133 @@
...
+using Moq;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace AWSSDK.UnitTests.S3.NetFramework.Custom
```
GarrettBeatty: this namespace is not right. should be AWSSDK.UnitTests

peterrsongg: yeah ^^ also, that means the dry run might not have run these tests?

philasmar: Will rename and rerun the dry run

sdk/src/Services/S3/Custom/Transfer/Internal/_async/MultipartUploadCommand.async.cs
```diff
@@ -69,6 +71,24 @@ public override async Task ExecuteAsync(CancellationToken cancellationToken)
                         cancellationToken.ThrowIfCancellationRequested();
 
                         var uploadRequest = ConstructUploadPartRequest(i, filePosition, initResponse);
+
+                        var expectedFileOffset = (i - 1) * this._partSize;
+                        var remainingBytes = this._contentLength - expectedFileOffset;
+                        var isLastPart = calculateIsLastPart(remainingBytes);
+                        if (isLastPart
+                            && _s3Client is Amazon.S3.Internal.IAmazonS3Encryption)
```
peterrsongg, line 84: can you explain this part in a comment in the code?

sdk/src/Services/S3/Custom/Transfer/Internal/_async/MultipartUploadCommand.async.cs
```diff
@@ -150,8 +170,50 @@ private async Task<UploadPartResponse> UploadPartAsync(UploadPartRequest uploadR
...
+                        var actualContentLength = uploadRequest.PartSize;
+                        if (actualContentLength != expectedUploadPart.ExpectedContentLength)
+                        {
+                            throw new InvalidOperationException($"Cannot complete multipart upload request. The expected content length of part {expectedUploadPart.PartNumber} " +
+                                $"does not equal the actual content length.");
+                        }
+
+                        if (expectedUploadPart.IsLastPart)
+                        {
+                            if (actualContentLength <= 0 ||
```
peterrsongg: isn't a 0 length upload still valid?

sdk/src/Services/S3/Custom/Transfer/Internal/_async/MultipartUploadCommand.async.cs
```diff
@@ -150,8 +170,50 @@ private async Task<UploadPartResponse> UploadPartAsync(UploadPartRequest uploadR
...
+                                actualContentLength > expectedUploadPart.ExpectedContentLength)
```
peterrsongg, line 199: I'm wondering if we can just do something like
```
actualContentLength != expectedUploadPart.ExpectedContentLength
```

sdk/src/Services/S3/Custom/Transfer/Internal/_async/MultipartUploadCommand.async.cs
philasmar: is it valid? Why would the last part have a 0 content length?

sdk/src/Services/S3/Custom/Transfer/Internal/_async/MultipartUploadCommand.async.cs
philasmar: Done

sdk/src/Services/S3/Custom/Transfer/Internal/_async/MultipartUploadCommand.async.cs
philasmar: After discussing offline, I removed the "or equal".

sdk/src/Services/S3/Custom/Transfer/Internal/_async/MultipartUploadCommand.async.cs
philasmar: Discussed that this check already exists and we are adding additional checks based on the design,.

peterrsongg (approved):
approving, assuming new dry run passes

philasmar:
> approving, assuming new dry run passes

New dry run passed

PR #4079, opened 2025-10-23
Add DownloadInitiated, Failed and Completed events

GarrettBeatty:
Stacked PRs:
 * #4109
 * __->__#4079

--- --- ---

## Description

This change adds progress listeners for "initiated", "complete", and "failed" lifecycle events for the DownloadCommand Consumers can subscribe to these callbacks to receive notifications when a simple upload starts, finishes successfully, or fails. Its similar to https://github.com/aws/aws-sdk-net/pull/4059 but for downloads

## Motivation and Context
1. To adhere to the SEP

## Testing
1. `f0cb4050-fb2c-4fb0-b489-280eb3313ca9` - pass
2. Integration tests

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project
- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [x] I have read the **README** document
- [x] I have added tests to cover my changes
- [x] All new and existing tests passed

## License

- [x] I confirm that this pull request can be released under the Apache 2 license

[issues]: https://github.com/aws/aws-sdk-net/issues
[license]: http://aws.amazon.com/apache2.0/
[cla]: http://en.wikipedia.org/wiki/Contributor_License_Agreement

PR #4085, opened 2025-10-29
add UploadWithResponseAsync api

GarrettBeatty:
Stacked PRs:
 * #4079
 * __->__#4085
 * #4076
 * #4087

--- --- ---

## Description
This change creates a new api `UploadWithResponseAsync`. It is the same as the existing `UploadAsync` api but it returns the `TransferUtilityUploadResponse` object.

## How
In order to re-use as much as the existing code as possible, I made `BaseCommand` take in a generic which is the return type of the `ExecuteAsync` function. All commands `UploadCommand`, `DownloadCommand`, etc will pass in their expected return type `TransferUtilityUploadResponse`, `TransferUTilityDownloadResponse`, etc. 

In order to make this change in one go, I had to create place holder classes for `TransferUtilityDownloadDirectoryResponse` and other responses. These will eventually be populated in future PRs 

## Motivation and Context
1. To adhere to the SEP

## Testing
1.dry run - d704b36b-7329-42dc-80e0-525a2f82bbcc pass
2. Integration tests

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/Transfer/TransferUtilityAbortMultipartUploadsResponse.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/TransferUtilityAbortMultipartUploadsResponse.cs
@@ -1,0 +1,35 @@
+/*******************************************************************************
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  Licensed under the Apache License, Version 2.0 (the "License"). You may not use
+ *  this file except in compliance with the License. A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ *  or in the "license" file accompanying this file.
+ *  This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ *  CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ *  specific language governing permissions and limitations under the License.
+ * *****************************************************************************
+ *    __  _    _  ___
+ *   (  )( \/\/ )/ __)
+ *   /__\ \    / \__ \
+ *  (_)(_) \/\/  (___/
+ *
+ *  AWS SDK for .NET
+ *  API Version: 2006-03-01
+ *
+ */
+
+using Amazon.Runtime;
+
+namespace Amazon.S3.Transfer
+{
+    /// <summary>
+    /// Response object for Transfer Utility abort multipart uploads operations.
+    /// Contains response metadata from abort multipart uploads operations.
+    /// </summary>
+    public class TransferUtilityAbortMultipartUploadsResponse
+    {
+        // Empty placeholder class - properties will be added in future iterations
+    }
+}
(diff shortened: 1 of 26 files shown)


sdk/src/Services/S3/Custom/Transfer/TransferUtilityAbortMultipartUploadsResponse.cs
```diff
@@ -0,0 +1,35 @@
+/*******************************************************************************
```
GarrettBeatty, line 1: placeholder class. not sure if we need to populate abort upload but i had to make it anyway

sdk/src/Services/S3/Custom/Transfer/TransferUtilityDownloadDirectoryResponse.cs
```diff
@@ -0,0 +1,26 @@
...
+ *  http://aws.amazon.com/apache2.0
+ * 
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+using Amazon.Runtime;
+
```
GarrettBeatty, line 17: this will be populated later on

sdk/src/Services/S3/Custom/Transfer/_async/TransferUtility.async.cs
```diff
@@ -88,6 +88,7 @@ public partial class TransferUtility : ITransferUtility
         ///     A cancellation token that can be used by other objects or threads to receive notice of cancellation.
         /// </param>
         /// <returns>The task object representing the asynchronous operation.</returns>
+        [Obsolete("Use UploadWithResponseAsync instead which allows you to upload files and also returns response metadata.", error: false)]
```
GarrettBeatty, line 91: copying phils question from other pr. 

"why do we need to mark these as obsolete? the obsolete reason is not really convincing enough to mark the method as obsolete."

-----

my answer

"
we want to encourage users to use the new functions. we are also creating DownloadWithResponse (Which will use multi part download) so having them use the "withResponse" functions of both upload/download makes sense to me
"

sdk/src/Services/S3/Custom/_async/AmazonS3Client.Extensions.cs
```diff
@@ -156,6 +153,15 @@ async Task ICoreAmazonS3.EnsureBucketExistsAsync(string bucketName)
             }
         }
         #endregion
+
+        #region Internal Helper Methods
+
+        private async Task UploadObjectInternalAsync(Amazon.S3.Transfer.TransferUtilityUploadRequest request, CancellationToken cancellationToken)
```
GarrettBeatty, line 159: @peterrsongg updated here to pass in request object

sdk/src/Services/S3/Custom/Transfer/_async/TransferUtility.async.cs
GarrettBeatty: ended up removing obsolete as per norm

PR #4087, opened 2025-10-29
Remove AmazonWebServiceResponse as base class for transfer utility repsonse objects.

GarrettBeatty:
--- --- ---
Remove AmazonWebServiceResponse as base class for transfer utility response objects. These shouldnt extend this class since its not the raw response from amazon web service.

Ran response mapper and transfer utility tests locally. 

Dry run - c4907d4e-948c-49c2-9b57-57f93dbde562 - in progress

### DIFF
--- a/sdk/test/Services/S3/UnitTests/Custom/ResponseMapperTests.cs
+++ b/sdk/test/Services/S3/UnitTests/Custom/ResponseMapperTests.cs
@@ -149,8 +149,7 @@ namespace AWSSDK.UnitTests
                 },
                 (sourceResponse, targetResponse) =>
                 {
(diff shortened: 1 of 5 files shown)


sdk/test/Services/S3/UnitTests/Custom/ResponseMapperTests.cs
```diff
@@ -524,12 +523,11 @@ public void MapCompleteMultipartUploadResponse_AllMappedProperties_WorkCorrectly
                 (sourceResponse) =>
                 {
                     sourceResponse.HttpStatusCode = HttpStatusCode.OK;
-                    sourceResponse.ContentLength = 2048;
+                    sourceResponse.ContentLength = 1024;
```
GarrettBeatty, line 646: to match the value of 1024 we are populating in the headers too

philasmar (approved):

PR #4088, opened 2025-10-30
[V4] Rename ErrorType exception properties to avoid hiding inherited member

muhammad-othman:
## Description
This change resolves a naming conflict in exception classes where the `ErrorType` property could hide the inherited `ErrorType` member from the base `AmazonServiceException` class.
Also removed previous customization approach at `AccountService` in favor of the generator-level fix, which caused the private property name to change but the pubic properties stayed that same.

## Motivation and Context
`DOTNET-8305`

## Testing
* Ran the generator and validated that only the `AccountService` was affect by this change.
* `DRY_RUN-3f60b43c-dedc-49b2-8e2d-1afc309f4889`

## Types of changes

## Checklist

## License


generator/ServiceClientGeneratorLib/Member.cs
```diff
@@ -482,6 +487,22 @@ public string DetermineType()
...
 
+        /// <summary>
+        /// Retrieves a property name override for specific member names that require special handling
+        /// during code generation to avoid naming conflicts or improve API consistency.
+        /// </summary>
+        /// <returns>
+        /// A string containing the overridden property name if a special case applies,
+        /// or null if the default property name should be used.
+        /// </returns>
+        private string GetPropertyNameOverride()
```
peterrsongg: same comment as V3 on the name of the method

PR #4093, opened 2025-11-03
Update AbortMultipartUploadcommand to use all fields

GarrettBeatty:
## Description
1. Create TransferUtilityAbortMultipartUploadRequest and new api that uses this request object so that all fields (Request player, expected bucket owner) are used when aborting a request.

## Motivation and Context
To adhere to the SEP

## Testing
1. Unit tests
2. Dry Run 6b6ec5e3-51da-4b4a-aad4-3ff0c894e70b PASS

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/Transfer/TransferUtilityAbortMultipartUploadRequest.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/TransferUtilityAbortMultipartUploadRequest.cs
@@ -1,0 +1,126 @@
+/*******************************************************************************
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  Licensed under the Apache License, Version 2.0 (the "License"). You may not use
+ *  this file except in compliance with the License. A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ *  or in the "license" file accompanying this file.
+ *  This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ *  CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ *  specific language governing permissions and limitations under the License.
+ * *****************************************************************************
+ *    __  _    _  ___
+ *   (  )( \/\/ )/ __)
+ *   /__\ \    / \__ \
+ *  (_)(_) \/\/  (___/
+ *
+ *  AWS SDK for .NET
+ *  API Version: 2006-03-01
+ *
+ */
+using System;
+using Amazon.S3.Model;
+
+namespace Amazon.S3.Transfer
+{
+    /// <summary>
+    /// Contains all the parameters that can be set when making a request to abort multipart uploads
+    /// with the <c>TransferUtility</c> method.
+    /// </summary>
+    public class TransferUtilityAbortMultipartUploadRequest
+    {
+        private string _bucketName;
+        private DateTime? _initiatedDate;
+        private string _expectedBucketOwner;
+        private RequestPayer _requestPayer;
+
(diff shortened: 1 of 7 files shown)


sdk/src/Services/S3/Custom/Transfer/TransferUtilityAbortMultipartUploadRequest.cs
```diff
@@ -0,0 +1,117 @@
...
+            return !string.IsNullOrEmpty(this._bucketName);
+        }
+
+        /// <summary>
+        /// Gets or sets the date before which the multipart uploads were initiated.
+        /// </summary>
+        /// <value>
+        /// The date before which the multipart uploads were initiated.
+        /// </value>
+        public DateTime InitiatedDate
```
GarrettBeatty: kept this as non optional since the original one wasnt DateTime? and i didnt want to break things

sdk/test/Services/S3/UnitTests/Custom/ResponseMapperTests.cs
```diff
@@ -260,6 +260,126 @@ public void MapUploadRequest_AbortMultipartRequest_AllMappedProperties_WorkCorre
                 });
         }
 
+        [TestMethod]
+        [TestCategory("S3")]
+        public void MapAbortMultipartUploadsCommand_ConstructAbortMultipartUploadRequest_AllMappedProperties_WorkCorrectly()
```
GarrettBeatty, line 265: had to manually check the fields for now since transferutilityabortmultipartuploadrequest doesnt contain `Key` (its derived in our helper logic from calling list uploads with associated timestamp)

sdk/src/Services/S3/Custom/Transfer/TransferUtilityAbortMultipartUploadRequest.cs
philasmar: i think the change to DateTime? would be ok. We're making things more permissive instead of restrictive, so should be fine?

GarrettBeatty: actually i think i know why its not optional. its not actually being set to aws. its only being used to compare in this line

```
                            if (upload.Initiated < this._request.InitiatedDate)

```

so i think keeping it as-is is fine

muhammad-othman: I agree that we shouldn't change the `public void AbortMultipartUploads(string bucketName, DateTime initiatedDate)` method and changing `DateTime` to `DateTime?` here would be a breaking change. However I believe that making `TransferUtilityAbortMultipartUploadRequest.InitiatedDate` of type `DateTime?` should be fine, since we aren't changing any existing interface and part can stay the same even with the nullable `DateTime`

Image

Though we will need to change `if (upload.Initiated < this._request.InitiatedDate)` to change if `InitiatedDate` is set, probably using `GetValueOrDefault()`.

GarrettBeatty: yeah i dont care either way which way we do it. is the pattern of using DateTime? only applicable for if we are sending that value to AWS though? or would it be for everything?

GarrettBeatty: updated in https://github.com/aws/aws-sdk-net/pull/4093/commits/0f56a43da01110bfd6a5010824f92632d2655b26 to be nullable and add validation

PR #4100, opened 2025-11-05
[V4] Re-introduce background refresh behavior

### LINKED ISSUE #4024: Refresh credentials in the background during preempt expiry
### Describe the feature

In V4, we attempted to perform a background refresh of credentials during their preempt expiry period (https://github.com/aws/aws-sdk-net/pull/3541), however we received several customer reports of expired credentials being used and had to revert the change (https://github.com/aws/aws-sdk-net/pull/4023).

The background refresh makes sense as a feature request (it was attempting to improve scenarios such as https://github.com/aws/aws-sdk-net/issues/2464), but we need to fix the proposed implementation before re-introducing the behavior.

dscpinheiro:
Resolves #4024 

## Description
This PR is the second attempt to update credentials to expire during their preempt expiry period. The approach used is similar to the previous behavior (https://github.com/aws/aws-sdk-net/pull/3541) but with some important differences:
- We added an expiration buffer to prevent credentials that are about to expire from being returned (currently 1 minute which should be enough for the current operation to complete). This is a similar approach to what other SDKs do (Java for example has a [stale and prefetch time](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/credential-caching.html), while we were using the preempt expiry for both)
- We also added a `CredentialsLoadState` enum to guarantee only one background refresh runs at a time

## Motivation and Context
`DOTNET-8326` (and a couple of internal support tickets too)

## Testing
Dry-runs (queued):
- .NET: `DRY_RUN-c85e29cd-d738-450d-8c5b-295646a0aa7d`
- PowerShell: `DRY_RUN-fbab3c8d-b969-441b-9523-fcfdc7eb1935`

In addition to the unit tests I added back, I also ran a couple of simple console apps locally (calling `GetCredentials` on a loop with `AssumeRoleAWSCredentials`):

- First app (second call to get credentials happened after they were already expired):
```
AssumeRoleAWSCredentials 40|2025-11-05T15:44:15.641Z|DEBUG|New credentials created for assume role that expire at 2025-11-05T15:59:15.0000000Z
RefreshingAWSCredentials 41|2025-11-05T16:04:15.648Z|INFO|Determined refreshing credentials should update. Expiration time: 2025-11-05T15:58:15.0000000Z, Current time: 2025-11-05T16:04:15.6485332Z
AssumeRoleAWSCredentials 46|2025-11-05T16:04:16.017Z|DEBUG|New credentials created for assume role that expire at 2025-11-05T16:19:15.0000000Z
```

- Second app (calls to get credentials happen consistently, and they're refreshed in the preempt expiry window):
```
AssumeRoleAWSCredentials 40|2025-11-05T16:22:53.596Z|DEBUG|New credentials created for assume role that expire at 2025-11-05T16:37:53.0000000Z
RefreshingAWSCredentials 41|2025-11-05T16:31:53.644Z|INFO|Determined refreshing credentials are in window for preempt expiration. Preempt time: 2025-11-05T16:31:53.0000000Z, Current time: 2025-11-05T16:31:53.6446562Z
AssumeRoleAWSCredentials 46|2025-11-05T16:31:53.990Z|DEBUG|New credentials created for assume role that expire at 2025-11-05T16:46:53.0000000Z
```

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Credentials/RefreshingAWSCredentials.cs
+++ b/sdk/src/Core/Amazon.Runtime/Credentials/RefreshingAWSCredentials.cs
@@ -114,6 +114,8 @@ namespace Amazon.Runtime
             Loading,
         }
 
+        // Note this is purposefuly marked as volatile since it is modified by multiple threads. Read the comments
+        // in the GetCredentials method for more information on the locking flow and how this is used.
         private volatile CredentialsLoadState currentLoadState;
 
         #endregion
@@ -166,6 +168,31 @@ namespace Amazon.Runtime
             // We save the currentState as it might be modified or cleared.
             var tempState = currentState;
 
+            // Before acquiring the lock, check if we need to refresh credentials. This is essentially the read only section
+            // and going into the if block is the costly write section. Majority of threads going through this path will stay
+            // in the read section avoiding blocking on the semaphore. Execution only goes into the write section when credentials
+            // are nearing expiration or already expired.
+            //
+            // There are two phases of credentials needing to be refreshed:
+            // 
+            // Credentials are expired. In that case the lock will be acquired and credential refresh will be blocking further 
+            // execution until new credentials are retrieved. Once credentials are retrieved and the lock is released any other
+            // threads that were blocked due to expired credentials will one by one acquire the lock and see that 
+            // new credentials are present and use those refreshed credentials. 
+            //
+            // Credential epiration in preempt window. This is the case credentials are still valid but to avoid a later
+            // blocking expiration refresh a background refresh is triggered. Only one background refresh will be triggered
+            // which is controlled by setting currentLoadState to Loading while the lock is held.
+            // Any other threads that come through in either the read or write section in the prempt window will see the 
+            // currentLoadState is Loading and not trigger another background refresh. The current credentials will be returned
+            // instead of waiting for the background refresh to complete since they are still valid.
+            //
+            // The background refresh task is in charge of reseting the currentLoadState to NotLoading once the background refresh is complete
+            // or an exception happened. In the case of an exception during the background refresh resettting the currentLoadState to NotLoading
+            // is not done within the scope of the lock. This is safe because the only other place currentLoadState is modified to a different value
+            // is within the block of code that initiates the background refresh. The execution will never go into that block while a background
+            // refresh is in progress because the currentLoadState was set to Loading which prevents starting a background refresh. Since 
+            // currentLoadState is potentially modified by multiple threads it has been marked as volatile to ensure the latest value is always read.
             if (IsExpired(tempState) || (currentLoadState != CredentialsLoadState.Loading && IsPreemptExpiryWindow(tempState)))
             {
                 _updateGeneratedCredentialsSemaphore.Wait();
@@ -203,6 +230,9 @@ namespace Amazon.Runtime
 
         public override sealed async Task<ImmutableCredentials> GetCredentialsAsync()
         {
+            // NOTICE: Before modifying any of the logic read the comments in the synchronous GetCredentials method to 
+            // understand the locking flow. If any changes are required be sure the comments in that method are also updated.
+
             // We save the currentState as it might be modified or cleared.
             var tempState = currentState;
 
@@ -243,6 +273,9 @@ namespace Amazon.Runtime
 
         private async Task BackgroundCredentialsRefreshAsync()
         {
+            // NOTICE: Before modifying any of the logic read the comments in the synchronous GetCredentials method to 
+            // understand the locking flow. If any changes are required be sure the comments in that method are also updated.
+
             try
             {
                 var newState = await GenerateNewCredentialsAsync().ConfigureAwait(false);
@@ -263,8 +296,13 @@ namespace Amazon.Runtime
             catch (Exception e)
             {
                 _logger.Error(e, "Exception occurred performing background credentials refresh.");
+
                 // If any exceptions occur during background refresh, reset the state to NotLoading
                 // so that future GetCredentials calls can attempt to refresh again.
+                //
+                // This is safe to modify outside of the lock because the only other place currentLoadState is modified
+                // to a different value is within the block of code that initiates the background refresh. The block
+                // can never be entered while this background refresh is in progress because currentLoadState was set to Loading.
                 currentLoadState = CredentialsLoadState.NotLoading;
                 throw;
             }


sdk/src/Core/Amazon.Runtime/Credentials/InstanceProfileAWSCredentials.cs
```diff
@@ -91,10 +91,7 @@ protected override CredentialsRefreshState GenerateNewCredentials()
                 // but try again to refresh them in 2 minutes.
                 if (null != _currentRefreshState)
                 {
-#pragma warning disable CS0612, CS0618 // Type or member is obsolete
-                    var newExpiryTime = AWSSDKUtils.CorrectedUtcNow + TimeSpan.FromMinutes(2);
-#pragma warning restore CS0612,CS0618 // Type or member is obsolete
-
+                    var newExpiryTime = _timeProvider.CorrectedUtcNow + TimeSpan.FromMinutes(2);
```
dscpinheiro, line 94: Norm had suggested an easier way to mock `AWSSDKUtils.CorrectedUtcNow` so I added this new time provider interface. It's supposed to be similar to https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Core/Amazon.Util/Internal/IEnvironmentVariableRetriever.cs (which to be fair we don't use it as much as we should...)

`InstanceProfileAWSCredentials` is the only class that inherits from `RefreshingAWSCredentials` (where I accept the new interface) that was calling `AWSSDKUtils.CorrectedUtcNow` directly.

sdk/test/NetStandard/UnitTests/Core/Credentials/RefreshingAWSCredentialsTests.cs
dscpinheiro, line 1: These tests were in the first PR, but I modified them to include more edge case scenarios (such as: in the preempt expiry window but after the expiration buffer).

normj (approved):

sdk/src/Core/Amazon.Runtime/Credentials/RefreshingAWSCredentials.cs
```diff
@@ -185,65 +275,81 @@ private void UpdateToGeneratedCredentials(CredentialsRefreshState state)
...
-        // Test credentials existence and expiration time
-        // should update if:
-        //  credentials have not been loaded yet
-        // it's past the expiration time. At this point currentState.Expiration may 
-        // have the PreemptExpiryTime baked into to the expiration from a call to 
-        // UpdateToGeneratedCredentials but it may not if this is new application 
-        // load.
-        private bool ShouldUpdateState(CredentialsRefreshState state)
+        /// <summary>
+        /// Test if the credentials are expired currently expired.
```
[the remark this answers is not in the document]
peterrsongg: seems like valid feedback

sdk/src/Core/Amazon.Runtime/Credentials/RefreshingAWSCredentials.cs
```diff
@@ -185,65 +275,81 @@ private void UpdateToGeneratedCredentials(CredentialsRefreshState state)
...
+            _logger.InfoFormat(
+                "Determined refreshing credentials should update. Expiration time: {0}, Current time: {1}",
+                state.Expiration.ToUniversalTime().ToString("yyyy-MM-ddTHH:mm:ss.fffffffK", CultureInfo.InvariantCulture),
+                _timeProvider.CorrectedUtcNow.ToString("yyyy-MM-ddTHH:mm:ss.fffffffK", CultureInfo.InvariantCulture)
+            );
+        }
+
+        /// <summary>
+        /// Test if the credentials are in the preempt expiry window.
+        /// <para />
```
peterrsongg, line 377: nit: is the `<para />` supposed to be here?

sdk/src/Core/Amazon.Runtime/Credentials/InstanceProfileAWSCredentials.cs
```diff
@@ -187,7 +181,7 @@ protected override async Task<CredentialsRefreshState> GenerateNewCredentialsAsy
                 _logger.InfoFormat(_receivedExpiredCredentialsFromIMDS);
 
                 // use a custom refresh time
-                var newExpiryTime = AWSSDKUtils.CorrectedUtcNow + TimeSpan.FromMinutes(new Random().Next(5, 11));
+                var newExpiryTime = _timeProvider.CorrectedUtcNow + TimeSpan.FromMinutes(new Random().Next(5, 11));
                 _currentRefreshState = new CredentialsRefreshState(newState.Credentials, newExpiryTime);
```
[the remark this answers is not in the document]
dscpinheiro: I ended up not changing `InstanceProfileAWSCredentials` since the tests interact with the parent class directly (and the default time provider uses `AWSSDKUtils.CorrectedUtcNow` anyway).

sdk/src/Core/Amazon.Runtime/Credentials/RefreshingAWSCredentials.cs
```diff
@@ -119,62 +150,121 @@ public TimeSpan PreemptExpiryTime
...
-                return tempState.Credentials;
             }
-            finally
+            catch (Exception e)
             {
-                _updateGeneratedCredentialsSemaphore.Release();
+                _logger.Error(e, "Exception occurred performing background credentials refresh.");
+                // If any exceptions occur during background refresh, reset the state to NotLoading
+                // so that future GetCredentials calls can attempt to refresh again.
+                currentLoadState = CredentialsLoadState.NotLoading;
```
boblodgett, line 306: This is setting `currentLoadState` while it is outside of the lock. I have not yet traced under what case this could cause an issue. However, it should only be set within the lock and should be moved to the finally on line 245 and remove the one one 243.

I also see that maybe you did this because ValidateGeneratedCredentials(newState); could throw an exception which would leave currentLoadState as "Loading" potentially if it is moved. 

Unsure how to fix this yet but it is still potentially unsafe to set the variable outside of the lock.

peterrsongg, line 306: Shouldn't we set `currentLoadState` here using a lock also? It seems like this setting of state should also be thread safe.

Could there be a scenario where a background refresh fails and sets the state to `NotLoading` then another thread enters `GetCredentials()` and sees `NotLoading` state while the exception from the background refresh is still propagating leading to an inconsistent state?

sdk/src/Core/Amazon.Runtime/Credentials/RefreshingAWSCredentials.cs
```diff
@@ -74,19 +85,23 @@ internal bool IsExpiredWithin(TimeSpan preemptExpiryTime)
...
 
-        /// <summary>
-        /// Semaphore to control thread access to GetCredentialsAsync method.
-        /// The semaphore will allow only one thread to generate new credentials and
-        /// update the current state.
-        /// </summary>
         private readonly SemaphoreSlim _updateGeneratedCredentialsSemaphore = new SemaphoreSlim(1, 1);
+        private enum CredentialsLoadState
+        {
+            NotLoading,
```
peterrsongg, line 108: can we explain in a comment so we know what each state represents? What exactly are we loading and not loading?

sdk/src/Core/Amazon.Runtime/Credentials/RefreshingAWSCredentials.cs
peterrsongg: I'm thinking we acquire the semaphore again here so that we synchronize the state across threads
```
    await _updateGeneratedCredentialsSemaphore.WaitAsync().ConfigureAwait(false);
    try
    {
        currentLoadState = CredentialsLoadState.NotLoading;
    }
    finally
    {
        _updateGeneratedCredentialsSemaphore.Release();
    }
```

sdk/test/NetStandard/UnitTests/Core/Credentials/RefreshingAWSCredentialsTests.cs
```diff
@@ -0,0 +1,226 @@
...
+        {
+            public DateTime CorrectedUtcNow { get; set; }
+        }
+
+        // using a hand-written mock in order to have access to the protected fields
+        private sealed class MockRefreshingAWSCredentials : RefreshingAWSCredentials
+        {
+            private readonly TimeSpan _credentialsLifetime;
+            private readonly ManualResetEventSlim _generateCredsEvent;
+            private int _tokenCounter;
```
peterrsongg, line 223: what about the `expirationBuffer` that exists in the base class, should we be testing scenario with that too or will it aways be 1 minute? 

For example, it would be nice if we could test with values greater than the `PreEmptExpiry` time to validate what the behavior would be.

```
    var mockCredentials = new MockRefreshingAWSCredentials(TimeSpan.FromMinutes(60), _mockProvider)
    {
        PreemptExpiryTime = TimeSpan.FromMinutes(5), 
        ExpirationBuffer = TimeSpan.FromMinutes(10)  
    };
   _mockProvider.CorrectedUtcNow = _baseTimeUtc;
    var initialCreds = mockCredentials.GetCredentials();
    
    // expire the credentials
    _mockProvider.CorrectedUtcNow = _baseTimeUtc + TimeSpan.FromMinutes(51);
    var creds = mockCredentials.GetCredentials();

    // this should be treated as expired right? since 51min > (60min - 10min buffer)
    Assert.NotEqual(initialCreds, creds);
```

sdk/src/Core/Amazon.Runtime/Credentials/RefreshingAWSCredentials.cs
normj: My thinking was I need the lock to control the number of threads performing the refresh action or when I needed to change both `currentLoadState` and `currentState` atomically. In this case I only need to change the  `currentLoadState` property and have marked that field is `volatile` to avoid compiler optimizations that would cause problems when multiple threads read and set variable.

sdk/test/NetStandard/UnitTests/Core/Credentials/RefreshingAWSCredentialsTests.cs
```diff
@@ -0,0 +1,226 @@
...
+            var mockCredentials = new MockRefreshingAWSCredentials(_lifetime, _mockProvider)
+            {
+                PreemptExpiryTime = TimeSpan.FromMinutes(15),
+            };
+
+            _mockProvider.CorrectedUtcNow = _baseTimeUtc;
+            var initialCreds = mockCredentials.GetCredentials();
+
+            _mockProvider.CorrectedUtcNow = _baseTimeUtc + TimeSpan.FromMinutes(instantInMinutes);
+            var previousState = mockCredentials.CurrentState;
```
peterrsongg, line 140: Should we also have a version of this similar to `ConcurrentCallsToGetCredentialsOnlyGeneratesNewCredentialsOnceAsync` that validates we only do one background refresh for credentials when there are multiple threads that are in the preEmpt window of 15 minutes?

sdk/src/Core/Amazon.Runtime/Credentials/RefreshingAWSCredentials.cs
dscpinheiro: Yes, it adds a line break between the two sentences in intellisense.

sdk/test/NetStandard/UnitTests/Core/Credentials/RefreshingAWSCredentialsTests.cs
dscpinheiro: Yeah, I can add a test case for a moment in the expiry window.

sdk/src/Core/Amazon.Runtime/Credentials/RefreshingAWSCredentials.cs
dscpinheiro: I think this is the same point Bo mentioned here: https://github.com/aws/aws-sdk-net/pull/4100#discussion_r2495929147

sdk/src/Core/Amazon.Runtime/Credentials/RefreshingAWSCredentials.cs
dscpinheiro: It's loading the credentials 🙃

I can add some comments here.

sdk/test/NetStandard/UnitTests/Core/Credentials/RefreshingAWSCredentialsTests.cs
dscpinheiro: I can add a test case for that too. It's hard-coded to 1 minute (I even discussed with Norm if the other credential providers should set a different value, but it should be enough to both complete whatever the current operation is and not be expired by the time this method completes).

normj: To be clear it is defaulted to 1 minute, subclasses could choose to change the value. I made the property protected instead of public because I wasn't sure if making it public would make the class to complicated for users in understanding the difference between ExpirationBuffer and PreemptiveTime.

dscpinheiro: @normj Oh, I actually changed it to public so I could use it in the tests...

Do you think I should revert it back to protected? It would show up for the subclasses so I can see why it could be confusing.

normj: I'm okay with it being public. I just double check the docs and I think they are clear what the 2 properties are used for.

peterrsongg (approved):

boblodgett (approved):

dscpinheiro:
Ran two new dry-runs (`DRY_RUN-720f3877-d44d-496f-b871-6e5de7d283f7` for SDK and `DRY_RUN-c471ec94-24c8-4b24-a84b-468ac426d1b0` for PowerShell) after rebasing and they succeeded.

PR #4090, opened 2025-10-31
Generate metrics configuration s3 operations

peterrsongg:
## Description

Generates 4 S3 operations.

Best way to review is by pulling up the custom version of the file on one side and the generated version of the file on the other side and compare. Although, now that we have fuzz testing I feel much more confident about the results.

## Testing

Dry run passes: DRY_RUN-dc08f946-112a-4d5a-b85a-f33695a422d6
I've reviewed the fuzz-tests results. nothing backwards incompatible

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project
- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [ ] I have read the **README** document
- [ ] I have added tests to cover my changes
- [x] All new and existing tests passed

## License

- [x] I confirm that this pull request can be released under the Apache 2 license

[issues]: https://github.com/aws/aws-sdk-net/issues
[license]: http://aws.amazon.com/apache2.0/
[cla]: http://en.wikipedia.org/wiki/Contributor_License_Agreement

PR #4101, opened 2025-11-05
Generate PutBucketMetricsConfiguration and add required check for query string parameters

peterrsongg:
…equired query string params are set. Remove visitor pattern and move custom logic for putlifecycleconfiguration marshaller to custom partial method

## Description

This PR generates the S3 operation PutBucketMetricsConfiguration. It also adds a check to see if a required query string member is set in the marshallers, which also affects other services as well. This is necessary so that a different operation isn't called if a required query string parameter isn't set. 

This also moves some code to partial methods for PutLifecycleConfigurationRequest, since I removed the ImplementsVisitorPattern customization hook since generating PutBucketMetricsConfiguration using that customization hook logic wasn't possible. I decided to just move that logic to a custom partial so we can have an exact 1:1 mapping of that previous logic.

Best way to review is by pulling up the custom version of the file on one side and the generated version of the file on the other side and compare.

## Testing

[https://github.com/aws/aws-sdk-net/pull/4090](https://github.com/aws/aws-sdk-net/pull/4090) dry run passed  (contains 4 generated operations and this branch is branched off of this PR)

fuzz-tests for the larger PR all passed.
See output of larger PR's AI prompt (all passed)

## Types of changes

## Checklist

## License

### DIFF
--- a/generator/ServiceModels/s3/s3.customizations.json
+++ b/generator/ServiceModels/s3/s3.customizations.json
@@ -784,6 +784,11 @@
             "modify" : [
                 {
                     "Id" :{"emitPropertyName" :"MetricsId"}
+                },
+                {
+                    "ExpectedBucketOwner":{
+                        "injectXmlIsSet":["return !String.IsNullOrEmpty(this._expectedBucketOwner);"]
+                    }
                 }
             ]
         },
--- a/sdk/src/Services/S3/Generated/Model/PutBucketMetricsConfigurationRequest.cs
+++ b/sdk/src/Services/S3/Generated/Model/PutBucketMetricsConfigurationRequest.cs
@@ -143,7 +143,7 @@ namespace Amazon.S3.Model
         // Check to see if ExpectedBucketOwner property is set
         internal bool IsSetExpectedBucketOwner()
         {
-            return this._expectedBucketOwner != null;
+            return !String.IsNullOrEmpty(this._expectedBucketOwner);
         }
 
         /// <summary>


sdk/test/Services/S3/UnitTests/Custom/S3AccessPointTests.cs
```diff
@@ -282,7 +282,8 @@ public void AbortMultipartUploadTest()
             var request = new AbortMultipartUploadRequest
             {
                 BucketName = accessPointArn,
-                Key = "foo.txt"
+                Key = "foo.txt",
+                UploadId = "testing"
```
peterrsongg, line 286: since we are validating query string params now we must populate `UploadId`. Also, this test was only validating the resource path and not that the operation worked. If this was actually to be called, it would've turned into a DeleteObjects call.

generator/ServiceClientGeneratorLib/Generators/Marshallers/BaseMarshaller.tt
```diff
@@ -259,6 +259,23 @@ using Amazon.Runtime.Internal.Util;
 		{
 			foreach(var member in operation.RequestQueryStringMembers)
 			{
+                if (member.IsRequired)
+                {
+                    if (member.Shape.IsString)
+                    {
+#>
+            if (string.IsNullOrEmpty(<#=variableName#>.<#=member.PropertyName#>))
+                throw new <#=this.Config.BaseException#>("Request object does not have required field <#=member.PropertyName#> set");
```
peterrsongg, line 268: Throwing the base exception to be consistent with what we do for resource path members

generator/ServiceClientGeneratorLib/Generators/Marshallers/BaseMarshaller.tt
```diff
@@ -259,6 +259,23 @@ using Amazon.Runtime.Internal.Util;
 		{
 			foreach(var member in operation.RequestQueryStringMembers)
 			{
+                if (member.IsRequired)
+                {
+                    if (member.Shape.IsString)
+                    {
+#>
+            if (string.IsNullOrEmpty(<#=variableName#>.<#=member.PropertyName#>))
```
GarrettBeatty, line 267: nit: can you make it consistent with having `{` for the if statements. im having a hard time figuring out where the `            if (<#=variableName#>.<#=member.PropertyName#> == null)` case is applying. if its inside an if or else block i cant tell easily

GarrettBeatty, line 267: is there any reason to not ship this change (query string required field)  as its own thing? in case there is some issue that arises.

peterrsongg: This is the code we will generate and  not the actual logic, so I kept the brackets off to be consistent with other places where we throw exceptions

peterrsongg: hmm, `PutBucketMetricsConfiguration` has a required query string so even this operation uses this logic. I do understand the concern though. We do really need to ship the change for `AbortMPU` though since that is the one where the operation can actually change, so I'd rather just ship it in this PR😅

peterrsongg:

Task Completed

```csharp
```

```csharp
```

```csharp
```

- __MetricsConfiguration__

## SUMMARY:

generator/ServiceClientGeneratorLib/Generators/Marshallers/BaseMarshaller.tt
GarrettBeatty: i  was thinking if its not strictly required, then could just do it separate to make it easier to review and rollback if ever required in the future. will let david/muhammad decide though

dscpinheiro: Who's David? 🤔

I prefer the smaller PRs, they're easier to review and make me more confident we're not shipping anything we shouldn't.

GarrettBeatty: lol woops daniel :)

peterrsongg: gotcha, just to be clear we're talking about shipping it for S3 then all other services separately?

I can make the change but I would have to
1. add a condition in the generator to only do this for S3
2. Have a separate PR that just removes that condition.

If we're okay with that I'll go ahead and do it

generator/ServiceClientGeneratorLib/ServiceModel.cs
```diff
@@ -562,6 +546,30 @@ public List<Operation> S3AllowListOperations
                         new Operation(this, "PutObjectAcl", DocumentRoot[OperationsKey]["PutObjectAcl"]),
                         new Operation(this, "UploadPartCopy", DocumentRoot[OperationsKey]["UploadPartCopy"]),
                         new Operation(this, "ListObjectsV2", DocumentRoot[OperationsKey]["ListObjectsV2"]),
+                        // phase 3 (including those moved from phase 1)
```
muhammad-othman, line 549: Maybe we should move to a skip list instead of an allow list.

generator/ServiceModels/s3/s3.customizations.json
```diff
@@ -778,6 +779,44 @@
...
             ]
+        },
+        "PutBucketMetricsConfigurationRequest":{
+            "modify" : [
+                {
+                    "Id" :{"emitPropertyName" :"MetricsId"}
+                },
+                {
+                    "ExpectedBucketOwner":{
+                        "injectXmlIsSet":["return !String.IsNullOrEmpty(this._expectedBucketOwner);"]
```
muhammad-othman, line 790: I understand that this preserves the previous behavior, still I'm curios if sending an empty string would make the request fail as the bucket owner isn't equal to `""`

muhammad-othman (approved):

generator/ServiceModels/s3/s3.customizations.json
peterrsongg: yeah sending an empty string does fail the request, but the previous behavior was the an empty string wouldn't fail the request because it wouldn't be sent at all and this is not a required property so i decided to leave it

GarrettBeatty (approved):

PR #4102, opened 2025-11-05
Generate GetBucketMetricsConfiguration

peterrsongg:
## Description

PR no.2 of phase 3

Continuation of [this PR](https://github.com/aws/aws-sdk-net/pull/4101)

Generates GetBucketMetricsConfiguration
## Testing

Dry run passed
## Types of changes

## Checklist

## License


peterrsongg:

Task Completed

```csharp
```

```csharp
```

```csharp
```

- __MetricsConfiguration__

## SUMMARY:

sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/GetBucketMetricsConfigurationRequestMarshaller.cs
```diff
@@ -0,0 +1,101 @@
...
+            request.AddSubResource("metrics");
+        
+            if (publicRequest.IsSetExpectedBucketOwner()) 
+            {
+                request.Headers["x-amz-expected-bucket-owner"] = publicRequest.ExpectedBucketOwner;
+            }
+            if (string.IsNullOrEmpty(publicRequest.BucketName))
+                throw new System.ArgumentException("BucketName is a required property and must be set before making this call.", "GetBucketMetricsConfigurationRequest.BucketName");
+            if (string.IsNullOrEmpty(publicRequest.MetricsId))
+                throw new AmazonS3Exception("Request object does not have required field MetricsId set");
```
dscpinheiro, line 70: I think I missed the stand-up where we discussed this, but we're using the service exception here and `System.ArgumentException` above.

We weren't validating the metrics ID before (https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Services/S3/Custom/Model/Internal/MarshallTransformations/GetBucketMetricsConfigurationRequestMarshaller.cs#L49), so would it make sense to use the same exception as missing bucket name?

dscpinheiro: I missed this in the other PRs but it applies to the list and delete operations marshallers as well.

sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/GetBucketMetricsConfigurationResponseUnmarshaller.cs
```diff
@@ -0,0 +1,126 @@
...
+            UnmarshallResult(context,response);
+            
+            PostUnmarshallCustomization(context, response);
+            return response;
+        }        
+
+        private static void UnmarshallResult(XmlUnmarshallerContext context, GetBucketMetricsConfigurationResponse response)
+        {
+            int originalDepth = context.CurrentDepth;
+            int targetDepth = originalDepth + 1;
```
dscpinheiro, line 58: (Forgot to publish this one...)

Do we have an integration test for this operation? The manual code had a section I don't see here (or the updated metrics unmarshaller): https://github.com/aws/aws-sdk-net/blob/eb138204fbefd914b30bb84c9eedf4a228108065/sdk/src/Services/S3/Custom/Model/Internal/MarshallTransformations/GetBucketMetricsConfigurationResponseUnmarshaller.cs#L58-L59

sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/GetBucketMetricsConfigurationRequestMarshaller.cs
peterrsongg: So the reason I made it a service-level exception is because from my testing, when these required parameters aren't set, the SDK will throw `AmazonS3Exception`. So to not make it a breaking change I purposely made it throw `AmazonS3Exception` (this is actually what we do for resource path params too but S3 is just weird and we don't follow that)

sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/GetBucketMetricsConfigurationResponseUnmarshaller.cs
peterrsongg: https://github.com/aws/aws-sdk-net/blob/main/sdk/test/Services/S3/IntegrationTests/StorageInsightsMetricsTests.cs#L65-L107\

yup we do have integ tests for this :)

peterrsongg: just to explain though, the difference stems from the custom code not using the `MetricsConfigurationUnmarshaller` and trying to unmarshall the objects inside `MetricsConfiguration`. so it has to go 2 levels deep. Since here we are calling `MetricsConfigurationUnmarshaller`, we only have to go one level deep to find the `MetricsConfiguration` element. Then once we find that element we just go into the `MetricsConfigurationUnmarshaller` logic

sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/GetBucketMetricsConfigurationRequestMarshaller.cs
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        /// <summary>
+        /// Marshaller the request object to the HTTP request.
+        /// </summary>  
+        /// <param name="publicRequest"></param>
+        /// <returns></returns>
+        public IRequest Marshall(GetBucketMetricsConfigurationRequest publicRequest)
+        {
+            var request = new DefaultRequest(publicRequest, "Amazon.S3");
```
boblodgett, line 58: This changed from "AmazonS3" to "Amazon.S3". Is this valid? I have not revisited what this is used for but can you investigate and respond here if it is used, if the value change matters, and if we have already made this change in other marshallers correctly or not.

sdk/src/Services/S3/Generated/Model/GetBucketMetricsConfigurationRequest.cs
```diff
@@ -71,68 +77,78 @@ namespace Amazon.S3.Model
...
+            set { this._expectedBucketOwner = value; }
         }
 
-        // Check to see if MetricsId property is set
-        internal bool IsSetMetricsId()
+        // Check to see if ExpectedBucketOwner property is set
+        internal bool IsSetExpectedBucketOwner()
         {
-            return this.metricsId != null;
+            return this._expectedBucketOwner != null;
```
boblodgett: This was changed from `return !String.IsNullOrEmpty(this.expectedBucketOwner);`. Please confirm why this is correct.

sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/GetBucketMetricsConfigurationRequestMarshaller.cs
```diff
@@ -0,0 +1,101 @@
...
+            PreMarshallCustomization(request, publicRequest);
+            request.HttpMethod = "GET";
+            request.AddSubResource("metrics");
+        
+            if (publicRequest.IsSetExpectedBucketOwner()) 
+            {
+                request.Headers["x-amz-expected-bucket-owner"] = publicRequest.ExpectedBucketOwner;
```
boblodgett, line 65: Does `S3Transforms.ToStringValue` still need to be called on publicRequest.ExpectedBucketOwner? Why or why not?

sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/GetBucketMetricsConfigurationRequestMarshaller.cs
```diff
@@ -0,0 +1,101 @@
...
+            }
+            if (string.IsNullOrEmpty(publicRequest.BucketName))
+                throw new System.ArgumentException("BucketName is a required property and must be set before making this call.", "GetBucketMetricsConfigurationRequest.BucketName");
+            if (string.IsNullOrEmpty(publicRequest.MetricsId))
+                throw new AmazonS3Exception("Request object does not have required field MetricsId set");
+            
+            if (publicRequest.IsSetMetricsId())
+                request.Parameters.Add("id", StringUtils.FromString(publicRequest.MetricsId));
```
boblodgett, line 73: This used to be added as a `AddSubResource`. Is this change valid? Why or why not?

sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/GetBucketMetricsConfigurationResponseUnmarshaller.cs
```diff
@@ -0,0 +1,126 @@
...
+            if (context.IsEmptyResponse)
+            {
+                return;
+            }
+            while (context.Read())
+            {
+                if (context.IsStartElement || context.IsAttribute)
+                {
+                    if (context.TestExpression("MetricsConfiguration", targetDepth))
```
boblodgett, line 67: It doesn't look like it used to check for this level at all. Were we wrong before or now?

sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/GetBucketMetricsConfigurationRequestMarshaller.cs
peterrsongg: for all previously released PRs, it has been this way. But based on my investigation it is used in exception messages for example
```
            WebServiceResponseEventArgs args = new WebServiceResponseEventArgs
            {
                RequestHeaders = request.Headers,
                Parameters = request.Parameters,
                ServiceName = request.ServiceName,
                Request = request.OriginalRequest,
                Endpoint = request.Endpoint,
                Response = response
            };
```

and in the function SetMetrics
```
        private static void SetMetrics(IRequestContext requestContext)
        {
            requestContext.Metrics.AddProperty(Metric.ServiceName, requestContext.Request.ServiceName);
            requestContext.Metrics.AddProperty(Metric.ServiceEndpoint, requestContext.Request.Endpoint);
            requestContext.Metrics.AddProperty(Metric.MethodName, requestContext.Request.RequestName);
        }
```

Not a breaking change from user perspective, but maybe this affects api metrics?

sdk/src/Services/S3/Generated/Model/GetBucketMetricsConfigurationRequest.cs
peterrsongg: talked offline and updated

sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/GetBucketMetricsConfigurationRequestMarshaller.cs
peterrsongg: nope this change is valid. I explained to daniel in a different pr but will paste it here

> yup, that is intentional. We were incorrectly adding these non static-query parameters to the subresources collection, even though they are modeled as querystring. 99% of the time it is fine, but when there are special characters in any of these query strings (for example a datetime), we wont percent encode it in the URL which is incorrect. Moving it to the Parameters section where it belongs ensures it is percent encoded in the url

sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/GetBucketMetricsConfigurationResponseUnmarshaller.cs
peterrsongg: that is because it looked two levels deep, and directly unmarshalled each member. Now that we are generting this we look 1 level deep then hand it off to the next unmarshaller. There are also integ tests for this operation

sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/GetBucketMetricsConfigurationRequestMarshaller.cs
peterrsongg: the difference is S3Transforms.ToStringValue uses `CultureInfo.InvariantCulture` but other than that it just sets the string.

sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/GetBucketMetricsConfigurationRequestMarshaller.cs
boblodgett: Please check with the metrics team on this one to confirm if we can keep going this way and or if a mix is acceptable. They at least need to be aware there are two values now.

sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/GetBucketMetricsConfigurationRequestMarshaller.cs
boblodgett: This can change the values in some cases. We need to discuss this one to determine if it is always valid to remove the CultureInfo.InvariantCulture.

sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/GetBucketMetricsConfigurationRequestMarshaller.cs
boblodgett: We decided we will go with Amazon.S3 and update the value in custom S3 code as well.

sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/GetBucketMetricsConfigurationRequestMarshaller.cs
boblodgett: This is a string and since this code is now using the normal generator logic which wasn't apparent from the PR it will use CultureInfo.InvariantCulture when appropriate.

PR #4103, opened 2025-11-05
Generate ListBucketMetricsConfiguration

peterrsongg:
## Description

Continuation of https://github.com/aws/aws-sdk-net/pull/4102
## Motivation and Context

Generates ListBucketMetricsConfiguration
## Testing

All testng done in larger PR with all 4 operations which this PR is based off of.
dry run passed
## Types of changes

## Checklist

## License


peterrsongg:

Task Completed

```csharp
```

```csharp
```

```csharp
```

- __MetricsConfiguration__

## SUMMARY:

sdk/test/Services/S3/UnitTests/Custom/S3AccessPointTests.cs
```diff
@@ -282,7 +282,8 @@ public void AbortMultipartUploadTest()
             var request = new AbortMultipartUploadRequest
             {
                 BucketName = accessPointArn,
-                Key = "foo.txt"
+                Key = "foo.txt",
+                UploadId = "testing"
```
dscpinheiro, line 286: Was this supposed to be in this PR?

sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/ListBucketMetricsConfigurationsRequestMarshaller.cs
```diff
@@ -0,0 +1,99 @@
...
+        
+            if (publicRequest.IsSetExpectedBucketOwner()) 
+            {
+                request.Headers["x-amz-expected-bucket-owner"] = publicRequest.ExpectedBucketOwner;
+            }
+            if (string.IsNullOrEmpty(publicRequest.BucketName))
+                throw new System.ArgumentException("BucketName is a required property and must be set before making this call.", "ListBucketMetricsConfigurationsRequest.BucketName");
+            
+            if (publicRequest.IsSetContinuationToken())
+                request.Parameters.Add("continuation-token", StringUtils.FromString(publicRequest.ContinuationToken));
```
dscpinheiro, line 71: Similar to the other PR, we were doing sub resources before: https://github.com/aws/aws-sdk-net/blob/62c261a9ac72493e5a02cdcbf33b3a60a311ff48/sdk/src/Services/S3/Custom/Model/Internal/MarshallTransformations/ListBucketMetricsConfigurationsRequestMarshaller.cs#L51

Is that intentional here too?

peterrsongg: yup, that is intentional. We were incorrectly adding these non static-query parameters to the `subresources` collection, even though they are modeled as querystring. 99% of the time it is fine, but when there are special characters in any of these query strings (for example a datetime), we wont percent encode it in the URL which is incorrect. Moving it to the `Parameters` section where it belongs ensures it is percent encoded in the url

sdk/test/Services/S3/UnitTests/Custom/S3AccessPointTests.cs
peterrsongg: no, that should only be in the very first PR. this change is in both branches so not sure why it is showing up here😅

peterrsongg: okay i fixed it

dscpinheiro (approved):

sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/ListBucketMetricsConfigurationsResponseUnmarshaller.cs
```diff
@@ -12,62 +12,64 @@
...
             while (context.Read())
             {
                 if (context.IsStartElement || context.IsAttribute)
                 {
-                    if (context.TestExpression("ContinuationToken", targetDepth))
+                    if (context.TestExpression("IsTruncated", targetDepth))
                     {
-                        response.Token = StringUnmarshaller.Instance.Unmarshall(context);
-
+                        var unmarshaller = NullableBoolUnmarshaller.Instance;
```
boblodgett, line 71: Isn't t his a runtime breaking change? Before it wouldn't be returned as null even though the type was bool? and now it can. I agree the new behavior is correct now but it is different than it used to be.

sdk/src/Services/S3/Generated/Model/ListBucketMetricsConfigurationsRequest.cs
```diff
@@ -76,71 +84,79 @@ namespace Amazon.S3.Model
...
-            set { this.token = value; }
+            get { return this._continuationToken; }
+            set { this._continuationToken = value; }
         }
 
         // Check to see if ContinuationToken property is set
         internal bool IsSetContinuationToken()
         {
-            return !(string.IsNullOrEmpty(this.ContinuationToken));
+            return this._continuationToken != null;
```
boblodgett: Was: `!(string.IsNullOrEmpty(this.ContinuationToken));` is this right?

sdk/src/Services/S3/Generated/Model/ListBucketMetricsConfigurationsRequest.cs
```diff
@@ -76,71 +84,79 @@ namespace Amazon.S3.Model
...
 
-        /// <summary>
-        /// Checks to see if ExpectedBucketOwner is set.
-        /// </summary>
-        /// <returns>true, if ExpectedBucketOwner property is set.</returns>
+        // Check to see if ExpectedBucketOwner property is set
         internal bool IsSetExpectedBucketOwner()
         {
-            return !String.IsNullOrEmpty(this.expectedBucketOwner);
+            return this._expectedBucketOwner != null;
```
boblodgett: Was `!String.IsNullOrEmpty(this.expectedBucketOwner);` is this right?

sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/ListBucketMetricsConfigurationsRequestMarshaller.cs
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        /// <summary>
+        /// Marshaller the request object to the HTTP request.
+        /// </summary>  
+        /// <param name="publicRequest"></param>
+        /// <returns></returns>
+        public IRequest Marshall(ListBucketMetricsConfigurationsRequest publicRequest)
+        {
+            var request = new DefaultRequest(publicRequest, "Amazon.S3");
```
boblodgett, line 58: Like the other PR this requires a discussion with the metrics team.

sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/ListBucketMetricsConfigurationsResponseUnmarshaller.cs
peterrsongg: wouldn't this still return the same value since the underlying default value is null now? This is what the BoolUnmarshaller does
```
        public static T Unmarshall(JsonUnmarshallerContext context, ref StreamingUtf8JsonReader reader)
        {
            context.Read(ref reader);
            string text = context.ReadText(ref reader);
            if (text == null)
                return default(T);

            return (T)Convert.ChangeType(text, typeof(T), CultureInfo.InvariantCulture);
        }
```

boblodgett: The `BoolUnmarshaller` that it was setup the simple type as:

```
public bool Unmarshall(XmlUnmarshallerContext context)
        {
            return SimpleTypeUnmarshaller<bool>.Unmarshall(context);
        }
```

It isn't a `bool?`, it is a `<bool>` and in that case it will return `False` for the default when the text is null.

The NullableBoolUnmarshaller doesn't even use Default(T). It does;

```
 public bool? Unmarshall(XmlUnmarshallerContext context)
        {
            context.Read();
            string text = context.ReadText();
            if (string.IsNullOrEmpty(text))
            {
                return null;
            }

            return bool.Parse(text);
        }
```

So in this case if the text is null it will return `null`. So this is a behavior change.

peterrsongg: Ah you’re right, would you be okay with just calling this out in the changelog as a breaking change instead of adding a customization for it? I could be convinced either way but given this is fixing something I lean towards the changelog entry

normj: To make sure of my understanding the public interface is already `bool?` but because we were using the non nullable marshaller we would always marshal a value. But the `IsTruncated` element would have to exist in the xml output without a value. Essentially having a `<IsTruncated />`. My understanding is S3 will always returned back the `IsTruncated` element in the out with a value. So user's using S3, not considering S3 compatible services, will not notice any change in behavior. My suggestion is if we have a simple way to add the customization we should use it but I don't want to add any significant complexity to the SDK and generation for this.

peterrsongg: yup and I just confirmed `IsTruncated` will always return a value
```
<?xml version="1.0" encoding="UTF-8"?>
<ListMetricsConfigurationsResult xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
<IsTruncated>false</IsTruncated>
</ListMetricsConfigurationsResult>
```

So actually, from the user's perspective it would be the same

boblodgett: Fair points Then it is likely safe to change it to NullableBoolUnmarshaller.

Issue #3915, opened 2025-07-16
Support interop with JsonNode and JsonElement when working with AWS.Runtime Documents


rhdunn:
### Describe the feature

The `Document` class in AWS.Runtime is a simple JSON model class that is used to e.g. pass the JSON schema and response for AWS bedrock tool use.

.NET natively supports `JsonNode` and `JsonDocument`/`JsonElement` for working with JSON. Currently it is difficult to interoperate with these classes:
- `Document.FromObject` only supports `JsonElement` -- for `JsonNode` you need to do `Document.FromObject(JsonDocument.Parse(node.ToJsonString()).RootElement);
- There is no way to convert from a `Document` to a `JsonNode` or `JsonElement`.

### Use Case

I have a `JsonNode` object that defines the JSON schema (via `GetJsonSchemaAsNode`) that I want to use to specify an AWS Bedrock tool as using (setting the `Json` property of the `ToolSpecification`). I also want to use that JSON schema object to deserialize the JSON back into a .NET class. I'm using `JsonNode` because I need it to work with other parts of the project.

### Proposed Solution

I propose for `Document.FromObject`:
- A `Document.FromObject(JsonDocument doc)` overload that evaluates to `Document.FromObject(doc.RootElement)`;
- A `Document.FromObject(JsonNode node)` overload that evaluates to `Document.FromObject(JsonDocument.Parse(node.ToJsonString()))` or equivalent.

I also propose the following converters to convert to the corresponding .NET types:
- `JsonDocument Document.AsJsonDocument()`;
- `JsonDocument Document.AsJsonElement()`;
- `JsonDocument Document.AsJsonNode()`.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS .NET SDK and/or Package version used

AWS.BedrockRuntime 4.0.1.1

### Targeted .NET Platform

.NET 8

### Operating System and version

Windows 11

normj:
I agree we need to improve the usability but one of the purposes of the Document class, which there are variants of this type across all the AWS SDKs, is agnostic of the underlying wire protocol. Just so happens that today all the usages are JSON but that won't necessary always be the case.

I suspect we should add some extension/utility methods for easy conversion between JSON and the Document but don't want to nest JSON logic inside the Document type.

rhdunn:
[...]

PR #4113, opened 2025-11-07
Support for ChatOptions.ResponseFormat in AWSSDK.Extensions.Bedrock.MEAI

### LINKED ISSUE #3911: Support for `ChatOptions.ResponseFormat` in AWSSDK.Extensions.Bedrock.MEAI
### Describe the feature

Support `ChatOptions.ResponseFormat` in the `AWSSDK.Extensions.Bedrock.MEAI` implementation of `IChatClient`.
Currently, the `ResponseFormat` property is ignored regardless of whether it's set to `Text`, `Json`, or a custom `ChatResponseFormatJson` with a schema. This makes it impossible to request structured responses from models via Bedrock using the standardized interface from `Microsoft.Extensions.AI`.

### Use Case

I'm integrating Bedrock models into a .NET application using the standard `Microsoft.Extensions.AI` abstractions.
When using other providers, setting the `ChatOptions.ResponseFormat` to `Json` or a custom schema ensures the model returns structured data, which is critical for reliably parsing responses in applications. However, when using the `AWSSDK.Extensions.Bedrock.MEAI` package, this setting is silently ignored, making Bedrock a non-drop-in replacement and limiting its usefulness in structured workflows.

Support for `ResponseFormat` would allow for consistent behavior across providers and improve the developer experience when using Amazon Bedrock through the standardized `IChatClient` interface.

### Proposed Solution

_No response_

### Other Information

https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.ai.chatoptions.responseformat

### Acknowledgements

AWSSDK.Extensions.Bedrock.MEAI 4.0.2
AWSSDK.BedrockAgentRuntime 4.0.4.8

### Targeted .NET Platform

.NET 8

Windows 11, AmazonLinux

AlexDaines:
# Add ChatOptions.ResponseFormat support for Bedrock MEAI

## Description
Implements support for `ChatOptions.ResponseFormat` in the AWSSDK.Extensions.Bedrock.MEAI implementation of `IChatClient`. When `ResponseFormat` is set to `Json` or `ForJsonSchema`, the client now uses Bedrock's tool mechanism to enforce structured JSON responses from models.

**Implementation approach:**
- Creates a synthetic tool called "generate_response" with the provided JSON schema
- Forces model to use this tool via `toolChoice`
- Extracts JSON response from the tool use output
- Converts Bedrock `Document` objects to standard JSON

**Why synthetic tool?**: 

Bedrock lacks a native responseFormat API; all AWS SDKs (boto3, Java, now .NET) use tool calling as the official mechanism for structured output—we inject a synthetic tool with the JSON schema to implement ChatOptions.ResponseFormat transparently.

**Key behavior:**
- `ResponseFormat.Json`: Requests JSON with generic object schema
- `ResponseFormat.ForJsonSchema`: Requests JSON conforming to custom schema
- `ResponseFormat.Text`: No changes to request (default behavior)
- Throws `ArgumentException` if `ResponseFormat` is used with user-provided tools (mutual exclusivity)
- Throws `NotSupportedException` for streaming requests (Bedrock limitation)

## Motivation and Context
Closes #3911

Users need consistent behavior when using `IChatClient` across different AI providers. Currently, the Bedrock implementation ignores `ChatOptions.ResponseFormat`, making it impossible to request structured responses through the standardized Microsoft.Extensions.AI interface. This prevents Bedrock from being a drop-in replacement for other providers in structured data workflows.

## Testing

**Dryrun**: 

.NET v4 Build: DRY_RUN-9edf05db-56d5-4398-902c-826d8573804d

- Added 2 core unit tests covering request creation with schemas and response extraction
- Created local sample application demonstrating the feature with real Bedrock API calls
- Verified JSON responses are correctly structured and parsed
- Confirmed error handling for invalid configurations (tools + ResponseFormat, streaming)

**Test coverage:**
- `ResponseFormat_Json_WithSchema_CreatesSyntheticToolWithCorrectSchema`: Validates synthetic tool creation with custom schema
- `ResponseFormat_Json_ModelReturnsToolUse_ExtractsJsonCorrectly`: Validates JSON extraction from tool use responses

## Types of changes

## Checklist

## License

### DIFF
--- a/extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
+++ b/extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
@@ -119,8 +119,6 @@ internal sealed partial class BedrockChatClient : IChatClient
                 // This gives the user clean JSON conforming to their schema
                 result.Contents.Add(new TextContent(structuredContent) { RawRepresentation = response.Output?.Message });
 
-                // Skip normal ContentBlock processing (Text, Image, etc.) since the model's
-                // entire response is contained in the ToolUseBlock we just extracted
                 if (DocumentToDictionary(response.AdditionalModelResponseFields) is { } responseFieldsDict)
                 {
                     result.AdditionalProperties = new(responseFieldsDict);
@@ -866,10 +864,6 @@ internal sealed partial class BedrockChatClient : IChatClient
     }
 
     /// <summary>Creates a <see cref="ToolConfiguration"/> from the specified options.</summary>
-    /// <remarks>
-    /// When ResponseFormat is specified, creates a synthetic tool to enforce structured output.
-    /// This conflicts with user-provided tools as Bedrock only supports a single ToolChoice value.
-    /// </remarks>
     private static ToolConfiguration? CreateToolConfig(ToolConfiguration? toolConfig, ChatOptions? options)
     {
         if (options?.Tools is { Count: > 0 } tools)


extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
```diff
@@ -35,18 +37,21 @@ internal sealed partial class BedrockChatClient : IChatClient
...
     /// <summary>Metadata describing the chat client.</summary>
     private readonly ChatClientMetadata _metadata;
 
-    /// <summary>
-    /// Initializes a new instance of the <see cref="BedrockChatClient"/> class.
-    /// </summary>
-    /// <param name="runtime">The <see cref="IAmazonBedrockRuntime"/> instance to wrap.</param>
-    /// <param name="defaultModelId">Model ID to use as the default when no model ID is specified in a request.</param>
+    /// <summary>Initializes a new instance of the <see cref="BedrockChatClient"/> class.</summary>
     public BedrockChatClient(IAmazonBedrockRuntime runtime, string? defaultModelId)
```
GarrettBeatty, line 56: why are these docs deleted

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
```diff
@@ -182,6 +258,14 @@ public async IAsyncEnumerable<ChatResponseUpdate> GetStreamingResponseAsync(
             throw new ArgumentNullException(nameof(messages));
         }
 
+        // Check if ResponseFormat is set - not supported for streaming yet
+        if (options?.ResponseFormat is ChatResponseFormatJson)
```
GarrettBeatty, line 247: just wondering why this is

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
```diff
@@ -872,6 +1008,96 @@ private static Document ToDocument(JsonElement json)
...
+                // Convert the Document back to JSON string
+                return DocumentToJsonString(toolUse.Input);
+            }
+        }
+
+        return null;
+    }
+
+    /// <summary>Converts a <see cref="Document"/> to a JSON string.</summary>
+    private static string DocumentToJsonString(Document document)
```
GarrettBeatty, line 1070: all of this stuff i would be surprised if it doesnt exist already in a jsonutils or utils file. either way it shouldnt be in this class

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
```diff
@@ -79,7 +84,34 @@ public async Task<ChatResponse> GetResponseAsync(
...
         request.AdditionalModelRequestFields = CreateAdditionalModelRequestFields(request.AdditionalModelRequestFields, options);
 
-        var response = await _runtime.ConverseAsync(request, cancellationToken).ConfigureAwait(false);
+        // Execute the request with proper error handling for ResponseFormat scenarios
+        ConverseResponse response;
+        try
+        {
+            response = await _runtime.ConverseAsync(request, cancellationToken).ConfigureAwait(false);
+        }
+        catch (AmazonBedrockRuntimeException ex) when (options?.ResponseFormat is ChatResponseFormatJson)
```
GarrettBeatty: why are we checking bedrockruntimeexception here but down below i see we are throwing InvalidOperationException when it fails?

extensions/test/BedrockMEAITests/BedrockChatClientTests.cs
```diff
@@ -1,11 +1,111 @@
...
+    public Task<ConverseStreamResponse> ConverseStreamAsync(ConverseStreamRequest request, CancellationToken cancellationToken = default)
+    {
+        CapturedStreamRequest = request;
+        throw new NotImplementedException("Stream testing not implemented in this mock");
+    }
+
+    public void Dispose() { }
+
+    // Unused interface members - all throw NotImplementedException
+    public IBedrockRuntimePaginatorFactory Paginators => throw new NotImplementedException();
```
peterrsongg: So anytime BedrockRuntime adds a new operation, we have to remember to go here and update this mock class so that the operation throws `NotImplementedException` right? I think we should go with a different approach to mocking the service, because if that's the case then this isn't very scalable

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
```diff
@@ -872,6 +1008,96 @@ private static Document ToDocument(JsonElement json)
         return toolConfig;
     }
 
+    /// <summary>Extracts JSON content from the synthetic ResponseFormat tool use, if present.</summary>
+    private static string? ExtractResponseFormatContent(Message? message)
```
peterrsongg: `string?` is redundant here. a string is always nullable

extensions/test/BedrockMEAITests/BedrockChatClientTests.cs
```diff
@@ -1,11 +1,111 @@
...
+using System.Linq;
+using System.Text.Json;
+using System.Threading;
+using System.Threading.Tasks;
 using Xunit;
 
 namespace Amazon.BedrockRuntime;
 
+// Mock implementation to capture requests and control responses
+internal sealed class MockBedrockRuntime : IAmazonBedrockRuntime
```
peterrsongg: So anytime BedrockRuntime adds a new operation, we have to remember to go here and update this mock class so that the operation throws NotImplementedException right? I think we should go with a different approach to mocking the service, because if that's the case then this isn't very scalable.

Also the types of responses you can test here are severely limited. The `ConverseAsync` just returns a default response, so these test cases are only testing the happy path. My suggestion is to mock the httpLayer so that you can test out edge cases and different responses returned by Converse

peterrsongg: Here is my suggestion (sorry it is a lot). This is what we did for fuzz testing
1. Create a pipeline customizer
```
namespace NetSdkMock
{
    public class PipelineCustomizer : IRuntimePipelineCustomizer
    {
        public string UniqueName => "MockPipeline";

        public void Customize(Type type, RuntimePipeline pipeline)
        {
            pipeline.ReplaceHandler<HttpHandler<System.Net.Http.HttpContent>>(new HttpHandler<HttpContent>(new MockHttpRequestMessageFactory(), new object()));
        }
    }

    public class MockHttpRequestMessageFactory : IHttpRequestFactory<HttpContent>
    {
        public IHttpRequest<HttpContent> CreateHttpRequest(Uri requestUri)
        {
            return new MockHttpRequest(new HttpClient(), requestUri, null);
        }

        public void Dispose()
        {
            throw new NotImplementedException();
        }
    }

    public class MockHttpRequest : HttpWebRequestMessage, IHttpRequest<HttpContent>
    {
        private IWebResponseData _webResponseData;

        public MockHttpRequest(HttpClient httpClient, Uri requestUri, IClientConfig config) : base(httpClient, requestUri, config)
        {
        }
        public new IWebResponseData GetResponse()
        {
            return this.GetResponseAsync(CancellationToken.None).Result;
        }

        public new void ConfigureRequest(IRequestContext requestContext)
        {
            _webResponseData = (IWebResponseData)((IAmazonWebServiceRequest)requestContext.OriginalRequest).RequestState["response"];
        }

        public new Task<IWebResponseData> GetResponseAsync(CancellationToken cancellationToken)
        {
            return Task.FromResult(_webResponseData);
        }
    }
}
```
2. Stub the web response data so that you can control the type of responses that you get
```
using Amazon.Runtime.Internal.Transform;
using System;
using System.Linq;
using System.Net;
using System.Text;
using System.Threading.Tasks;

namespace NetSdkMock
{
    public class StubWebResponseData : IWebResponseData
    {
        public StubWebResponseData(string jsonResponse, Dictionary<string, string> headers)
        {
            this.StatusCode = HttpStatusCode.OK;
            this.IsSuccessStatusCode = true;
            JsonResponse = jsonResponse;
            this.Headers = headers;

            _httpResponseBody = new HttpResponseBody(jsonResponse);
        }
        public Dictionary<string, string> Headers { get; set; }

        public string JsonResponse { get; }
        private IHttpResponseBody _httpResponseBody;
        public long ContentLength { get; set; }

        public string ContentType { get; set; }

        public HttpStatusCode StatusCode { get; set; }

        public bool IsSuccessStatusCode { get; set; }

        public IHttpResponseBody ResponseBody
        {
            get
            {
                return _httpResponseBody;
            }
        }

        public string[] GetHeaderNames()
        {
            return Headers.Keys.ToArray();
        }

        public bool IsHeaderPresent(string headerName)
        {
            return this.Headers.ContainsKey(headerName);
        }

        public string GetHeaderValue(string headerName)
        {
            if (this.Headers.ContainsKey(headerName))
                return this.Headers[headerName];
            else
                return null;
        }
    }
    public class HttpResponseBody : IHttpResponseBody
    {
        private readonly string _jsonResponse;
        private Stream stream;
        public HttpResponseBody(string jsonResponse)
        {
            _jsonResponse = jsonResponse;
        }

        public void Dispose()
        {
            stream.Dispose();
        }

        public Stream OpenResponse()
        {
            stream = new MemoryStream(UTF8Encoding.UTF8.GetBytes(_jsonResponse));
            return stream;
        }

        public Task<Stream> OpenResponseAsync()
        {
            return Task.FromResult(OpenResponse());
        }
    }
}

```

Now in your test you can call it as normal and pass in different types of responses:

```
using Amazon.BedrockRuntime;
using Amazon.BedrockRuntime.Model;
using Amazon.Runtime;
using Microsoft.Extensions.AI;
using Moq;
using System.Text.Json;
namespace NetSdkMock
{
    public class MockTests : IClassFixture<ClientFixture>
    {
        private readonly ClientFixture _fixture;
        public MockTests()
        {
            _fixture = new ClientFixture();
        }
        [Fact]
        public async Task Test1()
        {
            var messages = new[] { new ChatMessage(ChatRole.User, "Test") };
            // try to test different schemas too. 
            var schemaJson = """
            {
                "type": "object",
                "properties": {
                    "name": { "type": "string" },
                    "age": { "type": "number" }
                },
                "required": ["name"]
            }
            """;
            var schemaElement = JsonDocument.Parse(schemaJson).RootElement;
            var options = new ChatOptions
            {
                ResponseFormat = ChatResponseFormat.ForJsonSchema(schemaElement,
                    schemaName: "PersonSchema",
                    schemaDescription: "A person object")
            };
            var chatClient = _fixture.BedrockRuntimeClient.AsIChatClient("claude-3");
            ConverseRequest request = new ConverseRequest();
            var interfaceType = typeof(IAmazonWebServiceRequest);
            var requestStatePropertyInfo = interfaceType.GetProperty("RequestState");
            var requestState = (Dictionary<string, object>)requestStatePropertyInfo.GetValue(request);
            //var schemaElement = JsonDocument.Parse(schemaJson).RootElement;
            //  now you can test out all different types of json responses
            var jsonResponse = """
            {
                "name":"Bob",
                "age" : 15
            }
            """;

            ChatOptions options = new ChatOptions();
            options.RawRepresentationFactory = chatClient => request;

            var webResponseData = new StubWebResponseData(jsonResponse, new Dictionary<string, string>());
            // this is where we are injecting the stubbed web response data. 
            requestState["response"] = webResponseData;
            var response = await chatClient.GetResponseAsync(messages, options).ConfigureAwait(false);

        }
    }
    public class ClientFixture: IDisposable
    {
        public ClientFixture()
        {
            RuntimePipelineCustomizerRegistry.Instance.Register(new PipelineCustomizer());
            BedrockRuntimeClient = new AmazonBedrockRuntimeClient();
        }
        public IAmazonBedrockRuntime BedrockRuntimeClient { get; private set; }

        public void Dispose()
        {
            // Cleanup after all tests in this class
            BedrockRuntimeClient.Dispose();
        }
    }
}
```

I spent a few hours creating a test project to make sure this code will work and it does, so I think you can use this to create many more test cases and not just test the happy path.

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
peterrsongg: Agree with garret, it should live in a utils class at the least

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
```diff
@@ -79,7 +84,34 @@ public async Task<ChatResponse> GetResponseAsync(
...
+        try
+        {
+            response = await _runtime.ConverseAsync(request, cancellationToken).ConfigureAwait(false);
+        }
+        catch (AmazonBedrockRuntimeException ex) when (options?.ResponseFormat is ChatResponseFormatJson)
+        {
+            // Check if this is a ToolChoice validation error (model doesn't support it)
+            bool isToolChoiceNotSupported =
+                ex.ErrorCode == "ValidationException" &&
+                (ex.Message.IndexOf("toolChoice", StringComparison.OrdinalIgnoreCase) >= 0 ||
```
peterrsongg: is checking the error message the only way to achieve this? error messages aren't gauranteed to stay the same.

extensions/test/BedrockMEAITests/BedrockChatClientTests.cs
```diff
@@ -31,17 +131,135 @@ public void AsIChatClient_ReturnsInstance(string modelId)
...
+        {
+            await client.GetResponseAsync(messages, options);
+        }
+        catch
+        {
+            // We're testing request creation
+        }
+
+        // Assert
+        var tool = mock.CapturedRequest.ToolConfig.Tools[0];
```
peterrsongg: i'm a bit confused, and maybe it is because i don't understand who is supposed to return the response in the provided schema (bedrock or us), but this test case just seems to be asserting that the tool has the correct schema set on it. Is there no way to test the actual functionality?

peterrsongg (changes requested):
My long comment on testing different json responses returned by the service might not make sense depending on what this feature is supposed to do. So if this option is set, that tells Bedrock to return the response in a certain way? 

I still think we shouldn't have a MockBedrockRuntime that implements IAmazonBedrockRuntime. We will have to update this class every time a new operation is released.

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
dscpinheiro: I haven't looked at the entire PR, but we've had requests in the past to make the `Document` class interop better with JSON.

It's something we should do, but we have to be aware the document type is meant to be agnostic (the service could start returning CBOR tomorrow for example). See this comment from Norm: https://github.com/aws/aws-sdk-net/issues/3915#issuecomment-3119627796

It'd probably make more sense to include this functionality in Core, but now I'm even wondering if it's better to do that first (and separately) from this PR.

extensions/test/BedrockMEAITests/BedrockChatClientTests.cs
```diff
@@ -31,17 +64,1112 @@ public void AsIChatClient_ReturnsInstance(string modelId)
...
+    /// <summary>
+    /// Pipeline customizer that replaces the HTTP handler with a mock implementation
+    /// </summary>
+    private class MockPipelineCustomizer : IRuntimePipelineCustomizer
+    {
+        public string UniqueName => "BedrockMEAIMockPipeline";
+
+        public void Customize(Type type, RuntimePipeline pipeline)
+        {
+#if BCL
```
peterrsongg: We can just use `if #NETFRAMEWORK` and `else`. we don't have to define our own constants anymore
source: 
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives

extensions/test/BedrockMEAITests/BedrockChatClientTests.cs
```diff
@@ -31,17 +64,1112 @@ public void AsIChatClient_ReturnsInstance(string modelId)
...
+        // Arrange
+        var mockRuntime = new Mock<IAmazonBedrockRuntime>();
+        var client = mockRuntime.Object.AsIChatClient("claude-3");
+        var messages = new[] { new ChatMessage(ChatRole.User, "Test") };
+        var options = new ChatOptions { ResponseFormat = ChatResponseFormat.Json };
+
+        // Act & Assert
+        await Assert.ThrowsAsync<NotSupportedException>(async () =>
+        {
+            await foreach (var update in client.GetStreamingResponseAsync(messages, options))
```
peterrsongg: I get what you're trying to do here but i feel like this test in general isn't necessary. The reason I say this is because if we do add support for the streaming case, this test will fail even though we aren't making a backwards incompatible change.

extensions/test/BedrockMEAITests/BedrockChatClientTests.cs
AlexDaines: Great callout, fixed in latest revision.

extensions/test/BedrockMEAITests/BedrockChatClientTests.cs
AlexDaines: Definitely fragile, removed in latest revision

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
```diff
@@ -79,7 +91,29 @@ public async Task<ChatResponse> GetResponseAsync(
...
+        {
+            response = await _runtime.ConverseAsync(request, cancellationToken).ConfigureAwait(false);
+        }
+        // Transforms ValidationException to NotSupportedException when error message indicates model lacks tool use support (required for ResponseFormat).
+        // This detection relies on error message text which may change in future Bedrock API versions.
+        catch (AmazonBedrockRuntimeException ex) when (options?.ResponseFormat is ChatResponseFormatJson)
+        {
+            // Detect unsupported model: ValidationException with specific tool support error messages
+            if (ex.ErrorCode == "ValidationException" &&
+                (ex.Message.IndexOf("toolChoice is not supported by this model", StringComparison.OrdinalIgnoreCase) >= 0 ||
```
GarrettBeatty: can you share what the error messages look like? why do we need the `||`? are there different cases and how have we determined those?

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
```diff
@@ -79,7 +91,29 @@ public async Task<ChatResponse> GetResponseAsync(
...
+                 ex.Message.IndexOf("This model doesn't support tool use", StringComparison.OrdinalIgnoreCase) >= 0))
+            {
+                throw new NotSupportedException(
+                    $"The model '{request.ModelId}' does not support ResponseFormat. " +
```
GarrettBeatty: nit: can we use stringbuilder or something similar and not use +

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
```diff
@@ -872,6 +1004,42 @@ private static Document ToDocument(JsonElement json)
...
+    {
+        if (message?.Content is null)
+        {
+            return null;
+        }
+
+        foreach (var content in message.Content)
+        {
+            if (content.ToolUse is ToolUseBlock toolUse &&
+                toolUse.Name == ResponseFormatToolName &&
```
GarrettBeatty, line 1056: is it ever possible that a user could define their own tool with the same name and conflict with ours? how do other sdks handle it

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
```diff
@@ -872,6 +1004,42 @@ private static Document ToDocument(JsonElement json)
...
 
+    /// <summary>Extracts JSON content from the synthetic ResponseFormat tool use, if present.</summary>
+    private static string? ExtractResponseFormatContent(Message? message)
+    {
+        if (message?.Content is null)
+        {
+            return null;
+        }
+
+        foreach (var content in message.Content)
```
GarrettBeatty, line 1053: can you explain what exactly this is doing? i understand its going through the content and seems to be finding one where the tool name matches our tool name. and somehow that tool's input is the json formatted content? 


why is there multiple `content` and what happens if more than 1 content has the tool? main question is can you explain a bit how the content/what it is and whats in the tool in the comments

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
```diff
@@ -857,6 +939,56 @@ private static Document ToDocument(JsonElement json)
             }
         }
 
+        // Handle ResponseFormat by creating a synthetic tool
+        if (options?.ResponseFormat is ChatResponseFormatJson jsonFormat)
+        {
+            // Check for conflict with user-provided tools
+            if (toolConfig?.Tools?.Count > 0)
+            {
```
GarrettBeatty, line 909: is this a limitation of only being able to have 1 tool in bedrock?

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
GarrettBeatty: as discussed its because by supplying only one tool it forces bedrock to use our json one. please add comment explaining this and also link to other sdk with example

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
```diff
@@ -872,6 +1004,42 @@ private static Document ToDocument(JsonElement json)
...
+        }
+
+        foreach (var content in message.Content)
+        {
+            if (content.ToolUse is ToolUseBlock toolUse &&
+                toolUse.Name == ResponseFormatToolName &&
+                toolUse.Input != default)
+            {
+                // Convert the Document back to JSON string
+                return DocumentToJsonString(toolUse.Input);
```
GarrettBeatty: can we add a test to see what happens if input is not json here

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
```diff
@@ -20,6 +20,7 @@
 using System;
```
GarrettBeatty, line 20: please link repo/code for the test program in this PR

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
```diff
@@ -857,6 +939,56 @@ private static Document ToDocument(JsonElement json)
             }
         }
 
+        // Handle ResponseFormat by creating a synthetic tool
+        if (options?.ResponseFormat is ChatResponseFormatJson jsonFormat)
```
GarrettBeatty: nit: separate some of this code out into smaller functions

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
```diff
@@ -872,6 +1004,42 @@ private static Document ToDocument(JsonElement json)
         return toolConfig;
     }
 
+    /// <summary>Extracts JSON content from the synthetic ResponseFormat tool use, if present.</summary>
+    private static string? ExtractResponseFormatContent(Message? message)
```
GarrettBeatty: i think this function could be

1. extract tool input
2. and then in the caller function after get tool input, have extractjson from tool input function

extensions/test/BedrockMEAITests/BedrockMEAITests.NetFramework.csproj
```diff
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFrameworks>net472</TargetFrameworks>
+    <DefineConstants>$(DefineConstants);BCL</DefineConstants>
```
peterrsongg: we don't need to define this constant now that we aren't using it

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
```diff
@@ -794,84 +865,224 @@ private static Document ToDocument(JsonElement json)
...
+    /// Note: Document is a struct (value type), so circular references are structurally impossible.
+    /// </summary>
+    private static string DocumentToJsonString(Document document)
+    {
+        using var stream = new MemoryStream();
+        using (var writer = new Utf8JsonWriter(stream, new JsonWriterOptions { Indented = false }))
+        {
+            Amazon.Runtime.Documents.Internal.Transform.DocumentMarshaller.Instance.Write(writer, document);
+        }
+        return Encoding.UTF8.GetString(stream.GetBuffer(), 0, (int)stream.Position);
```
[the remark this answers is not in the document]
peterrsongg: hmm while what Copilot is saying is true about `GetBuffer()` returning the entire buffer which may be larger than the actual data written, as long as the stream's position is obtainable than this seems safe. However, just to be defensive, maybe we can do something like


`return (int)stream.Position == (int)stream.Length ? Encoding.UTF8.GetString(stream.GetBuffer(), 0, (int)stream.Position) : Encoding.UTF8.GetString(stream.ToArray());`

but i understand why you did it this way (to avoid the extra byte allocation)

here is what MS docs say
> To obtain only the data in the buffer, use the [ToArray](https://learn.microsoft.com/en-us/dotnet/api/system.io.memorystream.toarray?view=net-10.0) method; however, [ToArray](https://learn.microsoft.com/en-us/dotnet/api/system.io.memorystream.toarray?view=net-10.0) creates a copy of the data in memory.

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
```diff
@@ -91,29 +92,7 @@ public async Task<ChatResponse> GetResponseAsync(
         request.InferenceConfig = CreateInferenceConfiguration(request.InferenceConfig, options);
         request.AdditionalModelRequestFields = CreateAdditionalModelRequestFields(request.AdditionalModelRequestFields, options);
 
-        ConverseResponse response;
```
peterrsongg: wondering why this was removed?

GarrettBeatty: i suggested removing it because if we just throw the exception as-is that already has info about the tool choice not being supported. and it handles the case where service team can change their error message and what not. i didnt really see any point in converint ig to notsupportexception? @peterrsongg

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
```diff
@@ -89,6 +102,50 @@ public async Task<ChatResponse> GetResponseAsync(
             MessageId = Guid.NewGuid().ToString("N"),
         };
```
GarrettBeatty, line 104:
```
public async Task<ChatResponse> GetResponseAsync(
    IEnumerable<ChatMessage> messages, ChatOptions? options = null, CancellationToken cancellationToken = default)
{
    if (messages is null)
    {
        throw new ArgumentNullException(nameof(messages));
    }

    // Validate ResponseFormat usage early
    ValidateResponseFormatUsage(options, isStreaming: false);

    ConverseRequest request = options?.RawRepresentationFactory?.Invoke(this) as ConverseRequest ?? new();
    // ... rest of request setup ...

    ConverseResponse response = await _runtime.ConverseAsync(request, cancellationToken).ConfigureAwait(false);

    // Extract structured response if using ResponseFormat
    if (options?.ResponseFormat is ChatResponseFormatJson)
    {
        return TryExtractStructuredResponse(response, request.ModelId, options);
    }

    // Normal content processing...
    ChatMessage result = CreateChatMessageFromResponse(response, options);
    return CreateChatResponse(result, response);
}

```

i dont want to add more work to this PR but im thinking eventually we need to refactor this file. i was thinking something like this. @peterrsongg. this function is getting really big and we are adding onto it now

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
peterrsongg: ah i see that makes sense.

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
AlexDaines: add some additional XML docs to explain the `content` blocks that bedrock uses, hopefully its more clear now

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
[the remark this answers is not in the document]
AlexDaines: Went with the suggested ToArray. I found a lot of usage of ToArray throughout the SDK, so at the very least we'll have better code consistency.

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
```diff
@@ -794,84 +865,224 @@ private static Document ToDocument(JsonElement json)
         }
     }
 
-    /// <summary>Creates an <see cref="ToolConfiguration"/> from the specified options.</summary>
+    /// <summary>Creates a <see cref="ToolConfiguration"/> from the specified options.</summary>
+    /// <remarks>
```
GarrettBeatty: nit: remove these lines

```
/// <remarks>
    /// When ResponseFormat is specified, creates a synthetic tool to enforce structured output.
    /// This conflicts with user-provided tools as Bedrock only supports a single ToolChoice value.
    /// </remarks>
```

since we have in the other function docs

Issue #3806, opened 2025-05-12
S3 SDK is far slower than S3 CLI


alehechka:
### Describe the bug

At my company, we download and process millions of files per day from S3, so reducing download times is a key priority. Our tech stack for this service is written in .net8 using version `3.7.305.30` of the `AWSSDK.S3` package. A couple of months ago, we noticed through experimentation that the CLI was able to produce much quicker download times than the SDK and decided to write a small class that would execute the S3 CLI to download larger files. 

I ran a small test to compare the S3 CLI to the TransferUtility and got the following averages for download times. Version 1.0.0.542 is the one with the TransferUtility, while the other 4 are using the CLI downloader.
![Image](https://github.com/user-attachments/assets/f04d4582-9c66-4ed5-b10b-084038c7ac93)

I later performed a more precise test on a single file 5.8GB in size, in a bucket in us-east-1, ran on our most common EC2 instance size of c5.xlarge running in us-east-1c AZ. Receiving the following download times:
- From CLI: 39s
- From GetObjectAsync: 147s
- From TransferUtility: 144s

I tried creating an AWS Support case, but the rep was asking for specific S3 RequestIDs, which seemed irrelevant since these download times impact all of our downloads.

Let me know if there's anything obviously wrong with the implementation of each of the SDK downloaders; if not, hopefully, this is an issue that can be resolved within the dotnet SDK. Thanks.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

S3 SDK should have comparable download times to the S3 CLI.

### Current Behavior

S3 SDK is between 4-6 times slower to download large files than the S3 CLI.

### Reproduction Steps

For the S3 CLI, I used a simple `cp` command:
```sh
aws s3 cp s3://<bucket>/<key> . 
```

Code:
```csharp
using System.Diagnostics;
using Amazon.Runtime.CredentialManagement;
using Amazon.S3;
using Amazon.S3.Transfer;

var sharedFile = new SharedCredentialsFile();
sharedFile.TryGetProfile("default", out var profile);
if (!AWSCredentialsFactory.TryGetAWSCredentials(profile, sharedFile, out var credentials))
{
    throw new UnauthorizedAccessException();
}

var s3Client = new AmazonS3Client(credentials);
var transferUtility = new TransferUtility(s3Client);

// TODO: populate these fields with S3 Object target
var bucket = "";
var key = "";

Console.WriteLine($"Starting GetObjectAsync download...");

[...]

PR #4130, opened 2025-11-14
Multi Part Download + OpenStreamWithResponseAsync

GarrettBeatty:
This PR implements multipart downloads for new api OpenStreamWithResponseAsync, which download parts concurrently in the background while buffering them in memory to provide sequential ordered access 

`OpenStreamWithResponseAsync` will be a new api that returns metadata to the user about the response and implements multi part download behind the scenes. it looks like this for the user

```
using var response = await _transferUtility.OpenStreamWithResponseAsync(streamRequest);
// response contains metadata about the response and the actual response stream
// can access response.Etag and other metadata attributes now


var buffer = new byte[32 * 1024 * 1024]; // 32MB buffer
int bytesRead;

while ((bytesRead = await response.ResponseStream.ReadAsync(buffer, 0, buffer.Length)) > 0)
{
   // read response stream. notice the user needs to do response.ResponseStream now
}
```

compared to old api (no response metadata and no multipart download)

```

using var s3Stream = await _transferUtility.OpenStreamAsync(streamRequest);
// s3stream is the literal responsestream


var buffer = new byte[32 * 1024 * 1024]; // 32MB buffer
int bytesRead;

while ((bytesRead = await s3stream.ReadAsync(buffer, 0, buffer.Length)) > 0)
{
    // read response stream. notice user just does s3stream.ReadAsync
```

I am open to better naming suggestions than `OpenStreamWithResponseAsync` if people have better suggestions

## Description

Overview of the classes. To review this PR, i recommend reviewing locally and starting with __OpenStreamWithResponseCommand__  and following the flow of logic from there. 

__OpenStreamWithResponseCommand__ - Sets up the download and returns a response with the stream to the user.

__BufferedMultipartStream__ - The Stream object users read from; delegates all the work to other components.

__MultipartDownloadCoordinator__ - Figures out if the file needs multiple downloads and launches them concurrently in the background.

__PartBufferManager__ - Makes sure parts are read in the correct order (1, 2, 3...) even though they download concurrently.

__BufferedPartDataHandler__ - Downloads each part from S3 and stores it in memory using ArrayPool buffers. In future we will have __FilePartDataHandler__ or similar to download to a file instead of a buffer

__StreamPartBuffer__ - Holds the actual bytes for one part and returns the memory to the pool when done.

## Motivation and Context

## Testing
1. Unit Tests
3. Integration Tests
4. Manual Tests using https://github.com/awslabs/aws-crt-s3-benchmarks/pull/103

Trying with 5Gb file using `OpenStreamWithResponseAsync`

```
Total bytes per run: 5,368,709,120

Run:1 Secs:2.363672 Gb/s:18.170740
Run:2 Secs:1.281866 Gb/s:33.505587
Run:3 Secs:1.082785 Gb/s:39.665944
Run:4 Secs:1.057797 Gb/s:40.602941
Run:5 Secs:0.963468 Gb/s:44.578213
Run:6 Secs:1.000351 Gb/s:42.934607
Run:7 Secs:0.979741 Gb/s:43.837787
Run:8 Secs:0.979618 Gb/s:43.843296
Run:9 Secs:0.978436 Gb/s:43.896265
Run:10 Secs:0.934025 Gb/s:45.983449

```

vs old result `OpenStreamAsync`

```
Total bytes per run: 5,368,709,120

Run:1 Secs:53.916500 Gb/s:0.796596
Run:2 Secs:53.691880 Gb/s:0.799929
Run:3 Secs:53.697803 Gb/s:0.799840
Run:4 Secs:53.691046 Gb/s:0.799941
Run:5 Secs:53.690165 Gb/s:0.799954
Run:6 Secs:53.691007 Gb/s:0.799942
Run:7 Secs:53.720662 Gb/s:0.799500
Run:8 Secs:53.690802 Gb/s:0.799945
Run:9 Secs:53.690605 Gb/s:0.799948
Run:10 Secs:53.690745 Gb/s:0.799946
```

6. Dry run - d002a133-4e40-4cb6-a606-b4d80ebb4094 pass

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/Transfer/Internal/IPartBufferManager.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/Internal/IPartBufferManager.cs
@@ -39,6 +39,12 @@ namespace Amazon.S3.Transfer.Internal
         Task WaitForBufferSpaceAsync(CancellationToken cancellationToken);
         
         /// <summary>
+        /// Adds a part data source and signals readers when next expected part arrives.
+        /// </summary>
+        /// <param name="dataSource">The part data source to add.</param>
+        void AddDataSource(IPartDataSource dataSource);
+        
+        /// <summary>
         /// Adds a downloaded part buffer and signals readers when next expected part arrives.
         /// </summary>
         /// <param name="buffer">The downloaded part buffer to add.</param>
--- a/sdk/src/Services/S3/Custom/Transfer/Internal/PartBufferManager.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/Internal/PartBufferManager.cs
@@ -140,7 +140,7 @@ namespace Amazon.S3.Transfer.Internal
 
         // Signals when new parts are added or download completes.
         // Automatically resets after waking one waiting reader.
-        // Signaled by: AddDataSourceAsync when new part added, MarkDownloadComplete when finished.
+        // Signaled by: AddDataSource when new part added, MarkDownloadComplete when finished.
         // Waited on by: ReadFromCurrentPartAsync when expected part not yet available.
         // Example: Reader waits for part 4. When download task adds part 4, it signals
         // this event, immediately waking the reader to proceed with consumption.
@@ -247,8 +247,6 @@ namespace Amazon.S3.Transfer.Internal
         /// Adds a part data source to the dictionary and signals waiting consumers.
         /// </summary>
         /// <param name="dataSource">The <see cref="IPartDataSource"/> to add.</param>
-        /// <param name="cancellationToken">A token to cancel the operation.</param>
-        /// <returns>A task that completes when the data source has been added.</returns>
         /// <exception cref="ArgumentNullException">Thrown when <paramref name="dataSource"/> is null.</exception>
         /// <exception cref="InvalidOperationException">Thrown when attempting to add a duplicate part number.</exception>
         /// <remarks>
@@ -261,7 +259,7 @@ namespace Amazon.S3.Transfer.Internal
         /// - Each signals _partAvailable
         /// - Consumer waiting for part 1 wakes up when part 1 is added
         /// </remarks>
-        public Task AddDataSourceAsync(IPartDataSource dataSource, CancellationToken cancellationToken)
+        public void AddDataSource(IPartDataSource dataSource)
         {
             ThrowIfDisposed();
             
@@ -285,8 +283,6 @@ namespace Amazon.S3.Transfer.Internal
 
             // Signal that a new part is available
             _partAvailable.Set();
-            
-            return Task.CompletedTask;
         }
 
         /// <inheritdoc/>
@@ -299,7 +295,7 @@ namespace Amazon.S3.Transfer.Internal
 
             // Create a BufferedDataSource and add it
             var bufferedSource = new BufferedDataSource(buffer);
-            await AddDataSourceAsync(bufferedSource, cancellationToken).ConfigureAwait(false);
+            AddDataSource(bufferedSource);
         }
 
         /// <inheritdoc/>
--- a/sdk/test/Services/S3/UnitTests/Custom/MultipartDownloadManagerTests.cs
+++ b/sdk/test/Services/S3/UnitTests/Custom/MultipartDownloadManagerTests.cs
@@ -1120,10 +1120,6 @@ namespace AWSSDK.UnitTests
             await coordinator.StartDownloadsAsync(discoveryResult, CancellationToken.None);
             stopwatch.Stop();
             
-            // Assert - Should return immediately for single-part
-            Assert.IsTrue(stopwatch.ElapsedMilliseconds < 100, 
-                $"Single-part download should return immediately, took {stopwatch.ElapsedMilliseconds}ms");
-            
             // DownloadCompletionTask should be completed immediately (no background work)
             Assert.IsTrue(coordinator.DownloadCompletionTask.IsCompleted, 
                 "DownloadCompletionTask should be completed for single-part downloads");
--- a/sdk/test/Services/S3/UnitTests/Custom/PartBufferManagerTests.cs
+++ b/sdk/test/Services/S3/UnitTests/Custom/PartBufferManagerTests.cs
@@ -311,10 +311,10 @@ namespace AWSSDK.UnitTests
 
         #endregion
 
-        #region AddDataSourceAsync Tests
+        #region AddDataSource Tests
 
         [TestMethod]
-        public async Task AddDataSourceAsync_AddsToCollection()
+        public async Task AddDataSource_AddsToCollection()
         {
             // Arrange
             var config = MultipartDownloadTestHelpers.CreateBufferedDownloadConfiguration();
@@ -327,7 +327,7 @@ namespace AWSSDK.UnitTests
                 var dataSource = new BufferedDataSource(partBuffer);
 
                 // Act
-                await manager.AddDataSourceAsync(dataSource, CancellationToken.None);
+                manager.AddDataSource(dataSource);
 
                 // Assert - Should be able to read from part 1
                 byte[] readBuffer = new byte[512];
@@ -342,7 +342,7 @@ namespace AWSSDK.UnitTests
 
         [TestMethod]
         [ExpectedException(typeof(ArgumentNullException))]
-        public async Task AddDataSourceAsync_WithNullDataSource_ThrowsArgumentNullException()
+        public void AddDataSource_WithNullDataSource_ThrowsArgumentNullException()
         {
             // Arrange
             var config = MultipartDownloadTestHelpers.CreateBufferedDownloadConfiguration();
@@ -351,7 +351,7 @@ namespace AWSSDK.UnitTests
             try
             {
                 // Act
-                await manager.AddDataSourceAsync(null, CancellationToken.None);
+                manager.AddDataSource(null);
 
                 // Assert - ExpectedException
             }
@@ -363,7 +363,7 @@ namespace AWSSDK.UnitTests
 
         [TestMethod]
         [ExpectedException(typeof(InvalidOperationException))]
-        public async Task AddDataSourceAsync_WithDuplicatePartNumber_ThrowsInvalidOperationException()
+        public void AddDataSource_WithDuplicatePartNumber_ThrowsInvalidOperationException()
         {
             // Arrange
             var config = MultipartDownloadTestHelpers.CreateBufferedDownloadConfiguration();
@@ -375,7 +375,7 @@ namespace AWSSDK.UnitTests
                 byte[] testBuffer1 = ArrayPool<byte>.Shared.Rent(512);
                 var partBuffer1 = new StreamPartBuffer(1, testBuffer1, 512);
                 var dataSource1 = new BufferedDataSource(partBuffer1);
-                await manager.AddDataSourceAsync(dataSource1, CancellationToken.None);
+                manager.AddDataSource(dataSource1);
 
                 // Try to add duplicate part 1
                 byte[] testBuffer2 = ArrayPool<byte>.Shared.Rent(512);
@@ -383,7 +383,7 @@ namespace AWSSDK.UnitTests
                 var dataSource2 = new BufferedDataSource(partBuffer2);
 
                 // Act
-                await manager.AddDataSourceAsync(dataSource2, CancellationToken.None);
+                manager.AddDataSource(dataSource2);
 
                 // Assert - ExpectedException
             }


sdk/src/Services/S3/Custom/Transfer/Internal/StreamPartBuffer.cs
```diff
@@ -0,0 +1,169 @@
...
+        /// Gets or sets the ArrayPool buffer containing the downloaded part data.
+        /// Ownership is transferred to this StreamPartBuffer and will be returned to pool on disposal.
+        /// </summary>
+        public byte[] ArrayPoolBuffer { get; set; }
+
+        /// <summary>
+        /// Gets or sets the actual length of valid data in the ArrayPool buffer.
+        /// The buffer may be larger than this due to ArrayPool size rounding.
+        /// </summary>
+        public int ActualDataLength { get; set; }
```
GarrettBeatty: todo make this just length and remove property

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedPartDataHandler.cs
```diff
@@ -0,0 +1,165 @@
...
+using System.Threading.Tasks;
+using Amazon.S3.Model;
+
+namespace Amazon.S3.Transfer.Internal
+{
+    /// <summary>
+    /// Buffers downloaded parts in memory using ArrayPool and IPartBufferManager.
+    /// Implements current streaming behavior for multipart downloads.
+    /// </summary>
+    internal class BufferedPartDataHandler : IPartDataHandler
```
GarrettBeatty, line 36: for downloading to files in the future we can implement the IPartDataHandler logic in a `FilePartDataHandler` and just change the implementation of FilePartDataHandler's `ProcessPartAsync`  to write to a file instead of a buffer. This will let use re-use the multi part download coordinator in that implementation

sdk/src/Services/S3/Custom/Transfer/Internal/_async/BaseCommand.async.cs
```diff
@@ -31,58 +31,6 @@ internal abstract partial class BaseCommand<TResponse> where TResponse : class
         /// </summary>
         public abstract Task<TResponse> ExecuteAsync(CancellationToken cancellationToken);
 
-        /// <summary>
-        ///  Waits for all of the tasks to complete or till any task fails or is canceled.
-        /// </summary>        
-        protected static async Task<List<T>> WhenAllOrFirstExceptionAsync<T>(List<Task<T>> pendingTasks, CancellationToken cancellationToken)
```
GarrettBeatty, line 37: moved to TaskHelpers.cs so i can use from a non Command class

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedPartDataHandler.cs
```diff
@@ -0,0 +1,165 @@
...
+                    {
+                        partBuffer = ExpandPartBuffer(partBuffer, totalRead);
+                        bufferSpace = partBuffer.Length - totalRead;
+                    }
+                    
+                    // Read size is minimum of: remaining bytes needed, buffer space available, and configured buffer size
+                    int readSize = Math.Min(Math.Min(remainingBytes, bufferSpace), _config.BufferSize);
+                    
+                    // Read directly into final destination at correct offset
+                    int bytesRead = await response.ResponseStream.ReadAsync(
```
GarrettBeatty, line 147: reading from the http response here is where the current performance bottleneck is. open to suggestions on how to make this faster in future

sdk/src/Services/S3/Custom/Transfer/Internal/_async/OpenStreamCommand.async.cs
```diff
@@ -32,7 +32,6 @@ public override async Task<TransferUtilityOpenStreamResponse> ExecuteAsync(Cance
             var response = await _s3Client.GetObjectAsync(getRequest, cancellationToken)
                 .ConfigureAwait(continueOnCapturedContext: false);
             _responseStream = response.ResponseStream;
-            // TODO map and return response
```
GarrettBeatty, line 35: not needed here since this is old api which isnt returning response object

generator/ServiceClientGeneratorLib/Generators/SourceFiles/AssemblyInfo.cs
```diff
@@ -94,9 +94,10 @@ public override string TransformText()
...
 [assembly: InternalsVisibleTo(""AWSSDK.UnitTests.NetFramework, PublicKey=0024000004800000940000000602000000240000525341310004000001000100db5f59f098d27276c7833875a6263a3cc74ab17ba9a9df0b52aedbe7252745db7274d5271fd79c1f08f668ecfa8eaab5626fa76adc811d3c8fc55859b0d09d3bc0a84eecd0ba891f2b8a2fc55141cdcc37c2053d53491e650a479967c3622762977900eddbf1252ed08a2413f00a28f3a0752a81203f03ccb7f684db373518b4"")]
+[assembly: InternalsVisibleTo(""DynamicProxyGenAssembly2, PublicKey=0024000004800000940000000602000000240000525341310004000001000100c547cac37abd99c8db225ef2f6c8a3602f3b3606cc9891605d02baa56104f4cfc0734aa39b93bf7852f7d9266654753cc297e7d2edfe0bac1cdcf9f717241550e0a7b191195b7667bb4f64bcb8e2121380fd1d9d46ad2d92d2d15605093924cceaf74c4861eff62abf69b9291ed0a340e113be11e6a7d3113e92484cf7045cc7"")]
```
GarrettBeatty, line 97: had to add this so Moq could mock some things https://stackoverflow.com/questions/15405040/what-is-the-dynamicproxygenassembly2-assembly

sdk/test/Services/S3/IntegrationTests/TransferUtilityOpenStreamTests.cs
```diff
@@ -0,0 +1,474 @@
...
+            }
+            
+            Assert.AreEqual(totalSize, totalRead, "Should read expected number of bytes");
+            return result;
+        }
+
+        /// <summary>
+        /// Calculates the MD5 checksum of a byte array.
+        /// </summary>
+        private static string CalculateChecksum(byte[] data)
```
GarrettBeatty, line 463: since we dont support full object checksum for multi part download (since s3 doesnt return it on getobject requests for individual parts) i manually calculate the checksum to verify everything is ok

sdk/src/Services/S3/Custom/Transfer/_async/ITransferUtility.async.cs
```diff
@@ -237,6 +237,37 @@ public partial interface ITransferUtility : IDisposable
         /// <returns>The task object representing the asynchronous operation.</returns>
         Task<Stream> OpenStreamAsync(TransferUtilityOpenStreamRequest request, CancellationToken cancellationToken = default(CancellationToken));
```
GarrettBeatty, line 368: planning on marking OpenStreamAsync as deprecated in the future PR once we get usage out of this one

sdk/src/Services/S3/Custom/Transfer/Internal/BaseCommand.cs
```diff
@@ -38,43 +38,11 @@ internal abstract partial class BaseCommand<TResponse> where TResponse : class
     {
         internal GetObjectRequest ConvertToGetObjectRequest(BaseDownloadRequest request)
         {
-            GetObjectRequest getRequest = new GetObjectRequest()
```
GarrettBeatty, line 41: moved to RequestMapper.cs so i can call from a non command class (multipartdownloadcoordinator)

sdk/src/Services/S3/Custom/Transfer/Internal/_async/OpenStreamWithResponseCommand.async.cs
```diff
@@ -0,0 +1,64 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Amazon.S3.Transfer.Internal
+{
+    internal partial class OpenStreamWithResponseCommand : BaseCommand<TransferUtilityOpenStreamResponse>
+    {
+        public override async Task<TransferUtilityOpenStreamResponse> ExecuteAsync(CancellationToken cancellationToken)
+        {
+            var bufferedStream = BufferedMultipartStream.Create(_s3Client, _request, _config, this.RequestEventHandler);
```
GarrettBeatty, line 46: passing in this.RequestEventHandler as a way to set the user agent string to include OpenStreamWithResponseCommand as the caller (similar what we do with other commands, if you check BaseCommand.cs). passing this in felt ugly but it was the only easy way i could think of to solve things

see https://github.com/aws/aws-sdk-net/pull/4130/files#r2532234671

sdk/src/Services/S3/Custom/Transfer/Internal/BaseCommand.cs
GarrettBeatty, line 56: since we do this.GetType here that will set it as whatevever command name called it, which is why i pass this in from openstreamcommandasync

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
```diff
@@ -0,0 +1,498 @@
...
+                        throw new InvalidOperationException($"ETag mismatch detected for part {partNumber} - object may have been modified during download");
+                    }
+                }
+                finally
+                {
+                    _httpConcurrencySlots.Release();
+                }
+                
+                // Delegate data handling to the handler
+                await _dataHandler.ProcessPartAsync(partNumber, response, cancellationToken).ConfigureAwait(false);
```
GarrettBeatty, line 327: https://github.com/aws/aws-sdk-net/pull/4130/files#r2532203647 see here. right now we only implement for buffering parts to buffer but in future when we add downloading to a file we can implement this interface easily

sdk/src/Services/S3/Custom/Transfer/Internal/IPartBufferManager.cs
```diff
@@ -0,0 +1,79 @@
...
+        
+        /// <summary>
+        /// Releases buffer space to allow new parts to be downloaded.
+        /// </summary>
+        void ReleaseBufferSpace();
+        
+        /// <summary>
+        /// Next expected part number in the sequence.
+        /// </summary>
+        int NextExpectedPartNumber { get; }
```
GarrettBeatty, line 77: im only exposing this publicly to make the unit tests easier to read/write. in reality it could really be a private field. its easier to check that the next expected part number is getting updated in certain testing scenarios (such as when doing ReadAsync and it reads across multiple parts) rather than checking the individual buffers/bytes etc

sdk/src/Services/S3/Custom/Transfer/Internal/_async/OpenStreamWithResponseCommand.async.cs
```diff
@@ -0,0 +1,64 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Amazon.S3.Transfer.Internal
+{
+    internal partial class OpenStreamWithResponseCommand : BaseCommand<TransferUtilityOpenStreamResponse>
```
GarrettBeatty, line 28: theres no progress tracking implemented in this right now because the original openstreamcommand doesnt support it either. i was trying to think if it even made sense for progress tracking (especially for in progress bytes), because it could either be
1. progress tracking when the part is downloaded to the internal buffer
2. progress tracking when the part is actually read by the user (outside of the sdk).

it might only make sense in file based downloads. let me know your thoughts

sdk/src/Services/S3/Custom/Transfer/_async/ITransferUtility.async.cs
```diff
@@ -237,6 +237,37 @@ public partial interface ITransferUtility : IDisposable
...
+        /// 	The name of the bucket.
+        /// </param>
+        /// <param name="key">
+        /// 	The object key.
+        /// </param>
+        /// <param name="cancellationToken">
+        ///     A cancellation token that can be used by other objects or threads to receive notice of cancellation.
+        /// </param>
+        /// <returns>The task object representing the asynchronous operation with response metadata.</returns>
+        Task<TransferUtilityOpenStreamResponse> OpenStreamWithResponseAsync(string bucketName, string key, CancellationToken cancellationToken = default(CancellationToken));
```
GarrettBeatty, line 427: `OpenStreamWithResponse` will be a new api that returns metadata to the user about the response and implements multi part download behind the scenes. it looks like this for the user

```
using var response = await _transferUtility.OpenStreamWithResponseAsync(streamRequest);
// response contains metadata about the response and the actual response stream
// can access response.Etag and other metadata attributes now


var buffer = new byte[32 * 1024 * 1024]; // 32MB buffer
int bytesRead;

while ((bytesRead = await response.ResponseStream.ReadAsync(buffer, 0, buffer.Length)) > 0)
{
   // read response stream. notice the user needs to do response.ResponseStream now
}
```

compared to old api (no response metadata and no multipart download)

```

using var s3Stream = await _transferUtility.OpenStreamAsync(streamRequest);
// s3stream is the literal responsestream


var buffer = new byte[32 * 1024 * 1024]; // 32MB buffer
int bytesRead;

while ((bytesRead = await s3stream.ReadAsync(buffer, 0, buffer.Length)) > 0)
{
    // read response stream. notice user just does s3stream.ReadAsync
```



I am open to better naming suggestions than `OpenStreamWithResponse` if people have better suggestions

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadCoordinator.cs
```diff
@@ -0,0 +1,499 @@
...
+namespace Amazon.S3.Transfer.Internal
+{
+    /// <summary>
+    /// Coordinates multipart downloads using PART or RANGE strategies per SEP spec.
+    /// Handles discovery, concurrent downloads.
+    /// </summary>
+    internal class MultipartDownloadCoordinator : IDownloadCoordinator
+    {
+        private readonly IAmazonS3 _s3Client;
+        private readonly TransferUtilityOpenStreamRequest _request;
```
GarrettBeatty: this should actually be BaseDownloadRequest (so it works with openstream and regular file based download). will update in next revision assuming there is comments

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedPartDataHandler.cs
GarrettBeatty: see https://github.com/aws/aws-sdk-net/pull/4136/files#diff-e78b241e54cda851a8d45d80de6dc06080ac3a418d987079b0627867889d009e for example

sdk/src/Services/S3/Custom/Transfer/Internal/_async/OpenStreamWithResponseCommand.async.cs
GarrettBeatty: talked with @normj offline. we decided for now, not to have progress tracking for the openstream stuff. we will include it for file based downloads though

generator/.DevConfigs/9d07dc1e-d82d-4f94-8700-c7b57f872042.json
```diff
@@ -0,0 +1,11 @@
+{
+  "services": [
+    {
+      "serviceName": "S3",
+      "type": "minor",
+      "changeLogMessages": [
+        "Create OpenStreamWithResponseAsync API and implement multi part download"
```
normj: Call out that this method is added to the TransferUtility class. In general making more helpful for customers to understand and care because this is a pretty big deal. Below is my quick swag but feel free to alter.

```
Created new OpenStreamWithResponseAsync method on the Amazon.S3.Transfer.TransferUtility class. The new operation supports downloading in parallel parts of the S3 object in the background while reading from the stream for improved performance.
```

sdk/src/Services/S3/Custom/Transfer/_async/ITransferUtility.async.cs
```diff
@@ -237,6 +237,37 @@ public partial interface ITransferUtility : IDisposable
         /// <returns>The task object representing the asynchronous operation.</returns>
         Task<Stream> OpenStreamAsync(TransferUtilityOpenStreamRequest request, CancellationToken cancellationToken = default(CancellationToken));
 
+        /// <summary>
```
normj, line 370: Since this is the main API public documentation this should be expanded with details about the buffering and where to adjust things. With this description I wouldn't see this method any different then existing other then I get a response.

sdk/src/Services/S3/Custom/Transfer/_async/ITransferUtility.async.cs
```diff
@@ -237,6 +237,37 @@ public partial interface ITransferUtility : IDisposable
...
+        /// 	The object key.
+        /// </param>
+        /// <param name="cancellationToken">
+        ///     A cancellation token that can be used by other objects or threads to receive notice of cancellation.
+        /// </param>
+        /// <returns>The task object representing the asynchronous operation with response metadata.</returns>
+        Task<TransferUtilityOpenStreamResponse> OpenStreamWithResponseAsync(string bucketName, string key, CancellationToken cancellationToken = default(CancellationToken));
+
+        /// <summary>
+        /// 	Returns a stream to read the contents from Amazon S3 as
```
normj, line 430: Since this is the main API public documentation this should be expanded with details about the buffering and where to adjust things. With this description I wouldn't see this method any different then existing other then I get a response.

sdk/src/Services/S3/Custom/Transfer/_async/TransferUtility.async.cs
```diff
@@ -505,6 +505,53 @@ public partial class TransferUtility : ITransferUtility
             }
         }
 
+        /// <summary>
```
normj: Given my previous comment about how the docs should be expanded you might want to use the `inheritdoc` doc tag to avoid duplicating docs here.

sdk/src/Services/S3/Custom/Transfer/_bcl+netstandard/ITransferUtility.sync.cs
```diff
@@ -248,6 +248,35 @@ public partial interface ITransferUtility
         /// </returns>
         Stream OpenStream(TransferUtilityOpenStreamRequest request);
 
+        /// <summary>
```
normj, line 265: Since this is the main API public documentation this should be expanded with details about the buffering and where to adjust things. With this description I wouldn't see this method any different then existing other then I get a response.

sdk/src/Services/S3/Custom/Transfer/_bcl+netstandard/ITransferUtility.sync.cs
```diff
@@ -248,6 +248,35 @@ public partial interface ITransferUtility
...
+        /// </param>
+        /// <param name="key">
+        /// 	The object key.
+        /// </param>
+        /// <returns>
+        /// 	A response containing the stream and metadata from the specified Amazon S3 bucket and key.
+        /// </returns>
+        TransferUtilityOpenStreamResponse OpenStreamWithResponse(string bucketName, string key);
+
+        /// <summary>
```
normj, line 323: Since this is the main API public documentation this should be expanded with details about the buffering and where to adjust things. With this description I wouldn't see this method any different then existing other then I get a response.

sdk/src/Services/S3/Custom/Transfer/_async/ITransferUtility.async.cs
normj: I would recommend updating the docs of the existing method encouraging users to use the new operations. Talk about the new operations use parallel downloads from S3 and memory buffering to improve performance.

sdk/src/Services/S3/Custom/Transfer/Internal/StreamPartBuffer.cs
```diff
@@ -0,0 +1,153 @@
...
+        protected virtual void Dispose(bool disposing)
+        {
+            if (!_disposed && disposing)
+            {
+                try
+                {
+                    // Return ArrayPool buffer to shared pool
+                    if (ArrayPoolBuffer != null)
+                    {
+                        ArrayPool<byte>.Shared.Return(ArrayPoolBuffer);
```
normj, line 152: From a design of point of view this gets tricky with having this class be the one in returning the buffer when it wasn't the one that rented it. On top of that I think my reading of `BufferedPartDataHandler` shows that class is returning the rented buffer and never disposing of this class. I suspect this whole dispose can be removed.

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedPartDataHandler.cs
```diff
@@ -0,0 +1,165 @@
...
+                    {
+                        // Unexpected EOF
+                        break;
+                    }
+                    
+                    totalRead += bytesRead;
+                }
+
+                // Create ArrayPool-based StreamPartBuffer (no copying!)
+                downloadedPart = StreamPartBuffer.FromArrayPoolBuffer(
```
normj: I think there is some logic issues with the `partBuffer` renting and returning because the `partBuffer` is exposed on the `downloadPart` via the  `ArrayPoolBuffer`. The `downloadPart` instances is returned to be used else where in the flow reading from `ArrayPoolBuffer` but the underlying `partBuffer` byte array is being returned to the pool in the `finally` block.

sdk/src/Services/S3/Custom/Transfer/Internal/PartBufferManager.cs
```diff
@@ -0,0 +1,282 @@
...
+namespace Amazon.S3.Transfer.Internal
+{
+    /// <summary>
+    /// Manages part buffers with ArrayPool lifecycle and concurrency control.
+    /// Uses IPartDataSource for uniform handling of streaming and buffered data.
+    /// </summary>
+    internal class PartBufferManager : IPartBufferManager
+    {
+        private readonly ConcurrentDictionary<int, IPartDataSource> _partDataSources;
+        private readonly SemaphoreSlim _bufferSpaceAvailable;
```
normj, line 139: Using 3 different lock primitives makes this pretty confusing. I'm not saying they aren't needed but this type of complexity should be documented here. For example in recent revamp of the locking in refreshing credentials I gave an explanation in code how the locking was being used to help future readers. https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Core/Amazon.Runtime/Credentials/RefreshingAWSCredentials.cs#L171

sdk/src/Services/S3/Custom/Transfer/Internal/PartBufferManager.cs
```diff
@@ -0,0 +1,282 @@
...
+                }
+                
+                // Wait for a part to become available
+                await Task.Run(() => _partAvailable.WaitOne(), cancellationToken).ConfigureAwait(false);
+            }
+
+            // Get the data source for current part
+            if (!_partDataSources.TryGetValue(currentPartNumber, out var dataSource))
+            {
+                throw new InvalidOperationException($"Part {currentPartNumber} disappeared after availability check");
```
normj: That error message is more of a log message but a user catching the exception won't know what that means especially "disappeared after availability check".

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadCoordinator.cs
```diff
@@ -0,0 +1,499 @@
...
+        private readonly BaseDownloadRequest _request;
+        private readonly DownloadCoordinatorConfiguration _config;
+        private readonly IPartDataHandler _dataHandler;
+        private readonly SemaphoreSlim _httpConcurrencySlots;
+        private readonly RequestEventHandler _requestEventHandler;
+        
+        private Exception _downloadException;
+        private bool _disposed = false;
+        private bool _discoveryCompleted = false;
+        private readonly object _lockObject = new object();
```
normj: This is an internal class only called `BufferedMultipartStream.InitializeAsync`. Only a single thread will go through this class so you don't need to add locking in this class.

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadCoordinator.cs
```diff
@@ -0,0 +1,499 @@
...
+            return (startByte, endByte);
+        }
+
+        private long ExtractTotalSizeFromContentRange(string contentRange)
+        {
+            if (string.IsNullOrEmpty(contentRange))
+                throw new InvalidOperationException("Content-Range header is missing from range request response");
+
+            // Format: "bytes 0-{end}/{total-size}" or "bytes 0-{end}/*"
+            var parts = contentRange.Split('/');
```
normj: Since the comment above says the range could contain a `total-size` or `*` does that mean if `*` is returned the object can not be read, because as I read the code we will throw an exception and give up.

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadCoordinator.cs
```diff
@@ -0,0 +1,499 @@
...
+        private (long startByte, long endByte) CalculatePartRange(int partNumber, long objectSize)
+        {
+            var targetPartSize = _request.IsSetPartSize() ? _request.PartSize : _config.TargetPartSizeBytes;
+            
+            var startByte = (partNumber - 1) * targetPartSize;
+            var endByte = Math.Min(startByte + targetPartSize - 1, objectSize - 1);
+            return (startByte, endByte);
+        }
+
+        private long ExtractTotalSizeFromContentRange(string contentRange)
```
normj: This would be a good method to have unit tests for but given it is private I suspect there are none. Should probably make it `internal` and add unit tests.

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadCoordinator.cs
```diff
@@ -0,0 +1,499 @@
...
+            if (_requestEventHandler != null)
+            {
+                ((Amazon.Runtime.Internal.IAmazonWebServiceRequest)request)
+                    .AddBeforeRequestHandler(_requestEventHandler);
+            }
+            
+            return request;
+        }
+
+        private (long startByte, long endByte) CalculatePartRange(int partNumber, long objectSize)
```
normj: This would be a good method to have unit tests for but given it is private I suspect there are none. Should probably make it `internal` and add unit tests.

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedMultipartStream.cs
```diff
@@ -0,0 +1,215 @@
...
+        }
+
+
+        #region Stream Implementation
+
+        public override bool CanRead => true;
+        public override bool CanSeek => false;
+        public override bool CanWrite => false;
+
+        public override long Length => throw new NotSupportedException("Length not supported for multipart download streams");
```
normj: Length should be available from the `ObjectSize` of the discovery results.

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedMultipartStream.cs
```diff
@@ -0,0 +1,215 @@
...
+        public override long Position 
+        { 
+            get => throw new NotSupportedException("Position not supported for multipart download streams");
```
normj: The `get` should be implemented essentially returning the total number of bytes that have so far been returned from the  public Read methods. The `set` makes sense to not implement. But if we have the `Position` and `Length` users can track their progress.

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedPartDataHandler.cs
```diff
@@ -0,0 +1,165 @@
...
+                while (totalRead < expectedBytes)
+                {
+                    // Calculate how many bytes we still need to read
+                    int remainingBytes = (int)(expectedBytes - totalRead);
+                    int bufferSpace = partBuffer.Length - totalRead;
+                    
+                    // Expand buffer if needed BEFORE reading
+                    if (bufferSpace == 0)
+                    {
+                        partBuffer = ExpandPartBuffer(partBuffer, totalRead);
```
normj: I think I'm missing something which either something is wrong or needs to be commented. Since the partBuffer size is allocated to the `ContentLength` why would it ever need to be expanded?

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedPartDataHandler.cs
```diff
@@ -0,0 +1,165 @@
...
+        {
+            _partBufferManager.ReleaseBufferSpace();
+        }
+        
+        public void OnDownloadComplete(Exception exception)
+        {
+            _partBufferManager.MarkDownloadComplete(exception);
+        }
+        
+        public void Dispose()
```
normj, line 104: Does `IPartDataHandler` need to extend from `IDisposable` if there is nothing to dispose because of ownership?

normj (changes requested):

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedPartDataHandler.cs
GarrettBeatty: i think you are right, it was more of a safety measure. probably i can remove and re-run the tests and see if theres ever a scenario where allocating to ContentLength wouldnt work

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedPartDataHandler.cs
GarrettBeatty: for this PR, no, but in the future, I have utilized this dispose function https://github.com/aws/aws-sdk-net/pull/4136/files?diff=unified#diff-e78b241e54cda851a8d45d80de6dc06080ac3a418d987079b0627867889d009eR116 for file based downloading, for disposing of the `.tmp` file when necessary

normj:
Also try out the new API with the S3 encryption library as the service client.

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedPartDataHandler.cs
GarrettBeatty: ` but the underlying partBuffer byte array is being returned to the pool in the finally block.` - This is not true. The finally block in this file, on line 145, only returns it to the pool itself if the ownership was never transferred (due to some error/or something between time of ownership transfer and finally block)

```
finally
            {
                // Only return partBuffer if ownership wasn't transferred
                if (partBuffer != null)
                    ArrayPool<byte>.Shared.Return(partBuffer);
            }
```

This is the general flow

# Buffer Lifecycle

1. **BufferedPartDataHandler** rents buffer from ArrayPool
2. Downloads data into buffer
3. **Transfers ownership** to StreamPartBuffer, sets local ref to `null`
4. Finally block: `if (partBuffer != null)` → FALSE, does **NOT** return buffer
5. StreamPartBuffer passed to PartBufferManager
6. Wrapped in BufferedDataSource for reading
7. Data read from buffer multiple times (BufferedDataSource → consumers)
8. When part fully consumed: BufferedDataSource.Dispose() called
9. Calls StreamPartBuffer.Dispose()
10. **StreamPartBuffer returns buffer to pool**

sdk/src/Services/S3/Custom/Transfer/Internal/PartBufferManager.cs
GarrettBeatty: i will add comments here

sdk/src/Services/S3/Custom/Transfer/Internal/PartBufferManager.cs
GarrettBeatty: this is still an invalid state though. i will keep the exception but change it to something more generic and have a detailed log message then

GarrettBeatty:
@normj i addressed all of the current comments in https://github.com/aws/aws-sdk-net/pull/4130/commits/0b3ed248ad91f743162053526e2447882eed261b

i still need to test with the encryption client though, i will do that today

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedPartDataHandler.cs
GarrettBeatty: I updated this code to be more clear, by having StreamPartBuffer have a `Create` function which rents from the ArrayPool within that class, and the release also happens in the same class too.

sdk/src/Services/S3/Custom/Transfer/Internal/PartBufferManager.cs
GarrettBeatty: added some comments here (With the help of AI). wasnt sure if some of them were overkill or not. i tried to focus on keeping examples for how the locks are used in real scenarios in order to help readers

generator/.DevConfigs/9d07dc1e-d82d-4f94-8700-c7b57f872042.json
GarrettBeatty: fixed in latest commit

sdk/src/Services/S3/Custom/Transfer/_async/ITransferUtility.async.cs
GarrettBeatty: fixed in latest commit

sdk/src/Services/S3/Custom/Transfer/Internal/StreamPartBuffer.cs
GarrettBeatty: i refactored this in latest commit to have StreamPartBuffer rent from the array pool

sdk/src/Services/S3/Custom/Transfer/Internal/PartBufferManager.cs
GarrettBeatty: fixed in latest commit

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadCoordinator.cs
GarrettBeatty: fixed in latest commit

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadCoordinator.cs
GarrettBeatty: updated code and comments to be more clear. from my assumption, "*" only happens when the total size is not known, (i think thats an RFC standard), but S3 will always know the size (im assuming), so there should never be a * in that case

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadCoordinator.cs
GarrettBeatty: fixed in latest commit, same in other places

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedMultipartStream.cs
GarrettBeatty: fixed in latest commit, same in other place for position

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedPartDataHandler.cs
GarrettBeatty: removed in latest commit

GarrettBeatty:
this is ready for another review. the only thing i havent touched yet is anything encryption client related. will try out java's encryption client today to see if that works for part get

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
```diff
@@ -0,0 +1,483 @@
...
+
+            var downloadTasks = new List<Task>();
+            var internalCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+
+            try
+            {
+                // Process Part 1 from InitialResponse (applies to both single-part and multipart)
+                await _dataHandler.ProcessPartAsync(1, discoveryResult.InitialResponse, cancellationToken).ConfigureAwait(false);
+
+                if (discoveryResult.IsSinglePart)
```
normj, line 153: If we see it is a single part can we rework this so we just hand back the network stream instead of buffering the single part stream and then returning back to the user.

I think this is important to do before we release but okay doing this as a separate PR to the feature branch to stop this PR from growing.

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedMultipartStream.cs
```diff
@@ -0,0 +1,236 @@
...
+        private long _totalBytesRead = 0;
+
+        /// <summary>
+        /// Gets the discovery result containing metadata from the initial GetObject response.
+        /// Available after InitializeAsync completes successfully.
+        /// </summary>
+        public DownloadDiscoveryResult DiscoveryResult => _discoveryResult;
+
+        /// <summary>
+        /// Creates a new BufferedMultipartStream with dependency injection.
```
philasmar: nit: use doc references for `BufferedMultipartStream`

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedMultipartStream.cs
```diff
@@ -0,0 +1,236 @@
...
+        /// <param name="config">Configuration settings for the stream.</param>
+        public BufferedMultipartStream(IDownloadCoordinator downloadCoordinator, IPartBufferManager partBufferManager, BufferedDownloadConfiguration config)
+        {
+            _downloadCoordinator = downloadCoordinator ?? throw new ArgumentNullException(nameof(downloadCoordinator));
+            _partBufferManager = partBufferManager ?? throw new ArgumentNullException(nameof(partBufferManager));
+            _config = config ?? throw new ArgumentNullException(nameof(config));
+        }
+
+        /// <summary>
+        /// Factory method to create BufferedMultipartStream with default dependencies.
```
philasmar: nit: use doc references for BufferedMultipartStream

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedMultipartStream.cs
```diff
@@ -0,0 +1,236 @@
...
+        /// </summary>
+        /// <param name="s3Client">S3 client for making requests.</param>
+        /// <param name="request">Stream request parameters.</param>
+        /// <param name="transferConfig">Transfer utility configuration.</param>
+        /// <param name="requestEventHandler">Optional request event handler for user agent tracking.</param>
+        /// <returns>A new BufferedMultipartStream instance.</returns>
```
philasmar: nit: use doc references for BufferedMultipartStream

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedMultipartStream.cs
```diff
@@ -0,0 +1,236 @@
...
+        public static BufferedMultipartStream Create(IAmazonS3 s3Client, TransferUtilityOpenStreamRequest request, TransferUtilityConfig transferConfig, RequestEventHandler requestEventHandler = null)
+        {
+            if (s3Client == null) throw new ArgumentNullException(nameof(s3Client));
+            if (request == null) throw new ArgumentNullException(nameof(request));
+            if (transferConfig == null) throw new ArgumentNullException(nameof(transferConfig));
+            
+            // Determine target part size from request or use 8MB default
+            long targetPartSize = request.IsSetPartSize() 
+                ? request.PartSize 
+                : 8 * 1024 * 1024; // 8MB default
```
philasmar: Move this to a constants class

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedMultipartStream.cs
```diff
@@ -0,0 +1,236 @@
...
+        }
+
+        /// <summary>
+        /// Initialize the stream by discovering download strategy and setting up appropriate handlers.
+        /// </summary>
+        /// <param name="cancellationToken">Cancellation token for the initialization operation.</param>
+        public async Task InitializeAsync(CancellationToken cancellationToken)
+        {
+            if (_disposed)
+                throw new ObjectDisposedException(nameof(BufferedMultipartStream));
```
philasmar: nit: Add/use the ThrowIfDisposed method (reuse it)

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedMultipartStream.cs
```diff
@@ -0,0 +1,236 @@
...
+                await _downloadCoordinator.StartDownloadsAsync(_discoveryResult, cancellationToken)
+                    .ConfigureAwait(false);
+                
+                _initialized = true;
+            }
+            catch (Exception)
+            {
+                // Clean up on initialization failure
+                throw;
+            }
```
philasmar: Why is this code block there?

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedMultipartStream.cs
```diff
@@ -0,0 +1,236 @@
...
+        }
+
+        public override async Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            if (_disposed)
+                throw new ObjectDisposedException(nameof(BufferedMultipartStream));
```
philasmar: same comment as before

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedMultipartStream.cs
```diff
@@ -0,0 +1,236 @@
...
+            get 
+            {
+                if (!_initialized)
+                    throw new InvalidOperationException("Stream must be initialized before accessing Position");
+                return Interlocked.Read(ref _totalBytesRead);
+            }
+            set => throw new NotSupportedException("Position setter not supported for read-only streams"); 
+        }
+
+        public override void Flush() { }
```
philasmar, line 223: since these dont do anything, should we be throwing an error if they're called?

sdk/src/Services/S3/Custom/Transfer/Internal/PartBufferManager.cs
```diff
@@ -0,0 +1,542 @@
...
+    /// <summary>
+    /// Manages buffered parts for multipart downloads with memory flow control and sequential consumption.
+    /// Implements a producer-consumer pattern where download tasks produce buffered parts and the read stream consumes them in order.
+    /// </summary>
+    /// <remarks>
+    /// This class coordinates concurrent multipart downloads while ensuring sequential reading and bounded memory usage.
+    /// 
+    /// SYNCHRONIZATION PRIMITIVES AND THEIR PURPOSES:
+    /// 
+    /// 1. _lockObject (object lock)
```
normj: You can get rid of the `_lockObject` by 
* Combining `_downloadComplete` and `_downloadException` as a `Tuple<bool, Exception>` so they can be set atomically.
* Mark the new `Tuple` and `_nextExpectedPartNumber` as `volatile`
* When incrementing `_nextExpectedPartNumber` use `Interlocked.Increment(ref _nextExpectedPartNumber);`

This will remove the concern of having to keep getting the `_lockObject` and making sure the lock is always being used correctly. For example at the start of `ReadFromCurrentPartAsync` you are reading `_nextExpectedPartNumber` without a lock.

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedPartDataHandler.cs
```diff
@@ -0,0 +1,141 @@
...
+                var partBuffer = downloadedPart.ArrayPoolBuffer;
+                
+                int totalRead = 0;
+                
+                // Read response stream into buffer based on ContentLength
+                while (totalRead < expectedBytes)
+                {
+                    // Calculate how many bytes we still need to read
+                    int remainingBytes = (int)(expectedBytes - totalRead);
+                    int bufferSpace = partBuffer.Length - totalRead;
```
normj: Is relying on the partBuffer's `Length` safe to do? The `byte[]` allocated for partBuffer via the ArrayPool Rent call which doesn't guarantee exact capacity size. It guarantees it will be at least as big as the request capacity. So if we get something larger then the capacity size which is essentially the part size won't relying on the rented byte [] throw off the counts.

Is the fix to use `downloadedPart.Length`?

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedPartDataHandler.cs
```diff
@@ -0,0 +1,141 @@
...
+using System;
+using System.Buffers;
+using System.Threading;
+using System.Threading.Tasks;
+using Amazon.S3.Model;
+
+namespace Amazon.S3.Transfer.Internal
+{
+    /// <summary>
+    /// Buffers downloaded parts in memory using ArrayPool and IPartBufferManager.
```
philasmar: use doc references for ArrayPool and IPartBufferManager

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedPartDataHandler.cs
```diff
@@ -0,0 +1,141 @@
...
+    /// Implements current streaming behavior for multipart downloads.
+    /// </summary>
+    internal class BufferedPartDataHandler : IPartDataHandler
```
philasmar, line 36: Add documentation for all the public methods

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedPartDataHandler.cs
```diff
@@ -0,0 +1,141 @@
...
+                    // Read directly into final destination at correct offset
+                    int bytesRead = await response.ResponseStream.ReadAsync(
+                        partBuffer,     // Destination: final ArrayPool buffer
+                        totalRead,      // Offset: current position in buffer
+                        readSize,       // Count: optimal chunk size
+                        cancellationToken).ConfigureAwait(false);
+                    
+                    if (bytesRead == 0) 
+                    {
+                        // Unexpected EOF
```
philasmar: Does this situation not warrant an exception?

sdk/src/Services/S3/Custom/Transfer/Internal/DownloadCoordinatorConfiguration.cs
```diff
@@ -0,0 +1,64 @@
...
+ */
+using System;
+
+namespace Amazon.S3.Transfer.Internal
+{
+    /// <summary>
+    /// Base configuration for multipart download coordination.
+    /// Contains settings common to all download strategies.
+    /// </summary>
+    internal class DownloadCoordinatorConfiguration
```
philasmar: not a fan of `DownloadCoordinatorConfiguration`. maybe `DownloadManagerConfiguration`?

sdk/src/Services/S3/Custom/Transfer/Internal/DownloadCoordinatorConfiguration.cs
```diff
@@ -0,0 +1,64 @@
...
+        /// </summary>
+        public long TargetPartSizeBytes { get; set; }
+        
+        /// <summary>
+        /// Protected default constructor for derived classes.
+        /// </summary>
+        protected DownloadCoordinatorConfiguration() { }
+        
+        /// <summary>
+        /// Creates a DownloadCoordinatorConfiguration with the specified configuration values.
```
philasmar: nit: add doc ref to DownloadCoordinatorConfiguration

sdk/src/Services/S3/Custom/Transfer/Internal/IDownloadCoordinator.cs
```diff
@@ -0,0 +1,83 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Amazon.S3.Model;
+
+namespace Amazon.S3.Transfer.Internal
+{
+    /// <summary>
+    /// Coordinates multipart downloads including discovery, concurrent downloads, and progress reporting.
+    /// </summary>
+    internal interface IDownloadCoordinator : IDisposable
```
philasmar: nit: maybe change name to `IDownloadManager`?

sdk/src/Services/S3/Custom/Transfer/Internal/IDownloadManager.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// <summary>
+        /// Exception that occurred during downloads, if any.
+        /// </summary>
+        Exception DownloadException { get; }
+    }
+
+    /// <summary>
+    /// Download discovery results with metadata for determining download strategy.
+    /// </summary>
+    internal class DownloadDiscoveryResult
```
philasmar, line 61: nit: for better organization, could you move this to its own file? Thinking if all these config classes should be placed in a `Models` folder inside the `TransferUtility` folder

sdk/src/Services/S3/Custom/Transfer/Internal/IDownloadCoordinator.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// </summary>
+        public int TotalParts { get; set; }
+
+        /// <summary>
+        /// Total size of the object in bytes.
+        /// </summary>
+        public long ObjectSize { get; set; }
+
+        /// <summary>
+        /// GetObjectResponse from discovery with ResponseStream (complete for single-part, first part for multipart).
```
philasmar: nit: description could be updated without needing to mention `GetObjectResponse` in it

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadCoordinator.cs
```diff
@@ -0,0 +1,483 @@
...
+using Amazon.Runtime;
+using Amazon.S3.Model;
+
+namespace Amazon.S3.Transfer.Internal
+{
+    /// <summary>
+    /// Coordinates multipart downloads using PART or RANGE strategies per SEP spec.
+    /// Handles discovery, concurrent downloads.
+    /// </summary>
+    internal class MultipartDownloadCoordinator : IDownloadCoordinator
```
philasmar: nit: change name to `MultipartDownloadManager`?

philasmar: Add docs to all public methods

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
```diff
@@ -0,0 +1,483 @@
...
+                {
+                    // Single-part: Part 1 is the entire object
+                    _dataHandler.OnDownloadComplete(null);
+                    return;
+                }
+
+                // Multipart: Start concurrent downloads for remaining parts (Part 2 onwards)
+                for (int partNum = 2; partNum <= discoveryResult.TotalParts; partNum++)
```
philasmar, line 165: Do you need to separate the first part like that? Why not do the IsSinglePart check first (and its processing), and if not, start the for loop directly from part 1?

sdk/src/Services/S3/Custom/Transfer/Internal/IPartDataHandler.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// Pairs with WaitForCapacityAsync for backpressure management.
+        /// </summary>
+        void ReleaseCapacity();
+        
+        /// <summary>
+        /// Called when all downloads complete (successfully or with error).
+        /// Allows implementations to perform cleanup or commit operations.
+        /// </summary>
+        /// <param name="exception">Exception if download failed, null if successful</param>
+        void OnDownloadComplete(Exception exception);
```
philasmar, line 68: make the exception optional. No need to explicitly pass `null` when the download is successful

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
```diff
@@ -0,0 +1,483 @@
...
+                // Note: expectedTaskCount + 1 accounts for Part 1 being buffered during discovery
+                if (expectedTaskCount + 1 != discoveryResult.TotalParts)
+                {
+                    throw new InvalidOperationException(
+                        $"Request count mismatch. Expected {discoveryResult.TotalParts} parts, " +
+                        $"but sent {expectedTaskCount + 1} requests");
+                }
+
+                // Mark successful completion
+                _dataHandler.OnDownloadComplete(null);
```
philasmar: not a fan of needing to explicitly pass `null`. Exception should be optional

sdk/src/Services/S3/Custom/Transfer/Internal/OpenStreamWithResponseCommand.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.IO;
+using System.Text;
+
+using Amazon.S3;
+using Amazon.S3.Model;
+
+namespace Amazon.S3.Transfer.Internal
+{
+    /// <summary>
+    /// Enhanced OpenStream command that uses BufferedMultipartStream for improved multipart download handling.
```
philasmar, line 33: add doc refs

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
GarrettBeatty: Im going to create a backlog for this. This needs to apply to both the first part _and_ and other parts where applicable. For example, if the user sets concurrent downloads to 1, we would expect parts 1, 2, 3, etc all to be read in serial from the network stream directly. so this will require changes in multiple places to suppport. 

i have made DOTNET-8400

sdk/src/Services/S3/Custom/Transfer/Internal/PartBufferManager.cs
GarrettBeatty: im trying to remember why i actually used the lock for the next expected part number. i dont think its actually needed for it and it can just be a regular int. 


The tuple though must definitely be volatile

sdk/src/Services/S3/Custom/Transfer/Internal/RequestMapper.cs
```diff
@@ -0,0 +1,92 @@
...
+namespace Amazon.S3.Transfer.Internal
+{
+    /// <summary>
+    /// Utility class for mapping Transfer Utility request objects to S3 request objects.
+    /// Centralizes request creation logic to ensure consistency across different commands.
+    /// </summary>
+    internal static class RequestMapper
+    {
+        /// <summary>
+        /// Maps a BaseDownloadRequest to GetObjectRequest.
```
philasmar, line 35: nit: add doc refs

sdk/src/Services/S3/Custom/Transfer/Internal/StreamPartBuffer.cs
```diff
@@ -0,0 +1,175 @@
...
+ *  AWS SDK for .NET
+ *  API Version: 2006-03-01
+ *
+ */
+using System;
+using System.Buffers;
+using System.Diagnostics.CodeAnalysis;
+using Amazon.S3.Model;
+
+namespace Amazon.S3.Transfer.Internal
```
philasmar, line 27: add doc refs to all referenced classes

philasmar (changes requested):

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
GarrettBeatty: i didnt get what you mean by this. can you explain your example more?

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedMultipartStream.cs
GarrettBeatty: i think flush is needed. its just in our case its a no-op.

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
GarrettBeatty: i think its safer personally to keep explicit null because it prevents developers from accidentally not passing the exception and marking the download as complete when in reality they just forgot to pass the exception

sdk/src/Services/S3/Custom/Transfer/Internal/IDownloadManager.cs
GarrettBeatty: it makes sense to keep this here to me because its directly related to the interface functions. also i think the things in Models folder is for public things, so we shouldnt put it there

GarrettBeatty:
@normj - addressed your comments here. i made one commit for your comments, you can check. 


----

@philasmar also addressed your comments, made multiple commits for those

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedPartDataHandler.cs
GarrettBeatty: updated to have exception and unit tests

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedMultipartStream.cs
GarrettBeatty: fixed to use s3constants

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedPartDataHandler.cs
GarrettBeatty: i updated this logic to just be simpler

```
  // Calculate how many bytes we still need to read
int remainingBytes = (int)(expectedBytes - totalRead);

// Read in chunks up to BufferSize, but never exceed remaining bytes
int readSize = Math.Min(remainingBytes, _config.BufferSize);
```

sdk/src/Services/S3/Custom/Transfer/Internal/PartBufferManager.cs
GarrettBeatty: i ended up removing the lock/volatile logic for `_nextExpectedPartNumber`. its only ever being updated by the consumer which should be on a single thread.

I made updates to use the tuple for the exception and download complete

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedPartDataHandler.cs
GarrettBeatty: i actually found that there is this function https://github.com/aws/aws-sdk-net/pull/4136/files#diff-7d76ec6d3c33088a35cba25bbc63006459be9754c861936319892b11e3072652R921 while working on file based downloads. potentially we can even update this to be something like


```
// Create a MemoryStream wrapper around the pooled buffer
// writable: true allows WriteResponseStreamAsync to write to it
// The MemoryStream starts at position 0 and can grow up to initialBufferSize
using (var memoryStream = new MemoryStream(partBuffer, 0, initialBufferSize, writable: true))
{
    Logger.DebugFormat("BufferedPartDataHandler: [Part {0}] Reading response stream into buffer",
        partNumber);

    // Use GetObjectResponse's stream copy logic which includes:
    // - Progress tracking with events
    // - Size validation (ContentLength vs bytes read)
    // - Buffered reading with proper chunk sizes
    await response.WriteResponseStreamAsync(
        memoryStream,
        null, // destination identifier (not needed for memory stream)
        _config.BufferSize,
        cancellationToken,
        validateSize: true)
        .ConfigureAwait(false);
    
    int totalRead = (int)memoryStream.Position;
    
    Logger.DebugFormat("BufferedPartDataHandler: [Part {0}] Read {1} bytes from response stream",
        partNumber, totalRead);

    // Set the length to reflect actual bytes read
    downloadedPart.SetLength(totalRead);

    if (totalRead != expectedBytes)
    {
        Logger.Error(null, "BufferedPartDataHandler: [Part {0}] Size mismatch - Expected {1} bytes, read {2} bytes",
            partNumber, expectedBytes, totalRead);
    }
}
```


in the future?

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
```diff
@@ -0,0 +1,557 @@
...
+                    downloadTasks.Add(task);
+                }
+
+                // Store count before WhenAllOrFirstException (which modifies the list internally)
+                var expectedTaskCount = downloadTasks.Count;
+
+                Logger.DebugFormat("MultipartDownloadManager: Waiting for {0} download tasks to complete", expectedTaskCount);
+
+                // Wait for all downloads to complete (fails fast on first exception)
+                await TaskHelpers.WhenAllOrFirstExceptionAsync(downloadTasks, cancellationToken).ConfigureAwait(false);
```
GarrettBeatty: apparently there is a bug here. i was doing performance testing with a 30Gb file.  theres a deadlock situation here where if the buffer for the parts is full, it blocks the the consumer from reading, so we need to start this as a background task so we can return immediately and allow the consumer to read. working on a fix for it still

GarrettBeatty: Fixed in 0f82dfe5c0d28cfd19ae3388c9a92be7d26336dc @normj FYI wanted to make sure you saw this change too

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedMultipartStream.cs
normj: We should not throw an exception. It is okay to be a no-op.

sdk/src/Services/S3/Custom/Transfer/Internal/PartBufferManager.cs
```diff
@@ -0,0 +1,617 @@
...
+        /// This method is thread-safe and can be called concurrently by multiple download tasks.
+        /// After adding the part to the dictionary, it signals _partAvailable to wake any consumer
+        /// waiting for this specific part number.
+        /// 
+        /// Example: Download tasks for parts 3, 1, 5 all call this concurrently:
+        /// - Each adds to dictionary with their part number as key
+        /// - Each signals _partAvailable
+        /// - Consumer waiting for part 1 wakes up when part 1 is added
+        /// </remarks>
+        public Task AddDataSourceAsync(IPartDataSource dataSource, CancellationToken cancellationToken)
```
normj: Does this method and `AddBufferAsync` need to be async? I'm guess this is leftover from a previous version and they can be made sync.

sdk/test/Services/S3/UnitTests/Custom/MultipartDownloadManagerTests.cs
```diff
@@ -0,0 +1,1627 @@
...
+                InitialResponse = new GetObjectResponse()
+            };
+            
+            // Act
+            var stopwatch = System.Diagnostics.Stopwatch.StartNew();
+            await coordinator.StartDownloadsAsync(discoveryResult, CancellationToken.None);
+            stopwatch.Stop();
+            
+            // Assert - Should return immediately for single-part
+            Assert.IsTrue(stopwatch.ElapsedMilliseconds < 100,
```
GarrettBeatty: note to self, remove this line since its flaky

sdk/src/Services/S3/Custom/Transfer/Internal/PartBufferManager.cs
GarrettBeatty: updated to be sync

PR #4137, opened 2025-11-17
chore: add unit tests for ConvertToGetObjectRequest in BaseCommand

philasmar:
## Description

Setting both ResponseExpires and ResponseHeadersOverrides.Expires on GetObjectRequest now gives priority to ResponseExpires. Previously, ResponseHeadersOverrides.Expires was not being used even if it was set.
Add unit tests for ConvertToGetObjectRequest in BaseCommand for Transfer Utility
## Motivation and Context

Setting both ResponseExpires and ResponseHeadersOverrides.Expires on GetObjectRequest now gives priority to ResponseExpires. Previously, ResponseHeadersOverrides.Expires was not being used even if it was set.
Add unit tests for ConvertToGetObjectRequest in BaseCommand for Transfer Utility

## Testing

Ran tests locally
Ran dry run b15f3324-ec7d-4078-86dc-56df20a04599
## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/Model/Internal/MarshallTransformations/GetObjectRequestMarshaller.cs
+++ b/sdk/src/Services/S3/Custom/Model/Internal/MarshallTransformations/GetObjectRequestMarshaller.cs
@@ -96,10 +96,16 @@ namespace Amazon.S3.Model.Internal.MarshallTransformations
                 request.Parameters.Add("response-content-language", S3Transforms.ToStringValue(headerOverrides.ContentLanguage));
             if (headerOverrides.ContentType != null)
                 request.Parameters.Add("response-content-type", S3Transforms.ToStringValue(headerOverrides.ContentType));
+            
             if (getObjectRequest.IsSetResponseExpires())
+            {
                 request.Parameters.Add("response-expires", S3Transforms.ToStringValue(getObjectRequest.ResponseExpires.Value));
+            }
             else if (headerOverrides.Expires != null)
+            {
                 request.Parameters.Add("response-expires", S3Transforms.ToStringValue(headerOverrides.Expires));
+            }
+            
             if (getObjectRequest.IsSetVersionId())
                 request.AddSubResource("versionId", S3Transforms.ToStringValue(getObjectRequest.VersionId));
             if (getObjectRequest.IsSetPartNumber())


sdk/test/Services/S3/UnitTests/Custom/ResponseMapperTests.cs
```diff
@@ -268,6 +268,96 @@ public void MapAbortMultipartUploadsCommand_ConstructAbortMultipartUploadRequest
...
+                ServerSideEncryptionCustomerProvidedKeyMD5 = "md5",
+                ExpectedBucketOwner = "expected-owner",
+                IfMatch = "etag-match",
+                IfNoneMatch = "etag-notmatch",
+                ResponseHeaderOverrides = new ResponseHeaderOverrides { CacheControl = "no-cache" },
+                RequestPayer = RequestPayer.Requester,
+                ChecksumMode = ChecksumMode.ENABLED
+            };
+
+            var cmd = new DownloadCommand(null, tuDownloadRequest);
```
GarrettBeatty: is there a reason we cant use ValidateMappingTransferUtilityAndSdkRequests like in MapUploadRequest_AbortMultipartRequest_AllMappedProperties_WorkCorrectly?

GarrettBeatty: something like 

```
[TestMethod]
public void MapDownloadRequest_GetObjectRequest_AllMappedProperties_WorkCorrectly()
{
    ValidateMappingTransferUtilityAndSdkRequests<TransferUtilityDownloadRequest, GetObjectRequest>(
        new[] { "Conversion", "DownloadRequest", "GetObjectRequest" },
        (sourceRequest) =>
        {
            var downloadCommand = new DownloadCommand(null, sourceRequest);
            return downloadCommand.ConvertToGetObjectRequest(sourceRequest);
        },
        // ... rest of test setup
    );
}

```

sdk/test/Services/S3/UnitTests/Custom/ResponseMapperTests.cs
```diff
@@ -268,6 +268,96 @@ public void MapAbortMultipartUploadsCommand_ConstructAbortMultipartUploadRequest
...
+                IfNoneMatch = "etag-notmatch",
+                ResponseHeaderOverrides = new ResponseHeaderOverrides { CacheControl = "no-cache" },
+                RequestPayer = RequestPayer.Requester,
+                ChecksumMode = ChecksumMode.ENABLED
+            };
+
+            var cmd = new OpenStreamCommand(null, tuOpenStreamRequest);
+            var getRequest = cmd.ConvertToGetObjectRequest(tuOpenStreamRequest);
+
+            Assert.IsNotNull(getRequest, "ConvertToGetObjectRequest returned null");
```
GarrettBeatty: i see in openstreacommand there is ConstructRequest which then calls ConvertToGetObjectRequest. do we want to test that one as well

sdk/test/Services/S3/UnitTests/Custom/ResponseMapperTests.cs
philasmar: Fixed

sdk/test/Services/S3/UnitTests/Custom/ResponseMapperTests.cs
philasmar: As discussed, we will not add tests for that.

sdk/src/Services/S3/Custom/Model/Internal/MarshallTransformations/GetObjectRequestMarshaller.cs
```diff
@@ -98,6 +98,8 @@ public IRequest Marshall(GetObjectRequest getObjectRequest)
                 request.Parameters.Add("response-content-type", S3Transforms.ToStringValue(headerOverrides.ContentType));
             if (getObjectRequest.IsSetResponseExpires())
                 request.Parameters.Add("response-expires", S3Transforms.ToStringValue(getObjectRequest.ResponseExpires.Value));
+            else if (headerOverrides.Expires != null)
```
GarrettBeatty, line 104: ~isnt this still a breaking change, because its a change in behavior? even if responseheaderoverrides.expires was not being used, if user had it set to something it would now be used over the other value and potentially change things.~

~i think no matter what we do here the behavior would end up changing no?~

GarrettBeatty: @normj can you please confirm on this?

GarrettBeatty: actually i misread the code. as long as `getObjectRequest.ResponseExpires` is given priority over the new value, which it is here, then i believe its fine.

sdk/src/Services/S3/Custom/Model/Internal/MarshallTransformations/GetObjectRequestMarshaller.cs
```diff
@@ -98,6 +98,8 @@ public IRequest Marshall(GetObjectRequest getObjectRequest)
...
+                request.Parameters.Add("response-expires", S3Transforms.ToStringValue(headerOverrides.Expires));
```
normj, line 106: If both the request level expires and headers collection expires are set and then both attempted to be added the `Parameters` collection that would trigger an `ArgumentException` for duplicate key.

philasmar: We have an `else if` so only one would be checked. Also, I have a unit test that sets both at the same time and that only reads the object level property.

normj: Can you separate the `if` and `else if` into its own block of code with spacing between the long section if `if` blocks. You can see how I didn't notice the `else if` and somebody later could easily not notice and reordering things around and make a subtle breaking change.

philasmar: Done

PR #4146, opened 2025-11-20
Generate GetBucketIntelligentTieringConfiguration

peterrsongg:
## Description

Generate GetBucketIntelligentTieringConfiguration 
## Testing

Dry run for base branch passed
Fuzz testing passed.
AI Prompt run

NOTE: doing IntelligentTiering validation changed the behavior. Previously, not setting this would call ListBucketIntelligentTieringConfiguration. Will call this out in the changelog.

## Types of changes

## Checklist

## License


peterrsongg:
GetBucketInteligentTiering AI report
## Breaking Changes Analysis Report

I have analyzed all 14 files affected in this PR migration from custom to generated S3 code. Here are the findings:

### ❌ BREAKING CHANGE #1: GetBucketIntelligentTieringConfigurationRequest.cs

__File__: `sdk/src/Services/S3/Generated/Model/GetBucketIntelligentTieringConfigurationRequest.cs`

__Issue__: IsSet method changed for string properties

__Custom (deleted) code__:

```csharp
internal bool IsSetIntelligentTieringId()
{
    return !string.IsNullOrEmpty(this.intelligentTieringId);
}
```

__Generated code__:

```csharp
internal bool IsSetIntelligentTieringId()
{
    return this._intelligentTieringId != null;
}
```

__Impact__: Per instruction #5, for request objects with string properties, changing IsSet from `!string.IsNullOrEmpty` to `!= null` is a __BREAKING CHANGE__. This means an empty string ("") will now be considered "set" when previously it was not.

---

### ⚠️ POTENTIAL ISSUE #1: GetBucketIntelligentTieringConfigurationRequestMarshaller.cs

__Files__:

- Custom: `sdk/src/Services/S3/Custom/Model/Internal/MarshallTransformations/GetBucketIntelligentTieringConfigurationRequestMarshaller.cs`
- Generated: `sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/GetBucketIntelligentTieringConfigurationRequestMarshaller.cs`

__Issue__: Different methods used to add the `id` query parameter

__Custom (deleted) code__:

```csharp
request.AddSubResource("id", getBucketIntelligentTieringConfigurationRequest.IntelligentTieringId);
```

__Generated code__:

```csharp
if (publicRequest.IsSetIntelligentTieringId())
    request.Parameters.Add("id", StringUtils.FromString(publicRequest.IntelligentTieringId));
```

__Analysis__: The custom code used `AddSubResource()` while generated uses `request.Parameters.Add()`. Both should add query parameters, but:

1. Generated code adds conditional check with `IsSetIntelligentTieringId()`
2. Generated code explicitly converts via `StringUtils.FromString()`

__Recommendation__: Verify that `AddSubResource` and `request.Parameters.Add` produce identical query string behavior for the `id` parameter.

---

### ⚠️ POTENTIAL ISSUE #2: TieringUnmarshaller.cs

__File__: `sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/TieringUnmarshaller.cs`

__Issue__: Different unmarshaller used for Days property

__Custom (deleted) code__:

```csharp
tiering.Days = IntUnmarshaller.GetInstance().Unmarshall(context);
```

__Generated code__:

```csharp
var unmarshaller = NullableIntUnmarshaller.Instance;
unmarshalledObject.Days = unmarshaller.Unmarshall(context);
```

__Analysis__: The Tiering model's Days property is defined as `int?` (nullable int). The custom code used `IntUnmarshaller` which returns non-nullable int, while the generated correctly uses `NullableIntUnmarshaller` for nullable int. This is likely a __BUG FIX__ in the generated code rather than a breaking change, as the custom implementation was technically incorrect for a nullable property. However, this changes unmarshalling behavior.

---

### ✅ NO ISSUES FOUND:

The following files have no breaking changes:

1. __GetBucketIntelligentTieringConfigurationResponse.cs__ - IsSet methods use `!= null` for object properties (responses exempt per instructions)
2. __GetBucketIntelligentTieringConfigurationResponseUnmarshaller.cs__ - Functionally equivalent logic
3. __IntelligentTieringConfigurationUnmarshaller.cs__ - Functionally equivalent logic
4. __IntelligentTieringConfiguration.cs__ (model) - Properties are compatible
5. __Tiering.cs__ (model) - Properties are compatible with nullable Days

---

## Summary

__Files Analyzed__: 14 out of 14 (100%)

__Confirmed Breaking Changes__: 1

- IsSet method change for string property in request object

__Potential Issues Requiring Verification__: 2

- Different parameter addition methods in marshaller
- Different unmarshaller type for Days property (likely a fix, not a break)

__Files With No Issues__: 11

__Critical Action Required__: Fix the IsSet method for `IntelligentTieringId` in the generated request class to use `!string.IsNullOrEmpty` instead of `!= null` to maintain backwards compatibility.

sdk/src/Services/S3/Generated/Model/GetBucketIntelligentTieringConfigurationRequest.cs
```diff
@@ -105,29 +123,34 @@ internal bool IsSetBucketName()
...
-		{
-            get { return this.intelligentTieringId; }
-            set { this.intelligentTieringId = value; }
+        {
+            get { return this._intelligentTieringId; }
+            set { this._intelligentTieringId = value; }
         }
 
         // Check to see if IntelligentTieringId property is set
         internal bool IsSetIntelligentTieringId()
```
peterrsongg, line 150: not changing this to `String.ISNullOREmpty` because we check this in the marshaller and will throw an exception if the string is null or empty

sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/TieringUnmarshaller.cs
```diff
@@ -13,74 +13,85 @@
...
+                        var unmarshaller = StringUnmarshaller.Instance;
+                        unmarshalledObject.AccessTier = unmarshaller.Unmarshall(context);
                         continue;
                     }
-                    if (context.TestExpression("AccessTier", targetDepth))
+                    if (context.TestExpression("Days", targetDepth))
                     {
-                        tiering.AccessTier = StringUnmarshaller.GetInstance().Unmarshall(context);
-
+                        var unmarshaller = NullableIntUnmarshaller.Instance;
```
peterrsongg, line 67: this is fine because Days must be specified for the AccessTier when you put the intelligenttiering configuration, meaning that if the `AccessTier` xmlElement is present on the `Get` side then there will always be a non-empty `Days` xml element

PR #4149, opened 2025-11-20
add shapes to S3NeedsCustomUpdate and add devconfig

peterrsongg:
## Description

Adds complex shapes to the S3NeedsCustomUpdate. Since these shapes are excluded from generation (or heavily customized). I'm backfilling those shapes that I missed in previous PRs so that the generator will throw an exception if any of these shapes are updated. This is a temporary solution. The longer term solution will work as a validator or an acknowledgment in the build system so that we don't mess up our preview build failure metrics.

## Testing

Dry run for base branch passed

## Types of changes

## Checklist

## License

### DIFF
--- a/generator/ServiceClientGeneratorLib/GeneratorDriver.cs
+++ b/generator/ServiceClientGeneratorLib/GeneratorDriver.cs
@@ -1719,8 +1719,13 @@ namespace ServiceClientGenerator
             // a dictionary of shape name to number of members that shape has
             var customUpdateShapes = new Dictionary<string, int>
             {
-                {"LifecycleFilter", 5 },
-                {"MetricsFilter", 4 }
+                { "LifecycleFilter", 5 },
+                { "MetricsFilter", 4 },
+                { "AnalyticsFilter", 3 },
+                { "IntelligentTieringFilter", 3 },
+                { "MetricsAndOperator", 3 },
+                { "AnalyticsAndOperator", 2 },
+                { "IntelligentTieringAndOperator", 2}
             };
             if (customUpdateShapes.TryGetValue(shape.Name, out int membersCount))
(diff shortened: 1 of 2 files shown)


generator/ServiceClientGeneratorLib/GeneratorDriver.cs
```diff
@@ -1719,8 +1719,13 @@ private void S3NeedsCustomUpdate(Shape shape)
...
-                {"MetricsFilter", 4 }
+                { "LifecycleFilter", 5 },
+                { "MetricsFilter", 4 },
+                { "AnalyticsFilter", 3 },
+                { "IntelligentTieringFilter", 3 },
+                { "MetricsAndOperator", 3 },
+                { "AnalyticsAndOperator", 2 },
+                { "IntelligentTieringAndOperator", 2}
             };
             if (customUpdateShapes.TryGetValue(shape.Name, out int membersCount))
```
muhammad-othman, line 1730: Why we are only checking the `membersCount`? What if the docs changed or the changed something with the validation and made them more open? 
I know this isn't part of this PR and I'm not sure if I reviewed the first PR and missed it.

peterrsongg: hmm that is a good point, a doc change wouldn't trigger this validation. This is a short-term solution though. Longer term we need a trebuchet validator.

boblodgett: I am ok which this only until the customization flag is no longer automatically set on the feature for the reason you stated where we don't want to have validation failures. Please add a task to add the validator to the S3 project.

muhammad-othman (approved):

PR #4150, opened 2025-11-20
[V4] Update SDK to validate region is a valid host label

dscpinheiro:
## Description
Update the client config to validate the resolved region is a valid host label. The approach I chose is similar to what the JS SDK did: https://github.com/smithy-lang/smithy-typescript/commit/13c5cd95b682c9252e6cf15b771b5decdc860729

## Motivation and Context
Internal ticket: `V2006609487`

## Testing
Dry-runs:
- .NET: `DRY_RUN-60c0d999-fe1c-445f-97a2-c9b9be00e4d3`
- PowerShell: `DRY_RUN-d1df41ad-55df-42f3-891a-3751ac9ae0e2`

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/ClientConfig.cs
+++ b/sdk/src/Core/Amazon.Runtime/ClientConfig.cs
@@ -267,9 +267,11 @@ namespace Amazon.Runtime
             {
                 if (probeForRegionEndpoint)
                 {
-                    RegionEndpoint = GetDefaultRegionEndpoint();
+                    var defaultRegion = GetDefaultRegionEndpoint();
+                    ValidateRegion(defaultRegion?.SystemName);
+
+                    RegionEndpoint = defaultRegion;
                     this.probeForRegionEndpoint = false;
-                    ValidateRegion(this.regionEndpoint?.SystemName);
                 }
                 return this.regionEndpoint;
             }
--- a/sdk/test/UnitTests/Custom/Runtime/ClientConfigTests.cs
+++ b/sdk/test/UnitTests/Custom/Runtime/ClientConfigTests.cs
@@ -85,6 +85,11 @@ namespace AWSSDK.UnitTests
             {
                 Environment.SetEnvironmentVariable("AWS_REGION", region);
                 var config = new TestClientConfig();
+                
+                // Throws as expected here.
+                Assert.ThrowsException<AmazonClientException>(() => config.RegionEndpoint);
+
+                // This verifies a second get (which could happen in another component) also fails.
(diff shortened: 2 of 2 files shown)


sdk/src/Core/Amazon.Runtime/ClientConfig.cs
```diff
@@ -294,6 +295,21 @@ public RegionEndpoint RegionEndpoint
...
+        private static void ValidateRegion(string regionSystemName)
+        {
+            if (string.IsNullOrWhiteSpace(regionSystemName))
+            {
+                return;
+            }
+
+            if (!Internal.Endpoints.StandardLibrary.Fn.IsValidHostLabel(regionSystemName, true))
+            {
+                throw new AmazonClientException(string.Format(CultureInfo.InvariantCulture, "The specified '{0}' region is not a valid hostname component.", regionSystemName));
```
boblodgett, line 314: Will this work with third party systems? I have not traced where this is called from.

dscpinheiro: It's called from the `RegionEndpoint` and `AuthenticationRegion` setters. I did add one test to verify https://docs.localstack.cloud/aws/integrations/aws-sdks/net/#example does not break.

boblodgett (approved):

peterrsongg (approved):

sdk/src/Core/Amazon.Runtime/ClientConfig.cs
```diff
@@ -269,6 +269,7 @@ public RegionEndpoint RegionEndpoint
                 {
                     RegionEndpoint = GetDefaultRegionEndpoint();
                     this.probeForRegionEndpoint = false;
+                    ValidateRegion(this.regionEndpoint?.SystemName);
```
normj: nit: I would suggest calling `ValidateRegion` before setting `RegionEndpoint` or `probeForRegionEndpoint` so that if the `get` is called a second time after the first exception the user still gets the exception.

```csharp
var defaultRegion = GetDefaultRegionEndpoint();
ValidateRegion(defaultRegion?.SystemName);
RegionEndpoint = defaultRegion;
this.probeForRegionEndpoint = false;
```

dscpinheiro: That makes sense, I updated the logic (and the test to verify it) in https://github.com/aws/aws-sdk-net/pull/4150/commits/684e689424e9b1b83ecb3bf5a62fddef0c2abc7c

Will run another dry-run and merge the PR tomorrow.

PR #4151, opened 2025-11-20
add failure policy to download directory

philasmar:
## Description

Add FailurePolicy property to TransferUtilityDownloadDirectoryRequest to allow configuration of failure handling behavior during directory downloads. The default behavior is set to abort on failure. Users can now choose to either abort the entire operation or continue downloading remaining files when a failure occurs.

## Motivation and Context

Add FailurePolicy property to TransferUtilityDownloadDirectoryRequest to allow configuration of failure handling behavior during directory downloads. The default behavior is set to abort on failure. Users can now choose to either abort the entire operation or continue downloading remaining files when a failure occurs.

## Testing

Ran tests locally
Ran dry run 7c7a2b17-f08a-4bb4-b6cf-16480a90580c
## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/Transfer/Internal/AbortOnFailurePolicy.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/Internal/AbortOnFailurePolicy.cs
@@ -61,7 +61,7 @@ namespace Amazon.S3.Transfer.Internal
             catch (Exception ex)
             {
                 // Cancel all pending operations before propagating the exception
-                cancellationTokenSource.Cancel();
+                cancellationTokenSource?.Cancel();
                 
                 onFailure?.Invoke(ex);
                 
--- a/sdk/src/Services/S3/Custom/Transfer/Internal/_bcl+netstandard/DownloadDirectoryCommand.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/Internal/_bcl+netstandard/DownloadDirectoryCommand.cs
@@ -39,11 +39,6 @@ namespace Amazon.S3.Transfer.Internal
             : this(s3Client, request)
         {
             this._config = config;
-            this._errors = request.FailurePolicy == FailurePolicy.ContinueOnFailure ? new ConcurrentBag<Exception>() : null;
-            this._failurePolicy =
-                request.FailurePolicy == FailurePolicy.AbortOnFailure
-                    ? new AbortOnFailurePolicy()
-                    : new ContinueOnFailurePolicy(_errors);
         }
 
         public override async Task<TransferUtilityDownloadDirectoryResponse> ExecuteAsync(CancellationToken cancellationToken)
@@ -122,8 +117,7 @@ namespace Amazon.S3.Transfer.Internal
                                 new ObjectDownloadFailedEventArgs(
                                     this._request, 
                                     downloadRequest, 
-                                    ex,
-                                    internalCts));
+                                    ex));
                         };
                         
                         var isValid = await _failurePolicy.ExecuteAsync(
@@ -164,9 +158,9 @@ namespace Amazon.S3.Transfer.Internal
                 return new TransferUtilityDownloadDirectoryResponse
                 {
                     ObjectsDownloaded = _numberOfFilesDownloaded,
-                    ObjectsFailed = _errors?.Count ?? 0,
-                    Errors = _errors?.ToList(),
-                    Result = (_errors is null || _errors.Count == 0) ?
+                    ObjectsFailed = _errors.Count,
+                    Errors = _errors.ToList(),
+                    Result = _errors.Count == 0 ?
                         DirectoryResult.Success :
                         (_numberOfFilesDownloaded > 0 ?
                             DirectoryResult.PartialSuccess :
--- a/sdk/src/Services/S3/Custom/Transfer/TransferUtilityDownloadDirectoryRequest.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/TransferUtilityDownloadDirectoryRequest.cs
@@ -94,7 +94,7 @@ namespace Amazon.S3.Transfer
         /// <param name="args">The details of the failed object download.</param>
         internal void OnRaiseObjectDownloadFailedEvent(ObjectDownloadFailedEventArgs args)
         {
-            AWSSDKUtils.InvokeInBackground(ObjectDownloadFailedEvent, args, this);
+            ObjectDownloadFailedEvent?.Invoke(this, args);
         }
 
         /// <summary>
@@ -632,17 +632,14 @@ namespace Amazon.S3.Transfer
         /// <param name="directoryRequest">The original <see cref="TransferUtilityDownloadDirectoryRequest"/> that initiated the directory download.</param>
         /// <param name="objectRequest">The <see cref="TransferUtilityDownloadRequest"/> representing the individual object download that failed.</param>
         /// <param name="exception">The <see cref="Exception"/> that caused the object download to fail.</param>
-        /// <param name="cancellationTokenSource"><see cref="CancellationTokenSource"/> associated with the failed object's operation.</param>
         internal ObjectDownloadFailedEventArgs(
             TransferUtilityDownloadDirectoryRequest directoryRequest,
             TransferUtilityDownloadRequest objectRequest,
-            Exception exception,
-            CancellationTokenSource cancellationTokenSource)
+            Exception exception)
         {
             DirectoryRequest = directoryRequest;
             ObjectRequest = objectRequest;
             Exception = exception;
-            CancellationTokenSource = cancellationTokenSource;
         }
 
         /// <summary>
@@ -671,10 +668,5 @@ namespace Amazon.S3.Transfer
         /// on the failure mode.
         /// </value>
         public Exception Exception { get; private set; }
-        
-        /// <summary>
-        /// Gets the <see cref="CancellationTokenSource"/> associated with the failed object's operation, if provided.
-        /// </summary>
-        public CancellationTokenSource CancellationTokenSource { get; private set; }
     }
 }
--- a/sdk/src/Services/S3/Custom/Transfer/TransferUtilityDownloadDirectoryResponse.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/TransferUtilityDownloadDirectoryResponse.cs
@@ -36,7 +36,6 @@ namespace Amazon.S3.Transfer
 
         /// <summary>
         /// The collection of exceptions encountered when downloading individual objects.
-        /// Only populated when the <see cref="FailurePolicy"/> is <see cref="FailurePolicy.ContinueOnFailure"/>.
         /// </summary>
         public IList<Exception> Errors { get; set; }
 
--- a/sdk/src/Services/S3/Custom/Transfer/_bcl+netstandard/TransferUtility.async.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/_bcl+netstandard/TransferUtility.async.cs
@@ -209,7 +209,7 @@ namespace Amazon.S3.Transfer
                 CheckForBlockedArn(request.BucketName, "DownloadDirectory");
                 var command = new DownloadDirectoryCommand(this._s3Client, request, this._config);
                 command.DownloadFilesConcurrently = request.DownloadFilesConcurrently;
-                var response = await command.ExecuteAsync(cancellationToken).ConfigureAwait(false);
+                await command.ExecuteAsync(cancellationToken).ConfigureAwait(false);
             }
         }
         #endregion
--- a/sdk/test/Services/S3/UnitTests/Custom/FailurePolicyTests.cs
+++ b/sdk/test/Services/S3/UnitTests/Custom/FailurePolicyTests.cs
@@ -197,14 +197,14 @@ namespace AWSSDK.UnitTests
 
         [TestMethod]
         [TestCategory("S3")]
-        public async Task DownloadDirectory_ObjectDownloadFailedEvent_CancelInHandler_ContinueOnFailure_PreventsFurtherDownloads()
+        public async Task DownloadDirectory_ObjectDownloadFailedEvent_CancelInHandler_ContinueOnFailure_Throws()
         {
             var keys = new[] { "prefix/file1.txt", "prefix/file2.txt", "prefix/file3.txt" };
             var mockS3 = CreateMockS3(keys, k => k.EndsWith("file2.txt", StringComparison.Ordinal));
             string localDir = CreateTempDirectory();
             try
             {
-                var config = new TransferUtilityConfig();
+                var tu = new TransferUtility(mockS3.Object);
                 var request = CreateRequest(localDir, FailurePolicy.ContinueOnFailure);
                 // Make sequential to make behavior deterministic for the test.
                 request.DownloadFilesConcurrently = false;
@@ -213,23 +213,13 @@ namespace AWSSDK.UnitTests
                 request.ObjectDownloadFailedEvent += (sender, args) =>
                 {
                     handlerInvoked = true;
-                    // Cancel the per-object CTS to stop scheduling remaining object downloads.
-                    args.CancellationTokenSource.Cancel();
+                    throw new AmazonS3Exception("Stop processing immediately");
                 };
 
-                var command = new DownloadDirectoryCommand(mockS3.Object, request, config);
-                command.DownloadFilesConcurrently = request.DownloadFilesConcurrently;
-                var response = await command.ExecuteAsync(CancellationToken.None).ConfigureAwait(false);
+                var ex = await Assert.ThrowsExceptionAsync<AmazonS3Exception>(() => tu.DownloadDirectoryAsync(request));
+                Assert.IsTrue(ex.Message.Equals("Stop processing immediately"));
 
                 Assert.IsTrue(handlerInvoked, "ObjectDownloadFailedEvent handler was not invoked.");
-                Assert.IsNotNull(response);
-                // Allow multiple failures due to race; ensure at least one failure occurred and that one of the errors refers to file2
-                Assert.IsTrue(response.ObjectsFailed >= 1, "Expected at least one failed object.");
-                Assert.IsNotNull(response.Errors);
-                Assert.IsTrue(response.Errors.Count >= 1, "Expected at least one error recorded.");
-                Assert.IsTrue(response.Errors.Any(e => e.Message.Contains("file2.txt") || e.Message.Contains("file2")), "Expected at least one error to reference file2.");
-                Assert.IsTrue(File.Exists(Path.Combine(localDir, "file1.txt")), "Expected file1 to be downloaded.");
-                Assert.IsFalse(File.Exists(Path.Combine(localDir, "file2.txt")), "Expected failed file2 not to exist on disk.");
             }
             finally
             {
@@ -252,12 +242,11 @@ namespace AWSSDK.UnitTests
 
                 request.ObjectDownloadFailedEvent += (sender, args) =>
                 {
-                    // Cancel the per-object CTS in the handler; with AbortOnFailure the overall operation should throw.
-                    args.CancellationTokenSource.Cancel();
+                    throw new AmazonS3Exception("Stop processing immediately");
                 };
 
                 var ex = await Assert.ThrowsExceptionAsync<AmazonS3Exception>(() => tu.DownloadDirectoryAsync(request));
-                Assert.IsTrue(ex.Message.Contains("second.txt") || ex.Message.Contains("Simulated failure for"));
+                Assert.IsTrue(ex.Message.Equals("Stop processing immediately"));
             }
             finally
             {
@@ -302,8 +291,6 @@ namespace AWSSDK.UnitTests
                 Assert.IsTrue(evt.ObjectRequest.Key.EndsWith("b.txt", StringComparison.Ordinal));
                 Assert.IsNotNull(evt.Exception);
                 Assert.IsTrue(evt.Exception.Message.Contains("Simulated failure for"));
-                Assert.IsNotNull(evt.CancellationTokenSource);
-                Assert.IsFalse(evt.CancellationTokenSource.IsCancellationRequested);
             }
             finally
             {
@@ -343,7 +330,6 @@ namespace AWSSDK.UnitTests
                 Assert.IsTrue(evt.ObjectRequest.Key.EndsWith("y.txt", StringComparison.Ordinal));
                 Assert.IsNotNull(evt.Exception);
                 Assert.IsTrue(evt.Exception.Message.Contains("Simulated failure for"));
-                Assert.IsNotNull(evt.CancellationTokenSource);
             }
             finally
             {
--- a/sdk/src/Services/S3/Custom/Transfer/Internal/DownloadDirectoryCommand.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/Internal/DownloadDirectoryCommand.cs
@@ -38,7 +38,7 @@ namespace Amazon.S3.Transfer.Internal
     internal partial class DownloadDirectoryCommand : BaseCommand<TransferUtilityDownloadDirectoryResponse>
     {
         private IFailurePolicy _failurePolicy;
-        private ConcurrentBag<Exception> _errors;
+        private ConcurrentBag<Exception> _errors = new ConcurrentBag<Exception>();
         private readonly IAmazonS3 _s3Client;
         private readonly TransferUtilityDownloadDirectoryRequest _request;
         private readonly bool _skipEncryptionInstructionFiles;
@@ -56,7 +56,6 @@ namespace Amazon.S3.Transfer.Internal
             this._s3Client = s3Client;
             this._request = request;
             this._skipEncryptionInstructionFiles = s3Client is Amazon.S3.Internal.IAmazonS3Encryption;
-            _errors = request.FailurePolicy == FailurePolicy.ContinueOnFailure ? new ConcurrentBag<Exception>() : null;
             _failurePolicy =
                 request.FailurePolicy == FailurePolicy.AbortOnFailure
                     ? new AbortOnFailurePolicy()


sdk/src/Services/S3/Custom/Transfer/Model/FailurePolicy.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace AWSSDK.S3.Transfer.Model
```
GarrettBeatty: missing header

sdk/src/Services/S3/Custom/Transfer/Model/DirectoryResult.cs
```diff
@@ -0,0 +1,21 @@
+namespace AWSSDK.S3.Transfer.Model
```
GarrettBeatty: same here

sdk/src/Services/S3/Custom/Transfer/Internal/_bcl+netstandard/DownloadDirectoryCommand.cs
```diff
@@ -104,19 +111,69 @@ await asyncThrottler.WaitAsync(cancellationToken)
...
+                                errors.Add(ex);
+                            }
+                            // Swallow exception to allow other tasks to continue when not cancelling on failure.
+                            continue;
+                        }
+                    }
+                    finally
+                    {
+                        asyncThrottler.Release();
+                    }
```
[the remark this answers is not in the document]
GarrettBeatty: if there was/is no test to catch this issue, can we add it?

sdk/src/Services/S3/Custom/Transfer/Internal/_async/BaseCommand.async.cs
```diff
@@ -105,5 +106,40 @@ await command.ExecuteAsync(internalCts.Token)
                 throttler.Release();
             }
         }
+
+        /// <summary>
+        /// Executes a command and optionally cancels remaining work on failure.
+        /// If cancelOnFailure is false, exceptions are captured into the provided errors list (if any) and suppressed.
+        /// </summary>
+        protected static async Task ExecuteCommandAsync<T>(BaseCommand<T> command, CancellationTokenSource internalCts, SemaphoreSlim throttler, bool cancelOnFailure, ConcurrentBag<Exception> errors = null) where T : class
```
GarrettBeatty: something about modifying base command to add this function and catch things seems off to me. Mostly because i see in this function we are adding all the errors to a list and we are also doing some logic in download directory creating the list of errors too. 

in download directory why not do something like


```
public override async Task<TransferUtilityDownloadDirectoryResponse> ExecuteAsync(...)
{
    // Step 1:
    var errors = _failurePolicy == FailurePolicy.ContinueOnFailure 
        ? new ConcurrentBag<Exception>()  // 
        : null;
    
    // Step 2: Create strategy 
    IDownloadFailureStrategy strategy = 
        _failurePolicy == FailurePolicy.AbortOnFailure
            ? new AbortOnFailureStrategy()
            : new ContinueOnFailureStrategy(errors); // ← Pass our bag
    
    // Step 3: Execute downloads
    foreach (var s3o in objs)
    {
        var task = strategy.ExecuteAsync(
            async () => {
                var cmd = new DownloadCommand(...);
                await ExecuteCommandAsync(cmd, internalCts, asyncThrottler);
            },
            internalCts
        );
        pendingTasks.Add(task);
    }
    
    // Step 4: Wait for all to complete
    await WhenAllOrFirstExceptionAsync(pendingTasks, cancellationToken);
    
    // Step 5: Access our error bag to build response
    return new TransferUtilityDownloadDirectoryResponse
    {
        ObjectsDownloaded = _numberOfFilesDownloaded,
        ObjectsFailed = errors?.Count ?? 0,     
        Errors = errors?.ToList(),              
        Result = (errors is null || errors.Count == 0) ?
            DirectoryResult.Success :
            (_numberOfFilesDownloaded > 0 ?
                DirectoryResult.PartialSuccess :
                DirectoryResult.Failure)
    };
}

```


and that way it doesnt have two go into different code paths/duplicate code.  and theoretically for upload directory we can re-use the same `strategy` logic.

```

{

    internal class AbortOnFailureStrategy : IDownloadFailureStrategy
    {
   
        public async Task<bool> ExecuteAsync(
            Func<Task> action, 
            CancellationTokenSource cancellationTokenSource)
        {
            try
            {
                await action().ConfigureAwait(continueOnCapturedContext: false);
                return true;
            }
            catch (Exception)
            {
                // Cancel all pending operations before propagating the exception
                cancellationTokenSource.Cancel();
                throw;
            }
        }
    }
}
```

```
 internal class ContinueOnFailureStrategy : IDownloadFailureStrategy
    {
        private readonly ConcurrentBag<Exception> _errors;

 
        public ContinueOnFailureStrategy(ConcurrentBag<Exception> errors)
        {
            _errors = errors
        }

        public async Task<bool> ExecuteAsync(
            Func<Task> action, 
            CancellationTokenSource cancellationTokenSource)
        {
            try
            {
                await action().ConfigureAwait(continueOnCapturedContext: false);
                return true;
            }
            catch (Exception ex)
            {
                // Collect the exception for later reporting but don't throw it.
                // This allows other downloads to continue processing.
                _errors.Add(ex);
                return false;
            }
        }
```


that way we dont have to have branching logic/duplicate code for capturing the list of errors and everything.

sdk/src/Services/S3/Custom/Transfer/Internal/_bcl+netstandard/DownloadDirectoryCommand.cs
```diff
@@ -104,19 +111,69 @@ await asyncThrottler.WaitAsync(cancellationToken)
...
                     
                     pendingTasks.Add(task);
                 }
-                await WhenAllOrFirstExceptionAsync(pendingTasks, cancellationToken)
-                    .ConfigureAwait(continueOnCapturedContext: false);
+
+                if (_failurePolicy == FailurePolicy.ContinueOnFailure)
+                {
+                    // Drain all tasks; exceptions were suppressed and collected
+                    while (pendingTasks.Count > 0)
```
GarrettBeatty: having this task cancellation logic  in DownloadDirectoryClass seems off - mostly because this is low level stuff. and a lot of the current code in this class is more abstract. why not abstract this failure policy logic into its own class? see above comment

sdk/test/Services/S3/UnitTests/Custom/FailurePolicyTests.cs
```diff
@@ -0,0 +1,198 @@
...
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace AWSSDK.UnitTests
+{
+    [TestClass]
+    public class FailurePolicyTests
+    {
+        private static TransferUtilityDownloadDirectoryRequest CreateRequest(string localDir, FailurePolicy policy)
```
GarrettBeatty, line 21: do we have a test for

1. Test what happens when user cancels mid-download with `CancellationToken`

sdk/src/Services/S3/Custom/Transfer/Model/DirectoryResult.cs
philasmar: Fixed

sdk/src/Services/S3/Custom/Transfer/Internal/_async/BaseCommand.async.cs
philasmar: Fixed

sdk/test/Services/S3/UnitTests/Custom/FailurePolicyTests.cs
philasmar: Added

sdk/src/Services/S3/Custom/Transfer/Internal/_bcl+netstandard/DownloadDirectoryCommand.cs
philasmar: Fixed

sdk/src/Services/S3/Custom/Transfer/Model/FailurePolicy.cs
philasmar: Fixed

sdk/src/Services/S3/Custom/Transfer/Internal/_async/BaseCommand.async.cs
```diff
@@ -83,7 +83,7 @@ await completedTask
             }
         }
 
-        protected static async Task ExecuteCommandAsync<T>(BaseCommand<T> command, CancellationTokenSource internalCts, SemaphoreSlim throttler) where T : class
```
GarrettBeatty: assuming this is safe because uploaddirectory/downloaddirectory are the only commands that use this function signature

sdk/src/Services/S3/Custom/Transfer/Internal/ContinueOnFailurePolicy.cs
```diff
@@ -0,0 +1,104 @@
...
+        public async Task<bool> ExecuteAsync(Func<Task> action, Action<Exception> onFailure, CancellationTokenSource cancellationTokenSource)
+        {
+            try
+            {
+                await action().ConfigureAwait(false);
+
+                return true;
+            }
+            // If the operation was canceled via the provided token, propagate cancellation.
+            catch (OperationCanceledException ex) when (cancellationTokenSource?.IsCancellationRequested == true)
```
GarrettBeatty, line 80: lets double check with @normj on this for what the best user experience is. i think im fine with it

sdk/src/Services/S3/Custom/Transfer/Internal/AbortOnFailurePolicy.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// <param name="action">An asynchronous delegate that performs the work to execute under the policy.</param>
+        /// <param name="onFailure">An optional callback that will be invoked with the exception if <paramref name="action"/> fails.</param>
+        /// <param name="cancellationTokenSource">A <see cref="CancellationTokenSource"/> that will be canceled by this policy to signal termination
+        /// of related work when a failure occurs.</param>
+        /// <returns>
+        /// A <see cref="Task{Boolean}"/> that completes with <c>true</c> when <paramref name="action"/> completes successfully.
+        /// If <paramref name="action"/> fails, this method cancels <paramref name="cancellationTokenSource"/>, invokes
+        /// <paramref name="onFailure"/> if provided, and rethrows the original exception; it does not return <c>false</c>.
+        /// </returns>
+        public async Task<bool> ExecuteAsync(Func<Task> action, Action<Exception> onFailure, CancellationTokenSource cancellationTokenSource)
```
GarrettBeatty, line 53: nit: in the code i dont think we ever check the return value of `bool`. suggestion to remove it if we never check it

sdk/src/Services/S3/Custom/Transfer/TransferUtilityDownloadDirectoryRequest.cs
```diff
@@ -559,4 +599,82 @@ public override string ToString()
...
+    /// the per-object download request that failed (<see cref="TransferUtilityDownloadRequest"/>),
+    /// and the exception that caused the failure.
+    /// </remarks>
+    /// <example>
+    /// <code>
+    /// var request = new TransferUtilityDownloadDirectoryRequest { /* ... */ };
+    /// request.ObjectDownloadFailedEvent += (sender, args) =>
+    /// {
+    ///     // args.DirectoryRequest: original directory request
+    ///     // args.ObjectRequest: upload request for the failed object
```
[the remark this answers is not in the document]
GarrettBeatty: please fix this

sdk/src/Services/S3/Custom/Transfer/TransferUtilityDownloadDirectoryRequest.cs
```diff
@@ -56,6 +58,44 @@ public class TransferUtilityDownloadDirectoryRequest
...
+        /// <see cref="TransferUtilityDownloadRequest"/>, and the exception that caused the failure.
+        /// This event is raised on a background thread by the transfer utility.
+        /// </remarks>
+        /// <example>
+        /// request.ObjectDownloadFailedEvent += (sender, args) =>
+        /// {
+        ///     // inspect args.DirectoryRequest, args.ObjectRequest, args.Exception
+        /// };
+        /// </example>
+        public event EventHandler<ObjectDownloadFailedEventArgs> ObjectDownloadFailedEvent;
```
GarrettBeatty, line 89: nit: eventually we will have DownloadDirectoryFailedEvent i think (or something like that). do we just want to call this one DownloadFailedEvent, to be consistent with https://github.com/aws/aws-sdk-net/pull/4079/files#diff-c9ea44600a6e9cb8ca2b947856decd87ee5fd9ac8fc73a714080283c278f9038R188?

sdk/src/Services/S3/Custom/Transfer/TransferUtilityDownloadDirectoryResponse.cs
```diff
@@ -23,8 +25,24 @@ namespace Amazon.S3.Transfer
...
+        /// <summary>
+        /// The number of objects that failed to download. Zero if all succeeded.
+        /// </summary>
+        public long ObjectsFailed { get; set; }
+
+        /// <summary>
+        /// The collection of exceptions encountered when downloading individual objects.
+        /// Only populated when the <see cref="FailurePolicy"/> is <see cref="FailurePolicy.ContinueOnFailure"/>.
+        /// </summary>
+        public IList<Exception> Errors { get; set; }
```
GarrettBeatty, line 40: why dont we populate this when its abort on failure (Even if it will always be 1 error), just to be consistent?

sdk/src/Services/S3/Custom/Transfer/Internal/AbortOnFailurePolicy.cs
philasmar: We do check it in one instance and i think it's valid to have it:
```
var isValid = await _failurePolicy.ExecuteAsync(
                            () => {
                                //Ensure the target file is a rooted within LocalDirectory. Otherwise error.
                                if(!InternalSDKUtils.IsFilePathRootedWithDirectoryPath(downloadRequest.FilePath, _request.LocalDirectory))
                                {
                                    throw new AmazonClientException($"The file {downloadRequest.FilePath} is not allowed outside of the target directory {_request.LocalDirectory}.");
                                }

                                return Task.CompletedTask;
                            },
                            onFailure,
                            internalCts
                        ).ConfigureAwait(false);
                        if (!isValid) continue;
```

sdk/src/Services/S3/Custom/Transfer/TransferUtilityDownloadDirectoryRequest.cs
philasmar: I was going to do that initially, but for now since we only have 1, it will be confusing for users and they will assume this is a directory level event since it is on the Directory request. When I look at:
```
TransferUtilityDownloadDirectoryRequest.DownloadFailedEvent
```
Users could easily get confused and this it will get fired when the directory request fails

sdk/src/Services/S3/Custom/Transfer/TransferUtilityDownloadDirectoryResponse.cs
philasmar: because abort on failure means the process will throw an exception. The user will not get a response returned back and they will need to handle the thrown exception.

sdk/src/Services/S3/Custom/Transfer/TransferUtilityDownloadDirectoryRequest.cs
GarrettBeatty: ok thats fine. initially i didnt like the word `object` but i realized DownloadDirectoryResponse uses the word `object` (`objectsDownloaded`) so im fine with it now

sdk/src/Services/S3/Custom/Transfer/Internal/_bcl+netstandard/DownloadDirectoryCommand.cs
```diff
@@ -36,6 +39,11 @@ internal DownloadDirectoryCommand(IAmazonS3 s3Client, TransferUtilityDownloadDir
             : this(s3Client, request)
         {
             this._config = config;
+            this._errors = request.FailurePolicy == FailurePolicy.ContinueOnFailure ? new ConcurrentBag<Exception>() : null;
```
normj: nit: Initializing `_errors` and `_failurePolicy` is a duplicate of what is being done when the `this(s3Client, request)` constructor is called above.

sdk/src/Services/S3/Custom/Transfer/TransferUtilityDownloadDirectoryRequest.cs
```diff
@@ -56,6 +58,44 @@ public class TransferUtilityDownloadDirectoryRequest
...
+
+        /// <summary>
+        /// Internal helper used by the transfer implementation to raise the <see cref="ObjectDownloadFailedEvent"/>.
+        /// </summary>
+        /// <param name="args">The details of the failed object download.</param>
+        internal void OnRaiseObjectDownloadFailedEvent(ObjectDownloadFailedEventArgs args)
+        {
+            AWSSDKUtils.InvokeInBackground(ObjectDownloadFailedEvent, args, this);
```
normj: Since the failed event is being invoked on a background thread that does make it so if a user attempts to stop the download via throwing an exception the exception will be lost and not stop the download. I don't think in this case you need to run the callback in a background task.

The original reason we fired the events in the background were for events communicating the progress downloading from the stream. We didn't want the callback to block the performance of reading from the stream.

Can we be sure to add a test that proves that throwing an exception from the callback stops the download.

sdk/src/Services/S3/Custom/Transfer/TransferUtilityDownloadDirectoryRequest.cs
```diff
@@ -559,4 +599,82 @@ public override string ToString()
...
+        /// The exception thrown by the underlying download operation. Can be an <see cref="Amazon.S3.AmazonS3Exception"/>,
+        /// <see cref="Amazon.Runtime.AmazonClientException"/>, <see cref="IOException"/>, or other exception type depending
+        /// on the failure mode.
+        /// </value>
+        public Exception Exception { get; private set; }
+        
+        /// <summary>
+        /// Gets the <see cref="CancellationTokenSource"/> associated with the failed object's operation, if provided.
+        /// </summary>
+        public CancellationTokenSource CancellationTokenSource { get; private set; }
```
normj: Why are exposing this? This is an internal state variable that I don't think we should be exposing.

sdk/src/Services/S3/Custom/Transfer/TransferUtilityDownloadDirectoryResponse.cs
```diff
@@ -23,8 +25,24 @@ namespace Amazon.S3.Transfer
...
         public long ObjectsDownloaded { get; set; }
+
+        /// <summary>
+        /// The number of objects that failed to download. Zero if all succeeded.
+        /// </summary>
+        public long ObjectsFailed { get; set; }
+
+        /// <summary>
+        /// The collection of exceptions encountered when downloading individual objects.
+        /// Only populated when the <see cref="FailurePolicy"/> is <see cref="FailurePolicy.ContinueOnFailure"/>.
```
normj: Call out this could be null if there are no errors or make sure it is never null. I believe in the line `Errors = _errors?.ToList(),` in DownloadDirectoryCommand implies it could sometime be null. But I'm not sure because maybe `_errors` would only be null of Abort on failure was set which would mean you never get a response.

sdk/src/Services/S3/Custom/Transfer/_bcl+netstandard/TransferUtility.async.cs
```diff
@@ -209,7 +209,7 @@ public partial class TransferUtility : ITransferUtility
                 CheckForBlockedArn(request.BucketName, "DownloadDirectory");
                 var command = new DownloadDirectoryCommand(this._s3Client, request, this._config);
                 command.DownloadFilesConcurrently = request.DownloadFilesConcurrently;
-                await command.ExecuteAsync(cancellationToken).ConfigureAwait(false);
+                var response = await command.ExecuteAsync(cancellationToken).ConfigureAwait(false);
```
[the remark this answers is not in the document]
normj: You aren't doing anything with the `var response`. Either do not have it or you could be explicit that you don't care about the response `_ =  await command.ExecuteAsync(cancellationToken).ConfigureAwait(false);`

sdk/src/Services/S3/Custom/Transfer/Internal/AbortOnFailurePolicy.cs
```diff
@@ -0,0 +1,72 @@
...
+            try
+            {
+                await action().ConfigureAwait(false);
+
+                return true;
+            }
+            catch (Exception ex)
+            {
+                // Cancel all pending operations before propagating the exception
+                cancellationTokenSource.Cancel();
```
peterrsongg: if cancellationTokenSource is null this will throw an `NullReferenceException` can we update this to
`cancellationTokenSource?.Cancel()`?

sdk/src/Services/S3/Custom/Transfer/_bcl+netstandard/TransferUtility.async.cs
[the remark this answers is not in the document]
peterrsongg: i'm assuming this response will be used someplace else eventually?

sdk/src/Services/S3/Custom/Transfer/Internal/AbortOnFailurePolicy.cs
peterrsongg: any reason we didn't do 
` CancellationToken cancellationToken = default(CancellationToken)` in the constructor like we do in the others?

sdk/src/Services/S3/Custom/Transfer/Internal/ContinueOnFailurePolicy.cs
```diff
@@ -0,0 +1,104 @@
...
+        /// <param name="onFailure">A callback invoked with the exception when <paramref name="action"/> fails.</param>
+        /// <param name="cancellationTokenSource">A <see cref="CancellationTokenSource"/> used to determine and signal cancellation.
+        /// The policy will rethrow cancellations when the cancellation token was requested.</param>
+        /// <returns>
+        /// A <see cref="Task{Boolean}"/> that completes with <c>true</c> when the action completed successfully.
+        /// If the action threw a non-cancellation exception, the exception is added to the internal error bag,
+        /// <paramref name="onFailure"/> is invoked if provided, and the method completes with <c>false</c> to indicate
+        /// the action failed but the policy handled it and allowed processing to continue.
+        /// </returns>
+        public async Task<bool> ExecuteAsync(Func<Task> action, Action<Exception> onFailure, CancellationTokenSource cancellationTokenSource)
```
peterrsongg, line 71: similar comment, can we do
``` CancellationToken cancellationToken = default(CancellationToken)```

File: sdk/src/Services/S3/Custom/Transfer/Internal/_bcl+netstandard/DownloadDirectoryCommand.cs [hunk 20]
```diff
@@ -36,6 +39,11 @@ internal DownloadDirectoryCommand(IAmazonS3 s3Client, TransferUtilityDownloadDir
...
+            this._failurePolicy =
```
peterrsongg: spoke to phil offline about initializing the same fields twice. we agreed we should only do it once

sdk/src/Services/S3/Custom/Transfer/TransferUtilityDownloadDirectoryRequest.cs
GarrettBeatty: @normj wondering if you think for https://github.com/aws/aws-sdk-net/pull/4079 if we should change the Initiated, completed, and failed events to not fire in the background of it doesnt matter? In https://github.com/aws/aws-sdk-net/pull/4079's case, the user isnt interacting with the event, its just giving them data.

normj: @GarrettBeatty My thinking is probably best to not send those in a background event because it seems logical that the event callbacks would be called before the root calling function returns from it's awaited state. If you put it in the background I don't think you can guarantee that. What do you think of that logic?

GarrettBeatty: Thats what i was thinking too. In that case I will update github.com/aws/aws-sdk-net/pull/4139 to not fire in the background and also also update the code from https://github.com/aws/aws-sdk-net/pull/4079 in another pr

sdk/src/Services/S3/Custom/Transfer/_bcl+netstandard/TransferUtility.async.cs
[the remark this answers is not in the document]
philasmar: This was a left over from testing. I removed it.

sdk/src/Services/S3/Custom/Transfer/Internal/_bcl+netstandard/DownloadDirectoryCommand.cs
philasmar: Done

sdk/src/Services/S3/Custom/Transfer/TransferUtilityDownloadDirectoryRequest.cs
philasmar: Fixed

sdk/src/Services/S3/Custom/Transfer/TransferUtilityDownloadDirectoryResponse.cs
philasmar: Switched to always initializing it.

sdk/src/Services/S3/Custom/Transfer/TransferUtilityDownloadDirectoryRequest.cs
philasmar: Removed and changed to rely on exceptions to abort

sdk/src/Services/S3/Custom/Transfer/Internal/AbortOnFailurePolicy.cs
philasmar: Fixed

sdk/src/Services/S3/Custom/Transfer/Internal/ContinueOnFailurePolicy.cs
philasmar: As discussed offline, this is not the desired behavior

sdk/src/Services/S3/Custom/Transfer/Internal/_bcl+netstandard/DownloadDirectoryCommand.cs
philasmar: Fixed

PR #4160, opened 2025-11-24
Generate CreateBucketMetadataTableConfiguration

peterrsongg:
## Description

Generates `CreateBucketMetadataTableConfiguration`

```
AssemblyComparer AWSSDK.S3.New.dll: Error Amazon.S3.Model.RecordExpiration/MethodRemoved: Missing method System.Void set_Days(System.Int32) in Amazon.S3.Model.RecordExpiration
AssemblyComparer AWSSDK.S3.New.dll: Message Amazon.S3.Model.RecordExpiration/MethodAdded: New method System.Void set_Days(System.Nullable<System.Int32>) in Amazon.S3.Model.RecordExpiration
AssemblyComparer AWSSDK.S3.New.dll: Error Amazon.S3.Model.RecordExpiration/MethodRemoved: Missing method System.Int32 get_Days() in Amazon.S3.Model.RecordExpiration
AssemblyComparer AWSSDK.S3.New.dll: Message Amazon.S3.Model.RecordExpiration/MethodAdded: New method System.Nullable<System.Int32> get_Days() in Amazon.S3.Model.RecordExpiration
```

Amazon.S3.Model.RecordExpiration `Days` property changed from int to `int?` will call this out in changelog
## Testing

Dry run passed
Fuzz tests ran no change
## Types of changes

## Checklist

## License

### DIFF
--- a/generator/ServiceModels/s3/s3.customizations.json
+++ b/generator/ServiceModels/s3/s3.customizations.json
@@ -1105,7 +1105,7 @@
                     "InventoryFilter":{"injectXmlMarshallCode":["InventoryFilterCustomMarshall(publicRequest, xmlWriter);"]}
                 },
                 {
-                    "IsEnabled":{"injectXmlMarshallCode":["xmlWriter.WriteElementString(\"IsEnabled\", StringUtils.FromBool(publicRequest.InventoryConfiguration.IsEnabled.GetValueOrDefault()));"]}
+                    "IsEnabled":{"injectXmlMarshallCode":["CustomIsEnabledMarshall(publicRequest, xmlWriter);"]}
                 }
             ]
         },
@@ -1208,6 +1208,7 @@
                 }
             ]
         }
+        
     },
     "operationModifiers": {
         "CreateBucket": {
--- a/sdk/src/Services/S3/Custom/Model/Internal/MarshallTransformations/PutBucketInventoryConfigurationRequestMarshaller.cs
+++ b/sdk/src/Services/S3/Custom/Model/Internal/MarshallTransformations/PutBucketInventoryConfigurationRequestMarshaller.cs
@@ -50,5 +50,9 @@ namespace Amazon.S3.Model.Internal.MarshallTransformations
                 xmlWriter.WriteEndElement();
             }
         }
+        void CustomIsEnabledMarshall(PutBucketInventoryConfigurationRequest publicRequest, XmlWriter xmlWriter)
+        {
+            xmlWriter.WriteElementString("IsEnabled", StringUtils.FromBool(publicRequest.InventoryConfiguration.IsEnabled.GetValueOrDefault()));
+        }
     }
 }
--- a/sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/PutBucketInventoryConfigurationRequestMarshaller.cs
+++ b/sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/PutBucketInventoryConfigurationRequestMarshaller.cs
@@ -132,7 +132,7 @@ namespace Amazon.S3.Model.Internal.MarshallTransformations
                         }            
                         xmlWriter.WriteEndElement();            
                     }
-                    xmlWriter.WriteElementString("IsEnabled", StringUtils.FromBool(publicRequest.InventoryConfiguration.IsEnabled.GetValueOrDefault()));
+                    CustomIsEnabledMarshall(publicRequest, xmlWriter);
                     if (publicRequest.InventoryConfiguration.Schedule != null)
                     {
                         xmlWriter.WriteStartElement("Schedule");


peterrsongg:
## READ THIS
* Changing `int` to `int?` is something we've done in other PR's. This follows the nullability pattern for the sdk that we just missed here because it is handcoded.
----------------
## Files Analyzed: 9 out of 9 files changed in this PR

### BREAKING CHANGES FOUND:

#### 1. __RecordExpiration.cs__ - CRITICAL BREAKING CHANGE

__File__: `sdk/src/Services/S3/Generated/Model/RecordExpiration.cs`

__Issue__: The `Days` property type changed from non-nullable `int` to nullable `int?`

__Custom version (deleted)__:

```csharp
public int Days
{
    get { return this._days.GetValueOrDefault(); }
    set { this._days = value; }
}
```

__Generated version (new)__:

```csharp
public int? Days
{
    get { return this._days; }
    set { this._days = value; }
}
```

__Impact__: This is a breaking change to the public API. Code that previously worked with `int Days` will break when trying to use `int? Days`. Existing code that accesses `Days` without null checking will fail at runtime if the value is null, or will not compile if assigned to non-nullable int variables.

__Example Breaking Scenario__:

```csharp
// Old code that will break:
RecordExpiration expiration = new RecordExpiration();
int days = expiration.Days; // Compilation error: Cannot implicitly convert type 'int?' to 'int'

// Or runtime issue:
int daysValue = expiration.Days.Value; // Could throw InvalidOperationException if null
```

---

### NO OTHER BREAKING CHANGES DETECTED IN:

1. ✅ __CreateBucketMetadataConfigurationRequest.cs__ - All IsSet methods use `!String.IsNullOrEmpty()` as expected (matches customizations.json configuration)
2. ✅ __CreateBucketMetadataConfigurationResponse.cs__ - Empty response class, no breaking changes
3. ✅ __CreateBucketMetadataConfigurationRequestMarshaller.cs__ - No custom version existed, purely generated
4. ✅ __CreateBucketMetadataConfigurationResponseUnmarshaller.cs__ - No custom version existed, purely generated
5. ✅ __InventoryTableConfiguration.cs__ - Properties and IsSet methods match
6. ✅ __JournalTableConfiguration.cs__ - Properties and IsSet methods match
7. ✅ __MetadataConfiguration.cs__ - Properties and IsSet methods match
8. ✅ __MetadataTableEncryptionConfiguration.cs__ - Only private field naming changed (`kmsKeyArn` → `_kmsKeyArn`), public API unchanged

---

## Summary:

- __Total Files Changed__: 9
- __Files Analyzed__: 9
- __Breaking Changes Found__: 1 (RecordExpiration.Days property type change)
- __Status__: ❌ BREAKING CHANGES DETECTED - REQUIRES IMMEDIATE ATTENTION

sdk/src/Services/S3/Generated/Model/RecordExpiration.cs
```diff
@@ -47,9 +48,9 @@ public partial class RecordExpiration
         /// <c>365</c>. 
         /// </para>
         /// </summary>
-        public int Days
+        public int? Days
```
muhammad-othman, line 51: I remember that we had another instance of this change before, but I don't remember if we agreed it is ok to change it to nullable 😅

peterrsongg: sure let's discuss this as a team as well :) 

Yeah we have done this before, we just have to make sure to call it out as a breaking change in the changelog. in reality, we just missed this since S3 is handcoded so i think it is okay. In the long run, we want to remove these one-off customizations for s3 that deviate from our standard generation logic.

sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/CreateBucketMetadataConfigurationRequestMarshaller.cs
```diff
@@ -75,9 +76,7 @@ public IRequest Marshall(CreateBucketMetadataConfigurationRequest publicRequest)
             }
             if (!publicRequest.IsSetBucketName())
                 throw new AmazonS3Exception("Request object does not have required field BucketName set");
-            request.AddPathResource("{Bucket}", StringUtils.FromString(publicRequest.BucketName));
-            request.ResourcePath = "/{Bucket}";
-
+            request.ResourcePath = "/";
```
dscpinheiro, line 79: I don't see this called out in https://github.com/aws/aws-sdk-net/pull/4160#issuecomment-3569286788 (while it was called out in `GetBucketMetadataConfiguration`).

Does the same reason apply here?

peterrsongg: yup it is the same reason. Operation still works as expected

PR #4162, opened 2025-11-24
Optimize Part Stream for Multi part Download

GarrettBeatty:
Previously in https://github.com/aws/aws-sdk-net/pull/4130, we would buffer every part, including the first part. This was not optimal because the first part can technically be streamed directly to the user because it will always be in order. Similarly, there is another case where if a part finishes downloading and is the next expected part, we can stream that right to the user instead of buffering.

This change adds support so that in those cases the part is streamed directly to the user instead of buffering in memory.

## Changes
1. Add support for direct streaming
2. Also fix potential resource leak in existing code. I discovered these issues while working on this PR and i figure they are small enough so i included them here

## Motivation and Context

## Testing
1. Unit Tests which validate no additional array pools are created for this scenario and the part is streamed to the user directly
4. Existing integration tests pass

Re-ran performance tests and got similar results

```
Total bytes per run: 5,368,709,120

Run:1 Secs:2.661879 Gb/s:16.135098
Run:2 Secs:1.455994 Gb/s:29.498529
Run:3 Secs:1.204284 Gb/s:35.664076
Run:4 Secs:1.119768 Gb/s:38.355867
Run:5 Secs:1.057126 Gb/s:40.628725
Run:6 Secs:1.055039 Gb/s:40.709082
Run:7 Secs:1.060168 Gb/s:40.512150
Run:8 Secs:1.053934 Gb/s:40.751759
Run:9 Secs:1.062503 Gb/s:40.423092
Run:10 Secs:1.094352 Gb/s:39.246664
```

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/Transfer/Internal/BufferedMultipartStream.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/Internal/BufferedMultipartStream.cs
@@ -282,6 +282,8 @@ namespace Amazon.S3.Transfer.Internal
         /// <remarks>
         /// This method disposes the underlying <see cref="IDownloadManager"/> and <see cref="IPartBufferManager"/>,
         /// which in turn cleans up any buffered part data and returns ArrayPool buffers to the pool.
+        /// It also disposes the InitialResponse from the discovery result, which contains the HTTP connection
+        /// and network stream that must be explicitly disposed to return the connection to the pool.
         /// </remarks>
         [SuppressMessage("Design", "CA1031:Do not catch general exception types", Justification = "Dispose methods should not throw exceptions")]
         protected override void Dispose(bool disposing)
@@ -290,7 +292,13 @@ namespace Amazon.S3.Transfer.Internal
             {
                 try
                 {
-                    // Dispose modular dependencies
+                    // Dispose InitialResponse first (contains HTTP connection and network stream)
+                    // This is critical because GetObjectResponse holds unmanaged resources that
+                    // won't be cleaned up by GC - must be explicitly disposed to return HTTP
+                    // connection to the pool and close network streams
+                    _discoveryResult?.InitialResponse?.Dispose();
+                    
+                    // Then dispose modular dependencies
                     _downloadCoordinator?.Dispose();
                     _partBufferManager?.Dispose();
                 }
--- a/sdk/src/Services/S3/Custom/Transfer/Internal/BufferedPartDataHandler.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/Internal/BufferedPartDataHandler.cs
@@ -30,9 +30,26 @@ using Amazon.S3.Model;
 namespace Amazon.S3.Transfer.Internal
 {
     /// <summary>
-    /// Buffers downloaded parts in memory using <see cref="ArrayPool{T}"/> and <see cref="IPartBufferManager"/>.
-    /// Implements current streaming behavior for multipart downloads.
+    /// Handles multipart download data with intelligent stream-vs-buffer decision making.
+    /// Optimizes for sequential part arrival by streaming directly to consumer when possible,
+    /// while buffering out-of-order parts into memory using <see cref="ArrayPool{T}"/>.
     /// </summary>
+    /// <remarks>
+    /// <para><strong>Optimization Strategy:</strong></para>
+    /// <list type="bullet">
+    /// <item>Parts arriving in expected order (matching NextExpectedPartNumber) stream directly to consumer</item>
+    /// <item>Out-of-order parts buffer into ArrayPool memory for later sequential consumption</item>
+    /// <item>Best case: All parts in order → zero buffering → pure streaming</item>
+    /// <item>Worst case: All parts out of order → full buffering (original behavior)</item>
+    /// </list>
+    /// 
+    /// 
+    /// <para><strong>Response Ownership:</strong></para>
+    /// <list type="bullet">
+    /// <item>Streaming: StreamingDataSource takes ownership and disposes after reading</item>
+    /// <item>Buffering: Handler disposes response immediately after buffering completes</item>
+    /// </list>
+    /// </remarks>
     internal class BufferedPartDataHandler : IPartDataHandler
     {
         private readonly IPartBufferManager _partBufferManager;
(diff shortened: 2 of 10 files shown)


sdk/src/Services/S3/Custom/Transfer/Internal/BufferedPartDataHandler.cs
```diff
@@ -64,28 +81,62 @@ public Task PrepareAsync(DownloadDiscoveryResult discoveryResult, CancellationTo
...
+                // Buffer the part from the response stream into memory
+                var buffer = await BufferPartFromResponseAsync(
+                    partNumber, 
+                    response, 
+                    cancellationToken).ConfigureAwait(false);
 
-            Logger.DebugFormat("BufferedPartDataHandler: [Part {0}] Added to buffer manager",
-                partNumber);
+                // Response has been fully read and buffered - dispose it now
+                response?.Dispose();
```
GarrettBeatty, line 221: __Previously:__ `MultipartDownloadManager` had a `finally` block that always disposed the response after `ProcessPartAsync` returned, because the handler always buffered data into memory immediately.

__Now:__ The streaming optimization creates two code paths with different ownership:

1. __In-order parts (streaming)__: `StreamingDataSource` takes ownership of the response and keeps it alive to stream data directly to the consumer. The response is disposed later when streaming completes.

2. __Out-of-order parts (buffering)__: The handler reads the entire response into memory immediately, then disposes the response since it's no longer needed.

__Why the move:__ The old `finally` block would dispose the response immediately after `ProcessPartAsync` returns, which breaks the streaming path - it would dispose the response while `StreamingDataSource` is still reading from it. By moving disposal into the handler, each path can dispose at the appropriate time based on whether it's streaming (later) or buffering (now).

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedPartDataHandler.cs
```diff
@@ -64,28 +81,91 @@ public Task PrepareAsync(DownloadDiscoveryResult discoveryResult, CancellationTo
...
+                        partNumber, _partBufferManager.NextExpectedPartNumber);
+
+                    // Buffer the part from the response stream into memory
+                    var buffer = await BufferPartFromResponseAsync(
+                        partNumber, 
+                        response, 
+                        cancellationToken).ConfigureAwait(false);
+
+                    // Response has been fully read and buffered - dispose it now
+                    response?.Dispose();
```
GarrettBeatty: we have to do this here now because previously we had kept it in the multipartdownloadmanager finally block, it would've been disposed before StreamingDataSource could use it

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedPartDataHandler.cs
```diff
@@ -64,28 +81,156 @@ public Task PrepareAsync(DownloadDiscoveryResult discoveryResult, CancellationTo
...
+        /// <para>
+        /// Intelligently chooses between streaming and buffering based on part arrival order:
+        /// </para>
+        /// <list type="bullet">
+        /// <item>If partNumber matches NextExpectedPartNumber: Stream directly (no buffering)</item>
+        /// <item>Otherwise: Buffer into memory for later sequential consumption</item>
+        /// </list>
+        /// <para><strong>Response Ownership:</strong></para>
+        /// <para>
+        /// This method takes ownership of the response and is responsible for disposing it in ALL cases,
```
GarrettBeatty, line 94: the reason for this is because previously `MultipartDownloadManager` disposed of the response in the `finally` block. but with this change of adding `StreamingDataSource`, if we kept the dispose in the `MultipartDownloadManager`, the `response` would've been dispoed by the time  `StreamingDataSource` has to read it back to the user. Therefore we need to have  `StreamingDataSource` keep ownership of the response

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedPartDataHandler.cs
```diff
@@ -64,28 +81,156 @@ public Task PrepareAsync(DownloadDiscoveryResult discoveryResult, CancellationTo
...
-            Logger.DebugFormat("BufferedPartDataHandler: [Part {0}] Starting to buffer part from response stream - ContentLength={1}",
-                partNumber, response.ContentLength);
+            if (partNumber == _partBufferManager.NextExpectedPartNumber)
+            {
+                await ProcessStreamingPartAsync(partNumber, response, cancellationToken).ConfigureAwait(false);
+            }
+            else
+            {
+                await ProcessBufferedPartAsync(partNumber, response, cancellationToken).ConfigureAwait(false);
+            }
```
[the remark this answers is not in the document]
GarrettBeatty: this is not valid. The actual flow is 

```
Producer (Part 2):
├─ Check: partNumber (2) == NextExpectedPartNumber (2) ✓
├─ Decide to stream
├─ Call ProcessStreamingPartAsync(2, ...)
│  ├─ Create StreamingDataSource(2, response)
│  └─ AddBufferAsync(streamingDataSource) → Part 2 now in dictionary
└─ Done

Consumer:
├─ Wait for part 2 in dictionary (blocking until AddBufferAsync completes)
├─ Read part 2 from dictionary
├─ Part 2 complete → increment NextExpectedPartNumber to 3
└─ Continue to part 3

```

The consumer cannot increment past part 2 until part 2 is already in the dictionary. By that time, the producer has already made its streaming vs buffering decision and completed the ProcessPartAsync call.

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
```diff
@@ -222,20 +222,26 @@ public async Task StartDownloadsAsync(DownloadDiscoveryResult discoveryResult, C
...
-                Logger.DebugFormat("MultipartDownloadManager: Buffering Part 1 from discovery response");
-                await _dataHandler.ProcessPartAsync(1, discoveryResult.InitialResponse, cancellationToken).ConfigureAwait(false);
-
-                // Detach the event handler after processing to prevent memory leak
-                if (wrappedCallback != null)
+                finally
                 {
-                    discoveryResult.InitialResponse.WriteObjectProgressEvent -= wrappedCallback;
+                    // Always detach the event handler to prevent memory leak
+                    // This runs whether ProcessPartAsync succeeds or throws
```
GarrettBeatty, line 249: this is not related to the change at all, but one potential issue i found anyway, so fixing it here

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedMultipartStream.cs
```diff
@@ -290,7 +292,13 @@ protected override void Dispose(bool disposing)
             {
                 try
                 {
-                    // Dispose modular dependencies
+                    // Dispose InitialResponse first (contains HTTP connection and network stream)
+                    // This is critical because GetObjectResponse holds unmanaged resources that
+                    // won't be cleaned up by GC - must be explicitly disposed to return HTTP
+                    // connection to the pool and close network streams
+                    _discoveryResult?.InitialResponse?.Dispose();
```
GarrettBeatty, line 299: unrelated to the optimizing stream for multi part download, but i discovered this potential resource leak while working on this

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedPartDataHandler.cs
```diff
@@ -64,28 +81,163 @@ public Task PrepareAsync(DownloadDiscoveryResult discoveryResult, CancellationTo
...
+        /// </para>
+        /// </remarks>
         public async Task ProcessPartAsync(
             int partNumber,
             GetObjectResponse response, 
             CancellationToken cancellationToken)
         {
-            Logger.DebugFormat("BufferedPartDataHandler: [Part {0}] Starting to buffer part from response stream - ContentLength={1}",
-                partNumber, response.ContentLength);
+            if (partNumber == _partBufferManager.NextExpectedPartNumber)
```
GarrettBeatty, line 103: this "felt" kind of weird putting this logic here but i felt it was fine for now

generator/.DevConfigs/9d07dc1e-d82d-4f94-8700-c7b57f872045.json
```diff
@@ -0,0 +1,11 @@
+{
+  "services": [
+    {
+      "serviceName": "S3",
+      "type": "patch",
+      "changeLogMessages": [
+        "Optimized multipart download manager to stream responses directly where applicable."
```
philasmar: isn't this an optimization for a feature we haven't yet released? Maybe we don't need a changelog in this case?

GarrettBeatty: yeah you are right i can remove this dev config

GarrettBeatty: removed it

normj (approved):

PR #4163, opened 2025-11-25
Move maxInMemoryParts to request object

GarrettBeatty:
Moving `maxInMemoryParts`  to the individual request object instead of the TransferUtilityConfig.

Why? Previously i mistakenly  put it in  the `TransferUtilityConfig` thinking that `maxInMemoryParts` would be utilized in directory downloads (and therefore the value would apply at the config level). However, this is not the case, since maxInMemoryParts is only used for `OpenStreamWithResponse` which is not applicable for directories.

## Motivation and Context
To accurately reflect the scope of the maxInMemoryParts value.

## Testing
1. Re-ran unit tests
2. Re-ran integ tests

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/Transfer/TransferUtilityOpenStreamRequest.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/TransferUtilityOpenStreamRequest.cs
@@ -32,7 +32,21 @@ namespace Amazon.S3.Transfer
     /// </summary>
     public class TransferUtilityOpenStreamRequest : BaseDownloadRequest
     {
+        private int _maxInMemoryParts = 1024;
 
-
+        /// <summary>
+        /// Gets or sets the maximum number of parts to buffer in memory during multipart downloads.
+        /// The default value is 1024.
+        /// </summary>
+        /// <remarks>
+        /// This property controls memory usage during streaming downloads. When combined with the 
+        /// default part size of 8MB, the default value of 1024 parts allows up to 8GB of memory usage.
+        /// Adjust this value based on your application's memory constraints and performance requirements.
+        /// </remarks>
+        public int MaxInMemoryParts
+        {
+            get { return this._maxInMemoryParts; }
+            set { this._maxInMemoryParts = value; }
+        }
     }
 }
--- a/generator/.DevConfigs/19ed68ce-9f46-4e1e-a0ff-45a2b3641946.json
+++ b/generator/.DevConfigs/19ed68ce-9f46-4e1e-a0ff-45a2b3641946.json
@@ -4,7 +4,7 @@
       "serviceName": "S3",
       "type": "patch",
       "changeLogMessages": [
-        "Added MaxInMemoryParts property to TransferUtilityConfig for controlling memory usage during multipart downloads",
+        "Added MaxInMemoryParts property to TransferUtilityOpenStreamRequest for controlling memory usage during multipart downloads",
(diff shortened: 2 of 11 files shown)


generator/.DevConfigs/19ed68ce-9f46-4e1e-a0ff-45a2b3641947.json
```diff
@@ -0,0 +1,12 @@
+{
```
normj: Since we have never shipped this we shouldn't update the changelog with this message. I assume it was here to run the dry run but when you merge this do not include this.

GarrettBeatty:
> Seems like this should have affected some unit/integ tests. Do we have a some test coverage gaps?

Good point, there is a small gap for where Integration tests create `TransferUtilityConfig` but never set `MaxInMemoryParts` (it was always just using default values). ill add these unit tests/integration tests

generator/.DevConfigs/19ed68ce-9f46-4e1e-a0ff-45a2b3641947.json
GarrettBeatty: i updated the existing dev config instead

GarrettBeatty:
> Seems like this should have affected some unit/integ tests. Do we have a some test coverage gaps?

added unit and integ tests

sdk/src/Services/S3/Custom/Transfer/TransferUtilityOpenStreamRequest.cs
```diff
@@ -32,7 +32,27 @@ namespace Amazon.S3.Transfer
...
+        /// The default value is 1024.
+        /// </summary>
+        /// <remarks>
+        /// This property controls memory usage during streaming downloads. When combined with the 
+        /// default part size of 8MB, the default value of 1024 parts allows up to 8GB of memory usage.
+        /// Adjust this value based on your application's memory constraints and performance requirements.
+        /// </remarks>
+        public int MaxInMemoryParts
+        {
+            get { return this._maxInMemoryParts; }
```
GarrettBeatty, line 48: should we keep this logic? what do people think?

sdk/src/Services/S3/Custom/Transfer/TransferUtilityOpenStreamRequest.cs
```diff
@@ -32,7 +32,27 @@ namespace Amazon.S3.Transfer
...
+            set
+            {
+                if (value < 1)
```
philasmar: i don't like this logic here because we are ignoring the value the user is setting and they are not aware of it. There should be a validation step further down the line which throws an exception informing the user of wrong value they used.

GarrettBeatty: yeah we do have validation down the line where if its less than 0 we throw error so i can remove this check

GarrettBeatty: i updated to remove the check from here since we have it in https://github.com/aws/aws-sdk-net/blob/feature/transfermanager/sdk/src/Services/S3/Custom/Transfer/Internal/BufferedDownloadConfiguration.cs#L57

philasmar (approved):

normj (approved):

PR #4165, opened 2025-11-25
Fix canonical URLs to use direct page paths

AlexDaines:
## Description
- Changes canonical URLs from `index.html?page=X&tocid=Y` to direct paths like `items/S3/TS3Client.html`
- Adds `<meta name="aws-tocid">` for redirect logic
- Updates parentloader.js to read tocid from meta tag

## Motivation and Context
Canonical URLs pointed to frameset with query params. Search engines saw mismatch between canonical and actual page URL, hurting indexing.

## Testing
- Built doc generator
- Verified canonical URLs on class, namespace, and method pages
- Confirmed redirect behavior preserved via meta tag

## Types of changes

## Checklist

## License

### DIFF
--- a/docgenerator/SDKDocGeneratorLib/output-files/resources/parentloader.js
+++ b/docgenerator/SDKDocGeneratorLib/output-files/resources/parentloader.js
@@ -33,7 +33,8 @@
     else {
         //This content page is not in the expected frameset which contains the TOC so
         //reload the parent frame and cause it to reload this page in the frame.
-        //Read the tocid from the meta tag and build the frameset URL.
+
+        //First, try the new approach: read tocid from meta tag and build frameset URL
         var tocid = jQuery("meta[name='aws-tocid']").attr('content');
         var pathname = window.location.pathname;
         var itemsIndex = pathname.indexOf('/items/');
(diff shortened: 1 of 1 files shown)


GarrettBeatty:
can you share the exact testing steps/validations/screenshots that you tested?

docgenerator/SDKDocGeneratorLib/output-files/resources/parentloader.js
```diff
@@ -31,31 +31,21 @@
         }        
     }
     else {
-        //This content page is not in the expected frameset which contains the TOC so
```
GarrettBeatty, line 34: is this file generated? i see its in output-files folder.

docgenerator/SDKDocGeneratorLib/output-files/resources/parentloader.js
```diff
@@ -31,31 +31,21 @@
...
-            //Else do nothing because the link doesn't appear to be in a normal URL format where the 
-            //hostname ends with a / starting the relative URL path.
+        if (tocid && itemsIndex !== -1) {
+            //Build the frameset URL from the current path
+            var basePath = pathname.substring(0, itemsIndex);
+            var filename = pathname.substring(pathname.lastIndexOf('/') + 1);
+            //Reload the parent frameset and this content page. The inFrame part of this script
+            //will run on reload to replace the address bar URL with the actual friendly content
+            //frame page URL.
+            window.location.href = basePath + '/index.html?page=' + filename + '&tocid=' + tocid;
```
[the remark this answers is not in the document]
AlexDaines: The tocid parameter is the source of truth here. `TOC.html` uses the tocid to lookup the correct TOC (Table of Contents) item and extracts the full path from its href attribute. The page parameter is just a "hint", i.e. when there is a mismatch with the TOC derived path, then the fallback logic corrects it. This was intentional and leverages the existing redirect mechanism for bookmarks or deprecations etc.

docgenerator/SDKDocGeneratorLib/output-files/resources/parentloader.js
AlexDaines: Discussed offline but copying here for other reviewers. This file is just a static resource that our build system copies- it isnt generated per-page, per-service, per-anything

docgenerator/SDKDocGeneratorLib/output-files/resources/parentloader.js
```diff
@@ -31,31 +31,21 @@
...
-            if (relPathIndex != -1) {            
-                //Reload the parent frameset and this content page. Note the address bar URL will
-                //look like the canonical link in the address bar. The inFrame part of this script 
-                //will run on release to replace the address bar URL with the actual friendly content
-                //frame page URL.
-                window.location.href = pageLink.substring(relPathIndex);
-            }
-            //Else do nothing because the link doesn't appear to be in a normal URL format where the 
-            //hostname ends with a / starting the relative URL path.
+        if (tocid && itemsIndex !== -1) {
```
peterrsongg, line 42: so what happens if we don't get into this `if` statement? we dont set the `window.location.href`? Is that ever a valid case?

AlexDaines: Good question. This is the same behavior as the original code. It also had some cases where it would just not redirect (if the canonical url was empty or it couldnt parse the path etc.) . If we cant determine the tocid or path, the page just displays without the frameset which is not a huge deal. Content is still there just without the table of contents sidebar.

boblodgett:
I do not know exactly what is causing it but the old link formats that could be in people's bookmarks or in search results do not work with the new code and end up 404 erroring.

Example (When tried local):

http://localhost/sdkfornet/v4/apidocs/index.html?page=MS3GetObjectGetObjectRequest.html&tocid=Amazon_S3_AmazonS3Client

=> Give 404 error with the new code.

Existing V4 website we can see it works: https://docs.aws.amazon.com/sdkfornet/v4/apidocs/index.html?page=MS3GetObjectGetObjectRequest.html&tocid=Amazon_S3_AmazonS3Client

This is a must fix as we cannot break existing bookmarks or search result links.

boblodgett (approved):

PR #4168, opened 2025-11-26
Fix content language initialization and update to use headers backing field

GarrettBeatty:
no dev config since its already in https://github.com/aws/aws-sdk-net/pull/4074

## Description
update the  backing field for content language to be in the headers instead so that for TransferUtilityDownloadResponse we can easily just copy the headers only instead of the response.ContentLanguage field.

Also was a bug in https://github.com/aws/aws-sdk-net/pull/4074 where it was always setting the content language in one case because the `if` statement didnt have brackets. So this fixes that bug and also makes it so the headers is the backing field

## Testing
1. unit tests pass
2. need to run dry run

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/Model/GetObjectResponse.cs
+++ b/sdk/src/Services/S3/Custom/Model/GetObjectResponse.cs
@@ -69,7 +69,6 @@ namespace Amazon.S3.Model
         private string _checksumSHA1;
         private string _checksumSHA256;
         private ChecksumType _checksumType;
-        private string _contentLanguage;
 
         /// <summary>
         /// The date and time at which the object is no longer cacheable.
@@ -175,14 +174,8 @@ namespace Amazon.S3.Model
         /// </summary>
         public string ContentLanguage
         {
-            get { return this._contentLanguage; }
-            set { this._contentLanguage = value; }
-        }
-
-        // Check to see if ContentLanguage property is set
-        internal bool IsSetContentLanguage()
-        {
-            return this._contentLanguage != null;
+            get { return this.Headers.ContentLanguage; }
+            set { this.Headers.ContentLanguage = value; }
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


sdk/src/Services/S3/Custom/Model/GetObjectResponse.cs
```diff
@@ -175,14 +174,8 @@ internal bool IsSetContentRange()
...
-            get { return this._contentLanguage; }
-            set { this._contentLanguage = value; }
-        }
-
-        // Check to see if ContentLanguage property is set
-        internal bool IsSetContentLanguage()
-        {
-            return this._contentLanguage != null;
+            get { return this.Headers.ContentLanguage; }
+            set { this.Headers.ContentLanguage = value; }
```
GarrettBeatty, line 178: the headers collection doesnt have a setter https://github.com/aws/aws-sdk-net/pull/4168/files#diff-7d76ec6d3c33088a35cba25bbc63006459be9754c861936319892b11e3072652L117


so i believe we should be fine here

GarrettBeatty: ## With Get-Only: Guaranteed Single Instance

When `Headers` is get-only, you're guaranteed that there is __exactly one__ `HeadersCollection` instance throughout the lifetime of the `GetObjectResponse` object:

```csharp
var response = new GetObjectResponse();

// First access - creates the HeadersCollection
response.ContentLanguage = "en-US";
// Internally: response.Headers.ContentLanguage = "en-US"
//            ↓ This calls the getter, which creates headersCollection once

// Later access - uses THE SAME HeadersCollection
response.Headers.ContentType = "text/html";
//       ↑ Returns the same headersCollection instance

// Even later
Console.WriteLine(response.ContentLanguage); // "en-US" ✓
// Still reading from the SAME headersCollection instance
```

sdk/src/Services/S3/Custom/Model/Internal/MarshallTransformations/GetObjectResponseUnmarshaller.cs
```diff
@@ -86,7 +86,6 @@ private static void UnmarshallResult(XmlUnmarshallerContext context,GetObjectRes
             if (responseData.IsHeaderPresent("Content-Encoding"))
                 response.Headers.ContentEncoding = S3Transforms.ToString(responseData.GetHeaderValue("Content-Encoding"));
             if (responseData.IsHeaderPresent("Content-Language"))
-                response.ContentLanguage = S3Transforms.ToString(responseData.GetHeaderValue("Content-Language"));
```
peterrsongg, line 89: we are still marshalling `response.ContentLanguage` somewhere though right?

PR #4170, opened 2025-11-26
Update initiated, complete and failed events to not fire in background

GarrettBeatty:
Updating as per https://github.com/aws/aws-sdk-net/pull/4151#discussion_r2566722462

Theres no dev config here because theres already an existing one for adding these events in this branch

## Motivation and Context
if events are fired in the background there was no guarantee they would be executed before the original task finishes or not. so we update to synchronous to ensure that they are.

## Testing
1. existing unit tests pass
2. existing integ tests pass

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/test/Services/S3/IntegrationTests/TransferUtilityTests.cs
+++ b/sdk/test/Services/S3/IntegrationTests/TransferUtilityTests.cs
@@ -2125,11 +2125,7 @@ namespace AWSSDK_DotNet.IntegrationTests.Tests.S3
                 if (EventException != null)
                     throw EventException;
 
-                // Since  AWSSDKUtils.InvokeInBackground fires the event in the background it is possible that we check too early that the event has fired. In this case, we sleep and check again.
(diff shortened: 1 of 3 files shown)


sdk/test/Services/S3/IntegrationTests/TransferUtilityTests.cs
```diff
@@ -2125,11 +2125,7 @@ public void AssertEventFired()
                 if (EventException != null)
                     throw EventException;
 
-                // Since  AWSSDKUtils.InvokeInBackground fires the event in the background it is possible that we check too early that the event has fired. In this case, we sleep and check again.
-                for (int retries = 1; retries < 5 && !EventFired; retries++)
-                {
-                    Thread.Sleep(1000 * retries);
-                }
+                // Since events are now fired synchronously, we can check immediately without retries
```
GarrettBeatty, line 2128: this function was only used for validating the _**non**_ in progress events. theres a separate function that validates the in progress once which still has the retry logic since its a background thread

PR #4171, opened 2025-11-28
Add error message for multipart download when using Encryption Client

GarrettBeatty:
Add check where if encryption client is being used we give a friendly error to the user.

## Description
Out of the box the https://github.com/aws/amazon-s3-encryption-client-dotnet does not seem to work with PART GET. if the user attempts to use it with multi part download with PART get they get

```
ERROR: Failed to decrypt: mac check in GCM failed
Stack trace: at Amazon.Extensions.S3.Encryption.Util.AesGcmDecryptStream.ReadAsync(Byte[] buffer, Int32 offset, Int32 count, CancellationToken cancellationToken) in C:\dev\repos\aws-sdk-net\amazon-s3-encryption-client-dotnet\src\Util\AesGcmDecryptStream.cs:line 123
```

This change makes it so we display a friendly error message instead of the above one. 

We also have the alternative option of updating the encryption client itself to add a message https://github.com/aws/amazon-s3-encryption-client-dotnet/blob/7f38b9e07575afb524e2e5b03899834ac8ce5d2c/src/Internal/SetupEncryptionHandler.cs#L237 similar to RANGED GET not being supported, however, to be extra cautious i have made the update in the SDK itself. The reason being is that maybe there is some configuration/edge case where PART GET _does work_ (and somehow i have not been able to figure it out), i didnt want to break those customers. **If we are confident that PART GET does not work, and we think the better solution is to update the encryption client to throw an error message, please let me know and i will do that instead**

I also checked with the java encryption client, and they also do not support PART GET.

ive made backlog item DOTNET-8419 to track adding PART GET support

## Motivation and Context

## Testing
1. add unit tests

## Types of changes

- [] New feature (non-breaking change which adds functionality)

## Checklist

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
@@ -149,7 +149,7 @@ namespace Amazon.S3.Transfer.Internal
             // Validate that S3 encryption client is not being used for multipart downloads
             if (_s3Client is Amazon.S3.Internal.IAmazonS3Encryption)
             {
-                throw new NotSupportedException("Multipart download is not supported when using S3 encryption client. Please use the regular S3 encryption client for multipart download.");
+                throw new NotSupportedException("Multipart download is not supported when using Amazon.S3.Internal.IAmazonS3Encryption client. Please use the Amazon.S3.AmazonS3Client for multipart download.");
             }
 
             _requestEventHandler = requestEventHandler;
--- a/sdk/test/Services/S3/UnitTests/Custom/MultipartDownloadManagerTests.cs
+++ b/sdk/test/Services/S3/UnitTests/Custom/MultipartDownloadManagerTests.cs
@@ -186,7 +186,7 @@ namespace AWSSDK.UnitTests
             }
             catch (NotSupportedException ex)
             {
-                Assert.IsTrue(ex.Message.Contains("Multipart download is not supported when using S3 encryption client. Please use the regular S3 encryption client for multipart download."));
+                Assert.IsTrue(ex.Message.Contains("Multipart download is not supported when using Amazon.S3.Internal.IAmazonS3Encryption client. Please use the Amazon.S3.AmazonS3Client for multipart download."));
             }
         }
 


sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
```diff
@@ -76,12 +76,20 @@ private Logger Logger
...
         {
             _s3Client = s3Client ?? throw new ArgumentNullException(nameof(s3Client));
             _request = request ?? throw new ArgumentNullException(nameof(request));
             _config = config ?? throw new ArgumentNullException(nameof(config));
             _dataHandler = dataHandler ?? throw new ArgumentNullException(nameof(dataHandler));
+
+            // Validate that S3 encryption client is not being used for multipart downloads
+            if (_s3Client is Amazon.S3.Internal.IAmazonS3Encryption)
+            {
+                throw new NotSupportedException("Multipart download is not supported when using S3 encryption client. Please use the regular S3 encryption client for multipart download.");
```
philasmar: You meant to say "Please use the regular S3 client for multipart download."?

GarrettBeatty: yes will update

GarrettBeatty: fixed

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
```diff
@@ -76,12 +76,20 @@ private Logger Logger
...
         {
             _s3Client = s3Client ?? throw new ArgumentNullException(nameof(s3Client));
             _request = request ?? throw new ArgumentNullException(nameof(request));
             _config = config ?? throw new ArgumentNullException(nameof(config));
             _dataHandler = dataHandler ?? throw new ArgumentNullException(nameof(dataHandler));
+
+            // Validate that S3 encryption client is not being used for multipart downloads
+            if (_s3Client is Amazon.S3.Internal.IAmazonS3Encryption)
+            {
+                throw new NotSupportedException("Multipart download is not supported when using S3 encryption client. Please use the regular S3 client for multipart download.");
```
philasmar: nit: wondering if you should use namespaces instead of `S3 encryption client` and `regular S3 client` in case people don't know what you're referring to

GarrettBeatty: ill update the message to use namespace

PR #4174, opened 2025-11-28
[DO NOT MERGE] Update existing integration tests to use async service APIs

dscpinheiro:
Before we update the SDK service integration tests to target .NET Standard, the test methods need to be changed to use the `Async` service APIs (as the synchronous versions are only available for the .NET Framework).

This PR shows what the changes would look like for services with hand-written code and for services that have redundant tests (e.g. generated services - those are covered by protocol tests now).


sdk/test/Services/APIGateway/IntegrationTests/AWSSDK.IntegrationTests.APIGateway.NetFramework.csproj
```diff
@@ -28,10 +28,6 @@
     <PackageReference Include="MSTest.TestAdapter" Version="1.1.18" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.0.0" />
     <PackageReference Include="MSTest.TestFramework" Version="1.1.18" />
-
-    <PackageReference Include="System.Data.SQLite" Version="1.0.97" />
```
dscpinheiro, line 32: I don't know why these references where included here. The project doesn't use any of them.

sdk/test/Services/ApplicationDiscoveryService/IntegrationTests/ApplicationDiscoveryServiceTests.cs
```diff
@@ -1,42 +0,0 @@
...
-using Amazon.Runtime;
-using Amazon.ApplicationDiscoveryService;
-using Amazon.ApplicationDiscoveryService.Model;
-using Amazon;
-using System.IO;
-using System.Text;
-using Amazon.APIGateway.Model;
-using Amazon.APIGateway;
-
-namespace AWSSDK_DotNet.IntegrationTests.Tests
```
GarrettBeatty, line 17: "For services without hand-written code, I'll delete the tests" - are you using a script to identify this or something else

dscpinheiro: I did, but I don't have it saved it anywhere so will need to create it again :)

In my design doc there's a (somewhat outdated) list in the appendix, it's essentially services that have integration tests but nothing in their `Custom` folder (or just a `GlobalSuppressions` file).

PR #4176, opened 2025-12-01
DownloadDirectory Initiated, Failed and Completed Events

GarrettBeatty:
## Description
1. Add DownloadDirectory Initiated, Failed, and Completed events. These are similar to https://github.com/aws/aws-sdk-net/pull/4079 but for directory events. These events are also different than the `ObjectDownloadedFailedEvent` from https://github.com/aws/aws-sdk-net/pull/4151 (see explanation below)

__ObjectDownloadFailedEvent:__

- Fires every time an individual file fails to download
- Think of it as "this specific file couldn't be downloaded"
- Can happen multiple times during one directory download (once per failed file)
- Always includes details about which specific file failed and why

__DownloadDirectoryFailedEvent:__

- Fires once when the entire directory download operation fails
- Think of it as "the whole directory download operation is now stopping"
- Only happens once per directory download attempt
- Includes summary information about the overall operation (how many files succeeded/failed, total progress, etc.)

## Key Scenarios:

1. __With "Continue on Failure" policy:__

   - Individual files fail → `ObjectDownloadFailedEvent` fires for each one
   - Directory download continues and eventually completes
   - `DownloadDirectoryFailedEvent` does NOT fire (operation succeeded overall)

2. __With "Abort on Failure" policy:__

   - First file fails → `ObjectDownloadFailedEvent` fires
   - Operation immediately aborts → `DownloadDirectoryFailedEvent` also fires
   - Both events fire in this case

3. __Pre-download failures (validation, listing S3 objects, etc.):__

   - No individual files have been attempted yet
   - Only `DownloadDirectoryFailedEvent` fires
   - `ObjectDownloadFailedEvent` never fires because no specific file failed

The important distinction is: `ObjectDownloadFailedEvent` is about individual file failures, while `DownloadDirectoryFailedEvent` is about the overall operation failing.

## Motivation and Context
To comply with the SEP. 

## Testing
1. Integration Tests

## Types of changes

## Checklist

## License

### DIFF
--- a/generator/.DevConfigs/7f23582e-3225-487b-83e7-167cf17cb234.json
+++ b/generator/.DevConfigs/7f23582e-3225-487b-83e7-167cf17cb234.json
@@ -1,0 +1,11 @@
+{
+  "services": [
+    {
+      "serviceName": "S3",
+      "type": "minor",
+      "changeLogMessages": [
+      "Added DownloadDirectoryInitiatedEvent, DownloadDirectoryCompletedEvent, and DownloadDirectoryFailedEvent for Amazon.S3.Transfer.TransferUtility.DownloadDirectory."
+      ]
+    }
+  ]
+}


sdk/src/Services/S3/Custom/Transfer/Internal/_async/DownloadCommand.async.cs
```diff
@@ -32,10 +32,10 @@ internal partial class DownloadCommand : BaseCommand<TransferUtilityDownloadResp
     {
         public override async Task<TransferUtilityDownloadResponse> ExecuteAsync(CancellationToken cancellationToken)
         {
-            ValidateRequest();
-            
             FireTransferInitiatedEvent();
```
GarrettBeatty, line 35: moved this to the top just to be consistent

sdk/test/Services/S3/IntegrationTests/TransferUtilityDownloadDirectoryLifecycleTests.cs
```diff
@@ -0,0 +1,342 @@
...
+            };
+
+            request.DownloadDirectoryFailedEvent += eventValidator.OnEventFired;
+
+            try
+            {
+                transferUtility.DownloadDirectory(request);
+                Assert.Fail("Expected an exception to be thrown for invalid bucket name");
+            }
+            catch (Exception ex)
```
dscpinheiro, line 120: Nit: You can use `Assert.ThrowsExceptionAsync` instead of the try/catch block.

sdk/test/Services/S3/IntegrationTests/TransferUtilityDownloadDirectoryLifecycleTests.cs
```diff
@@ -0,0 +1,342 @@
...
+                    Assert.IsNotNull(args.Request);
+                    Assert.IsNotNull(args.Response);
+                    
+                    // Verify progress information is available in completed event
+                    Assert.IsTrue(args.TotalFiles > 0, "TotalFiles should be greater than 0");
+                    Assert.AreEqual(args.TransferredFiles, args.TotalFiles, "All files should be transferred");
+                    Assert.IsTrue(args.TotalBytes > 0, "TotalBytes should be greater than 0");
+                    Assert.AreEqual(args.TransferredBytes, args.TotalBytes, "All bytes should be transferred");
+                    
+                    Console.WriteLine($"DownloadDirectoryCompleted - TotalFiles: {args.TotalFiles}");
```
dscpinheiro: I left similar feedback in one of Phil's PRs: I don't think we should use `Console.WriteLine` in tests. It's OK for debugging locally but our build system will not display them.

If we need to validate something let's use `Asserts` instead.

GarrettBeatty: ill remove the console.writelines

GarrettBeatty: we have asserts for these things in other places

GarrettBeatty: updated to remove console.writeline

PR #4178, opened 2025-12-01
[V4] Update existing service tests to use async operations

dscpinheiro:
Follow up for https://github.com/aws/aws-sdk-net/pull/4177, this PR updates the tests for services with hand-written code to use the `Async` operations. Most of the changes here are similar to `Client.GetRestApi` -> `await Client.GetRestApiAsync` but some important things to call out:
- I did not include S3 or DDB since those are the majority of tests (and there's parallel work on those services). I'll address them in their own PR later
- If the tests were not validating the custom code, I deleted the test project altogether. For example, the hand-written code for CloudFormation is about updating a stack (https://github.com/aws/aws-sdk-net/tree/main/sdk/src/Services/CloudFormation/Custom), but we were testing other operations such as `ValidateTemplate` instead (with a `aws-vs-toolkit.s3.amazonaws.com` URL...)
- If the test was broken and already ignored, I deleted it as well. For example, Bedrock had a test for event streaming but a) the model identifier is not supported anymore, and b) newer models require a different syntax.
- (This is the most controversial one) If the tests are not being run in our build system today, I got rid of them as well. There are some tests that need long-term credentials (i.e. an IAM user) and are marked with a specific category (which we ignore during releases). Note most of them aren't for custom code but I thought I might as well include them in the refactoring.

## Testing
Dry-run: `DRY_RUN-f2020ce8-4c7b-4f82-80aa-24b63d923a66`

## License


sdk/test/Services/APIGateway/IntegrationTests/APIGateway.cs
```diff
@@ -93,37 +87,43 @@ public void TestResources()
             });
         }
 
-        private IEnumerable<Resource> GetResources()
+        private async Task<IEnumerable<Resource>> GetResources()
```
ashishdhingra, line 90: nit: Should we rename this to `GetResourcesAsync()`?

sdk/test/Services/APIGateway/IntegrationTests/APIGateway.cs
```diff
@@ -171,32 +171,39 @@ public void TestRestApiCalls()
             });
         }
 
-        private IEnumerable<RestApi> GetRestApis()
+        private async Task<IEnumerable<RestApi>> GetRestApis()
```
ashishdhingra, line 174: nit: Same as above. Should we rename this to `GetRestApisAsync()`?

sdk/test/Services/APIGateway/IntegrationTests/APIGateway.cs
```diff
@@ -208,25 +215,33 @@ public void TestOtherOperations()
...
 
-            Client.DeleteClientCertificate(new DeleteClientCertificateRequest
+            await Client.DeleteClientCertificateAsync(new DeleteClientCertificateRequest
             {
                 ClientCertificateId = clientCert.ClientCertificateId
             });
         }
 
-        private static IEnumerable<ClientCertificate> GetAllCerts()
+        private static async Task<IEnumerable<ClientCertificate>> GetAllCerts()
```
ashishdhingra, line 227: nit: Rename method to `GetAllCertsAsync()`.

sdk/test/Services/APIGateway/IntegrationTests/APIGateway.cs
dscpinheiro: I could, but I don't think it really matters since this is a private method in a test file (there's a warning during build if I forget to `await`).

Let me know if you feel strongly about it. The larger projects that I haven't addressed yet (DDB and S3) are full of these "helper" methods so it'd increase their (already large) diffs later.

sdk/test/Services/CognitoSync/UnitTests/Custom/CacheItemsTests.cs
```diff
@@ -1,36 +1,13 @@
...
     {
-        [TestCleanup]
-        public void Cleanup()
-        {
-        }
-
-        [ClassCleanup]
-        public static void ClassCleanup()
-        {
-            BaseClean();
```
GarrettBeatty, line 31: was baseclean not doing anything?

dscpinheiro: So this was an integration test that was validating properties via reflection... https://github.com/aws/aws-sdk-net/blob/main/sdk/test/Services/CognitoSync/IntegrationTests/CognitoSync.cs

Since I moved it to an unit test instead there's no need for a `TestBase` anymore, but for reference it just disposes the service client: https://github.com/aws/aws-sdk-net/blob/d7d3513146e381d147745c9f3ab6c85e8e80d6ea/sdk/test/IntegrationTests/Tests/TestBase.cs#L34-L41

sdk/test/Services/S3/IntegrationTests/KMSTests.cs
```diff
@@ -268,9 +265,10 @@ public void TestKmsDssePresignedUrls()
         [TestCategory("S3")]
         public void SpecificKeyTests()
         {
-            var keyId = KeyManagementService.Client.CreateKey(new Amazon.KeyManagementService.Model.CreateKeyRequest
+            var kms = new AmazonKeyManagementServiceClient(Client.Config.RegionEndpoint);
+            var keyId = kms.CreateKey(new Amazon.KeyManagementService.Model.CreateKeyRequest
```
ashishdhingra, line 269: Should this be using `async` version?

sdk/test/Services/APIGateway/IntegrationTests/APIGateway.cs
ashishdhingra: Don't feel strongly about it since most test methods are also named the same way. Although it would have been good to rename these to reflect async version.

sdk/test/Services/S3/IntegrationTests/KMSTests.cs
dscpinheiro: So do not ask me why, but this is a test in the S3 project that was using the client from KMS... I plan to make the same update I made here (sync -> async) for S3 and DDB in their own PRs since those are the largest services (and with the highest number of tests).

ashishdhingra: Sounds good.

ashishdhingra (approved):

GarrettBeatty (approved):

PR #4184, opened 2025-12-02
Update IChatClient with support from latest bedrock runtime / M.E.AI

stephentoub:
## Description

Update IChatClient with support from latest bedrock runtime / M.E.AI

## Motivation and Context

- Adds support for multi-modal tool returns.
- Adds support for citations with URIs.
- Adds a ton of tests verifying IChatClient behavior around the underlying IAmazonBedrockRuntime.

## Testing

The PR includes a lot of tests for the IChatClient implementation, which previously had few-to-none.

## Types of changes

## Checklist

## License


normj:
@stephentoub I refreshed your branch and added a devconfig for the change log. 

There is a class of unit tests failing though converting the `ChatOptions.AdditionalProperties` to `ConverseRequest.AdditionalModelRequestFields`. Looking at the code for `BedrockChatClient.GetResponseAsync` I don't see any code to handle the conversion. Is there a commit missing from your side?

stephentoub:
> @stephentoub I refreshed your branch and added a devconfig for the change log.
> 
> There is a class of unit tests failing though converting the `ChatOptions.AdditionalProperties` to `ConverseRequest.AdditionalModelRequestFields`. Looking at the code for `BedrockChatClient.GetResponseAsync` I don't see any code to handle the conversion. Is there a commit missing from your side?

https://github.com/aws/aws-sdk-net/pull/4226 was merged a couple weeks after I put up this PR, which adds a bunch of tests including for what that PR changed. I'll just remove those tests from here.

stephentoub:
@normj, updated, thanks.

stephentoub:
Updated again, for M.E.AI 10.3.0

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
```diff
@@ -1098,4 +1136,82 @@ private static InferenceConfiguration CreateInferenceConfiguration(InferenceConf
...
+                ReasoningEffort.High => 16384,
+                _ => 32768, // ExtraHigh
+            };
+            inferenceConfig.MaxTokens = budgetTokens * 4;
+        }
+
+        var thinkingConfig = new Document(new Dictionary<string, Document>
+        {
+            ["type"] = new Document("enabled"),
+            ["budget_tokens"] = new Document(budgetTokens),
```
jdaigle: I know that Opus 4.6 properly supports "effort" when using [adaptive thinking](https://docs.aws.amazon.com/bedrock/latest/userguide/claude-messages-adaptive-thinking.html). But it only works when using Opus 4.6, and I suppose there's no good way to check for the model here.

And per documentation, "budget_tokens" could be removed eventually.

Curious if you have any thoughts of how we might support both extended thinking and adaptive thinking in the future, besides the user doing themself with `RawRepresentationFactory`?

stephentoub: I wish AWS was better about smoothing over the experience across models. Having to do completely different things for different models, and having it be an error when you do something the target model doesn't understand, makes it really challenging. You end up either needing to hardcode every possible model into the client, which is both brittle over time and also insufficient when you factor in fine-tuning and the like that enables custom named models, or you push the challenge to every consumer.

In short, I do not have a good answer.

jdaigle: FWIW, I _think_ it's more Anthropic's fault than AWS 😏 

But thanks anyway for your response!

dscpinheiro:
@stephentoub Sorry this one has fallen off our radar, but we're finally getting back to it. One thing that came up while I was discussing this with Norm though is that the `MockBedrockRuntime` class will break our build every time the service adds a new operation (and cause our daily release to fail when that happens).

We definitely want to increase our test coverage, so do you think it'd be possible to leverage the `Moq` dependency the project already uses?

stephentoub:
@dscpinheiro, @normj, updated to use Moq. Also bumped the dependency again, this time to 10.4.0.

dscpinheiro (approved):

GarrettBeatty (approved):

dscpinheiro:
Thanks again for the contribution! This PR will be included in the next release of the SDK.

stephentoub:
Excellent. Thanks.

GarrettBeatty:
we had to temporarily revert this change because the update of the dependencies caused some other issues with our build system. going to discuss with the team during standup on the path to get this merged in

stephentoub:
Ok, if I can help in some way, let me know.

GarrettBeatty:
Making a note to myself and @dscpinheiro  we should reopen this after your other PR is merged in to fix the build dependencies Daniel

GarrettBeatty:
i made new pr https://github.com/aws/aws-sdk-net/pull/4376 but i need to rebase on dev

dscpinheiro:
@stephentoub Apologies for the delay but we finally re-introduced this PR in the latest version of https://www.nuget.org/packages/AWSSDK.Extensions.Bedrock.MEAI in today's release (https://github.com/aws/aws-sdk-net/releases/tag/4.0.243.0).

Norm talked about it in https://github.com/aws/aws-sdk-net/pull/4373#pullrequestreview-4094398495, but in summary the issue was updating `Microsoft.Extensions.AI.Abstractions` also brought in a new version of `System.Text.Json` transitively. In addition to publishing to NuGet, we create a ZIP artifact with all AWS service DLLs (for customers in isolated environments), and the logic for creating that file was not handling this scenario properly.

stephentoub:
Thanks!

PR #4180, opened 2025-12-02
Fix download directory concurrency issue and refactor

GarrettBeatty:
## Description

Fixes the issue where in the previous code this happened

```
 // Creates semaphore with the right limit
fileOperationThrottler = DownloadFilesConcurrently ?
    new SemaphoreSlim(this._config.ConcurrentServiceRequests) :  // e.g., 4
    new SemaphoreSlim(1);

foreach (S3Object s3o in objs) // Say we have 100 files
{
    await fileOperationThrottler.WaitAsync(); // Wait for slot (max 4 waiting)
    
    try
    {
        var task = _failurePolicy.ExecuteAsync(/* download logic */);
        pendingTasks.Add(task); // Queue the task
    }
    finally
    {
        fileOperationThrottler.Release(); // ⚠️ RELEASE IMMEDIATELY!
    }
}

// ⚠️ ALL 100 DOWNLOADS RUN SIMULTANEOUSLY!
await TaskHelpers.WhenAllOrFirstExceptionAsync(pendingTasks, cancellationToken);

```

this was accidentally introduced in https://github.com/aws/aws-sdk-net/pull/4151

I also refactored download directory to be easier to read

## Motivation and Context
https://github.com/aws/aws-sdk-net/issues/3806

## Testing
1. made unit tests which pass on development branch but failed on feature/transfermanager branch before this change. Then after i made the fix they pass

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] My code follows the code style of this project
- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [ ] I have read the **README** document
- [ ] I have added tests to cover my changes
- [x] All new and existing tests passed

## License

- [x] I confirm that this pull request can be released under the Apache 2 license

[issues]: https://github.com/aws/aws-sdk-net/issues
[license]: http://aws.amazon.com/apache2.0/
[cla]: http://en.wikipedia.org/wiki/Contributor_License_Agreement

PR #4186, opened 2025-12-02
refactor upload directory and fix concurrency bug

GarrettBeatty:
same as https://github.com/aws/aws-sdk-net/pull/4180 but for upload directory.

Phils branch already has tests for concurrency which get fixed with this PR


GarrettBeatty:
ill probably have to rebase this anyway since the download pr also has the taskHelpers change too.

PR #4195, opened 2025-12-04
Return search Metadata and Query Insights

irina-herciu:
## Description
Adds support for returning DynamoDB Query/Scan response metadata from DocumentModel Search operations.

## Motivation and Context
This enables callers to access pagination and capacity-related information that is already returned by DynamoDB, without switching to low-level request APIs.

 Query with metadata

```csharp
var search = table.Query(new QueryOperationConfig
{
    HashKey = pk,
    ReturnConsumedCapacity = ReturnConsumedCapacity.TOTAL
});

var docs = await search.GetNextSetAsync();

// Access metadata from the last request
var metadata = search.SearchMetadata;

var lastEvaluatedKey = metadata.LastEvaluatedKey;
var capacityUnits = metadata.ConsumedCapacity?.CapacityUnits;
```

## Types of changes

## Checklist

## License


sdk/src/Services/DynamoDBv2/Custom/DocumentModel/Search.cs
```diff
@@ -797,8 +849,99 @@ internal void Reset()
...
+
+        /// <summary>
+        /// Gets the history of consumed capacity details for all operations performed during the search.
+        /// </summary>
+        public IReadOnlyList<ConsumedCapacity> ConsumedCapacityHistory => _history;
+
+        /// <summary>
+        /// Total capacity units accumulated.
+        /// </summary>
+        public double? TotalCapacityUnits { get; internal set; }
```
GarrettBeatty: nit: for each of these can you add the corresponding href to the metric? and then say something like This is the total capacity units <href-abc> across all operations performed during the search. 

and similar to the other ones. so users can know easily see what each metric is

GarrettBeatty: e.g. for the consumed capacity one can add link to https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/metrics-dimensions.html#ConsumedReadCapacityUnits (assuming this is correct)

sdk/src/Services/DynamoDBv2/Custom/DocumentModel/DocumentOperationRequest.cs
```diff
@@ -124,8 +131,9 @@ public class ScanDocumentOperationRequest : DocumentOperationRequest
         public ScanDocumentOperationRequest()
         {
             Limit = Int32.MaxValue;
+            ConsistentRead = false;
             Select = SelectValues.AllAttributes;
-            TotalSegments = 1;
```
normj, line 128: Why was setting `TotalSegments` to 1 removed and then added `ConsistentRead` to false? Isn't letting `TotalSegments` default to 0 a behavioral change that might be surprising? Keep in mind I don't remember why it was default to 1 to begin with.

sdk/src/Services/DynamoDBv2/Custom/DocumentModel/Search.cs
```diff
@@ -797,8 +849,99 @@ internal void Reset()
...
+
+    /// <summary>
+    /// Provides aggregated metrics and capacity usage information for a multi-page search or query operation.
+    /// </summary>
+    /// <remarks>This class exposes read-only properties that summarize capacity consumption, item counts, and
+    /// scan statistics across all pages retrieved during a search or query. Instances are typically returned by
+    /// operations that support capacity reporting, such as paginated database queries. All properties reflect the
+    /// cumulative or most recent values as appropriate, and are updated as additional pages are processed. This type is
+    /// not intended to be instantiated directly.</remarks>
+    public sealed class SearchMetrics
```
normj: This class seems big enough to put its own `SearchMetrics.cs` file and this `Search.cs` is already big enough.

sdk/src/Services/DynamoDBv2/Custom/DocumentModel/DocumentOperationRequest.cs
irina-herciu: TotalSegments = 1; was my mistake when ScanDocumentOperationRequest was added as I misinterpreted the range constraints. as for ConsistentRead = false it aligns existing default values from ScanOperationConfig.
updated the docs for TotalSegments and Segment to match the ScanOperationConfig.

sdk/src/Services/DynamoDBv2/Custom/DocumentModel/Search.cs
irina-herciu: done.

normj (dismissed):

normj:
Started internal validation build, if it passes I'll approve the PR.

PR #4196, opened 2025-12-04
Fix DownloadPartProgressEventCallback race condition

GarrettBeatty:
Fixes the following race conditon.

**Old Code**

### Thread A (Part 1 finishes) - Arrives First
```
// GetObjectResponse fires: e.IncrementTransferred = 0, e.IsCompleted = true
private void DownloadPartProgressEventCallback(object sender, WriteObjectProgressArgs e)
{
    // e.IncrementTransferred = 0 (stream finished, no new bytes)
    long transferredBytes = Interlocked.Add(ref _totalTransferredBytes, 0);
    // transferredBytes = 20 MB (unchanged)
    
    bool isComplete = false;
    if (transferredBytes >= _totalObjectSize)  // 20 MB >= 20 MB? YES
    {
        // Try to set _completionEventFired from 0 to 1
        int originalValue = Interlocked.CompareExchange(ref _completionEventFired, 1, 0);
        // originalValue = 0 (we're first!)
        // _completionEventFired is now 1
        
        if (originalValue == 0)  // TRUE - we won
        {
            isComplete = true;  // ✅ We will fire completion
        }
    }
    
    // Fire event
    var aggregatedArgs = CreateProgressArgs(0, 20MB, true);  // completed = true
    _userProgressCallback?.Invoke(this, aggregatedArgs);
    // ✅ COMPLETION EVENT FIRED
}

```

### Thread B (Part 2 finishes) - Arrives Microseconds Later

```
// GetObjectResponse fires: e.IncrementTransferred = 0, e.IsCompleted = true
private void DownloadPartProgressEventCallback(object sender, WriteObjectProgressArgs e)
{
    // e.IncrementTransferred = 0 (stream finished, no new bytes)
    long transferredBytes = Interlocked.Add(ref _totalTransferredBytes, 0);
    // transferredBytes = 20 MB (unchanged)
    
    bool isComplete = false;
    if (transferredBytes >= _totalObjectSize)  // 20 MB >= 20 MB? YES
    {
        // Try to set _completionEventFired from 0 to 1
        int originalValue = Interlocked.CompareExchange(ref _completionEventFired, 1, 0);
        // originalValue = 1 (Thread A already set it!)
        // _completionEventFired stays 1
        
        if (originalValue == 0)  // FALSE - we lost
        {
            isComplete = true;  // This line never executes
        }
        // isComplete is still false
        // ⚠️ BUT WE DON'T RETURN - WE CONTINUE!
    }
    
    // Fire event - THIS IS THE BUG
    var aggregatedArgs = CreateProgressArgs(0, 20MB, false);  // completed = false!
    _userProgressCallback?.Invoke(this, aggregatedArgs);
    // ❌ POST-COMPLETION EVENT FIRED: increment=0, total=20MB, completed=FALSE
    // This happens AFTER Thread A already fired the completion event!
}

```

__Thread B should have returned early__ after detecting that `_completionEventFired` was already set, but the old code had no early return. It continued to fire an event with `isComplete=false`, which violated the contract that no events should fire after completion. 

To answer the question on "Why did Thread B fire an event even though all the bytes  have been transferred?" its because in https://github.com/aws/aws-sdk-net/blob/cf2f0787d8149dfcfced66968bf713d5468f72b5/sdk/src/Services/S3/Custom/Model/GetObjectResponse.cs#L770 it will fire a final event at the end (0 bytes transferred). 

----
New  Code

```
private void DownloadPartProgressEventCallback(object sender, WriteObjectProgressArgs e)
{
    long transferredBytes = Interlocked.Add(ref _totalTransferredBytes, e.IncrementTransferred);
    
    // ✅ NEW: Check if completion already fired - exit immediately if so
    if (Interlocked.Read(ref _completionEventFired) == 1)
    {
        return;  // Don't fire any event - we're done!
    }
    
    bool isComplete = false;
    if (transferredBytes == _totalObjectSize)
    {
        int originalValue = Interlocked.CompareExchange(ref _completionEventFired, 1, 0);
        if (originalValue == 0)
        {
            isComplete = true;
        }
        else
        {
            // ✅ NEW: Also return here if we lost the race
            return;
        }
    }
    
    // Only fire event if we haven't completed yet
    var aggregatedArgs = CreateProgressArgs(e.IncrementTransferred, transferredBytes, isComplete);
    _userProgressCallback?.Invoke(this, aggregatedArgs);
}

```

## Motivation and Context

## Testing
I reran the affected test

```
Failed MultipartDownloadProgressTest [1 s]
  Error Message:
   Assert.Fail failed. A progress event was received after completion.

  Stack Trace:
   at AWSSDK_DotNet.IntegrationTests.Tests.S3.TransferUtilityTests.ProgressValidator`1.AssertOnCompletion() in C:\codebuild\tmp\output\src2637311874\src\aws-sdk-net\sdk\test\Services\S3\IntegrationTests\TransferUtilityTests.cs:line 2254
   at AWSSDK_DotNet.IntegrationTests.Tests.S3.TransferUtilityTests.<MultipartDownloadProgressTest>d__62.MoveNext() in C:\codebuild\tmp\output\src2637311874\src\aws-sdk-net\sdk\test\Services\S3\IntegrationTests\TransferUtilityTests.cs:line 1490
```

multiple times and it passed every time

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
@@ -63,7 +63,7 @@ namespace Amazon.S3.Transfer.Internal
         // Atomic flag to ensure completion event fires exactly once
         // Without this, concurrent parts completing simultaneously can both see
         // transferredBytes == _totalObjectSize and fire duplicate completion events
-        // Uses int instead of bool because Interlocked.CompareExchange requires reference types
+        // Uses long instead of bool for compatibility with Interlocked operations
         private long _completionEventFired = 0;  // 0 = false, 1 = true
 
         private readonly Logger _logger = Logger.GetLogger(typeof(MultipartDownloadManager));


sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
```diff
@@ -64,7 +64,7 @@ internal class MultipartDownloadManager : IDownloadManager
         // Without this, concurrent parts completing simultaneously can both see
         // transferredBytes >= _totalObjectSize and fire duplicate completion events
         // Uses int instead of bool because Interlocked.CompareExchange requires reference types
-        private int _completionEventFired = 0;  // 0 = false, 1 = true
+        private long _completionEventFired = 0;  // 0 = false, 1 = true
```
GarrettBeatty, line 67: interlocked.read required a `long`

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
```diff
@@ -733,26 +733,40 @@ private WriteObjectProgressArgs CreateProgressArgs(long incrementTransferred, lo
...
+            // the aggregated completion event has already been fired
+            if (Interlocked.Read(ref _completionEventFired) == 1)
+            {
+                return; // Already completed, don't fire any more events
+            }
+            
             // Use atomic CompareExchange to ensure only first thread fires completion
             bool isComplete = false;
-            if (transferredBytes >= _totalObjectSize)
+            if (transferredBytes == _totalObjectSize)
```
GarrettBeatty, line 752: updated to `==` just to be more clear. the old code `>=` still would've worked fine but may have confused people since transferred bytes can never be over total object size

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
```diff
@@ -733,26 +733,40 @@ private WriteObjectProgressArgs CreateProgressArgs(long incrementTransferred, lo
...
+        /// Uses atomic flag to ensure completion event fires exactly once and prevents any events after completion.
         /// </summary>
         private void DownloadPartProgressEventCallback(object sender, WriteObjectProgressArgs e)
         {
             long transferredBytes = Interlocked.Add(ref _totalTransferredBytes, e.IncrementTransferred);
             
+            // Check if completion was already fired - if so, skip this event entirely
+            // This prevents the race condition where per-part completion events arrive after
+            // the aggregated completion event has already been fired
+            if (Interlocked.Read(ref _completionEventFired) == 1)
```
philasmar, line 745: id say make the check > 0 if there's any situation where _completionEventFired was incremented twice

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
philasmar: maybe defensively add that check and throw an exception?

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
GarrettBeatty: _completionEventFired  doesnt get incremented. its either 0 or 1 since we use `exchange`

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
GarrettBeatty: we have integration tests that assert transferredbytes is never greater than total object size already so i dont think its needed to throw exception

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
```diff
@@ -733,26 +733,40 @@ private WriteObjectProgressArgs CreateProgressArgs(long incrementTransferred, lo
...
+            {
+                return; // Already completed, don't fire any more events
+            }
+            
             // Use atomic CompareExchange to ensure only first thread fires completion
             bool isComplete = false;
-            if (transferredBytes >= _totalObjectSize)
+            if (transferredBytes == _totalObjectSize)
             {
```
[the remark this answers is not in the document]
GarrettBeatty: if we feel its necessary i can add a log but this should theoretically never happen

normj (approved):

PR #4203, opened 2025-12-08
[V4] Update DynamoDB tests to use async operations

dscpinheiro:
Another follow-up for https://github.com/aws/aws-sdk-net/pull/4178, this updates the DDB .NET Framework integration tests to use async operations when possible. I tried to split one file in each commit to make the review more manageable.

## Testing
- Dry-run: `DRY_RUN-82596693-e13f-425a-9515-31a47d1cc7f6`

## License

### DIFF
--- a/sdk/test/Services/DynamoDBv2/IntegrationTests/Cache.cs
+++ b/sdk/test/Services/DynamoDBv2/IntegrationTests/Cache.cs
@@ -34,9 +34,9 @@ namespace AWSSDK_DotNet.IntegrationTests.Tests.DynamoDB
         [TestCleanup]
         public async Task Cleanup()
         {
-            var allTables = DynamoDBTests.Client.ListTables().TableNames;
-            var tableExists = allTables.Contains(TABLENAME);
-            if (tableExists)
+            var allTables = (await DynamoDBTests.Client.ListTablesAsync()).TableNames;
+            
+            if (allTables.Contains(TABLENAME))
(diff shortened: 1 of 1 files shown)


sdk/test/Services/DynamoDBv2/IntegrationTests/CSMTest.cs
dscpinheiro, line 1: This class is not marked with the `TestClass` attribute so the test never runs... I deleted it since it wasn't doing anything special, we have a separate project for CSM tests: https://github.com/aws/aws-sdk-net/tree/main/sdk/test/CSMTest

sdk/test/Services/DynamoDBv2/IntegrationTests/ServiceTests.cs
```diff
@@ -1,228 +1,70 @@
...
 {
     public partial class DynamoDBTests : TestBase<AmazonDynamoDBClient>
     {
         [TestMethod]
         [TestCategory("DynamoDBv2")]
-        public void TestTableCalls()
-        {
-
-            // Only run these tests if we are not reusing tables
-            if (ReuseTables)
```
dscpinheiro, line 22: `ReuseTables` is hard-coded to true so this test is always skipped. Also deleted since several other tests call `Create / DeleteTable` themselves.

peterrsongg (approved):

PR #4204, opened 2025-12-09
Generate DeleteBucketMetadataConfiguration

peterrsongg:
Stacked PRs:
 * #4209
 * #4208
 * #4207
 * #4206
 * #4205
 * __->__#4204

--- --- ---

## Description

Generates `DeleteBucketMetadataConfiguration`

Assembly Comparison Output (empty)
```
C:\Scripts> .\assembly-comparison.ps1

Build succeeded in 16.2s
Build succeeded in 13.2s

Workload updates are available. Run `dotnet workload list` for more information.
```

Fuzz Tests yielded no breaking change
## Testing

DRY_RUN-e581806f-264b-46d0-a331-1318e27c2273
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] My code follows the code style of this project
- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [ ] I have read the **README** document
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed

## License

- [ ] I confirm that this pull request can be released under the Apache 2 license

[issues]: https://github.com/aws/aws-sdk-net/issues
[license]: http://aws.amazon.com/apache2.0/
[cla]: http://en.wikipedia.org/wiki/Contributor_License_Agreement

PR #4205, opened 2025-12-09
Generate ListObjects

peterrsongg:
Stacked PRs:
 * #4209
 * #4208
 * #4207
 * #4206
 * __->__#4205
 * #4204

--- --- ---

## Description

Generates `ListObjects`

Assembly Comparison Output (empty)
```

Build succeeded in 1.1s

Build succeeded in 1.0s
```

Fuzz Tests yielded no breaking change
## Testing

DRY_RUN-e581806f-264b-46d0-a331-1318e27c2273
## Types of changes

## Checklist

## License


peterrsongg:
## READ THIS

None of the aforementioned breaking changes are actually breaking. The bucketName because we check `String.IsNullOrEmpty` when validating required uri params. The BucketName added to the resource path is a hallucination. It got foncused on which is the old and which is the new. None of them ever added BucketName to the resource path.



# Breaking Changes Analysis for Commit ca9c530733713ecbac55b4d163f1f939122774f1

## CRITICAL BREAKING CHANGES FOUND: 2

## POTENTIAL BREAKING CHANGES: 1

---

## __File 1: `sdk/src/Services/S3/Generated/Model/ListObjectsRequest.cs`__

### ❌ BREAKING CHANGE #1: BucketName IsSet Method Logic Changed

__Location:__ `IsSetBucketName()` method

__OLD (Custom):__

```csharp
internal bool IsSetBucketName()
{
    return !String.IsNullOrEmpty(this.bucketName);
}
```

__NEW (Generated):__

```csharp
internal bool IsSetBucketName()
{
    return this._bucketName != null;
}
```

__Impact:__ This is a breaking change for request objects with string properties. The old code treated empty strings (`""`) as NOT SET, while the new code treats empty strings as SET:

- __Before:__ `IsSetBucketName()` returns `false` for both `null` and `""`
- __After:__ `IsSetBucketName()` returns `true` for `""`, `false` only for `null`

This changes validation behavior and could cause API calls with empty bucket names to be processed differently.

---

### ✅ NO ISSUE: Other String Property IsSet Methods Preserved

The following IsSet methods correctly preserve null-only checking (which matches the old behavior for non-bucket properties):

- `IsSetDelimiter()` - `return this._delimiter != null;` (same as before)
- `IsSetMarker()` - `return this._marker != null;` (same as before)
- `IsSetPrefix()` - `return this._prefix != null;` (same as before)
- `IsSetExpectedBucketOwner()` - Correctly customized via injection to use `!String.IsNullOrEmpty`

---

## __File 2: `sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/ListObjectsRequestMarshaller.cs`__

### 🚨 CRITICAL BREAKING CHANGE #2: Bucket Name NOT Added to Request Path

__Location:__ `Marshall()` method

__OLD (Custom - deleted file):__

```csharp
if (!publicRequest.IsSetBucketName())
    throw new AmazonS3Exception("Request object does not have required field BucketName set");
request.AddPathResource("{Bucket}", StringUtils.FromString(publicRequest.BucketName));
request.ResourcePath = "/{Bucket}";
```

__NEW (Generated):__

```csharp
if (string.IsNullOrEmpty(publicRequest.BucketName))
    throw new System.ArgumentException("BucketName is a required property and must be set before making this call.", "ListObjectsRequest.BucketName");
// ... headers and parameters ...
request.ResourcePath = "/";
```

__Impact:__ __THIS IS A CRITICAL FAILURE.__ The bucket name is completely missing from the request path. The API call will be malformed and will FAIL at runtime.

__Expected Request Path:__ `/{BucketName}?delimiter=...&marker=...`\
__Actual Request Path:__ `/?delimiter=...&marker=...`

__Custom Partial Check:__ I verified there are NO custom partial implementations of `ListObjectsRequestMarshaller` with `PreMarshallCustomization` or `PostMarshallCustomization` methods to add the bucket name back. The generated marshaller has the hooks but nothing implements them.

__Result:__ Every call to `ListObjects` will fail because the bucket name is not in the request URL.

---

### ✅ POSITIVE: Query Parameters Properly Marshalled

The generated marshaller correctly adds all query parameters:

- `delimiter`, `encoding-type`, `marker`, `max-keys`, `prefix` - All properly added
- Headers (`x-amz-expected-bucket-owner`, `x-amz-optional-object-attributes`, `x-amz-request-payer`) - All properly added

---

## __File 3: `sdk/src/Services/S3/Generated/Model/ListObjectsResponse.cs`__

### ✅ NO BREAKING CHANGES - Custom Logic Preserved

__Property Name Mapping Preserved:__

- `Contents` renamed to `S3Objects` via customization (matches old behavior)
- `EncodingType` mapped to `Encoding` property (not used in old code, but configuration exists)

__Custom NextMarker Getter Preserved:__

```csharp
public string NextMarker
{
    get { return NextMarkerCustomGetter(); }
    set { this._nextMarker = value; }
}
```

The custom `NextMarkerCustomGetter()` method in the Custom partial file correctly implements the fallback logic:

- If `NextMarker` is not provided by service AND list is truncated AND there are objects, it uses the last object's Key as NextMarker
- This matches the old behavior exactly

---

## __File 4: `sdk/src/Services/S3/Custom/Model/ListObjectsResponse.cs`__

### ✅ NO BREAKING CHANGES - Custom Partial Preserved

The custom partial file correctly implements:

```csharp
private string NextMarkerCustomGetter()
{
    if (String.IsNullOrEmpty(_nextMarker) &&
        _isTruncated.GetValueOrDefault() &&
        (this.S3Objects?.Count > 0))
    {
        int lastObjIdx = this.S3Objects.Count - 1;
        _nextMarker = this.S3Objects[lastObjIdx].Key;
    }
    return this._nextMarker;
}
```

This preserves the original logic that was in the old custom code.

---

## __File 5: `sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/ListObjectsResponseUnmarshaller.cs`__

### ⚠️ POTENTIAL BREAKING CHANGE #3: Exception Construction Method Changed

__Location:__ `UnmarshallException()` method

__OLD (Custom - partially deleted):__

```csharp
if (errorResponse.Code != null && errorResponse.Code.Equals("NoSuchBucket"))
{
    return new NoSuchBucketException(errorResponse.Message);
}
return new AmazonS3Exception(errorResponse.Message, innerException, errorResponse.Type, 
    errorResponse.Code, errorResponse.RequestId, statusCode);
```

__NEW (Generated):__

```csharp
if (errorResponse.Code != null && errorResponse.Code.Equals("NoSuchBucket"))
{
    return NoSuchBucketExceptionUnmarshaller.Instance.Unmarshall(contextCopy, errorResponse);
}
return base.ConstructS3Exception(context, errorResponse, innerException, statusCode);
```

__Impact:__ The exception construction has changed:

1. `NoSuchBucket` exceptions now use an unmarshaller instead of direct instantiation
2. Other exceptions use `ConstructS3Exception()` instead of direct `AmazonS3Exception` instantiation

This could affect error handling if the unmarshaller or `ConstructS3Exception()` constructs exceptions differently. However, this is likely just a refactoring to use common patterns. Needs verification.

---

### ✅ POSITIVE: Custom Contents Unmarshalling Preserved

The generated unmarshaller correctly calls the custom method:

```csharp
if (context.TestExpression("Contents", targetDepth))
{
    ContentsCustomUnmarshall(context, response);
    continue;
}
```

The custom partial file implements `ContentsCustomUnmarshall()` which adds the bucket name to each S3Object:

```csharp
private static void ContentsCustomUnmarshall(XmlUnmarshallerContext context, ListObjectsResponse response)
{
    if (response.S3Objects == null)
    {
        response.S3Objects = new List<S3Object>();
    }
    var s3Object = ContentsItemUnmarshaller.Instance.Unmarshall(context);
    s3Object.BucketName = response.Name;  // CRITICAL: Adds bucket name for PowerShell pipelining
    response.S3Objects.Add(s3Object);
}
```

This preserves the important behavior of populating `BucketName` on each `S3Object` for better PowerShell experience.

---

## __File 6: `sdk/src/Services/S3/Custom/Model/Internal/MarshallTransformations/ListObjectsResponseUnmarshaller.cs`__

### ✅ NO BREAKING CHANGES

The custom partial file is correctly preserved with the `ContentsCustomUnmarshall` method that is called by the generated unmarshaller.

---

## __Files 7-8: Generator Configuration Files__

### ✅ NO BREAKING CHANGES

__Files:__

- `generator/ServiceClientGeneratorLib/ServiceModel.cs`
- `generator/ServiceModels/s3/s3.customizations.json`

These files add configuration to:

- Add ListObjects to the allow list
- Configure property name mappings (`Contents` → `S3Objects`, `EncodingType` → `Encoding`)
- Add custom unmarshalling hooks
- Configure CommonPrefixes type swap to `List<string>`
- Add ExpectedBucketOwner IsSet customization

No breaking changes in configuration.

---

## SUMMARY

__Files Analyzed: 9 out of 9 total files in the commit__

### Breaking Changes by Severity:

1. __CRITICAL (1):__ Missing bucket name in request path - All API calls will fail
2. __BREAKING (1):__ IsSetBucketName logic changed for empty strings
3. __POTENTIAL (1):__ Exception construction method changed (likely just refactoring)

### Files with Issues:

- ❌ `ListObjectsRequest.cs` - 1 breaking change (IsSetBucketName)
- 🚨 `ListObjectsRequestMarshaller.cs` - 1 critical breaking change (missing bucket name in path)
- ⚠️ `ListObjectsResponseUnmarshaller.cs` - 1 potential breaking change (exception construction)

### Files Without Issues:

- ✅ `ListObjectsResponse.cs` (Generated) - Custom logic preserved via injection
- ✅ `ListObjectsResponse.cs` (Custom partial) - Preserved NextMarker custom getter
- ✅ `ListObjectsResponseUnmarshaller.cs` (Custom partial) - Preserved Contents custom unmarshalling
- ✅ `ServiceModel.cs` - Configuration only
- ✅ `s3.customizations.json` - Configuration only

__RECOMMENDATION:__ The missing bucket name in the request path (Breaking Change #2) MUST be fixed before this code can be merged. This is a blocking issue that will cause ALL ListObjects API calls to fail with 404 or similar errors.

__Note:__ The custom unmarshalling logic for Contents and NextMarker was correctly preserved through customization hooks, which is excellent. The marshaller issue needs the same treatment - either the bucket name needs to be added in generated code, or a custom partial implementation must be created.

sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/ListObjectsRequestMarshaller.cs
```diff
@@ -0,0 +1,120 @@
...
+            request.HttpMethod = "GET";
+        
+            if (publicRequest.IsSetExpectedBucketOwner()) 
+            {
+                request.Headers["x-amz-expected-bucket-owner"] = publicRequest.ExpectedBucketOwner;
+            }
+        
+            if (publicRequest.IsSetOptionalObjectAttributes()) 
+            {
+                        request.Headers["x-amz-optional-object-attributes"] = StringUtils.FromList(publicRequest.OptionalObjectAttributes);
```
dscpinheiro, line 69: Nit but the indentation on this line looks odd (more spaces than the other branches)

sdk/src/Services/S3/Generated/Model/ListObjectsRequest.cs
```diff
@@ -12,22 +12,43 @@
...
 using Amazon.Runtime;
 using Amazon.Runtime.Internal;
 
+#pragma warning disable CS0612,CS0618,CS1570
 namespace Amazon.S3.Model
 {
     /// <summary>
     /// Container for the parameters to the ListObjects operation.
-    /// <note> 
+    /// <important>
```
dscpinheiro: I had never seen service models use `<important>`... Does the doc generator handle it properly? (you can check by getting the docs from the dry-run)

sdk/src/Services/S3/Custom/Model/Internal/MarshallTransformations/ListObjectsResponseUnmarshaller.cs
```diff
@@ -12,153 +12,36 @@
...
             }
-                            
-
 
+            // adding the bucket name into the S3Object instance enables
+            // a better pipelining experience in PowerShell
+            var s3Object = ContentsItemUnmarshaller.Instance.Unmarshall(context);
+            s3Object.BucketName = response.Name;
+            response.S3Objects.Add(s3Object);
             return;
```
dscpinheiro: I know it was there before but this return is redundant.

sdk/src/Services/S3/Generated/Model/ListObjectsResponse.cs
```diff
@@ -0,0 +1,307 @@
...
+        /// </para>
+        ///  <note> 
+        /// <para>
+        /// When using the URL encoding type, non-ASCII characters that are used in an object's
+        /// key name will be percent-encoded according to UTF-8 code values. For example, the
+        /// object <c>test_file(3).png</c> will appear as <c>test_file%283%29.png</c>.
+        /// </para>
+        ///  </note>
+        /// </summary>
+        public EncodingType EncodingType
```
dscpinheiro: Since this was missing from the hand-written class, do you think it'd be worth to customize it to `Encoding` as well? (to match the request class)

I don't feel strongly about either option.

sdk/src/Services/S3/Generated/Model/ListObjectsRequest.cs
peterrsongg: Image

doesn't seem to... i think we should add a backlog for it

sdk/src/Services/S3/Generated/Model/ListObjectsResponse.cs
peterrsongg: I fixed this.. idk why it isn't showing up in the file diff but it's in the latest commit:
https://github.com/aws/aws-sdk-net/commit/cdbd7d2bf4a702a90ec7afbb4ec86e4f4f30b279

sdk/src/Services/S3/Custom/Model/Internal/MarshallTransformations/ListObjectsResponseUnmarshaller.cs
peterrsongg: fixed this too here: https://github.com/aws/aws-sdk-net/commit/cdbd7d2bf4a702a90ec7afbb4ec86e4f4f30b279

muhammad-othman (approved):

sdk/src/Services/S3/Generated/Model/ListObjectsRequest.cs
peterrsongg: i added a backlog item for it

sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/ListObjectsRequestMarshaller.cs
peterrsongg: I tried fixing this but it ends up changing a lot of other services so i didn't. I will change it in a separate PR

dscpinheiro (approved):

PR #4208, opened 2025-12-09
generate RestoreObject

peterrsongg:
Stacked PRs:
 * #4209
 * __->__#4208
 * #4207
 * #4206
 * #4205
 * #4204

--- --- ---

## Description

Generates `RestoreObject`

Assembly Comparison Output (new properties and types added, which isn't a breaking change)
```
AssemblyComparer AWSSDK.S3.New.dll: Message Amazon.S3.Model.RestoreRequest/TypeAdded: New type Amazon.S3.Model.RestoreRequest
AssemblyComparer AWSSDK.S3.New.dll: Message Amazon.S3.Model.GlacierJobParameters/TypeAdded: New type Amazon.S3.Model.GlacierJobParameters
AssemblyComparer AWSSDK.S3.New.dll: Message Amazon.S3.Model.ObjectAlreadyInActiveTierErrorException/TypeAdded: New type Amazon.S3.Model.ObjectAlreadyInActiveTierErrorException
```

The fuzz test caught one issue where I made a mistake in the TierCustomMarshall, but that has been fixed.
## Testing

DRY RUN: DRY_RUN-e581806f-264b-46d0-a331-1318e27c2273
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] My code follows the code style of this project
- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [ ] I have read the **README** document
- [x] I have added tests to cover my changes
- [x] All new and existing tests passed

## License

- [x] I confirm that this pull request can be released under the Apache 2 license

[issues]: https://github.com/aws/aws-sdk-net/issues
[license]: http://aws.amazon.com/apache2.0/
[cla]: http://en.wikipedia.org/wiki/Contributor_License_Agreement

PR #4207, opened 2025-12-09
generate ListMultipartUploads

peterrsongg:
Stacked PRs:
 * #4209
 * #4208
 * __->__#4207
 * #4206
 * #4205
 * #4204

--- --- ---

## Description

Generates `DeleteBucketMetadataConfiguration`

Assembly Comparison Output ( new property added, which isn't a breaking change)
```
AssemblyComparer AWSSDK.S3.New.dll: Message Amazon.S3.Model.ListMultipartUploadsResponse/MethodAdded: New method System.Void set_EncodingType(Amazon.S3.EncodingType) in Amazon.S3.Model.ListMultipartUploadsResponse
AssemblyComparer AWSSDK.S3.New.dll: Message Amazon.S3.Model.ListMultipartUploadsResponse/MethodAdded: New method Amazon.S3.EncodingType get_EncodingType() in Amazon.S3.Model.ListMultipartUploadsResponse
```

Fuzz Tests yielded no breaking change
## Testing

DRY RUN: DRY_RUN-e581806f-264b-46d0-a331-1318e27c2273
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] My code follows the code style of this project
- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [ ] I have read the **README** document
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed

## License

- [ ] I confirm that this pull request can be released under the Apache 2 license

[issues]: https://github.com/aws/aws-sdk-net/issues
[license]: http://aws.amazon.com/apache2.0/
[cla]: http://en.wikipedia.org/wiki/Contributor_License_Agreement

PR #4206, opened 2025-12-09
Generate UpdateBucketMetadataTableConfiguration

peterrsongg:
Stacked PRs:
 * #4209
 * #4208
 * #4207
 * __->__#4206
 * #4205
 * #4204

--- --- ---

## Description

Generates `UpdateBucketMetadataTableConfiguration`. There are breaking changes in this, but that is because the operation wasn't working before. This will be called out in the changelog as necessary to make the operation work.

Assembly Comparison Output (empty)
```
AssemblyComparer AWSSDK.S3.New.dll: Error Amazon.S3.Model.JournalTableConfigurationUpdates/MethodRemoved: Missing method Amazon.S3.JournalConfigurationState get_ConfigurationState() in Amazon.S3.Model.JournalTableConfigurationUpdates
AssemblyComparer AWSSDK.S3.New.dll: Error Amazon.S3.Model.JournalTableConfigurationUpdates/MethodRemoved: Missing method System.Void set_ConfigurationState(Amazon.S3.JournalConfigurationState) in Amazon.S3.Model.JournalTableConfigurationUpdates
AssemblyComparer AWSSDK.S3.New.dll: Error Amazon.S3.Model.JournalTableConfigurationUpdates/MethodRemoved: Missing method System.Void set_EncryptionConfiguration(Amazon.S3.Model.MetadataTableEncryptionConfiguration) in Amazon.S3.Model.JournalTableConfigurationUpdates
AssemblyComparer AWSSDK.S3.New.dll: Message Amazon.S3.Model.JournalTableConfigurationUpdates/MethodAdded: New method Amazon.S3.Model.RecordExpiration get_RecordExpiration() in Amazon.S3.Model.JournalTableConfigurationUpdates
AssemblyComparer AWSSDK.S3.New.dll: Message Amazon.S3.Model.JournalTableConfigurationUpdates/MethodAdded: New method System.Void set_RecordExpiration(Amazon.S3.Model.RecordExpiration) in Amazon.S3.Model.JournalTableConfigurationUpdates
AssemblyComparer AWSSDK.S3.New.dll: Error Amazon.S3.Model.JournalTableConfigurationUpdates/MethodRemoved: Missing method Amazon.S3.Model.MetadataTableEncryptionConfiguration get_EncryptionConfiguration() in Amazon.S3.Model.JournalTableConfigurationUpdates

Workload updates are available. Run `dotnet workload list` for more information.
```

Fuzz tests don't apply here since the request fails
## Testing

[...]

PR #4209, opened 2025-12-09
Generate CompleteMultipartUpload

peterrsongg:
Stacked PRs:
 * __->__#4209
 * #4208
 * #4207
 * #4206
 * #4205
 * #4204

--- --- ---

## Description

Generates `CompleteMultipartUpload`

Before in `s3.customizations.json` we were excluding `MultipartUpload` and injecting `PartETags`. Since s3 was not generated, this change didn't really matter. This would only matter when we generate the simple methods for an operation from the s3.customizations.simpleforms.json but in this case there is no simple method that has `PartETags` as a param. The issue with this approach is that if the service team updates `MultipartUpload` and adds member, the now generated S3 will miss this update. Instead, we take the correct approach which is to flatten `MultipartUpload` so that its members are added to the top-level shape. This way, any updates to `MultipartUpload` are generated as well.

Assembly Comparison Output (MpuObject size changed to a nulalble int) will call this out in changelog
```
AssemblyComparer AWSSDK.S3.New.dll: Error Amazon.S3.Model.CompleteMultipartUploadRequest/MethodRemoved: Missing method System.Int64 get_MpuObjectSize() in Amazon.S3.Model.CompleteMultipartUploadRequest
AssemblyComparer AWSSDK.S3.New.dll: Message Amazon.S3.Model.CompleteMultipartUploadRequest/MethodAdded: New method System.Nullable<System.Int64> get_MpuObjectSize() in Amazon.S3.Model.CompleteMultipartUploadRequest
AssemblyComparer AWSSDK.S3.New.dll: Error Amazon.S3.Model.CompleteMultipartUploadRequest/MethodRemoved: Missing method System.Void set_MpuObjectSize(System.Int64) in Amazon.S3.Model.CompleteMultipartUploadRequest
AssemblyComparer AWSSDK.S3.New.dll: Message Amazon.S3.Model.CompleteMultipartUploadRequest/MethodAdded: New method System.Void set_MpuObjectSize(System.Nullable<System.Int64>) in Amazon.S3.Model.CompleteMultipartUploadRequest
AssemblyComparer AWSSDK.S3.New.dll: Message Amazon.S3.Model.CompletedMultipartUpload/TypeAdded: New type Amazon.S3.Model.CompletedMultipartUpload
```

## Testing

DRY RUN: DRY_RUN-e581806f-264b-46d0-a331-1318e27c2273
## Types of changes

## Checklist

## License


sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/CompleteMultipartUploadRequestMarshaller.cs
```diff
@@ -0,0 +1,212 @@
...
+            if (publicRequest.IsSetUploadId())
+                request.Parameters.Add("uploadId", StringUtils.FromString(publicRequest.UploadId));
+            request.ResourcePath = "/{Key+}";
+            var stringWriter = new XMLEncodedStringWriter(CultureInfo.InvariantCulture);
+            using (var xmlWriter = XmlWriter.Create(stringWriter, new XmlWriterSettings() { Encoding = System.Text.Encoding.UTF8, OmitXmlDeclaration = true, NewLineHandling = NewLineHandling.Entitize }))
+            {   
+                SortPartsList(publicRequest);
+                    xmlWriter.WriteStartElement("CompleteMultipartUpload", "http://s3.amazonaws.com/doc/2006-03-01/");
+                    var publicRequestPartETags = publicRequest.PartETags;
+                    if (publicRequestPartETags != null && (publicRequestPartETags.Count > 0 || !AWSConfigs.InitializeCollections))
```
dscpinheiro: I haven't looked at all changes yet, but I had a similar comment in another PR: Why not use the IsSet method that already has this logic?

Also, the indentation on this method is really odd (and much harder to read than the hand-written file).

generator/ServiceModels/s3/s3.customizations.json
```diff
@@ -43,16 +43,22 @@
             ]
         },
         "CompleteMultipartUploadRequest": {
-            "exclude": [
```
dscpinheiro, line 46: Can you update the PR description explaining this? I don't follow why we were excluding something before but not it's in `flattenShapes`.

peterrsongg: added in the PR description

sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/CompleteMultipartUploadRequestMarshaller.cs
peterrsongg: that's a good question. For some reason the rest-xml payload marshalling logic doesn't use `IsSet` and does these `if` statements. This logic has existed before any of the S3 generation logic. I didn't want to change it because it is out of scop and will potentially affect all rest-xml services

sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/CompleteMultipartUploadRequestMarshaller.cs
```diff
@@ -0,0 +1,212 @@
...
+                                xmlWriter.WriteElementString("ChecksumSHA1", StringUtils.FromString(publicRequestPartETagsValue.ChecksumSHA1));
+                            if(publicRequestPartETagsValue.IsSetChecksumSHA256())
+                                xmlWriter.WriteElementString("ChecksumSHA256", StringUtils.FromString(publicRequestPartETagsValue.ChecksumSHA256));
+                            if(publicRequestPartETagsValue.IsSetETag())
+                                xmlWriter.WriteElementString("ETag", StringUtils.FromString(publicRequestPartETagsValue.ETag));
+                            if(publicRequestPartETagsValue.IsSetPartNumber())
+                                xmlWriter.WriteElementString("PartNumber", StringUtils.FromInt(publicRequestPartETagsValue.PartNumber.Value));
+                            xmlWriter.WriteEndElement();
+                        }
+                        }
```
dscpinheiro, line 171: I know I'm being annoying, but after we're done generating S3 we should really fix the indentation in these classes...

sdk/src/Services/S3/Custom/Model/Internal/MarshallTransformations/CompleteMultipartUploadRequestMarshaller.cs
```diff
@@ -31,162 +31,20 @@ namespace Amazon.S3.Model.Internal.MarshallTransformations
...
-            if (string.IsNullOrEmpty(completeMultipartUploadRequest.Key))
-                throw new System.ArgumentException("Key is a required property and must be set before making this call.", "CompleteMultipartUploadRequest.Key");
-            request.AddPathResource("{Key+}", S3Transforms.ToStringValue(completeMultipartUploadRequest.Key));
-            request.ResourcePath = "/{Key+}";
-
-            request.AddSubResource("uploadId", S3Transforms.ToStringValue(completeMultipartUploadRequest.UploadId));
-
-            var stringWriter = new XMLEncodedStringWriter(System.Globalization.CultureInfo.InvariantCulture);
-            using (var xmlWriter = XmlWriter.Create(stringWriter, new XmlWriterSettings() { Encoding = Encoding.UTF8, OmitXmlDeclaration = true, NewLineHandling = NewLineHandling.Entitize }))
+            if (publicRequest.PartETags != null && (publicRequest.PartETags.Count > 0 || !AWSConfigs.InitializeCollections))
```
dscpinheiro, line 38: Could use `IsSet` here now, right?

sdk/src/Services/S3/Generated/Model/CompletedMultipartUpload.cs
```diff
@@ -0,0 +1,68 @@
...
+using Amazon.Runtime;
+using Amazon.Runtime.Internal;
+
+#pragma warning disable CS0612,CS0618,CS1570
+namespace Amazon.S3.Model
+{
+    /// <summary>
+    /// The container for the completed multipart upload details.
+    /// </summary>
+    public partial class CompletedMultipartUpload
```
dscpinheiro: Where is this class used? I don't see it referenced in the request or response... The request has its own `private List<PartETag> _partETags` field.

sdk/src/Services/S3/Custom/Model/Internal/MarshallTransformations/CompleteMultipartUploadRequestMarshaller.cs
peterrsongg: yup, rebased on dev and ran the generator so that this uses the `IsSet` methods.

sdk/src/Services/S3/Generated/Model/CompletedMultipartUpload.cs
peterrsongg: It actually isn't used anywhere at the moment, since the custom code had flattened this shape and moved it up to the response. 

We could remove it altogether by including it in the `excludeShapes` array, but I left it in on purpose in the off chance that this shape is used in another operation later down the line. If another operation were to use this shape then the generator would fail at runtime.

sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/CompleteMultipartUploadRequestMarshaller.cs
peterrsongg: No worries, I'll backlog it for now as an LHF and open a separate PR to fix the spacing

peterrsongg: ended up just fixing it here:

https://github.com/aws/aws-sdk-net/pull/4256

sdk/src/Services/S3/Generated/Model/CompletedMultipartUpload.cs
peterrsongg: I went and checked and decided to remove it. Looking at the docs it looks like it is clearly stated as a container for a multipart upload, so it shouldn't be used anywhere else.

PR #4210, opened 2025-12-09
Generate PutBucketTagging

peterrsongg:
## Description

Generates `PutBucketTagging`

Assembly Comparison Output (MpuObject size changed to a nulalble int) will call this out in changelog
```

```

## Testing

DRY RUN: DRY_RUN-e581806f-264b-46d0-a331-1318e27c2273
## Types of changes

## Checklist

## License


sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/PutObjectTaggingRequestMarshaller.cs
```diff
@@ -99,16 +99,13 @@ public IRequest Marshall(PutObjectTaggingRequest publicRequest)
                     {
                         xmlWriter.WriteStartElement("TagSet");
                         foreach (var publicRequestTaggingTagSetValue in publicRequestTaggingTagSet) 
-                        {
-                        if (publicRequestTaggingTagSetValue != null)
```
peterrsongg, line 103: this change is intentional. I tested that the behavior is the same service side if a null value is set for `TagSet` but the original custom code uses this custom method

https://github.com/aws/aws-sdk-net/blob/4.0.0.0/sdk/src/Services/S3/Custom/Model/Internal/MarshallTransformations/PutObjectTaggingRequestMarshaller.cs#L70

which eventually also does this same thing. Behavior wise, nothing changed but since PutBucketTagging also needs to do the same thing and it shares this shape. Adding the customization `skipXmlIsSet` affects other operations too.

sdk/src/Services/S3/Generated/Model/Internal/MarshallTransformations/RestoreObjectRequestMarshaller.cs
```diff
@@ -128,16 +128,13 @@ public IRequest Marshall(RestoreObjectRequest publicRequest)
                                 {
                                     xmlWriter.WriteStartElement("TagSet");
                                     foreach (var publicRequestOutputLocationS3TaggingTagSetValue in publicRequestOutputLocationS3TaggingTagSet) 
-                                    {
-                                    if (publicRequestOutputLocationS3TaggingTagSetValue != null)
```
peterrsongg, line 132: same comment as above. behavior is the same

PR #4218, opened 2025-12-10
fix http concurrency

GarrettBeatty:
Stacked PRs:
 * #4224
 * #4220
 * #4219
 * __->__#4218

--- --- ---

## Description
Fix HTTP concurrency for multi part download. Previously we were releasing the http semaphore slot after we got the http response but _before_ we read the response body (wrote it to a file/stream). This meant that it was possible for there to be more active `ConcurrentServiceRequests` than the user actually wanted.

The fix was to move the releasing of the http semaphore slot _after_ processing the part. This ensured that its only released once its written to the file/stream and ensures only `ConcurrentServiceRequests` are active at a time.

## Motivation and Context

## Testing
1. I added unit tests such as `HttpSemaphore_HeldThroughProcessPartAsync` which verify that the http semaphore is only released after process part is executed. In order to ensure this test actually checks the issue, i ran it before my fix and the test failed (as expected). After my fix, the test passes. 
2. dry run 

```
"build_id": "7e8b2922-9aab-4b1c-8cf2-ee8c6ad6b8d1",
    "build_type": "DRY_RUN",
    "build_status": "SUCCEEDED",
    "created_time": "2025-12-09T19:44:19.400Z",
    "last_modified_time": "2025-12-09T22:47:24.680Z",
```

Before (we can see concurrency is over the set 10)
image

After, we can see concurrency stays at or below 10
image

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
@@ -173,6 +173,41 @@ namespace Amazon.S3.Transfer.Internal
             }
         }
 
+        /// <summary>
+        /// Discovers the download strategy (single-part vs multipart) by making an initial GetObject request.
+        /// </summary>
+        /// <param name="cancellationToken">Cancellation token to cancel the discovery operation.</param>
+        /// <returns>
+        /// A <see cref="DownloadDiscoveryResult"/> containing information about the object size, part count,
+        /// and the initial GetObject response.
+        /// </returns>
+        /// <remarks>
+        /// <para><strong>IMPORTANT - HTTP Semaphore Lifecycle:</strong></para>
+        /// <para>
+        /// This method acquires an HTTP concurrency slot from the configured semaphore and downloads Part 1.
+        /// The semaphore slot is <strong>HELD</strong> until <see cref="StartDownloadsAsync"/> completes processing Part 1.
+        /// Callers <strong>MUST</strong> call <see cref="StartDownloadsAsync"/> after this method to release the semaphore.
+        /// Failure to call <see cref="StartDownloadsAsync"/> will cause the semaphore slot to remain held indefinitely,
+        /// potentially blocking other downloads and causing deadlocks.
+        /// </para>
+        /// <para><strong>Concurrency Implications:</strong></para>
+        /// <para>
+        /// With limited HTTP concurrency (e.g., <c>ConcurrentServiceRequests=1</c> for shared throttlers in directory downloads),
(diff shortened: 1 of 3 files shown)


sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
```diff
@@ -173,6 +173,41 @@ public Exception DownloadException
             }
         }
 
+        /// <summary>
```
GarrettBeatty, line 176: i think we should eventually refactor this to get rid of the extra discoverdownloadstrategy function. since its a little confusing that the semaphore releases go across functions (between discoverdownloadstrategy and startdownload). i plan on doing that after we fix all of the bugs

sdk/test/Services/S3/UnitTests/Custom/MultipartDownloadManagerTests.cs
```diff
@@ -1951,74 +1951,6 @@ public async Task DiscoverUsingRangeStrategy_CallsWaitForCapacityAsync()
             mockDataHandler.Verify(x => x.WaitForCapacityAsync(It.IsAny<CancellationToken>()), Times.Once);
         }
 
-        [TestMethod]
-        public async Task DiscoverUsingPartStrategy_AcquiresAndReleasesHttpSlot()
```
GarrettBeatty, line 1955: these are no longer applicable since they asserted old (and incorrect behavior)

sdk/test/Services/S3/UnitTests/Custom/MultipartDownloadManagerTests.cs
```diff
@@ -2044,16 +1976,22 @@ public async Task MultipleDownloads_WithSharedHttpThrottler_RespectsLimits()
             var coordinator1 = new MultipartDownloadManager(mockClient1.Object, request1, config, mockDataHandler1.Object, null, sharedThrottler);
             var coordinator2 = new MultipartDownloadManager(mockClient2.Object, request2, config, mockDataHandler2.Object, null, sharedThrottler);
 
-            // Act - Start both discoveries concurrently
-            var task1 = coordinator1.DiscoverDownloadStrategyAsync(CancellationToken.None);
-            var task2 = coordinator2.DiscoverDownloadStrategyAsync(CancellationToken.None);
+            var discovery1 = await coordinator1.DiscoverDownloadStrategyAsync(CancellationToken.None);
```
GarrettBeatty, line 1979: had to update these because now with new change the semaphore is held through start download, so callers need to do discover + start download in order to get through the whole http cycle

normj (approved):

PR #4220, opened 2025-12-10
use max size semaphore

GarrettBeatty:
Stacked PRs:
 * #4224
 * __->__#4220

--- --- ---

## Description
I noticed during testing using AI that it said the maxinmemoryparts being used from the logs was 1026 even though i had configured it to 1024. this means i was:
1. Releasing the semaphore twice
2. Not enforcing a maximum amount for the sempahore

This change updates the code to fix the double release and also ensure the semaphores enforce a maximum.

## Motivation and Context

## Testing
1. Reran all unit tests and they pass
4. Reran performance tests and they have same performance
6. dry run in progress 16f30d8d-ec62-40bb-ab69-5f50383ed393 pass

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/Transfer/Internal/BufferedPartDataHandler.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/Internal/BufferedPartDataHandler.cs
@@ -153,9 +153,6 @@ namespace Amazon.S3.Transfer.Internal
                 // If ReleaseBufferSpace() throws, we no longer own the data source, so we won't dispose it
                 streamingDataSource = null;
 
-                // Release capacity immediately since we're not holding anything in memory
-                _partBufferManager.ReleaseBufferSpace();
-
                 _logger.DebugFormat("BufferedPartDataHandler: [Part {0}] StreamingDataSource added and capacity released",
                     partNumber);
             }
--- a/sdk/test/Services/S3/UnitTests/Custom/BufferedPartDataHandlerTests.cs
+++ b/sdk/test/Services/S3/UnitTests/Custom/BufferedPartDataHandlerTests.cs
@@ -463,10 +463,6 @@ namespace AWSSDK.UnitTests
                 // Part 3 should use buffering path (ArrayPool allocation)
                 Assert.AreEqual(1, bufferedPartNumbers.Count, "Expected exactly 1 part to be buffered");
                 Assert.AreEqual(3, bufferedPartNumbers[0], "Part 3 should be buffered");
-                
-                // Verify ReleaseBufferSpace was called for streaming path (immediate capacity release)
-                mockBufferManager.Verify(m => m.ReleaseBufferSpace(), Times.Once, 
-                    "Streaming path should release capacity immediately");
             }
             finally
             {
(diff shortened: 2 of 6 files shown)


sdk/test/Services/S3/UnitTests/Custom/BufferedPartDataHandlerTests.cs
```diff
@@ -463,10 +463,6 @@ public async Task ProcessPartAsync_InOrderVsOutOfOrder_VerifyStreamingVsBufferin
                 // Part 3 should use buffering path (ArrayPool allocation)
                 Assert.AreEqual(1, bufferedPartNumbers.Count, "Expected exactly 1 part to be buffered");
                 Assert.AreEqual(3, bufferedPartNumbers[0], "Part 3 should be buffered");
-                
-                // Verify ReleaseBufferSpace was called for streaming path (immediate capacity release)
-                mockBufferManager.Verify(m => m.ReleaseBufferSpace(), Times.Once,
```
GarrettBeatty, line 468: this isnt true anymore since we hold it until the reader consumes it which is why i removed it

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedPartDataHandler.cs
```diff
@@ -153,9 +153,6 @@ private void ProcessStreamingPart(
                 // If ReleaseBufferSpace() throws, we no longer own the data source, so we won't dispose it
                 streamingDataSource = null;
 
-                // Release capacity immediately since we're not holding anything in memory
-                _partBufferManager.ReleaseBufferSpace();
```
GarrettBeatty, line 157: this was being released here and then also when the reader read the part later on. so we remove this to fix the double release.

Side note: 
The original reason i had this hear was because my intent was not count streaming to the user as "holding the part in memory". But for simplicitys sake and this bug fix, we now consider a part that gets directly streamed to the user as "taking up memory". in the future we could probably optimize/change this if we want.

sdk/src/Services/S3/Custom/Transfer/Internal/BufferedPartDataHandler.cs
```diff
@@ -153,9 +153,6 @@ private void ProcessStreamingPart(
...
-
                 _logger.DebugFormat("BufferedPartDataHandler: [Part {0}] StreamingDataSource added and capacity released",
                     partNumber);
```
GarrettBeatty, line 157: this was being released here and then also when the reader read the part later on. so we remove this to fix the double release.

Side note: 
The original reason i had this hear was because my intent was not count streaming to the user as "holding the part in memory". But for simplicity's takes and this bug fix we now consider a part that gets directly streamed to the user as "taking up memory". in the future we could probably optimize/change this if we want.

sdk/test/Services/S3/UnitTests/Custom/MultipartDownloadManagerTests.cs
```diff
@@ -601,25 +601,19 @@ public async Task DiscoverUsingRangeStrategy_CalculatesPartCount()
         public async Task StartDownloadsAsync_SinglePart_ReturnsImmediately()
         {
             // Arrange
-            var mockClient = MultipartDownloadTestHelpers.CreateMockS3Client();
+            var mockResponse = MultipartDownloadTestHelpers.CreateSinglePartResponse(1024);
```
GarrettBeatty, line 604: had to update all of these tests to call DiscoverDownloadStrategyAsync/reservere buffer space before doing any action that would release buffer space. otherwise it was getting exceeded semaphore maximum error (which is expected now)

PR #4222, opened 2025-12-10
Reduce object allocations in UserCrypto.Decrypt by 98% for .NET8+

peterrsongg:
## Description

This optimization is primarily for windows users. `UserCrypto.Decrypt` is a hot path for windows machines, and it was doing unnecessary string allocations. Here is the code snippet that was doing the allocations.
```
List<byte> dataIn = new List<byte>();
for (int i = 0; i < encrypted.Length; i = i + 2)
{
    byte data = Convert.ToByte(encrypted.Substring(i, 2), 16);
    dataIn.Add(data);
}
```
For each iteration we were: allocating a new 2-char string via Substring(i, 2), call Convert.ToByte(string, 16) which parses that string. So if the hex string is length ~9,760, the loop runs ~4,880 times and creates 4,880 tiny strings. (which was my case). This method is called when decrypting any setting stored in the persistence manager. 
call-stack

In newer versions of .NET (.NET 5+) we can just use `Convert.FromHexString` and do it all in one byte array and much faster.

In my profiling of looping a PutObject call 10 times, I saw the allocations for this specific method go down by 98% and in total by 23%. 

## Motivation and Context

Performance improvement for windows users
## Testing

Dry run passed

Pre Optimization
user-crypto-optimization
Post Optimization

User-Crypto-post-optimization

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Internal/Settings/UserCrypto.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Settings/UserCrypto.cs
@@ -35,7 +35,7 @@ namespace Amazon.Runtime.Internal.Settings
         {
             byte[] dataIn;
 
-#if NET8_0_OR_GREATER
+#if NET5_0_OR_GREATER
             dataIn = Convert.FromHexString(encrypted);
 #else
     // Legacy behavior preserved for older TFMs


sdk/src/Core/Amazon.Runtime/Internal/Settings/UserCrypto.cs
```diff
@@ -33,21 +33,28 @@ public static class UserCrypto
...
-            {
-                byte data = Convert.ToByte(encrypted.Substring(i, 2), 16);
-                dataIn.Add(data);
-            }
+            byte[] dataIn;
+
+#if NET8_0_OR_GREATER
+            dataIn = Convert.FromHexString(encrypted);
+#else
+    // Legacy behavior preserved for older TFMs
```
dscpinheiro, line 41: Nit: Fix the indentation for the legacy behavior here.

generator/.DevConfigs/a3466402-916b-4e40-8822-79b28ab5de33.json
```diff
@@ -0,0 +1,9 @@
+{
+  "core": {
+    "updateMinimum": true,
+    "type": "patch",
+    "changeLogMessages": [
+      "optimize user crypto decrypt"
```
dscpinheiro: What do you think of using the PR title (`Reduce object allocations in UserCrypto.Decrypt by 98% for .NET8+`) here instead? I think customers would find that easier to parse :)

dscpinheiro (approved):

generator/.DevConfigs/a3466402-916b-4e40-8822-79b28ab5de33.json
peterrsongg: i like that better, updated :)

sdk/src/Core/Amazon.Runtime/Internal/Settings/UserCrypto.cs
peterrsongg: done

normj:
Out of curiosity do you have any part of your profile stored in the legacy .NET store that uses UserCrypto? I'm wondering why the code even went down this path into .NET legacy store.

peterrsongg:
> Out of curiosity do you have any part of your profile stored in the legacy .NET store that uses UserCrypto? I'm wondering why the code even went down this path into .NET legacy store.

Not that i'm aware of? I just used the default profile in `~/.aws/credentials`

peterrsongg:
> Out of curiosity do you have any part of your profile stored in the legacy .NET store that uses UserCrypto? I'm wondering why the code even went down this path into .NET legacy store.

call-stack-user-crypto


Here is the call stack, looks like it is getting invoked when trying to get the profile from the credentialprofilestorechain, when checking if the user has enabled servicespecificendpoints. Eventually `IsUserCryptoAvailable` is called which is defined as 

```
        public static bool IsUserCryptAvailable
        {
            get
            {
#if NET8_0_OR_GREATER
                if (!OperatingSystem.IsWindows())
                {
                    return false;
                }
#endif
                if(!_isUserCryptAvailable.HasValue)
                {
                    try
                    {
                        Encrypt("test");
                        _isUserCryptAvailable = true;
                    }
                    catch(Exception e)
                    {
                        var logger = Logger.GetLogger(typeof(UserCrypto));
                        logger.InfoFormat("UserCrypto is not supported.  This may be due to use of a non-Windows operating system or Windows Nano Server, or the current user account may not have its profile loaded. {0}", e.Message);
                        _isUserCryptAvailable = false;
                    }
                }
                return _isUserCryptAvailable.Value;
            }
        }
```

maybe it's worth looking at the `IsUserCryptoAvailable` logic?

peterrsongg:
merged for now, but i think it's worth looking at why this is getting called

PR #4224, opened 2025-12-10
Content Language Bug Fix

GarrettBeatty:
## Description

This PR fixes two issues related to the `Content-Language` HTTP header in the AWS SDK for .NET:

1. **Request Signing Issue**: The `Content-Language` header was not properly handled as a content header in `HttpRequestMessageFactory`, causing signature mismatch errors. This was is also happening on `development` branch too.

2. **GetObjectMetadataResponse Backing Field Consistency**: Changed `GetObjectMetadataResponse.ContentLanguage` to use `Headers.ContentLanguage` as the backing field, matching the pattern used by `GetObjectResponse.ContentLanguage`.

### Request Signing Fix

**Symptom:**
When setting `Content-Language` on a PutObject request, the SDK throws `SignatureDoesNotMatch`:

**Root Cause:**
The SDK includes `Content-Language` in signature calculation, but .NET's `HttpClient` wasn't actually sending it because the header wasn't being placed on `HttpContent.Headers`.

**Evidence from AWS error response:**

| Component | Canonical Request Content |
|-----------|--------------------------|
| **SDK signed with** | `content-language:en-US` |
| **AWS received** | `content-language:` (empty) |

The `Content-Language` header value was lost because:
1. `HttpWebRequestMessage.ContentHeaderNames` didn't include `Content-Language`
2. `SetRequestHeaders()` attempted to add it to `HttpRequestMessage.Headers` (request headers)
3. Per RFC 7231, `Content-Language` is a content header - .NET's `HttpClient` rejects it as a request header
4. `WriteContentHeaders()` had no code to add `Content-Language` to `HttpContent.Headers`
5. Result: Header signed but never sent → signature mismatch

**Fix:**
1. Added `HeaderKeys.ContentLanguageHeader` to `ContentHeaderNames` HashSet
2. Added `Content-Language` handling in `WriteContentHeaders()` method

### GetObjectMetadataResponse Backing Field Fix

Changed `GetObjectMetadataResponse.ContentLanguage` to delegate to `Headers.ContentLanguage` as the backing store instead of a private field, matching the pattern in `GetObjectResponse.ContentLanguage`.

## Testing
1. Integ Test
2. Dry run e27c9b4e-0984-4201-9fc9-fb5ab5f28ac0

## Checklist

## License

### DIFF
--- a/sdk/test/Services/S3/IntegrationTests/PutObjectTests.cs
+++ b/sdk/test/Services/S3/IntegrationTests/PutObjectTests.cs
@@ -658,6 +658,7 @@ namespace AWSSDK_DotNet.IntegrationTests.Tests.S3
             // Verify via GetObjectMetadata
             var metadata = Client.GetObjectMetadata(bucketName, key);
             Assert.AreEqual(contentLanguage, metadata.Headers.ContentLanguage);
+            Assert.AreEqual(contentLanguage, metadata.ContentLanguage);
         }
 
         private void PutObjectWithContentEncoding()


sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_netstandard/HttpRequestMessageFactory.cs
```diff
@@ -731,6 +732,10 @@ private void WriteContentHeaders(IDictionary<string, string> contentHeaders)
                 _request.Content.Headers.TryAddWithoutValidation(HeaderKeys.ContentEncodingHeader,
                     contentEncodingHeader);
 
+            if (contentHeaders.TryGetValue(HeaderKeys.ContentLanguageHeader, out var contentLanguageHeader))
```
peterrsongg, line 735: Should we also add `CacheControl` while we're at it? Or is there some nuance i'm missing here about "Content" headers versus this one which is "CacheControl"

peterrsongg: And is there a similar rational for not using the content length?

GarrettBeatty: i havent checked cachecontrol or the other ones yet. either way if we were to update them i think that should be in another PR

PR #4227, opened 2025-12-11
fix retrying valid IO exceptions

philasmar:
## Description

The Atomic File Handler was retrying all IOExceptions 100 times instead of only the case when the temp file already exists. This fix makes sure to throw all other exceptions.

## Motivation and Context

DOTNET-8434

## Testing

Ran dry run aba5fa13-6e47-4b5d-a861-f9d78429a9a2

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/Transfer/Internal/AtomicFileHandler.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/Internal/AtomicFileHandler.cs
@@ -33,6 +33,7 @@ namespace Amazon.S3.Transfer.Internal
     {
         private string _tempFilePath;
         private bool _disposed = false;
+        private static readonly object _fileLock = new object();
 
         /// <summary>
         /// Creates a temporary file with unique identifier for atomic operations.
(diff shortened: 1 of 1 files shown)


sdk/src/Services/S3/Custom/Transfer/Internal/AtomicFileHandler.cs
```diff
@@ -68,10 +69,24 @@ public string CreateTemporaryFile(string destinationPath)
                     _tempFilePath = tempPath;
                     return tempPath;
                 }
-                catch (IOException) when (attempt < 99)
+                catch (IOException)
```
peterrsongg, line 72: The PR description says that it was retrying all IO exceptions, and now this is fixiing that so it only retries when the file exists, but here you're still catching IOException which can be for any number of reasons. Instead of using a lock here and checking, couldn't you just do something like
```
catch (IOException ex) when (ex.HResult == ex.HResult == -2147024816) 
{
   continue;
}
catch (IOException) //any other reason we throw
{
  throw;
}
```

-2147024816 is the specific error for file existing. This is also more performant because it removes a `File.Exists` call


I feel like this approach is better because it addresses the real issue which is when the file exists without having to use a lock and it removes a File System read to check if the file exists which is an expensive operation

PR #4228, opened 2025-12-11
Multi part download: Stop queueing more tasks if in progress one fails

GarrettBeatty:
## Description
Previously if an error occurred when for a part, we would still queue the next parts, even though a previous part failed. 

This would result in the behavior where, part 2 failed, queues part 3, part 3 fails, queues part 4, etc.. 

This change adds a check so that if a part fails with some exception, it will cancel the internal cancellation token to prevent other parts from getting queued.

Now the new behavior is part 2 failed, dont queue part 3.

## Motivation and Context

## Testing
1. Unit test

Before my fix the unit test failed on 

```
                    $"Part 10 should NOT be queued after Part 3 fails. Queued parts: {string.Join(", ", queuedPartsList)}");

```

after my fix, the test passes.

2. I also re-ran this fix on the ec2 instance where i originally saw this error and i verified that the program was successfully cancelled immediately instead of going through all of the parts.

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
@@ -46,8 +46,6 @@ namespace Amazon.S3.Transfer.Internal
         private readonly SemaphoreSlim _httpConcurrencySlots;
         private readonly bool _ownsHttpThrottler;
         private readonly RequestEventHandler _requestEventHandler;
-        
-        private Exception _downloadException;
         private bool _disposed = false;
         private bool _discoveryCompleted = false;
         private Task _downloadCompletionTask;
@@ -166,15 +164,6 @@ namespace Amazon.S3.Transfer.Internal
             }
         }
 
-        /// <inheritdoc/>
-        public Exception DownloadException
-        { 
-            get 
-            { 
-                return _downloadException;
-            }
-        }
-
         /// <summary>
         /// Discovers the download strategy and starts concurrent downloads in a single unified operation.
         /// This eliminates resource leakage by managing HTTP slots and buffer capacity internally.
@@ -259,7 +248,6 @@ namespace Amazon.S3.Transfer.Internal
             }
             catch (Exception ex)
             {
-                _downloadException = ex;
                 _logger.Error(ex, "MultipartDownloadManager: Discovery failed");
                 throw;
             }
@@ -336,7 +324,6 @@ namespace Amazon.S3.Transfer.Internal
             }
             catch (Exception ex)
             {
-                _downloadException = ex;
                 _logger.Error(ex, "MultipartDownloadManager: Download failed");
                 
                 HandleDownloadError(ex, internalCts);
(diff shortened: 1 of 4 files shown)


sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
```diff
@@ -377,37 +378,75 @@ public async Task StartDownloadsAsync(DownloadDiscoveryResult discoveryResult, E
...
+                                    
+                                    // Add failure detection to immediately cancel internal token on first error
+                                    // This prevents the for loop from queuing additional parts after a failure
+                                    _ = task.ContinueWith(t =>
+                                    {
+                                        if (t.IsFaulted && !internalCts.IsCancellationRequested)
+                                        {
+                                            // Capture and log the ORIGINAL exception immediately
+                                            var originalException = t.Exception?.GetBaseException();
+                                            if (_downloadException == null && originalException != null)
```
peterrsongg: cline helped me review this pr and i think this is a valid concern. Here is the race condition it pointed out:
2. Race Condition in Exception Storage
```

if (_downloadException == null && originalException != null)
{
    _downloadException = originalException;
}
```
Problem: Multiple tasks can fail simultaneously, and without synchronization, there's a race condition where the "first" exception might not be the one stored.

Recommendation: Use Interlocked.CompareExchange for atomic exception storage:
```
Interlocked.CompareExchange(ref _downloadException, originalException, null);
```

sdk/test/Services/S3/UnitTests/Custom/MultipartDownloadManagerTests.cs
```diff
@@ -3597,5 +3598,140 @@ public async Task StartDownloadsAsync_CancellationRacesWithDispose_HandlesGracef
...
 
         #endregion
+
+        #region Issue #2 - Continued Queuing After Failure Tests
+
+        [TestMethod]
+        public async Task StartDownloadsAsync_PartFailsDuringDownload_StopsQueuingNewParts()
+        {
+
+            AWSConfigs.LoggingConfig.LogTo = LoggingOptions.Console;
```
peterrsongg: `AWSConfigs` is shared across the whole SDK, if you set it here it will be turned on for every test. I'd consider adding it in a try block then resetting it in a finally block or removing it altogether.

peterrsongg (changes requested):

sdk/test/Services/S3/UnitTests/Custom/MultipartDownloadManagerTests.cs
GarrettBeatty: i actually forgot to remove this. i will udpate

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
GarrettBeatty: i will do this

GarrettBeatty: ended up removing it since the variable is not needed

PR #4236, opened 2025-12-15
add case sensitivity check to download directory command

philasmar:
## Description

A security fix has been applied to the Transfer Utility Download Directory command which now checks if the directory is case-sensitive before proceeding with the download. This prevents potential overwriting of files in case-insensitive file systems. If you would like to revert to the previous behavior, you can set the 'SkipDirectoryCaseSensitivityCheck' property on the 'TransferUtilityConfig' to true.

## Motivation and Context

DOTNET-8392
## Testing

Ran tests locally
Ran dry run
## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/Transfer/Internal/_bcl+netstandard/DownloadDirectoryCommand.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/Internal/_bcl+netstandard/DownloadDirectoryCommand.cs
@@ -66,10 +66,13 @@ namespace Amazon.S3.Transfer.Internal
 
                 // Step 3: Filter to actual files (exclude directory markers)
                 var objectsToDownload = FilterObjectsToDownload(s3Objects);
-
+                
                 // Step 4: Setup resources and execute downloads
                 using (var resources = CreateDownloadResources(cancellationToken))
                 {
+                    await CheckIfKeysDifferByCase(objectsToDownload, prefixLength, resources.InternalCancellationTokenSource)
+                        .ConfigureAwait(false);
+                    
                     await ExecuteParallelDownloadsAsync(
                         objectsToDownload,
                         prefixLength,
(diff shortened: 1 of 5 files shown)


sdk/src/Services/S3/Custom/Util/FileSystemHelper.cs
```diff
@@ -0,0 +1,89 @@
...
+            if (string.IsNullOrEmpty(directoryPath))
+            {
+                throw new ArgumentNullException(nameof(directoryPath));
+            }
+
+            if (!Directory.Exists(directoryPath))
+            {
+                throw new DirectoryNotFoundException($"The directory '{directoryPath}' does not exist.");
+            }
+
```
GarrettBeatty, line 49: just wondering if theres any other alternative to making a temp file to check

sdk/src/Services/S3/Custom/Transfer/Internal/_bcl+netstandard/DownloadDirectoryCommand.cs
```diff
@@ -93,6 +96,63 @@ await ExecuteParallelDownloadsAsync(
...
+                
+                await _failurePolicy.ExecuteAsync(
+                    () =>
+                    {
+                        string currentKey = s3Object.Key;
+
+                        // Check the key against the hash set for case-insensitive duplicates
+                        if (!hashSet.Add(currentKey))
+                        {
+                            // A case-insensitive match was found.
```
GarrettBeatty, line 140: is there any way to do this where we are not repeating the logic for constructing the failure policy and creating the request.

GarrettBeatty: why cant we add a check in DownloadSingleFileAsync for each file?

sdk/src/Services/S3/Custom/Util/FileSystemHelper.cs
philasmar: I couldn't find any. Talked to Norm also and this is where we landed. .NET doesnt have a Native way to check this. Also, case sensitivity can be directory specific and not necessarily OS or File System specific. Like Windows, you can make 1 directory case sensitive while the rest of the OS stays insensitive.

GarrettBeatty: okay temp file is fine then

sdk/src/Services/S3/Custom/Transfer/Internal/_bcl+netstandard/DownloadDirectoryCommand.cs
philasmar: I am trying to detect a collision before actually starting the download. we have the ability to do so, so why let the download start and then have an exception thrown.

GarrettBeatty: current code

```
ExecuteAsync:
  │
  ├─ Step 1: Validate & setup directory
  │
  ├─ Step 2: List S3 objects
  │
  ├─ Step 3: Filter to actual files
  │
  ├─ Step 4: Create download resources
  │     │
  │     ├─► CheckIfKeysDifferByCase()          ← SEPARATE PHASE
  │     │     └─ for each s3Object:
  │     │          └─ await _failurePolicy.ExecuteAsync(...)  ← Async + failure policy per object
  │     │
  │     └─► ExecuteParallelDownloads()
  │           └─ for each s3Object (concurrent):
  │                └─ DownloadSingleFileAsync()
  │                      ├─ Validate path
  │                      └─ Execute download
  │
  └─ Step 5: Build response

```

suggested

```
ExecuteAsync:
  │
  ├─ Step 1: Validate & setup directory
  │
  ├─ Step 2: List S3 objects
  │
  ├─ Step 3: Filter to actual files
  │
  ├─ Step 4: Create download resources
  │     │
  │     ├─► IdentifyCaseConflictingKeys()      ← SYNC, simple loop, no failure policy
  │     │     └─ Build HashSet of conflicting keys (O(n) one pass)
  │     │
  │     └─► ExecuteParallelDownloads()
  │           └─ for each s3Object (concurrent):
  │                └─ DownloadSingleFileAsync()
  │                      ├─ Validate path
  │                      ├─ ValidateCaseConflict()  ← CHECK MOVED HERE (with failure policy)
  │                      └─ Execute download
  │
  └─ Step 5: Build response

```

normj (approved):

PR #4272, opened 2026-01-06
Improve ArrayPool chunk disposal

afroz429:
## Description

Enhances error handling in the `ChunkedBufferStream.Dispose` method to ensure individual ArrayPool chunk cleanup failures don't prevent the disposal of remaining chunks.

## Motivation and Context

There is potentially a memory leak in the `ChunkedBufferStream.Dispose` method. This PR addresses the potential issue.

## Testing

Dry run succeeded: DRY_RUN-9d6f40e2-2000-4b79-996e-ca0b56bfee36
## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/Transfer/Internal/ChunkedBufferStream.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/Internal/ChunkedBufferStream.cs
@@ -26,6 +26,7 @@ using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
+using Amazon.Runtime.Internal.Util;
 
 namespace Amazon.S3.Transfer.Internal
 {
@@ -79,6 +80,7 @@ namespace Amazon.S3.Transfer.Internal
         private readonly List<byte[]> _chunks;
         private readonly int _chunkSize;
         private readonly long _maxStreamSize;
+        private readonly Logger _logger = Logger.GetLogger(typeof(ChunkedBufferStream));
         private long _length = 0;
         private long _position = 0;
         private bool _isReadMode = false;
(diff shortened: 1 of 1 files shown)


GarrettBeatty (approved):

sdk/src/Services/S3/Custom/Transfer/Internal/ChunkedBufferStream.cs
```diff
@@ -385,20 +385,26 @@ protected override void Dispose(bool disposing)
...
                 {
-                    // Return all chunks to ArrayPool
-                    foreach (var chunk in _chunks)
+                    try
                     {
                         ArrayPool<byte>.Shared.Return(chunk);
                     }
+                    catch (Exception)
+                    {
+                        // Suppress exceptions in Dispose - continue cleanup
```
normj, line 399: That seems really unusual for us to get an exception here, it might be worth adding an SDK debug log statement. Have we seen any exceptions from the Return method?

GarrettBeatty: we havent seen any issues, but we just added this as a precaution.

normj: Okay, I would be surprised if this ever through an exception but it won't hurt.

Seabizkit: i have very little context here, but the scope seems really small...

i gave this a go in Ai as all the try stuff seems werid and not returning early 
and it seems to me its should be something like

````

[SuppressMessage("Design", "CA1065:Do not raise exceptions in unexpected locations",
    Justification = "Dispose is non-throwing by design per .NET guidance")]
protected override void Dispose(bool disposing)
{
    // Idempotent guard: if already disposed, do nothing.
    if (_disposed) return;

    if (disposing)
    {
        // Managed cleanup: return arrays to the pool.
        // If buffers can contain sensitive data, consider clearArray: true.
        foreach (var chunk in _chunks)
        {
            ArrayPool<byte>.Shared.Return(chunk /*, clearArray: true */);
        }

        _chunks.Clear();
    }

    // (No unmanaged resources in this class; if you add any, release them here.)

    _disposed = true;
    base.Dispose(disposing);
}


````

which make more sense to me, unless this is coverying a specific known case. then i missed it

boblodgett: I don't agree with the empty catch blocks it can mask serious bugs especially in something that shouldn't be throwing an exception. It is also not best practice.

At a minimum a log statement should be added as norm stated or use Seabizkit's suggested approach.

PR #4273, opened 2026-01-07
Generate GetBucketWebsite

peterrsongg:
## Testing

DRY_RUN-98a2b771-b69c-4c45-8285-4ce70f26b273
AssemblyComparison Output:
empty

Fuzz Testing: empty
## Types of changes

## Checklist

## License


peterrsongg:
# BREAKING CHANGES ANALYSIS FOR COMMIT 848ee7d661147787bb0d8503f6adac425a03f851

After thorough analysis of all 22 files changed in this commit, examining the migration from custom to generated S3 code for GetBucketWebsite operation:

## FILES ANALYZED: 22 of 22

### ANALYSIS BY CATEGORY:

#### **Generator Infrastructure Changes (7 files)** - ✅ NO ISSUES
- `Customizations.cs` - Added InjectXmlUnmarshallResultCode support
- `GeneratorDriver.cs` - Added IndexDocument/ErrorDocument to exception list
- `RestXmlResponseUnmarshaller.cs` - Added support for custom unmarshall result code injection  
- `RestXmlResponseUnmarshaller.tt` - Template changes for customization
- `ServiceModel.cs` - Added TryGetShapeModifier method
- `s3.customizations.json` - Added customizations for GetBucketWebsite generation

#### **Request/Response Models (2 files)** - ✅ NO ISSUES
- **GetBucketWebsiteRequest.cs** (Generated) - New generated file, correctly implements all properties including:
  - `BucketName` property with IsSetBucketName() using `!= null` check
  - `ExpectedBucketOwner` property with IsSetExpectedBucketOwner() using `!String.IsNullOrEmpty()` check
  - ✅ IsSet implementation matches expected pattern for request objects

- **GetBucketWebsiteResponse.cs** (Custom + Generated partials) - ✅ NO ISSUES
  - Custom partial: Contains `WebsiteConfiguration` property (preserved)
  - Generated partial: Empty partial class
  - Combined: Full functionality maintained through partial class pattern

#### **Marshallers (2 files)** - ✅ NO ISSUES  
- **GetBucketWebsiteRequestMarshaller** (Custom + Generated partials):
  - Custom partial: `PreMarshallCustomization` sets `Suppress404Exceptions = true` ✅ PRESERVED
  - Generated partial: Main marshalling logic with calls to PreMarshallCustomization/PostMarshallCustomization
  - ✅ All previous behavior preserved

- **GetBucketWebsiteResponseUnmarshaller** (Custom + Generated partials):
  - Custom partial: `RoutingRuleCustomUnmarshall` initializes RoutingRules list ✅ PRESERVED
  - Generated partial: Main unmarshalling logic with call to RoutingRuleCustomUnmarshall
  - ✅ All previous behavior preserved

#### **RoutingRule Model Classes (6 files)** - ✅ NO ISSUES

Comparing deleted Custom models with new Generated models:

**RoutingRule.cs:**
- Custom (deleted): Had `Condition` and `Redirect` properties with proper getters/setters
- Generated (new): Has same `Condition` and `Redirect` properties with identical semantics
- IsSet methods: Both use `!= null` check ✅ IDENTICAL

**RoutingRuleCondition.cs:**
- Custom (deleted): Had `HttpErrorCodeReturnedEquals` and `KeyPrefixEquals` string properties  
- Generated (new): Has same properties with identical implementation
- IsSet methods: Both use `!= null` check ✅ IDENTICAL

**RoutingRuleRedirect.cs:**
- Custom (deleted): Had `HostName`, `HttpRedirectCode`, `Protocol`, `ReplaceKeyPrefixWith`, `ReplaceKeyWith` properties
- Generated (new): Has same 5 properties with identical implementation
- IsSet methods: All use `!= null` check ✅ IDENTICAL

#### **Unmarshallers for Routing Rules (3 files)** - ✅ NO ISSUES

**RoutingRuleUnmarshaller.cs:**
- Refactored from old style to new generated pattern
- ✅ All unmarshalling logic preserved
- Added partial method hook: `XmlStructureUnmarshallCustomization`

**RoutingRuleConditionUnmarshaller.cs:**
- Refactored from old style to new generated pattern
- ✅ All unmarshalling logic for HttpErrorCodeReturnedEquals and KeyPrefixEquals preserved

**RoutingRuleRedirectUnmarshaller.cs:**
- Refactored from old style to new generated pattern  
- ✅ All unmarshalling logic for all 5 properties preserved

---

## FINAL VERDICT: ✅ NO BREAKING CHANGES DETECTED

### Summary:
- **Total Files Changed**: 22
- **Files Analyzed**: 22
- **Breaking Changes Found**: 0
- **Files with Issues**: 0

### Key Findings:
1. **Partial Class Pattern**: Custom and Generated partial classes combine correctly to preserve all functionality
2. **IsSet Methods**: All string property IsSet methods use consistent `!= null` checks (no `!string.IsNullOrEmpty` changes detected)
3. **Public Properties**: All public properties maintained with identical implementations
4. **Custom Logic**: All custom marshalling/unmarshalling logic preserved through PreMarshall/PostMarshall/CustomUnmarshall partial methods
5. **Response Data**: WebsiteConfiguration property fully preserved through partial class pattern
6. **Routing Rules**: All three routing rule model classes properly migrated with identical property implementations

The migration from custom to generated code for GetBucketWebsite operation has been executed correctly with full backward compatibility maintained.

generator/ServiceModels/s3/s3.customizations.json
```diff
@@ -1829,6 +1829,37 @@
...
+        "GetBucketWebsiteOutput":{
+            "modify":[
+                {
+                    "ErrorDocument":{"injectXmlUnmarshallCode":["response.WebsiteConfiguration.ErrorDocument = StringUnmarshaller.Instance.Unmarshall(context);"], "skipContextTestExpressionUnmarshallingLogic" : true}
+                },
+                {
+                    "IndexDocument":{"emitPropertyName":"IndexDocumentSuffix"}
+                },
+                {
+                    "IndexDocumentSuffix":{"injectXmlUnmarshallCode":["response.WebsiteConfiguration.IndexDocumentSuffix = StringUnmarshaller.Instance.Unmarshall(context);"],"skipContextTestExpressionUnmarshallingLogic" : true}
```
muhammad-othman, line 1849: I don't understand why we need this, does the generator still use the shape unmarshaller even after you added a customization to set its type to a string?

generator/ServiceModels/s3/s3.customizations.json
```diff
@@ -2288,6 +2325,21 @@
                 "Marshaller": "StringUtils.FromString",
                 "Unmarshaller": "StringUnmarshaller"
             }
+        },
+        "GetBucketWebsiteOutput":{
+            "ErrorDocument":{
+                "Type":"String",
```
muhammad-othman, line 2358: does it make a difference when we use `String` vs `string`  type?

peterrsongg: nope no difference other than difference in casing

generator/ServiceModels/s3/s3.customizations.json
peterrsongg: ah so this isn't so that it uses the right unmarshaller but to make sure we set the value on `response.WebsiteConfiguration`
1. We want to set it on `response.WebsiteConfiguration` since `WebsiteConfiguration` isn't a modeled member on the output shape, if we don't do this it will incorrectly try to set it on `response.IndexDocumentSuffix` instead

here's the modeled output (notice how there is no `WebsiteConfiguration`
```
    "GetBucketWebsiteOutput":{
      "type":"structure",
      "members":{
        "RedirectAllRequestsTo":{
          "shape":"RedirectAllRequestsTo",
          "documentation":"<p>Specifies the redirect behavior of all requests to a website endpoint of an Amazon S3 bucket.</p>"
        },
        "IndexDocument":{
          "shape":"IndexDocument",
          "documentation":"<p>The name of the index document for the website (for example <code>index.html</code>).</p>"
        },
        "ErrorDocument":{
          "shape":"ErrorDocument",
          "documentation":"<p>The object key name of the website error document to use for 4XX class errors.</p>"
        },
        "RoutingRules":{
          "shape":"RoutingRules",
          "documentation":"<p>Rules that define when a redirect is applied and the redirect behavior.</p>"
        }
      }
    },
```

So without this customization it will do something like
```
response.IndexDocumentSuffix = unmarshaller.Unmarshall(context);
```

generator/ServiceModels/s3/s3.customizations.json
muhammad-othman: nit: would prefer if we can stick to one casing.

muhammad-othman (approved):

generator/ServiceClientGeneratorLib/GeneratorDriver.cs
```diff
@@ -1726,7 +1726,10 @@ private void S3NeedsCustomUpdate(Shape shape)
                 { "AnalyticsAndOperator", 2 },
                 { "IntelligentTieringAndOperator", 2},
                 //TODO: Add runbook entry as to why this excluded and how to do a customization for this.
```
boblodgett: What does this TODO apply to and is it done? This doesn't seem like it belongs in the code as is as it can be an offline reminder or it can be applied as part of this task.

sdk/src/Services/S3/Generated/Model/PutBucketLoggingRequest.cs
```diff
@@ -0,0 +1,281 @@
...
+        public S3BucketLoggingConfig LoggingConfig
+        {
+            get { return this._loggingConfig; }
+            set { this._loggingConfig = value; }
+        }
+
+        // Check to see if LoggingConfig property is set
+        internal bool IsSetLoggingConfig()
+        {
+            return this._loggingConfig != null && this._loggingConfig.IsSetTargetBucketName();
```
boblodgett, line 277: `& this._loggingConfig.IsSetTargetBucketName();` changes the condition from what it was before. Is this correct? Ultimately it will possibly change what is marshalled in the request which may not be breaking because it is a request to the server as long as it is correct.

peterrsongg: This is correct because in the[ custom marshaller pre-generation](https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Services/S3/Custom/Model/Internal/MarshallTransformations/PutBucketLoggingRequestMarshaller.cs#L70-L73) we only write the `LoggingEnabled` XML element if TargetBucketName is also set. So to have the same effect I updated the `IsSet` here

generator/ServiceClientGeneratorLib/GeneratorDriver.cs
peterrsongg: I removed the todo, Its basically just to update the runbook on what to do if the generator fails because these shapes have been updated. I added runbook entry already for `GlacierJobParameters` so the TODO was out of date. I also added runbook entries for the two shapes below.

PR #4279, opened 2026-01-07
make maxinmemorypartsnullable

GarrettBeatty:
make maxinmemoryparts nullable in case we want to auto calculate it later on

4367fa3e-b2e5-4110-81be-82d23caff658 dry run

### DIFF
--- a/sdk/src/Services/S3/Custom/Transfer/TransferUtilityOpenStreamRequest.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/TransferUtilityOpenStreamRequest.cs
@@ -32,22 +32,17 @@ namespace Amazon.S3.Transfer
     /// </summary>
     public class TransferUtilityOpenStreamRequest : BaseDownloadRequest
     {
-        private int _maxInMemoryParts = 1024;
-
         /// <summary>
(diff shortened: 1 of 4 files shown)


sdk/src/Services/S3/Custom/Transfer/TransferUtilityOpenStreamRequest.cs
```diff
@@ -32,22 +32,17 @@ namespace Amazon.S3.Transfer
...
         /// default part size of 8MB, the default value of 1024 parts allows up to 8GB of memory usage.
         /// Adjust this value based on your application's memory constraints and performance requirements.
+        /// Set to null to use the default value of 1024.
         /// </remarks>
-        public int MaxInMemoryParts
-        {
-            get { return this._maxInMemoryParts; }
-            set { this._maxInMemoryParts = value; }
-        }
+        public int? MaxInMemoryParts { get; set; }
```
GarrettBeatty, line 45: https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Services/S3/Custom/Transfer/TransferUtilityOpenStreamRequest.cs#L35 this property is brand new (not on main) so its safe to update

PR #4281, opened 2026-01-08
Stale connections retry

GarrettBeatty:
# Stale Connection Retry Support

## Description
This PR implements automatic retry logic for stale connection errors. When HTTP clients reuse pooled connections that have been closed by the server, they encounter socket errors like "Broken pipe" (Shutdown), "Connection reset", or "Connection aborted". These errors are transient and should be retried immediately without penalizing the application's normal retry budget.

**Key Changes:**
- Added `MaxStaleConnectionRetries` configuration property to `ClientConfig` and `IClientConfig` (default: 3)
- Enhanced `RetryHandler` to detect and handle stale connection errors (SocketError.Shutdown, ConnectionReset, ConnectionAborted)
- Stale connection retries execute immediately and  don't count against `MaxErrorRetry`
- Properly unwraps nested exceptions to detect underlying socket errors
- After exhausting stale connection retries, falls back to normal retry logic for persistent issues

## Motivation and Context
While testing S3 50TB upload with large part sizes (5GB) we frequently  encountered socket errors. Even with the current SDK retry logic we ran out of attempts.

DOTNET-8432

**Solution:** This PR treats stale connection errors specially - they get their own retry budget (MaxStaleConnectionRetries) and retry immediately without backoff, providing transparent recovery from connection pool issues while preserving normal retry logic for genuine failures.

## Testing
1. **Unit tests** 
2. **Integration tests**
3. **Dry run:** `8224bf25-6830-4919-84d9-787f196ed04a` pass 

## Types of changes

## Checklist
- [x] I have updated the documentation accordingly (XML comments on public API)
- [x] I have added tests to cover my changes (unit + integration tests)

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs
+++ b/sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs
@@ -283,8 +283,8 @@ namespace Amazon.Runtime
                 }
 
 #if NETSTANDARD
-                // .NET Core/.NET 7+ fallback: HttpClient may throw HttpIOException when the response
-                // ends prematurely due to stale connections. HttpIOException is .NET 7+ - we check by
+                // .NET 8+ fallback: HttpClient may throw HttpIOException when the response
+                // ends prematurely due to stale connections. HttpIOException is .NET 8+ - we check by
                 // type name to avoid compilation issues on older targets that support NetStandard.
                 if (currentException.GetType().Name == "HttpIOException")
                 {


dscpinheiro:
I haven't looked at all changes, but are you adding this for a specific reason? We should probably check what the retry SEP says about stale connections (and that doc that was reviewed in the design meeting about updates to the retry mode in 2026).

generator/.DevConfigs/7f23582e-3225-487b-83e7-167cf17cb236.json
```diff
@@ -0,0 +1,13 @@
+{
+  "core": {
+    "changeLogMessages": [
+      "Update the retry behavior of the SDK to automatically retry on stale connections (e.g Broken pipe), which do not count against the standard retry amount.",
+      "Add MaxStaleConnectionRetries to Amazon.Runtime.ClientConfig to allow users to customize the amount of retries available for stale connections."
+    ],
+    "type": "patch",
+    "updateMinimum": true,
+    "backwardIncompatibilitiesToIgnore" : [
+      "Amazon.Runtime.IClientConfig"
```
GarrettBeatty, line 10: i need to double check this with  new service + old core, and old service package + new core

extensions/src/AWSSDK.Extensions.NETCore.Setup/Schema/ConfigurationSchema.json
```diff
@@ -91,6 +91,10 @@
           "type": "integer",
           "description": "How many retry HTTP requests an SDK should make for a single SDK operation invocation before giving up."
         },
+        "MaxStaleConnectionRetries": {
```
GarrettBeatty, line 94: i got an error in the dry run which said i needed to add this change to defaultclientconfig, configurationextensions, clientfactory and this json so thats why ive updated these

sdk/src/Core/Amazon.Runtime/ClientConfig.cs
```diff
@@ -588,6 +589,31 @@ public bool IsMaxErrorRetrySet
...
+        /// the server has closed. These retries don't count against the MaxErrorRetry limit.
+        /// Default value is 3.
+        /// </summary>
+        public int MaxStaleConnectionRetries
+        {
+            get
+            {
+                if (!this.maxStaleConnectionRetries.HasValue)
+                {
+                    return 3;
```
GarrettBeatty: i put the default as 3 retries @normj

generator/.DevConfigs/7f23582e-3225-487b-83e7-167cf17cb236.json
GarrettBeatty: made a test program that tested these and it seemed ok


```
  <ProjectReference Include="..\..\src\Core\AWSSDK.Core.NetStandard.csproj" />

// tried with this version (4.0.15.1, old version of service) and also a new version (4.0.17.1, which is a later version than my local commit)
    <PackageReference Include="AWSSDK.S3" Version="4.0.15.1"> </PackageReference>
```

```
using Amazon;
using Amazon.Runtime;
using Amazon.S3;
using Amazon.S3.Model;

namespace CrossVersionCompatibilityTest;

/// <summary>
/// Simple console app to test cross-version compatibility between Core and S3 packages.
/// 
/// </summary>
class Program
{
    static async Task Main(string[] args)
    {
        Console.WriteLine("=== AWS SDK Cross-Version Compatibility Test ===");
        Console.WriteLine();

        // Print assembly versions for debugging
        PrintAssemblyVersions();

        // Test 1: Verify ClientConfig properties
        TestClientConfigProperties();

        // Test 2: Try a simple S3 operation (requires valid credentials)
        if (args.Length > 0 && args[0] == "--live")
        {
            await TestLiveS3Operations();
        }
        else
        {
            Console.WriteLine();
            Console.WriteLine("Skipping live S3 tests. Run with --live flag to test against real S3.");
            Console.WriteLine("Example: dotnet run -- --live");
        }

        Console.WriteLine();
        Console.WriteLine("=== All tests completed ===");
    }

    static void PrintAssemblyVersions()
    {
        Console.WriteLine("Assembly Versions:");
        Console.WriteLine("------------------");

        var coreAssembly = typeof(AmazonServiceClient).Assembly;
        Console.WriteLine($"  AWSSDK.Core: {coreAssembly.GetName().Version}");
        Console.WriteLine($"    Location: {coreAssembly.Location}");

        var s3Assembly = typeof(AmazonS3Client).Assembly;
        Console.WriteLine($"  AWSSDK.S3:   {s3Assembly.GetName().Version}");
        Console.WriteLine($"    Location: {s3Assembly.Location}");
        Console.WriteLine();
    }

    static void TestClientConfigProperties()
    {
        Console.WriteLine("Test 1: ClientConfig Properties");
        Console.WriteLine("--------------------------------");

        try
        {
            var config = new AmazonS3Config
            {
                RegionEndpoint = RegionEndpoint.USEast1
            };

            // Test that basic properties work
            Console.WriteLine($"  RegionEndpoint: {config.RegionEndpoint}");
            Console.WriteLine($"  MaxErrorRetry: {config.MaxErrorRetry}");

            // Test the new MaxStaleConnectionRetries property
            // This will fail at compile time if using an old Core that doesn't have this property
            Console.WriteLine($"  MaxStaleConnectionRetries (default): {config.MaxStaleConnectionRetries}");

            // Try setting a custom value
            config.MaxStaleConnectionRetries = 5;
            Console.WriteLine($"  MaxStaleConnectionRetries (after set to 5): {config.MaxStaleConnectionRetries}");

            // Reset to default
            config.MaxStaleConnectionRetries = 3;
            Console.WriteLine($"  MaxStaleConnectionRetries (reset to 3): {config.MaxStaleConnectionRetries}");

            Console.WriteLine("  [PASS] ClientConfig properties work correctly");
        }
        catch (Exception ex)
        {
            Console.WriteLine($"  [FAIL] Error: {ex.Message}");
        }
        Console.WriteLine();
    }

    static async Task TestLiveS3Operations()
    {
        Console.WriteLine("Test 2: Live S3 Operations");
        Console.WriteLine("---------------------------");

        try
        {
            // Create client with custom MaxStaleConnectionRetries
            var config = new AmazonS3Config
            {
                RegionEndpoint = RegionEndpoint.USEast1,
                MaxStaleConnectionRetries = 5  // Custom value
            };

            using var client = new AmazonS3Client(config);

            Console.WriteLine("  Listing buckets...");
            var response = await client.ListBucketsAsync();

            Console.WriteLine($"  Found {response.Buckets.Count} buckets:");
            foreach (var bucket in response.Buckets.Take(5))
            {
                Console.WriteLine($"    - {bucket.BucketName}");
            }
            if (response.Buckets.Count > 5)
            {
                Console.WriteLine($"    ... and {response.Buckets.Count - 5} more");
            }

            Console.WriteLine("  [PASS] S3 operations work correctly");
        }
        catch (AmazonServiceException ex)
        {
            Console.WriteLine($"  [FAIL] AWS Error: {ex.Message}");
            Console.WriteLine($"         Error Code: {ex.ErrorCode}");
        }
        catch (Exception ex)
        {
            Console.WriteLine($"  [FAIL] Error: {ex.Message}");
        }
        Console.WriteLine();
    }
}

```

sdk/test/NetStandard/IntegrationTests/IntegrationTests/Core/StaleConnectionRetryTests.cs
```diff
@@ -0,0 +1,431 @@
...
+using System.Net;
+using System.Net.Http;
+using System.Net.Sockets;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Amazon.DNXCore.IntegrationTests.Core
+{
+    /// <summary>
```
GarrettBeatty, line 15: i try to cover all scenarios here, please double check that im not missing any

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryHandler.cs
```diff
@@ -92,6 +95,23 @@ public override void InvokeSync(IExecutionContext executionContext)
                 }
                 catch (Exception exception)
                 {
+                    // Check for stale connection retry first
+                    if (ShouldRetryForStaleConnection(exception, executionContext, 
+                        staleConnectionRetries, maxStaleConnectionRetries))
+                    {
+                        staleConnectionRetries++;
+                        PrepareForRetry(requestContext);
```
GarrettBeatty, line 103: the reason i had to put this logic in this class because of being able to utilize the `PrepareForRetry` function which resets/rewinds the stream. i think in stale connection case all errors occurs before it reaches service and on the client, but i assumed there is some edge case where maybe the stream is read partially (at some low level tcp/http level) which means we need to always reset it to be safe.

sdk/test/NetStandard/UnitTests/ClientConfigTests.cs
```diff
@@ -30,6 +30,7 @@ public class ClientConfigTests
             "LogResponse",
             "AWSTokenProvider",
             "BufferSize",
+            "MaxStaleConnectionRetries",
```
GarrettBeatty, line 33: this test was checking that any propertties in clientconfig which get attached to httpclient are included in the CreateConfigUniqueString function. 

but in our case since our property does not affect the http client, we add it to this array to ignore

sdk/src/Core/Amazon.Runtime/ClientConfig.cs
```diff
@@ -479,9 +480,9 @@ public string AuthenticationServiceName
         /// 
         /// Valid values are a list of one or more of the following:
         /// <list type="bullet">
-        /// <item><b>sigv4</b> � Signature Version 4 (fastest performance, single-region)</item>
```
GarrettBeatty, line 482: there was some wonky non readable character here which visual studio was complaining about

sdk/src/Core/Amazon.Runtime/IClientConfig.cs
```diff
@@ -173,9 +173,9 @@ public partial interface IClientConfig
         /// 
         /// Valid values are a list of one or more of the following:
         /// <list type="bullet">
-        /// <item><b>sigv4</b> � Signature Version 4 (fastest performance, single-region)</item>
```
GarrettBeatty, line 176: same here this was some weird encoded character

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryHandler.cs
GarrettBeatty: if we feel that `but i assumed there is some edge case where maybe the stream is read partially (at some low level tcp/http level)` will never happen, we could do this instead https://github.com/aws/aws-sdk-net/pull/4230#discussion_r2612319722 but i wasnt confident enough that we can ignore this

sdk/src/Core/Amazon.Runtime/ClientConfig.cs
normj: I'm confused why we want the comment to now use `â€”`

sdk/src/Core/Amazon.Runtime/ClientConfig.cs
normj: I'm not sure how to figure out the right number but my gut says 3 is low. My logic is we should be failing fast on all these attempts so it doesn't really hurt to try more and with a lot of concurrency there could be quite a few stale connections. Again I don't know what is the right but unscientifically for a nice round round number I'll say `10`. Open for push back through.

sdk/src/Core/Amazon.Runtime/IClientConfig.cs
normj: I'm confused why we want the comment to now use `â€”`

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs
```diff
@@ -256,6 +256,38 @@ public virtual bool IsThrottlingError(Exception exception)
...
+        /// </summary>
+        /// <param name="exception">The exception to check.</param>
+        /// <returns>true if the exception indicates a stale connection error, else false.</returns>
+        public virtual bool IsStaleConnectionError(Exception exception)
+        {
+            // Walk the exception chain looking for SocketException with known stale connection error codes
+            var currentException = exception;
+            while (currentException != null)
+            {
+                if (currentException is System.Net.Sockets.SocketException socketException)
```
normj, line 272: Does looking at the `SocketException` work for both .NET Core and .NET Framework?

sdk/src/Core/Amazon.Runtime/IClientConfig.cs
GarrettBeatty: should be fixed now. it is a hypen character (the one next to 0 on the keyboard)

sdk/src/Core/Amazon.Runtime/ClientConfig.cs
GarrettBeatty: should be fixed now to use hyphen character `-`

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs
GarrettBeatty: you are right there is some inconsistency - mainly because of `HttpWebRequest` being used rather than `HttpClient`. i need to add some additional logic to check if its .net framework to handle this

GarrettBeatty: https://gist.github.com/GarrettBeatty/993a9a8acfb41732afc56cdfff79eea4  i made this test program which tries to mimic the different error scenarios i encountered.

I updated in https://github.com/aws/aws-sdk-net/pull/4281/commits/ca72b8370c11fd6a1ddec0077bcac042c9050178 to have fallback in .net framework to account for these. it seems .net framework doesnt expose the socket error since HttpWebRequest is used so we have to look for specific io exception "The connection was closed"


i wanted to add an integration test for net framework but from my understanding i cant inject an IHttpClientFactory  since its not supported in .net framework. if you have other suggestions for how i can add an integration test for .net framework please let me know

sdk/src/Core/Amazon.Runtime/ClientConfig.cs
GarrettBeatty: i updated it to use 10

GarrettBeatty:
c77a45c9-0e1c-43e3-aa22-c61ea37e1caa latest dry run

sdk/test/UnitTests/Custom/Runtime/RetryHandlerStaleConnectionTests.cs
```diff
@@ -0,0 +1,371 @@
...
+            };
+
+            var request = CreateTestContext();
+            IsolatedPipeline.InvokeSync(request);
+
+            Assert.AreEqual(1, Tester.CallCount, "Should have made only 1 attempt (successful)");
+            Assert.AreEqual(0, request.RequestContext.Retries, "Should have 0 retries");
+        }
+
+#if BCL
```
dscpinheiro, line 315: Is this required? (All .NET Frameworks targets support async)

generator/.DevConfigs/7f23582e-3225-487b-83e7-167cf17cb236.json
dscpinheiro: FYI I think we shouldn't ignore the entire class since it could "hide" some unexpected breaking changes (at least until we improve the assembly comparer).

But no need to change here.

dscpinheiro (approved):

GarrettBeatty:
c77a45c9-0e1c-43e3-aa22-c61ea37e1caa  passed going to merge in

PR #4283, opened 2026-01-09
[V4] Update S3 integration tests to have initialize and cleanup be async

dscpinheiro:
First PR updating the S3 integration tests to be async (so that the project can also target .NET Standard). To make the review manageable, only changes to the `TestInitialize` / `TestCleanup` / `ClassCleanup` methods are included.

The most important file to review is `S3TestUtils`, where I added async versions for methods that already existed. The implementations are pretty similar to the sync methods, I left comments where they're different.

I also didn't delete the existing methods yet (there are test methods that call `CreateBucket` directly - those will be addressed in later PRs) and removed a couple of classes where all tests were ignored.

## Testing
- Dry-run: `DRY_RUN-aa267279-4ae1-43d6-9d6c-1df9f8921950`
- Ran integration tests locally and confirmed they succeeded

## License

### DIFF
--- a/generator/.DevConfigs/b809525a-4308-4b3e-93f8-a0dc9cac0b84.json
+++ b/generator/.DevConfigs/b809525a-4308-4b3e-93f8-a0dc9cac0b84.json
@@ -1,9 +1,0 @@
-{
-  "core": {
-    "updateMinimum": false,
-    "type": "patch",
-    "changeLogMessages": [
-      "Placeholder file for dry-run"
-    ]
-  }
-}


sdk/test/Services/S3/IntegrationTests/S3TestUtils.cs
```diff
@@ -21,25 +22,28 @@ public static class S3TestUtils
         private const int MAX_SPIN_LOOPS = 100;
         private const string TEST_MRAP_NAME = UtilityMethods.SDK_TEST_PREFIX + "-for-mrap-tests";
 
-        public static string CreateBucket(IAmazonS3 s3Client)
```
dscpinheiro, line 24: This overload without the request object is not used anymore.

sdk/test/Services/S3/IntegrationTests/S3TestUtils.cs
```diff
@@ -222,26 +315,24 @@ public static string GetOrCreateTestMRAP(IAmazonS3Control s3ControlClient, IAmaz
         /// <param name="accountId">AWS account ID that is expected to contain the MRAP</param>
         /// <param name="mrapName">Name of the multi-region access point</param>
         /// <returns>The MRAP's ARN if it exists, an empty string otherwise</returns>
-        private static string CheckIfMRAPExists(IAmazonS3Control s3ControlClient, string accountId, string mrapName)
+        private static async Task<string> CheckIfMRAPExists(IAmazonS3Control s3ControlClient, string accountId, string mrapName)
         {
             var request = new ListMultiRegionAccessPointsRequest { AccountId = accountId };
-            var paginatorForAccessPoints = s3ControlClient.Paginators.ListMultiRegionAccessPoints(request);
-            
-            foreach (var response in paginatorForAccessPoints.Responses)
```
dscpinheiro, line 230: I couldn't get the async paginators to work here (without changing the language version of the project), so decided to re-write it as a loop instead.

sdk/test/Services/S3/IntegrationTests/S3TestUtils.cs
```diff
@@ -340,6 +431,49 @@ public static T WaitForConsistency<T>(Func<T> loadFunction)
...
+                        }
+
+                        return result;
+                    }
+                }
+                catch
+                {
+                }
+
+                await Task.Delay(0);
```
dscpinheiro, line 468: This method does look odd, but it's the same behavior as the sync method...

sdk/test/Services/S3/IntegrationTests/ObjectLockConfigurationTests.cs
```diff
@@ -41,19 +40,19 @@ public class ObjectLockConfigurationTests : TestBase<AmazonS3Client>
...
-                ObjectLockEnabledForBucket = true                
+                ObjectLockEnabledForBucket = true
             });
         }
 
         [ClassCleanup]
-        public static void ClassCleanup()
+        public static async Task ClassCleanup()
         {
             DeleteBucketObjectsIncludingLocked(Client, bucketName);
```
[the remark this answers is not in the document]
dscpinheiro: Fixed in https://github.com/aws/aws-sdk-net/pull/4283/commits/b454d70635ce347059c3665ca79889b1b8019cec

sdk/test/Services/S3/IntegrationTests/AWSSDK.IntegrationTests.S3.NetFramework.csproj
```diff
@@ -21,10 +21,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-
 	<Compile Remove="**/obj/**" />
 	<None Remove="**/obj/**" />
-	<Compile Include="../../KeyManagementService/IntegrationTests/KeyManagementService.cs" />
```
peterrsongg, line 27: what is this for? and would this mean we don't run this test anymore?

dscpinheiro: This file doesn't exist anymore (https://github.com/aws/aws-sdk-net/blob/main/sdk/test/Services/KeyManagementService/IntegrationTests/KeyManagementService.cs), I actually removed it in a previous PR.

The reason this didn't cause build failures before is because the release uses https://github.com/aws/aws-sdk-net/blob/main/sdk/test/IntegrationTests/AWSSDK.IntegrationTests.NetFramework.csproj instead.

PR #4285, opened 2026-01-12
S3 tests update - Part 1

dscpinheiro:
First batch of updating S3 test classes to be async (this is a follow-up of https://github.com/aws/aws-sdk-net/pull/4283)

## Testing
Dry-run: `DRY_RUN-48975a94-5550-448b-a0be-6c15d88e4668`

## License

### DIFF
--- a/sdk/test/Services/S3/IntegrationTests/BucketOwnershipControlsTests.cs
+++ b/sdk/test/Services/S3/IntegrationTests/BucketOwnershipControlsTests.cs
@@ -26,7 +26,7 @@ namespace AWSSDK_DotNet.IntegrationTests.Tests.S3
     [TestCategory("S3")]
     public class BucketOwnershipControlsTests : TestBase<AmazonS3Client>
     {
-        public static string bucketName;
+        private static string bucketName;
         private IAmazonS3 s3Client;
 
         [TestInitialize]


sdk/test/IntegrationTests/Utils/UtilityMethods.cs
```diff
@@ -258,22 +258,6 @@ await WaitUntilAsync(async () =>
             return result;
         }
 
-        public static async Task WaitUntilExceptionAsync(Func<Task> asyncAction, int sleepSeconds = 5, int maxWaitSeconds = 300)
```
dscpinheiro, line 261: I added this method, but it doesn't work the same way as its sync version. It's not being used at the moment, so I deleted it.

I'll re-add it if needed, but see `sdk/test/Services/S3/IntegrationTests/BucketPolicyTests.cs` for an example of how I changed the sync method not to use this wait anymore.

sdk/test/Services/S3/IntegrationTests/AccelerateTests.cs
```diff
@@ -78,170 +77,168 @@ public void TestAccelerateEnabledClient()
...
+            using (var explicitAccelerateEndpointClient = new AmazonS3Client(new AmazonS3Config
             {
                 RegionEndpoint = TestRegionEndpoint,
                 UseAccelerateEndpoint = true
             }))
             {
-                TestAccelerateUnsupportedOperations(explicitAccelerateEndpointClient);
-                TestControlPlaneOperations(explicitAccelerateEndpointClient);
-                TestDataPlaneOperations(explicitAccelerateEndpointClient);
+                await TestAccelerateUnsupportedOperations(explicitAccelerateEndpointClient);
```
ashishdhingra, line 119: nit: Do you want to rename methods with `Async` suffix just to be consistent with .NET standard?

sdk/test/Services/S3/IntegrationTests/AccessPointTests.cs
```diff
@@ -59,7 +60,7 @@ await s3ControlClient.DeleteAccessPointAsync(new DeleteAccessPointRequest
         }
 
         [TestMethod]
-        public void PutAndGetObject()
+        public async Task PutAndGetObject()
```
ashishdhingra, line 63: nit: Same as before. Perhaps rename with `Async` suffix.

sdk/test/Services/S3/IntegrationTests/AccelerateTests.cs
dscpinheiro: Previous discussion here: https://github.com/aws/aws-sdk-net/pull/4178#discussion_r2585982634 (TL;DR: I can, but does make the reviews larger and this is test code that customers won't see)

PR #4290, opened 2026-01-13
S3 tests update - Part 3

dscpinheiro:
Third batch of updating S3 test classes to be async (this is a follow-up of https://github.com/aws/aws-sdk-net/pull/4288). Most of the files have the same changes, but two classes were refactored further:
* `sdk/test/Services/S3/IntegrationTests/GeneratePreSignedUrlTests.cs`: Similar to the previous PR, this class had tests that were skipped due to using session credentials. I also added a setup method so only the minimum amount of buckets are created.
* `sdk/test/Services/S3/IntegrationTests/KMSTests.cs`: Updated to also only create new buckets when needed; Removed redundant tests (with logic already validated in this same class); And moved tests that don't make service calls to the unit test project.

## Testing
Dry-run: `DRY_RUN-e897593e-a33a-4992-b772-672ae6410d06`

## License

### DIFF
--- a/sdk/test/Services/S3/IntegrationTests/GeneratePreSignedUrlTests.cs
+++ b/sdk/test/Services/S3/IntegrationTests/GeneratePreSignedUrlTests.cs
@@ -91,6 +91,8 @@ namespace AWSSDK_DotNet.IntegrationTests.Tests.S3
 
             await AmazonS3Util.DeleteS3BucketWithObjectsAsync(_euNorth1Client, _euNorth1BucketName);
             _euNorth1Client.Dispose();
+
+            BaseClean();
         }
 
         [TestMethod]


sdk/test/Services/S3/IntegrationTests/GeneratePreSignedUrlTests.cs
```diff
@@ -18,166 +18,171 @@
...
+                BucketName = _euNorth1BucketName,
+                Key = TestKey,
+                ContentBody = TestContent
+            });
+        }
+
+        [ClassCleanup]
+        public static async Task ClassCleanup()
+        {
+            await AmazonS3Util.DeleteS3BucketWithObjectsAsync(_usEast1Client, _usEast1BucketName);
```
peterrsongg, line 88: should this be a try, finally so that if for whatever reason any of these calls fail it will still dispose?

sdk/test/Services/S3/IntegrationTests/S3TestUtils.cs
```diff
@@ -495,31 +489,31 @@ internal static void PutAndGetObjectTestHelper(IAmazonS3 s3Client, string bucket
...
             {
                 BucketName = bucketName,
                 Key = keyName,
                 Verb = HttpVerb.GET,
                 Expires = DateTime.UtcNow + TimeSpan.FromDays(5)
             });
 
-            var httpRequest = HttpWebRequest.Create(presignedUrl);
-            using (var httpResponse = httpRequest.GetResponse())
+            var httpRequest = WebRequest.Create(presignedUrl);
```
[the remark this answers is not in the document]
dscpinheiro: Similar to https://github.com/aws/aws-sdk-net/pull/4290#discussion_r2688191739 (which I added a TODO for)

sdk/test/Services/S3/IntegrationTests/KMSTests.cs
```diff
@@ -13,171 +12,123 @@
...
+        [TestMethod]
+        public async Task TestKmsOverHttp()
         {
-            var client = new AmazonS3Client(RegionEndpoint.USWest2);
-            var defaultEndpointClient = new AmazonS3Client(RegionEndpoint.USEast1);
+            var config = new AmazonS3Config
+            {
+                RegionEndpoint = Client.Config.RegionEndpoint,
+                UseHttp = true
+            };
```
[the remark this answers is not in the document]
dscpinheiro: This is only a test that validates using HTTP will throw an exception. The default is false so it doesn't propagate to customer's applications.

sdk/test/Services/S3/IntegrationTests/GeneratePreSignedUrlTests.cs
dscpinheiro: This is what a lot of the existing tests do: https://github.com/aws/aws-sdk-net/blob/0d372debac4e929417a20505e819b3e70055f173/sdk/test/Services/S3/IntegrationTests/PutObjectTests.cs#L45-L50

One thing I do want to do later though (when we start using xUnit) is to have a central place create these buckets so not every class has to do it (using https://xunit.net/docs/shared-context).

peterrsongg: ah I see, so should we call `BaseClean` instead of `dispose` directly then?

dscpinheiro: Fixed here: https://github.com/aws/aws-sdk-net/pull/4290/commits/52487884cc0f4afd73cfef342f3726a2740c3709 (needs both since the `_usEast1Client` only exists in this class)

peterrsongg (approved):

PR #4292, opened 2026-01-14
Generate paginators for S3

peterrsongg:
## Description

This PR removes the custom paginators and generates paginators for S3. To accomplish this, some of the custom interfaces and classes for paginators had to be made partial. To prevent changes to every single paginator, I put conditionals in the generator so that only S3 has partial interfaces. ListPartsPaginator had to be excluded because the input and output token to paginate are not the same type which the custom code did correctly but the generated code didn't. So for that operation I left the custom paginator and added a customization hook to skip paginators. 

## Testing

Build id: 3032cb7e-6ce4-4afc-9556-c84d1046c50d
Assembly Comparison outputs only new types added (ListDirectoryBuckets paginator added)

## Types of changes

## Checklist

## License

### DIFF
--- a/generator/.DevConfigs/92e08a0f-9103-4349-b102-d071e25ca3c1.json
+++ b/generator/.DevConfigs/92e08a0f-9103-4349-b102-d071e25ca3c1.json
@@ -1,0 +1,11 @@
+{
+  "services": [
+    {
+      "serviceName": "S3",
+      "type": "patch",
+      "changeLogMessages": [
+        "Generate paginators for S3"
+      ]
+    }
+  ]
+}


generator/ServiceModels/s3/s3.customizations.json
```diff
@@ -1944,6 +1944,14 @@
         },
         "GetBucketReplication": {
             "documentation": "Retrieves the replication configuration for the given Amazon S3 bucket."
+        },
+        "ListVersions":{
+            "excludeResultKey":[
+                "DeleteMarkers"
```
peterrsongg, line 1950: DeleteMarkers is not defined on` ListVersions` due to how it was written in the custom class. This result key needs to be excluded as such.

generator/ServiceClientGeneratorLib/Generators/SourceFiles/ServicePaginatorFactoryInterface.tt
```diff
@@ -14,7 +14,20 @@ namespace <#=this.Config.Namespace#>.Model
     /// <summary>
     /// Paginators for the <#=this.Config.ServiceNameRoot#> service
     ///</summary>
+<#
+if (this.Config.ServiceId == "S3")
```
dscpinheiro, line 18: This is fine for now but in the new generator we should probably make all interfaces partial (to avoid hard-coding service IDs...)

sdk/src/Services/S3/Generated/Model/_bcl+netstandard/ListObjectsV2Paginator.cs
```diff
@@ -71,15 +76,15 @@ IEnumerable<ListObjectsV2Response> IPaginator<ListObjectsV2Response>.Paginate()
                 continuationToken = response.NextContinuationToken;
                 yield return response;
             }
-            while (response.IsTruncated.GetValueOrDefault());
```
dscpinheiro, line 74: I _think_ this is equivalent based on the docs (https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html):
> NextContinuationToken is sent when isTruncated is true, which means there are more keys in the bucket that can be listed.

But I don't see it called out in the PR description (also we don't run integration tests for paginators, so you if wanted to double check that needs to be done manually).

generator/ServiceClientGeneratorLib/Generators/SourceFiles/ServicePaginatorFactoryInterface.tt
peterrsongg: yeah agreed. Will do it as a follow up

muhammad-othman: +1

muhammad-othman (approved):

PR #4294, opened 2026-01-14
S3 tests update - Part 5

dscpinheiro:
Another batch of S3 tests updates (following https://github.com/aws/aws-sdk-net/pull/4293). 

Apologies for the larger PR, this includes everything that was left except for the Transfer Utility tests (which will be the final PR later):
- For tests targeting S3 Express, there were assumptions about running in `us-east-1` (such as using `use1-az2`) so I created a client for that region (later, I want to update all tests to use the region configured in the environment unless necessary to validate a specific behavior)
- In `sdk/test/Services/S3/IntegrationTests/PutObjectTests.cs`, I removed a couple of tests that were either redundant (e.g. validating marshalling was done properly) or belonged somewhere else (e.g. testing unusual keys - which are now in `sdk/test/Services/S3/IntegrationTests/KeyNameTests.cs`)
- Also removed a couple of tests from `sdk/test/Services/S3/IntegrationTests/S3ExpressTests.cs`, since there were separate cases for uploading an object via MPU and another one for the TransferUtility (which would already exercise the same behavior)

## Testing
Dry-run: `DRY_RUN-5247086c-1ecc-4042-9a7c-bfacc3f7dee7`

## License

### DIFF
--- a/sdk/test/Services/S3/IntegrationTests/S3ExpressTests.cs
+++ b/sdk/test/Services/S3/IntegrationTests/S3ExpressTests.cs
@@ -1,10 +1,11 @@
-using Amazon;
+﻿using Amazon;
 using Amazon.S3;
 using Amazon.S3.Model;
 using Amazon.S3.Transfer;
 using Amazon.S3.Util;
 using Amazon.SecurityToken;
 using Amazon.SecurityToken.Model;
+using Amazon.Util;
 using AWSSDK_DotNet.IntegrationTests.Utils;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using System;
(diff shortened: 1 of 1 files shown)


sdk/test/Services/S3/IntegrationTests/S3ExpressTests.cs
```diff
@@ -265,338 +242,89 @@ public void Test_CopyObject_BetweenRegularBucket_And_S3ExpressBucket()
...
-                    File.Delete(filePath);
-                if (File.Exists(retrievedFilepath))
-                    File.Delete(retrievedFilepath);
-            }
-        }
-
-        [TestMethod]
-        [TestCategory("S3")]
-        public void Test_TransferUtility()
+        public async Task Test_TransferUtility()
```
muhammad-othman, line 453: I don't think we should remove these tests.
If I remember correctly, these tests were test cases in the SEP, and they helped us catching few bugs.

dscpinheiro: Re-introduced the 2 test cases in https://github.com/aws/aws-sdk-net/pull/4294/commits/7b064de37801fc01a90c4c5ab79e5765925b7cc7

Issue #1822, opened 2021-03-30
Rethrowing exceptions hides system traces, use ExceptionDispatchInfo.Throw instead


dave-yotta:
https://github.com/aws/aws-sdk-net/blob/475822dec5e87954b7a47ac65995714ae1f1b115/sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_netstandard/HttpRequestMessageFactory.cs#L547

Here you are rethrowing, which loses the trace into HttpClient.  This prevents me from investigating the cause of specific IOException by looking through the .net source.  Please use `ExceptionDispatchInfo.Throw` so that the trace is not replaced with a trace to the throw statment in the SDK.

dave-yotta:
The issue has not been resolved.

normj:
Version 4.0.3.9 of AWSSDK.Core was released today with this change. Note it was only done for the .NET 8+ targets due to the support level of `ExceptionDispatchInfo.Throw`.

PR #4296, opened 2026-01-15
Use ExceptionDispatchInfo.Throw 

normj:
## Description
In the HttpRequestMessageFactory we had several instances we were rethrowing the inner exception. This was causing the stacktrace for the inner exceptions to be reset from the point we were rethrowing the exception. To help debugging it would be helpful to maintain the original context so updated the code to use `ExceptionDispatchInfo.Throw`.

## Motivation and Context
https://github.com/aws/aws-sdk-net/issues/1822

## Testing
Dry Run: Success (DRY_RUN-bc01d7ad-b32c-45d3-b050-6816c157f56e)

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_netstandard/HttpRequestMessageFactory.cs
+++ b/sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_netstandard/HttpRequestMessageFactory.cs
@@ -563,9 +563,9 @@ namespace Amazon.Runtime
         }
 #endif
 
-                        /// <summary>
-                        /// Aborts the HTTP request.
-                        /// </summary>
+        /// <summary>
+        /// Aborts the HTTP request.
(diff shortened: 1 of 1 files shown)


sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/_netstandard/HttpRequestMessageFactory.cs
```diff
@@ -552,9 +563,9 @@ public IWebResponseData GetResponse()
         }
 #endif
 
-        /// <summary>
-        /// Aborts the HTTP request.
-        /// </summary>
+                        /// <summary>
```
dscpinheiro: Is this summary change intentional?

normj: Not intentional and reverted

Issue #4036, opened 2025-10-11
DynamoDB property attribute  converter is ignored if global type converter is already registered


Keksilton:
### Describe the bug

Global DynamoDB converters registered with `context.ConverterCache.Add` take precedence over converters specified using property attribute.

I'm suspecting that this is a duplicate, but I was unable to find issues regarding property converter precedence

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

I would expect the property specific converter to take precedence over any globally registered converters

### Current Behavior

Converters registered with `context.ConverterCache.Add` take precedence over converters specified with `DynamoDBProperty` attribute

### Reproduction Steps

1. Create 2 hashset converters with different comparers
2. Register one of them globally and another one with `DynamoDBProperty` attribute
3. Read the object
4. Verify that converter specified with `DynamoDBProperty` attribute was used instead of the global converter.

### Possible Solution

Looking into the issue I found that this is the likely source of the issue:

https://github.com/aws/aws-sdk-net/blob/cf2f0787d8149dfcfced66968bf713d5468f72b5/sdk/src/Services/DynamoDBv2/Custom/DataModel/InternalModel.cs#L254C13-L258C14

```cs
IPropertyConverter converter;
if (context.ConverterCache.TryGetValue(MemberType, out converter) && converter != null)
{
    this.Converter = converter;
}
```

Replacing it with the following code resolves my issue, the property converter takes precedence.

```cs
if (ConverterType == null)
{
    IPropertyConverter converter;
    if (context.ConverterCache.TryGetValue(MemberType, out converter) && converter != null)
    {
        this.Converter = converter;
    }
}
```

I did not test this "fix" thoroughly. Current behavior was introduced more than 10 years ago and even if we could agree that this is a bug, there might be software that relies on it.

### Additional Information/Context

Simplified code reproducing this issue:

```cs

public class SetOfStringsConverter : IPropertyConverter
{
    private readonly IEqualityComparer<string>? _comparer;

    public SetOfStringsConverter()
    {
        
    }

    public SetOfStringsConverter(IEqualityComparer<string>? comparer)
    {
        _comparer = comparer;
    }
    
    public DynamoDBEntry ToEntry(object value)
    {
        if (value is not HashSet<string> ls)
            throw new Exception("");


        var res = new PrimitiveList();

        foreach (var e in ls)
        {
            res.Add(new Primitive(e));
        }

        return res;
    }

[...]

PR #4323, opened 2026-02-03
Fix DynamoDB property converter precedence when a global converter is…

aanton-git:
## Description
There is a bug in the AWS .NET SDK’s DynamoDB object persistence model where property-specific converters (defined via the DynamoDBProperty(Converter = …) attribute) are ignored if a global converter is already registered in the DynamoDBContext.ConverterCache. The global converter takes precedence even when a property has its own converter defined, which is unexpected behavior and contrary to how attribute-based converters should work. ([https://github.com/aws/aws-sdk-net/issues/4036][1])

## Motivation and Context
**Current Behavior**
When using:
- A global converter registered via context.ConverterCache.Add(typeof(T), converter), and
- A per-property converter specified with the DynamoDBProperty(Converter = typeof(...)) attribute,
the SDK always picks the global converter first, ignoring the more specific property converter. ([https://github.com/aws/aws-sdk-net/issues/4036][1])

 **Expected Behavior**
Property-specific converters should take precedence over globally registered converters. That is:
1. If a property has a converter attribute defined, use that.
2. Otherwise, fall back to a global converter from the converter cache. ([[GitHub](https://github.com/aws/aws-sdk-net/issues/4036?utm_source=chatgpt.com)][1])

## Types of changes

## Checklist

- [x ] I have added tests to cover my changes
- [ x] All new and existing tests passed

## License

### DIFF
--- a/generator/.DevConfigs/214e45e8-60ef-4e07-8f70-7fb4dc554186.json
+++ b/generator/.DevConfigs/214e45e8-60ef-4e07-8f70-7fb4dc554186.json
@@ -4,7 +4,7 @@
       "serviceName": "DynamoDBv2",
       "type": "patch",
       "changeLogMessages": [
-        "Fix DynamoDB property converter precedence when a global converter is registered."
+        "[Breaking Change] Fix DynamoDB property converter precedence when a global converter is registered."
       ]
     }
   ]


sdk/test/Services/DynamoDBv2/UnitTests/Custom/DocumentModel/PropertyStorageTests.cs
```diff
@@ -244,12 +240,16 @@ public void Validate_AllowsIsAutoGeneratedTimestampOnDateTime()
         [ExpectedException(typeof(InvalidOperationException))]
         public void Validate_ThrowsIsAutoGeneratedTimestampOnNonDateTime()
         {
+            var mockClient = new Mock<IAmazonDynamoDB>();
```
irina-herciu: can this be moved to test class Setup?

generator/.DevConfigs/214e45e8-60ef-4e07-8f70-7fb4dc554186.json
```diff
@@ -0,0 +1,11 @@
+{
+  "services": [
+    {
+      "serviceName": "DynamoDBv2",
+      "type": "patch",
+      "changeLogMessages": [
+        "Fix DynamoDB property converter precedence when a global converter is registered."
```
normj: There is the small possibility users possibly unintentionally depended on the current behavior. I think making the fix is the right choice but for transparency can you add `[Breaking Change] ` as a prefix to the log message.

aanton-git: Updated

normj:
Sent branch into internal validation build. Assuming it passes then I'll approve the PR.

PR #4327, opened 2026-02-05
Add unified platform map for cross-platform API documentation

### LINKED ISSUE #3938: ReadWriteTimeout property missing from AmazonBedrockRuntimeConfig
### Describe the bug

When attempting to configure AmazonBedrockRuntimeConfig with the ReadWriteTimeout property as shown in the official documentation, the compiler throws an error stating that the property does not exist. The [AWS SDK documentation](https://docs.aws.amazon.com/sdkfornet/v4/apidocs/items/BedrockRuntime/TBedrockRuntimeConfig.html) clearly states this property should be inherited from Amazon.Runtime.ClientConfig, but it's not accessible in practice.

### Regression Issue

### Expected Behavior

The ReadWriteTimeout property should be available on AmazonBedrockRuntimeConfig as documented:
- Property should be settable with System.Nullable<System.TimeSpan> values
- Should inherit from Amazon.Runtime.ClientConfig like other AWS service configurations
- Should allow fine-grained control over read/write timeouts for Bedrock operations

```csharp
var config = new AmazonBedrockRuntimeConfig
{
    RegionEndpoint = RegionEndpoint.USEast1,
    Timeout = TimeSpan.FromMinutes(20),
    MaxErrorRetry = 3,
    RetryMode = Amazon.Runtime.RequestRetryMode.Adaptive,
    ReadWriteTimeout = TimeSpan.FromMinutes(20), // This should work
};
```

### Current Behavior

Compilation fails with the following error:
```
CS0117: 'AmazonBedrockRuntimeConfig' does not contain a definition for 'ReadWriteTimeout'
```

The property is completely unavailable, forcing developers to remove it and rely only on the general Timeout property, which may not provide adequate control for long-running AI model inference requests.

### Reproduction Steps

1. Create a new .NET project
2. Install `AWSSDK.BedrockRuntime` package (version 4.0.2 or 4.0.2.1)
3. Add the following code:

```csharp
using Amazon;
using Amazon.BedrockRuntime;

var config = new AmazonBedrockRuntimeConfig
{
    RegionEndpoint = RegionEndpoint.USEast1,
    Timeout = TimeSpan.FromMinutes(10),
    ReadWriteTimeout = TimeSpan.FromMinutes(20), // Compilation error here
};

var client = new AmazonBedrockRuntimeClient(config);
```

### Possible Solution

_No response_

_No response_

- **Primary Package**: `AWSSDK.BedrockRuntime` version `4.0.2.1`
- **Core Package**: `AWSSDK.Core` version `4.0.0.17`
- **Issue reproduced with**: Both `4.0.2` and `4.0.2.1` versions of BedrockRuntime

### Targeted .NET Platform

.NET 9

Windows 11

AlexDaines:
## Description

Adds PlatformAvailabilityMap infrastructure for tracking API availability across .NET platforms (net472, netstandard2.0, netcoreapp3.1, net8.0). Enables documentation of platform-exclusive APIs like H2 eventstream methods that only exist in net8.0.

## Motivation and Context

Fixes #3938 - Customers see documented APIs that don't exist on their target platform, wasting time trying to use APIs that won't compile.

This PR ships the infrastructure only. Badge UX deferred to follow-up PR pending design review.

## Testing
Dryruns (03/31/26): 
dotnetv4 (passed) -  a4511f2c-a7d8-49ce-813a-f685abc46ed0 
powershell5 (passed) - 6c0420ce-829e-42f9-a3a3-560b5dd86684 

- Built doc generator: `dotnet build SDKDocGenerator/SDKDocGenerator.csproj`
- Generated test docs for BedrockRuntime and S3 services
- Verified H2 exclusive method pages generated (InvokeModelWithBidirectionalStreamAsync)
- Verified Version Information section unchanged
- No new build errors (only expected obsolete warnings)

None. This is additive infrastructure with no changes to existing doc output behavior.

TTranscribeStreamingClient.html methods section:
Before (live docs):
image

After (artifacts from dryrun of this branch):
image

Example of New Individual Method Page:
image

Beyond Compare
Comparing the latest release build's doc build output to the doc build output from the dryrun of this feature branch:
image

## Types of changes

## Checklist

## License


docgenerator/SDKDocGeneratorLib/PlatformMap/MemberSignature.cs
```diff
@@ -0,0 +1,193 @@
...
+    ///
+    /// All signature generation delegates to NDocUtilities to ensure consistency
+    /// with XML documentation lookup signatures.
+    /// </summary>
+    public static class MemberSignature
+    {
+        /// <summary>
+        /// Length of the NDoc member-type prefix (e.g., "T:", "M:", "P:").
+        /// </summary>
+        private const int MemberTypePrefixLength = 2;
```
GarrettBeatty: nit: add new line between this

docgenerator/SDKDocGeneratorLib/PlatformMap/PlatformMapBuilder.cs
```diff
@@ -0,0 +1,373 @@
...
+            var loadedContexts = new List<PlatformAssemblyContext>();
+
+            // Primary platform is what the Options specify (usually net472)
+            var primaryPlatform = _options.Platform;
+
+            try
+            {
+                // First pass: Scan primary platform to establish baseline signatures
+                var primaryAssemblyPath = Path.GetFullPath(Path.Combine(_options.SDKAssembliesRoot, primaryPlatform, assemblyName + ".dll"));
+                if (File.Exists(primaryAssemblyPath))
```
GarrettBeatty: any reason why it wouldnt exist here? should we be throwing error if it doesnt exist instead

docgenerator/SDKDocGeneratorLib/PlatformMap/PlatformMapBuilder.cs
```diff
@@ -0,0 +1,373 @@
...
+                        {
+                            loadedContexts.Add(context);
+                            scannedPlatforms.Add(platform);
+                            if (_options.Verbose)
+                                Trace.WriteLine($"  Scanned {platform}: {memberIndex.Count} unique signatures total");
+                        }
+                    }
+                    catch (Exception ex)
+                    {
+                        Trace.WriteLine($"  WARNING: Failed to scan {platform}: {ex.Message}");
```
GarrettBeatty: suggestion: although this works, one thing thats kind of funky here is having separate logic for primary vs other ones.

why not have one loop (similar to the for each loop that you have `foreach (var platform in platforms)`

and then just have `isPrimary` set on each loop by checking if its equal to `_options.Platform`?


also the second thing that is kind of weird is the function takes in a list of platforms but then uses some things set from options.Platform to get the primary one. wondering if it makes sense to instead update platformmap builder signature to

```
public PlatformAvailabilityMap BuildMap(
            string serviceName,
            string assemblyName,
            IEnumerable<string> platformsToScan,
            string primaryPlatform,
            boolean isVerbose
        )
        {
```

and then you dont even need to pass in the `GeneratorOptions` into this class.

docgenerator/SDKDocGeneratorLib/Writers/ClassWriter.cs
```diff
@@ -1,4 +1,5 @@
-﻿using SDKDocGenerator.Syntax;
+﻿using SDKDocGenerator.PlatformMap;
+using SDKDocGenerator.Syntax;
 using System.Collections.Generic;
```
GarrettBeatty, line 2: question/suggestion: i think currently the way this works is something like

1. (`GenerationManifest.Generate()` → `WriteType()` → `new ClassWriter(this, version, type).Write()`) generates the class page with only net472 methods.

2. (`GenerationManifest.GenerateExclusivePagesFromMap()` → `new ClassWriter(this, version, primaryType, exclusiveMethods).Write()`) generates the __same class page again__, overwriting the file, now with exclusive methods merged in.


and then classwriter is called in two ways 

```
/ 1st cakk — normal, no supplemental methods
var writer = new ClassWriter(this, version, type);
writer.Write();

// 2nd call — with supplemental methods
var classWriter = new ClassWriter(this, version, primaryType, exclusiveMethods);
classWriter.Write();
```

but inside AddMethods theres some hidden branching with supplementalMethods branching.

wondering if its possible to instead just do 

```
// Caller assembles the complete method list
var methods = type.GetMethodsToDocument().ToList();
var exclusive = PlatformMap?.GetExclusiveMethodsForType(type.FullName) 
                ?? Enumerable.Empty<MethodInfoWrapper>();
methods.AddRange(exclusive);

// ClassWriter just renders what it's given
var writer = new ClassWriter(this, version, type, methods);
writer.Write();
```

and then you dont need to do a second pass

GarrettBeatty: the second reason for this question/suggestion is because im thinking classwriter should do exactly that - write stuff to the docs.

but in current state it has some business logic of deciding what to write, instead of just directly writing

docgenerator/SDKDocGeneratorLib/PlatformMap/MemberSignature.cs
AlexDaines: done

docgenerator/SDKDocGeneratorLib/PlatformMap/PlatformMapBuilder.cs
AlexDaines: was going back and forth on this. i changed it to throw FileNotFoundException for the primary assembly since its a requirement for the baseline docs, while still skipping supplemental platforms if needed.

docgenerator/SDKDocGeneratorLib/PlatformMap/PlatformMapBuilder.cs
AlexDaines: this was a really helpful comment. merged into a single loop with `isPrimary` derived from the platform name, and BuildMap now takes `primaryPlatform` and `isVerbose` directly instead of the whole options object (SDKAssembliesRoot stays on the constructor since it doesn't change per time called)

GarrettBeatty:
can you include screenshots of the before/after docs that get generated

docgenerator/SDKDocGeneratorLib/PlatformMap/PlatformMapBuilder.cs
```diff
@@ -0,0 +1,372 @@
...
+    /// 4. For exclusive members, store the MethodInfoWrapper for page generation
+    /// 5. Return map with contexts - caller disposes when done
+    ///
+    /// Memory Model:
+    /// - Assemblies stay loaded after BuildMap() returns
+    /// - Map holds references to all assembly contexts
+    /// - Caller must call map.Dispose() to release resources
+    /// - For typical builds (3-4 platforms), memory usage is acceptable
+    /// </summary>
+    public class PlatformMapBuilder
```
GarrettBeatty: i think this class could probably be static, and you can remove the sdk assemblies root from constructor and just pass it into the buildmap function.

just a suggestion

GarrettBeatty (commented):
one other small comment

docgenerator/SDKDocGeneratorLib/PlatformMap/PlatformMapBuilder.cs
AlexDaines: good call.. once the options were method params there was no instance state left. Made the class and all methods static. aslo took the opportunity to prune leftover API surface from a descoped badges feature

dscpinheiro:
Is this in the dry-run output expected?
> EXEC : warning : Failed to scan netcoreapp3.1: Unable to load one or more of the requested types. [C:\codebuild\tmp\output\src974\src\s3\00\aws-sdk-net\buildtools\doc-build.proj]

GarrettBeatty (dismissed):
talked via slack and showed screenshot of the diff. alex will upload to pr description.

can you re-run the dry run before merging

AlexDaines:
> Is this in the dry-run output expected?
> 
> > EXEC : warning : Failed to scan netcoreapp3.1: Unable to load one or more of the requested types. [C:\codebuild\tmp\output\src974\src\s3\00\aws-sdk-net\buildtools\doc-build.proj]

Yes expected. New from this pr. The platform scanner loads DLLs to reflect over types, (unlike the current generator which only reads XML doc files). netcoreapp3.1 throws a ReflectionTypeLoadException because of missing dependencies in the build env. The catch block at PlatformMapBuilder.cs:117 handles it and moves on. net8.0 and net472 both load normally

docgenerator/SDKDocGeneratorLib/PlatformMap/PlatformMapBuilder.cs
```diff
@@ -0,0 +1,366 @@
...
+        public static PlatformAvailabilityMap BuildMap(
+            string serviceName,
+            string assemblyName,
+            IEnumerable<string> platformsToScan,
+            string primaryPlatform,
+            string sdkAssembliesRoot,
+            bool isVerbose)
+        {
+            if (string.IsNullOrEmpty(serviceName))
+                throw new ArgumentNullException(nameof(serviceName));
```
boblodgett, line 49: Non-blocking but ArgumentNullException is misleading because it can also be empty on all of these except platformsToScan. ArgumentException with information that it could be null or empty would have been more accurate.

boblodgett (approved):

GarrettBeatty (approved):

PR #4330, opened 2026-02-06
Switch to using sorted parameter list in `GetParametersAsString` in AWSSDKUtils

danielmarbach:
See discussion in https://github.com/aws/aws-sdk-net/issues/1922#issuecomment-3862627555

// cc @peterrsongg 

## Breaking Changes Assessment

1. Identify all breaking changes including the following details:
    * What functionality was changed?
    * How will this impact customers?
    * Why does this need to be a breaking change and what are the most notable non-breaking alternatives?
    * Are best practices being followed?
    * How have you tested this breaking change?
2. Has a senior/+ engineer been assigned to review this PR?

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project
- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [x] I have read the **README** document
- [ ] I have added tests to cover my changes
- [x] All new and existing tests passed

## License

- [x] I confirm that this pull request can be released under the Apache 2 license

[issues]: https://github.com/aws/aws-sdk-net/issues
[license]: http://aws.amazon.com/apache2.0/
[cla]: http://en.wikipedia.org/wiki/Contributor_License_Agreement

Issue #1922, opened 2021-09-30
SES service allocates huge amounts of memory 


dziedrius:
## Description
We've noticed huge memory traffic in our app after we started to send larger attachments with SES, so I've composed small app to see what is going on. To send ~5MB email message SES SDK allocates > 250MB of memory, which is 50 times more than message itself, to send 4 such messages in parallel we need to allocate more than 1GB of RAM, which in my opinion is not right.

Screenshots are taken from dotMemory profiler, I was running console app below with allocations recording.
In total sending email allocated:
![image](https://user-images.githubusercontent.com/772162/135398845-c7a2b870-3f7d-4af6-85fd-48a506a4228f.png)
The call tree is very deep, hence did not fit in single screenshot, but main parts are two, CreateWebRequest:
![image](https://user-images.githubusercontent.com/772162/135399004-e05b48aa-cf3e-41a1-98d7-54313286601f.png)
and signing:
![image](https://user-images.githubusercontent.com/772162/135399112-27cd56ee-dbba-4f3a-940a-04a55f29f869.png)

Some strange things, I have to admit that I haven't checked the source code:
1. Seems that GetParametersAsString was called twice, hence allocating twice.
2. Why you need to sign request - I'm pretty sure it travels through HTTPS.
3. Why you need to encode everything as url and why it takes so much memory.

## Reproduction Steps
Project:
```
<Project Sdk="Microsoft.NET.Sdk">

    <PropertyGroup>
        <OutputType>Exe</OutputType>
        <TargetFramework>net5.0</TargetFramework>
    </PropertyGroup>

    <ItemGroup>
      <PackageReference Include="AWSSDK.SimpleEmail" Version="3.7.0.73" />
      <PackageReference Include="MimeKit" Version="2.15.1" />
    </ItemGroup>

</Project>
```
Program.cs
```
using System;
using System.IO;
using System.Threading.Tasks;
using Amazon;
using Amazon.Runtime;
using Amazon.SimpleEmail;
using Amazon.SimpleEmail.Model;
using MimeKit;

namespace SesMemory
{
    class Program
    {
        static async Task Main(string[] args)
        {
            var options = FormatOptions.Default.Clone();
            
            var sesKey = "my-ses-key";
            var sesSecret = "my-ses-secret";
            var region = "us-west-2";
            var amazonSimpleEmailService = new AmazonSimpleEmailServiceClient(
                new BasicAWSCredentials(sesKey, sesSecret),
[...]

PR #4331, opened 2026-02-06
Optimize sorting by leveraging ArrayPool for reduced allocations in ParameterCollection.

danielmarbach:
Alternative to #4330 

Commit `37a9afe` optimized `GetParametersAsString` to use `GetParametersEnumerable()` directly instead of `GetSortedParametersList().ToList()`. While this avoided the intermediate `List<KeyValuePair<string,string>>` allocation, it exposed the pre-existing mutation bug where the original lists were being sorted in-place on every call.

Use `ArrayPool<T>.Shared.Rent()` to obtain temporary arrays for sorting, then return them to the pool. This approach:
- Zero heap allocations for the sorting operation (just rents from the pool)
- Prevents mutation of the original lists
- Maintains performance comparable to the buggy in-place sort

## Benchmark Results

BenchmarkDotNet v0.15.0 shows the `ArrayPool.Rent` approach is competitive with alternatives:

image

- `ArrayPool.Rent` has identical performance to `new List<T>(copy)` but with ~40% less memory allocation
- `ArrayPool.Rent` is ~25% faster than LINQ OrderBy with ~3x less memory allocation
- The fix maintains the originally intended v4 optimization benefits while preventing the mutation bug

Although the span approach doesn't work due to the iterator. It is just here for comparison

Benchmark
<details>
```csharp
using System;
using System.Buffers;
using System.Globalization;
using System.Linq;
using System.Runtime.CompilerServices;

using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Columns;
using BenchmarkDotNet.Configs;
using BenchmarkDotNet.Diagnosers;
using BenchmarkDotNet.Exporters;
using BenchmarkDotNet.Jobs;
using BenchmarkDotNet.Reports;
using BenchmarkDotNet.Running;

namespace MutationBenchmark
{
    [MemoryDiagnoser]
    [SimpleJob]
    [Config(typeof(Config))]
    public class ParameterSortingBenchmarks
    {
        private List<string> _stringList;
        private List<double> _doubleList;

        [Params(3, 10, 50)]
        public int ListSize { get; set; }

        [GlobalSetup]
        public void GlobalSetup()
        {
            _stringList = GenerateStringList(ListSize);
            _doubleList = GenerateDoubleList(ListSize);
        }

        [Benchmark(Baseline = true, Description = "ArrayPool.Rent (FIX)")]
        public int ArrayPoolApproach()
        {
            int count = 0;

            var stringArray = ArrayPool<string>.Shared.Rent(_stringList.Count);
            try
            {
                _stringList.CopyTo(stringArray);
                Array.Sort(stringArray, 0, _stringList.Count, StringComparer.Ordinal);
                for (int i = 0; i < _stringList.Count; i++)
                {
                    count += stringArray[i].Length;
                }
            }
            finally
            {
                ArrayPool<string>.Shared.Return(stringArray);
            }

            var doubleArray = ArrayPool<double>.Shared.Rent(_doubleList.Count);
            try
            {
                _doubleList.CopyTo(doubleArray);
                Array.Sort(doubleArray, 0, _doubleList.Count);
                for (int i = 0; i < _doubleList.Count; i++)
                {
                    count += doubleArray[i].ToString(CultureInfo.InvariantCulture).Length;
                }
            }
            finally
            {
                ArrayPool<double>.Shared.Return(doubleArray);
            }

            return count;
        }

        [Benchmark(Description = "new List<T>(copy)")]
        public int ListCopyApproach()
        {
            int count = 0;

            var sortedStringList = new List<string>(_stringList);
            sortedStringList.Sort(StringComparer.Ordinal);
            foreach (var item in sortedStringList)
            {
                count += item.Length;
            }

            var sortedDoubleList = new List<double>(_doubleList);
            sortedDoubleList.Sort();
            foreach (var item in sortedDoubleList)
            {
                count += item.ToString(CultureInfo.InvariantCulture).Length;
            }

            return count;
        }

        [Benchmark(Description = "LINQ OrderBy")]
        public int LinqOrderByApproach()
        {
            int count = 0;

            foreach (var item in _stringList.OrderBy(x => x, StringComparer.Ordinal))
            {
                count += item.Length;
            }

            foreach (var item in _doubleList.OrderBy(x => x))
            {
                count += item.ToString(CultureInfo.InvariantCulture).Length;
            }

            return count;
        }

        [Benchmark(Description = "Span (stackalloc doubles)")]
        public int SpanApproach()
        {
            int count = 0;

            var stringArray = ArrayPool<string>.Shared.Rent(_stringList.Count);
            try
            {
                _stringList.CopyTo(stringArray);
                Array.Sort(stringArray, 0, _stringList.Count, StringComparer.Ordinal);
                for (int i = 0; i < _stringList.Count; i++)
                {
                    count += stringArray[i].Length;
                }
            }
            finally
            {
                ArrayPool<string>.Shared.Return(stringArray);
            }

            if (_doubleList.Count <= 256)
            {
                Span<double> doubleSpan = stackalloc double[_doubleList.Count];
                for (int i = 0; i < _doubleList.Count; i++)
                {
                    doubleSpan[i] = _doubleList[i];
                }

                doubleSpan.Sort();
                foreach (var item in doubleSpan)
                {
                    count += item.ToString(CultureInfo.InvariantCulture).Length;
                }
            }
            else
            {
                var doubleArray = ArrayPool<double>.Shared.Rent(_doubleList.Count);
                try
                {
                    _doubleList.CopyTo(doubleArray);
                    Array.Sort(doubleArray, 0, _doubleList.Count);
                    for (int i = 0; i < _doubleList.Count; i++)
                    {
                        count += doubleArray[i].ToString(CultureInfo.InvariantCulture).Length;
                    }
                }
                finally
                {
                    ArrayPool<double>.Shared.Return(doubleArray);
                }
            }

            return count;
        }

        [MethodImpl(MethodImplOptions.NoInlining)]
        private static List<string> GenerateStringList(int size)
        {
            var list = new List<string>(size);
            var random = new Random(42);
            var chars = "abcdefghijklmnopqrstuvwxyz";

            for (int i = 0; i < size; i++)
            {
                var value = new char[random.Next(5, 15)];
                for (int j = 0; j < value.Length; j++)
                {
                    value[j] = chars[random.Next(chars.Length)];
                }

                list.Add(new string(value));
            }

            return list;
        }

        [MethodImpl(MethodImplOptions.NoInlining)]
        private static List<double> GenerateDoubleList(int size)
        {
            var list = new List<double>(size);
            var random = new Random(42);

            for (int i = 0; i < size; i++)
            {
                list.Add(random.NextDouble() * 1000);
            }

            return list;
        }

        private class Config : ManualConfig
        {
            public Config()
            {
                AddExporter(MarkdownExporter.GitHub);
            }
        }
    }

    public static class Program
    {
        public static void Main(string[] args)
        {
            BenchmarkRunner.Run<ParameterSortingBenchmarks>(
                DefaultConfig.Instance
                    .WithOption(ConfigOptions.JoinSummary, true)
                    .WithOption(ConfigOptions.DisableOptimizationsValidator, true));
        }
    }
}
```
</details>

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/test/NetStandard/UnitTests/Core/AWSSDKUtilsTests.cs
+++ b/sdk/test/NetStandard/UnitTests/Core/AWSSDKUtilsTests.cs
@@ -50,7 +50,6 @@ namespace UnitTests.NetStandard.Core
                 {"key4", new List<string> { "value5", "value4" } },
                 {"key1", new List<string>{"value1", "value2"} }
             };
-            //check that the original StringParameterListValue isn't mutated.
             var result = AWSSDKUtils.GetParametersAsString(parameters);
             Assert.Equal("key1=value1&key1=value2&key4=value4&key4=value5", result);
         }


sdk/src/Core/Amazon.Runtime/ParameterCollection.cs
```diff
@@ -90,21 +91,43 @@ internal IEnumerable<KeyValuePair<string, string>> GetParametersEnumerable()
...
+                        try
+                        {
+                            stringList.CopyTo(stringArray);
+                            Array.Sort(stringArray, 0, stringList.Count, StringComparer.Ordinal);
+                            for (int i = 0; i < stringList.Count; i++)
+                                yield return new KeyValuePair<string, string>(name, stringArray[i]);
+                        }
+                        finally
+                        {
+                            ArrayPool<string>.Shared.Return(stringArray);
```
danielmarbach, line 106: Safer to clear?

danielmarbach: Even with clearing it is still reasonably fast

image

danielmarbach:
Using the pool here seems safe since all generated enumerator paths need to dispose of the enumerator, which triggers the finally. If someone were to acquire the enumerator manually and not dispose of it, then the finally would not be called. Given this is a wrong usage of the enumerator API and the method is internal, I would not be concerned about this. Additionally, there is also this hint

> Failure to return a rented buffer is not a fatal error. However, it may lead to decreased application performance, as the pool may need to create a new buffer to replace the lost one.

Which indicates it is not the end of the world. But again, if someone manually acquires the enumerator and doesn't follow the disposal guidelines, I guess that abuse of the internal method deserves to be punished with a small perf hit ;)

danielmarbach:
@peterrsongg This is ready for review

peterrsongg:
> @peterrsongg This is ready for review

Just doing our due-diligence on this. Someone else on the team is actively looking into this and has this assigned to them as I got pulled into something else. But I will be the second reviewer

sdk/src/Core/Amazon.Runtime/ParameterCollection.cs
```diff
@@ -90,21 +91,43 @@ internal IEnumerable<KeyValuePair<string, string>> GetParametersEnumerable()
                         yield return new KeyValuePair<string, string>(name, stringParameterValue.Value);
                         break;
                     case StringListParameterValue stringListParameterValue:
-                        var sortedStringListParameterValue = stringListParameterValue.Value;
-                        sortedStringListParameterValue.Sort(StringComparer.Ordinal);
-                        foreach (var listValue in sortedStringListParameterValue)
-                            yield return new KeyValuePair<string, string>(name, listValue);
+                    {
```
muhammad-othman, line 94: I checked the values of the parameters dictionary for the reproduction code provided in https://github.com/aws/aws-sdk-net/issues/1922 and none of them are `StringListParameterValue` or `DoubleListParameterValue`

Image

All of them are `StringParameterValue`, which means this change wont affect the issue and we wont make any improvements for the memory allocation.
This change looks useful to me and it would improve the allocations for cases where there are `ListParameterValue`s, but I wouldn't recommend merging it until we figure out the root cause of the issue as we might need to rework how `GetParametersEnumerable` work. 
Maybe revert to using `GetSortedParametersList` for now? What do you think @danielmarbach ?

danielmarbach: Overall when I did the small improvements here in the past I deliberately left out any redesigns. The entire parameter valua collection being mutable would be one thing to look at besides others.

Judging by the benchmark I ran this seems still giving good allocation gains that benefit's all SDK clients that run through this path I'd say it is worth and safe to keep.

muhammad-othman:
Hi @danielmarbach,
Sorry for the late review, I got distracted by other tasks. Still planning on getting back to the allocations xml issue but I wanted to focus on the json issue first as it got worse after we moved to v4.

danielmarbach:
No worries. There is no rush

danielmarbach:
The commit will no longer be co-authored?

muhammad-othman:
@danielmarbach I rewrote your commit using the same commands that I usually use for rewriting my commits which added me as the author, then it took me a while to figure out how to add you back as the author, didn't mean to take the credit for your change :sweat_smile:

I'm currently running a dry run for this PR in our internal build system, will approve it when it passes.

danielmarbach:
@muhammad-othman no worries. I'm not easily offended 😜 I just figured I'll ask. Happy coding!

muhammad-othman (dismissed):

sdk/test/NetStandard/UnitTests/Core/AWSSDKUtilsTests.cs
```diff
@@ -29,6 +29,20 @@ public void GetParametersAsString()
...
+            var parameters = new ParameterCollection
+            {
+                { "key3", "value3" },
+                { "key1", "value1" },
+                { "key2", "value2" },
+            };
+
+            var result = AWSSDKUtils.GetParametersAsString(parameters);
+            Assert.Equal("key1=value1&key2=value2&key3=value3", result);
+        }
```
peterrsongg, line 44: Since your pr is touching the StringListParameterValue code path, can we add a test like so to this file, which tests that the original List isn't mutated and that its sorted correctly? I know we have one above, but I think this covers the mutation concern and an unsorted list too.

```suggestion
        }
        [Fact]
        public void GetParametersAsStringWithStringListParameterValueCorrectlySortsOutput()
        {
            var parameters = new ParameterCollection
            {
                {"key4", new List<string> { "value5", "value4" } },
                {"key1", new List<string>{"value1", "value2"} }
            };
            //check that the original StringParameterListValue isn't mutated.
            var originalKey4 = new List<string>(((StringListParameterValue)parameters["key4"]).Value);
            var result = AWSSDKUtils.GetParametersAsString(parameters);
            Assert.Equal("key1=value1&key1=value2&key4=value4&key4=value5", result);
            Assert.Equal(originalKey4, ((StringListParameterValue)parameters["key4"]).Value);
        }
```

I've tested this locally and it compiles and runs btw.

peterrsongg (changes requested):
I just added a suggestion for a unit test that I think would cover making sure the original list isn't mutated as you intended in this PR. Do you mind adding that to the PR and then I'd be happy to hit approve. Thank you for the PR

sdk/test/NetStandard/UnitTests/Core/AWSSDKUtilsTests.cs
danielmarbach: With the whitespace or without?

danielmarbach: Fyi this PR has edit rights for maintainers so you could also just push if you prefer that

peterrsongg: ah the whitespace after the end bracket on line 44 wasn't intentional

peterrsongg: fixed it, i'll commit the suggestion and hit approve / merge

sdk/test/NetStandard/UnitTests/Core/AWSSDKUtilsTests.cs
```diff
@@ -42,6 +42,20 @@ public void GetParametersAsString_ProducesCorrectlySortedOutput()
             var result = AWSSDKUtils.GetParametersAsString(parameters);
             Assert.Equal("key1=value1&key2=value2&key3=value3", result);
         }
+        [Fact]
+        public void GetParametersAsStringWithStringListParameterValueCorrectlySortsOutput()
```
muhammad-othman, line 46: I don't like this test because it’s too tied to the current implementation and doesn’t check for the intended behavior. 
Right now, the implementation doesn’t mutate the list, but if we ever find a more efficient approach that does, we should use it. Adding this test, though, makes it seem like we’re against mutating the list, which I don’t think is the case.

sdk/test/NetStandard/UnitTests/Core/AWSSDKUtilsTests.cs
```diff
@@ -29,6 +29,34 @@ public void GetParametersAsString()
...
+            var parameters = new ParameterCollection
+            {
+                {"key4", new List<string> { "value5", "value4" } },
+                {"key1", new List<string>{"value1", "value2"} }
+            };
+            //check that the original StringParameterListValue isn't mutated.
+            var originalKey4 = new List<string>(((StringListParameterValue)parameters["key4"]).Value);
+            var result = AWSSDKUtils.GetParametersAsString(parameters);
+            Assert.Equal("key1=value1&key1=value2&key4=value4&key4=value5", result);
+            Assert.Equal(originalKey4, ((StringListParameterValue)parameters["key4"]).Value);
```
peterrsongg:
```suggestion
```

sdk/test/NetStandard/UnitTests/Core/AWSSDKUtilsTests.cs
```diff
@@ -29,6 +29,33 @@ public void GetParametersAsString()
...
+        [Fact]
+        public void GetParametersAsStringWithStringListParameterValueCorrectlySortsOutput()
+        {
+            var parameters = new ParameterCollection
+            {
+                {"key4", new List<string> { "value5", "value4" } },
+                {"key1", new List<string>{"value1", "value2"} }
+            };
+            //check that the original StringParameterListValue isn't mutated.
+            var originalKey4 = new List<string>(((StringListParameterValue)parameters["key4"]).Value);
```
peterrsongg:
```suggestion
```

sdk/test/NetStandard/UnitTests/Core/AWSSDKUtilsTests.cs
```diff
@@ -29,6 +29,32 @@ public void GetParametersAsString()
...
+        }
+        [Fact]
+        public void GetParametersAsStringWithStringListParameterValueCorrectlySortsOutput()
+        {
+            var parameters = new ParameterCollection
+            {
+                {"key4", new List<string> { "value5", "value4" } },
+                {"key1", new List<string>{"value1", "value2"} }
+            };
+            //check that the original StringParameterListValue isn't mutated.
```
peterrsongg:
```suggestion
```

PR #4335, opened 2026-02-18
Optimize MemoryStream Base64 serialization in JSON marshallers

muhammad-othman:
## Description
This PR reduces memory allocations when serializing `MemoryStream` (blob) properties in JSON-based service marshallers by eliminating an unnecessary intermediate Base64 string allocation.

When marshalling `MemoryStream` properties (e.g., raw email data in `SESV2`), the generated code previously called:
`context.Writer.WriteStringValue(StringUtils.FromMemoryStream(requestObject.Data));`

This created a two-step process:

1. `FromMemoryStream()` converts the entire stream to a __Base64 `string`__
2. `WriteStringValue()` re-encodes that string to UTF-8 bytes in the JSON output

This intermediate string allocation is entirely unnecessary since `Utf8JsonWriter` natively supports writing Base64 directly from binary data via `WriteBase64StringValue(ReadOnlySpan<byte>)` which is what was add in this PR.
This improvement applies to __all JSON-based services__ that marshal `MemoryStream`/blob properties, not just `SESV2`.

Comparing the old `FromMemoryStream` + `WriteStringValue` approach vs the new `WriteBase64StringValue` approach for serializing a 5MB MIME message:
| Method                                | Job           | Runtime       | Mean      | Error     | StdDev     | Ratio | RatioSD | Gen0      | Gen1      | Gen2      | Allocated | Alloc Ratio |
|-------------------------------------- |-------------- |-------------- |----------:|----------:|-----------:|------:|--------:|----------:|----------:|----------:|----------:|------------:|
| Old_FromMemoryStream_WriteStringValue | .NET 6.0      | .NET 6.0      | 29.147 ms | 0.6410 ms |  1.8391 ms |  1.00 |    0.09 | 2000.0000 | 2000.0000 | 2000.0000 | 195.37 MB |        1.00 |
| New_WriteBase64StringValue            | .NET 6.0      | .NET 6.0      |  7.105 ms | 0.2020 ms |  0.5829 ms |  0.24 |    0.02 | 1000.0000 | 1000.0000 | 1000.0000 |  49.13 MB |        0.25 |
|                                       |               |               |           |           |            |       |         |           |           |           |           |             |
| Old_FromMemoryStream_WriteStringValue | .NET 8.0      | .NET 8.0      | 37.141 ms | 4.7770 ms | 14.0850 ms |  1.16 |    0.65 | 2000.0000 | 2000.0000 | 2000.0000 | 195.37 MB |        1.00 |
| New_WriteBase64StringValue            | .NET 8.0      | .NET 8.0      |  6.491 ms | 0.2993 ms |  0.8685 ms |  0.20 |    0.08 |         - |         - |         - |  33.12 MB |        0.17 |
|                                       |               |               |           |           |            |       |         |           |           |           |           |             |
| Old_FromMemoryStream_WriteStringValue | .NET Core 3.1 | .NET Core 3.1 | 16.626 ms | 0.3524 ms |  0.9882 ms |  1.00 |    0.08 | 1000.0000 | 1000.0000 | 1000.0000 |  61.59 MB |        1.00 |
| New_WriteBase64StringValue            | .NET Core 3.1 | .NET Core 3.1 |  6.498 ms | 0.3580 ms |  1.0498 ms |  0.39 |    0.07 | 1000.0000 | 1000.0000 | 1000.0000 |  34.22 MB |        0.56 |

__On .NET 8.0__: 75% faster, 83% less memory allocated, and zero GC pressure (Gen0/1/2 all eliminated).

Comparing SES V1 vs V2 sending a 5MB email with attachment __before this fix__:

| Method                   | Mean     | Error    | StdDev    | Median   | Ratio | RatioSD | Gen0      | Gen1      | Gen2      | Allocated | Alloc Ratio |
|------------------------- |---------:|---------:|----------:|---------:|------:|--------:|----------:|----------:|----------:|----------:|------------:|
| SimpleEmail_SendRawEmail | 792.0 ms | 23.96 ms |  66.39 ms | 794.4 ms |  1.01 |    0.12 | 2000.0000 | 2000.0000 | 2000.0000 |  277.6 MB |        1.00 |
| SimpleEmailV2_SendEmail  | 829.8 ms | 85.64 ms | 252.50 ms | 745.5 ms |  1.05 |    0.33 | 1000.0000 | 1000.0000 | 1000.0000 |   67.5 MB |        0.24 |

Same benchmark __after applying this fix__:

| Method                   | Mean     | Error    | StdDev   | Ratio | RatioSD | Gen0      | Gen1      | Gen2      | Allocated | Alloc Ratio |
|------------------------- |---------:|---------:|---------:|------:|--------:|----------:|----------:|----------:|----------:|------------:|
| SimpleEmail_SendRawEmail | 752.5 ms | 15.61 ms | 44.02 ms |  1.00 |    0.08 | 2000.0000 | 2000.0000 | 2000.0000 | 277.61 MB |        1.00 |
| SimpleEmailV2_SendEmail  | 541.9 ms | 16.71 ms | 47.68 ms |  0.72 |    0.08 |         - |         - |         - |  33.24 MB |        0.12 |

__After the fix, SES V2 allocates only 33.24 MB (down from 67.5 MB — a 51% reduction), is 28% faster, and generates zero GC pressure.__ Compared to V1, V2 now allocates 88% less memory.

> __Note:__ The V1 and V2 benchmarks were run separately to avoid assembly referencing issues between the two SDK versions.

## Motivation and Context
https://github.com/aws/aws-sdk-net/issues/1922

## Testing
- Benchmarked using `BenchmarkDotNet` comparing old vs new serialization approaches with a 5MB MIME email payload
- `DRY_RUN-7b49ff35-34e6-4522-903a-9f9804867aad`.
- We have protocol tests that was affected by this change and did succeed in the dry run.
- Added unit tests to test `WriteBase64StringValue` compared to `FromMemoryStream`.

## Types of changes

## Checklist

## License

### DIFF
--- a/generator/.DevConfigs/5ba72673-5d59-4b0a-9206-6237d869c0f0.json
+++ b/generator/.DevConfigs/5ba72673-5d59-4b0a-9206-6237d869c0f0.json
@@ -1,0 +1,9 @@
+{
+  "core": {
+    "updateMinimum": true,
+    "type": "patch",
+    "changeLogMessages": [
+      "Optimize MemoryStream Base64 serialization in JSON marshallers (Fix: https://github.com/aws/aws-sdk-net/issues/1922)."
+    ]
+  }
+}
--- a/sdk/src/Core/Amazon.Runtime/Internal/Util/StringUtils.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Util/StringUtils.cs
@@ -22,6 +22,7 @@ using Amazon.Util;
 using System.Linq;
 using System.Diagnostics.CodeAnalysis;
 using System.Buffers;
+using System.Text.Json;
 
 namespace Amazon.Runtime.Internal.Util
 {
@@ -61,6 +62,46 @@ namespace Amazon.Runtime.Internal.Util
                 {
                     value.Read(array, 0, (int)value.Length);
                     return Convert.ToBase64String(array, 0, (int)value.Length);
+                }
+                finally
+                {
+                    ArrayPool<byte>.Shared.Return(array);
+                }
+            }
+        }
+
+        /// <summary>
+        /// Writes the contents of a MemoryStream as a Base64-encoded JSON string value
+        /// directly to a Utf8JsonWriter, avoiding the intermediate Base64 string allocation
+        /// that would occur when using <see cref="FromMemoryStream"/> with WriteStringValue.
+        /// </summary>
+        /// <remarks>
+        /// This method always encodes the entire contents of the MemoryStream, regardless of
+        /// the current <see cref="MemoryStream.Position"/>. When <see cref="MemoryStream.TryGetBuffer"/>
+        /// succeeds, the underlying buffer is accessed directly without modifying the stream.
+        /// When TryGetBuffer fails (e.g., for MemoryStreams wrapping a byte array), the stream's
+        /// <see cref="MemoryStream.Position"/> is reset to 0 before reading. In this fallback case,
+        /// the caller's stream position will be modified as a side effect.
+        /// </remarks>
+        /// <param name="writer">The Utf8JsonWriter to write to.</param>
+        /// <param name="value">The MemoryStream containing the binary data to encode.</param>
+        public static void WriteBase64StringValue(Utf8JsonWriter writer, MemoryStream value)
+        {
+            if (value.TryGetBuffer(out var buffer))
+            {
+                writer.WriteBase64StringValue(new ReadOnlySpan<byte>(buffer.Array, buffer.Offset, buffer.Count));
+            }
+            else
+            {
+                var array = ArrayPool<byte>.Shared.Rent((int)value.Length);
+                try
+                {
+                    value.Position = 0;
+                    // MemoryStream.Read is backed by an in-memory byte array and always returns the full requested byte count in a single call,
+                    // so there is no need to handle partial reads in a loop.
+                    // See: https://github.com/dotnet/runtime/blob/960dca4391a731a20b25a92cdb500ef737bfcbbd/src/libraries/System.Private.CoreLib/src/System/IO/MemoryStream.cs#L320
+                    value.Read(array, 0, (int)value.Length);
+                    writer.WriteBase64StringValue(new ReadOnlySpan<byte>(array, 0, (int)value.Length));
                 }
                 finally
                 {
--- a/generator/ServiceClientGeneratorLib/Generators/Marshallers/JsonRPCStructureMarshaller.cs
+++ b/generator/ServiceClientGeneratorLib/Generators/Marshallers/JsonRPCStructureMarshaller.cs
@@ -18,7 +18,7 @@ namespace ServiceClientGenerator.Generators.Marshallers
     /// Class to produce the template output
     /// </summary>
     
-    #line 1 "C:\codebase\v4\aws-sdk-net-v4\generator\ServiceClientGeneratorLib\Generators\Marshallers\JsonRPCStructureMarshaller.tt"
+    #line 1 "C:\repos\aws-sdk-net-v4\generator\ServiceClientGeneratorLib\Generators\Marshallers\JsonRPCStructureMarshaller.tt"
     [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.VisualStudio.TextTemplating", "17.0.0.0")]
     public partial class JsonRPCStructureMarshaller : BaseRequestMarshaller
     {
@@ -29,7 +29,7 @@ namespace ServiceClientGenerator.Generators.Marshallers
         public override string TransformText()
         {
             
-            #line 6 "C:\codebase\v4\aws-sdk-net-v4\generator\ServiceClientGeneratorLib\Generators\Marshallers\JsonRPCStructureMarshaller.tt"
+            #line 6 "C:\repos\aws-sdk-net-v4\generator\ServiceClientGeneratorLib\Generators\Marshallers\JsonRPCStructureMarshaller.tt"
 
 	AddLicenseHeader();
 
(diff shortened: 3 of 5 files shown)


sdk/src/Core/Amazon.Runtime/Internal/Util/StringUtils.cs
```diff
@@ -69,6 +70,35 @@ public static string FromMemoryStream(MemoryStream value)
...
+            {
+                writer.WriteBase64StringValue(new ReadOnlySpan<byte>(buffer.Array, buffer.Offset, buffer.Count));
+            }
+            else
+            {
+                var array = ArrayPool<byte>.Shared.Rent((int)value.Length);
+                try
+                {
+                    value.Position = 0;
+                    value.Read(array, 0, (int)value.Length);
```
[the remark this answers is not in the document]
muhammad-othman: Documented the side effect.

sdk/src/Core/Amazon.Runtime/Internal/Util/StringUtils.cs
```diff
@@ -69,6 +70,35 @@ public static string FromMemoryStream(MemoryStream value)
...
+                    writer.WriteBase64StringValue(new ReadOnlySpan<byte>(array, 0, (int)value.Length));
+                }
+                finally
+                {
+                    ArrayPool<byte>.Shared.Return(array);
+                }
+            }
+        }
```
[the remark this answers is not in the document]
muhammad-othman: Added unit tests to cover these cases except the last one, as we don't except to get a memory stream with a changed position otherwise the original `FromMemoryStream` would have failed, I just added `value.Position = 0;` to be extra protective

generator/.DevConfigs/5ba72673-5d59-4b0a-9206-6237d869c0f0.json
```diff
@@ -0,0 +1,9 @@
+{
+  "core": {
+    "updateMinimum": true,
+    "type": "patch",
+    "changeLogMessages": [
+      "Optimize MemoryStream Base64 serialization in JSON marshallers"
```
peterrsongg: nit, could do "Fix #1922" as part of the changelog message, that would link the issue too

sdk/src/Core/Amazon.Runtime/Internal/Util/StringUtils.cs
```diff
@@ -69,6 +70,43 @@ public static string FromMemoryStream(MemoryStream value)
...
+                writer.WriteBase64StringValue(new ReadOnlySpan<byte>(buffer.Array, buffer.Offset, buffer.Count));
+            }
+            else
+            {
+                var array = ArrayPool<byte>.Shared.Rent((int)value.Length);
+                try
+                {
+                    value.Position = 0;
+                    value.Read(array, 0, (int)value.Length);
+                    writer.WriteBase64StringValue(new ReadOnlySpan<byte>(array, 0, (int)value.Length));
```
peterrsongg, line 104: I think there are two issues here.

1. Shouldn't we use the actual `bytesRead` here instead of `(int)value.Length`? The `Read` operation can read less than you requested. It's actually quite common for very large memory stream reads. 

2. Also, we should keep reading until the bytesRead is zero, or else we may do an incomplete read of the stream and only partially write

peterrsongg: suggested code in the `try` block
```
        try
        {
            value.Position = 0;
            int totalBytesRead = 0;
            int bytesToRead = (int)value.Length;
            
            while (totalBytesRead < bytesToRead)
            {
                int bytesRead = value.Read(array, totalBytesRead, bytesToRead - totalBytesRead);
                if (bytesRead == 0)
                    throw new IOException($"Stream ended prematurely: expected {bytesToRead} bytes, read {totalBytesRead}");
                totalBytesRead += bytesRead;
            }
            
            writer.WriteBase64StringValue(new ReadOnlySpan<byte>(array, 0, totalBytesRead));
        }
```

peterrsongg (changes requested):

sdk/src/Core/Amazon.Runtime/Internal/Util/StringUtils.cs
muhammad-othman: Good point about `Stream.Read` potentially returning fewer bytes than requested. 

However the original `FromMemoryStream` method uses the same single-read pattern `value.Read(array, 0, (int)value.Length)` followed by `Convert.ToBase64String(array, 0, (int)value.Length)`, and we didn't face partial reads so far.
The reason for non partial reads so far is that in practice, `MemoryStream.Read` always returns the full requested bytes. Looking at the code https://github.com/dotnet/runtime/blob/960dca4391a731a20b25a92cdb500ef737bfcbbd/src/libraries/System.Private.CoreLib/src/System/IO/MemoryStream.cs#L320, the `Read` does a single `Buffer.BlockCopy` at the end, it never produces partial reads since it's purely an in-memory operation backed by a byte array.

The `Stream.Read` contract does allow partial reads, and the official docs state: `An implementation is free to return fewer bytes than requested even if the end of the stream has not been reached.` So while the current `MemoryStream` implementation won't exhibit this behavior, relying on it is technically relying on an implementation detail rather than the documented contract.

But since both methods (`FromMemoryStream` and `WriteBase64StringValue`) accept `MemoryStream` (not `Stream`), and the parameter type constrains this to the known implementation, I think the current approach is safe.

peterrsongg: Okay, taking a look at the source code for MemoryStream makes me feel better, but i am still a bit concerned b/c [the docs](https://learn.microsoft.com/en-us/dotnet/api/system.io.memorystream.read?view=net-10.0) state that if the data isn't fully available then it wouldn't do that. Also, good point about it being that way originally, but if we can make it more resilient i don't see a harm in that. @normj what do you think?

peterrsongg (approved):

GarrettBeatty (approved):

normj (approved):

Issue #4110, opened 2025-11-07
NuGet packages are not deterministic


martincostello:
### Describe the bug

Viewing [AWSSDK.Core 4.0.2 in NuGet Package Explorer](https://nuget.info/packages/AWSSDK.Core/4.0.2) shows various warnings about the build of the package with respect to Source Link and determinism:

<img width="450" height="190" alt="Image" src="https://github.com/user-attachments/assets/818ddb0a-28e7-409f-8b29-4cefea7f75e2" />

The NuGet packages should be built with `ContinuousIntegrationBuild=true`, `Deterministic=true` and `EmbedUntrackedSources=true` to resolve this.

.NET global tools such as dotnet-validate and meziantou.framework.nugetpackagevalidation.tool can be used to verify this.

For Amazon.Lambda.Core at least, this doesn't appear to be a regression and these have never been enabled.=true` to resolve this.

.NET global tools such as dotnet-validate and meziantou.framework.nugetpackagevalidation.tool can be used to verify this.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

Packages should be deterministic and have no untracked sources.

### Current Behavior

NuGet packages are not deterministic and have untracked sources.

### Reproduction Steps

View NuGet packages in [NuGet Package Explorer](https://nuget.info/packages/AWSSDK.Core/4.0.2).

### Possible Solution

Set `ContinuousIntegrationBuild=true`, `Deterministic=true` and `EmbedUntrackedSources=true=true` MSBuild properties when building the NuGet packages for publishing to NuGet.org.

### AWS .NET SDK and/or Package version used

I checked these two, but I imagine all the packages produced from this repository have the same issue:

AWSSDK.Core 4.0.2
AWSSDK.Lambda 4.0.6.1

### Targeted .NET Platform

Any

### Operating System and version

Any

dscpinheiro:
Yeah, this happens for all packages in the SDK today. 

This is something we should do, but never got prioritized (similar feature requests from a few years ago... https://github.com/aws/aws-sdk-net/issues/1921 and https://github.com/aws/aws-sdk-net/issues/1777).

martincostello:
This should be pretty trivial to enable.

For example, this is what I do for my own libraries:

```xml
<PropertyGroup>
  <ContinuousIntegrationBuild>$([MSBuild]::ValueOrDefault('$(CI)', 'false'))</ContinuousIntegrationBuild>
[...]

PR #4337, opened 2026-02-21
Enable deterministic builds and embed untracked sources in NuGet packages

slang25:
## Description

Enable `Deterministic` and `EmbedUntrackedSources` MSBuild properties across all projects that produce NuGet packages.

- **`Deterministic`**: Ensures the compiler produces byte-identical output for the same inputs, enabling reproducible builds.
- **`EmbedUntrackedSources`**: Embeds source files not tracked by source control (e.g. auto-generated code) into the PDB. This complements the existing SourceLink support so that consumers can debug into generated sources that can't be fetched from GitHub.

Properties are added to `sdk/Directory.Build.props`, `extensions/Directory.Build.props`, and the 5 per-extension `Directory.Build.props` files that shadow the parent.

## Motivation and Context

The NuGet packages published by this repository already include SourceLink, allowing sources to be mapped back to their GitHub repo source files. However:

1. Untracked/generated sources are not embedded in the PDB, so stepping into generated code during debugging fails.
2. The `Deterministic` flag is not explicitly set, which means builds may not be reproducible across machines.

image

## Testing

These are declarative MSBuild property changes. They can be verified by:

1. Building any SDK or extension project and inspecting the PDB with `dotnet tool run sourcelink print-json` to confirm SourceLink metadata.
2. Checking the PDB with a tool like `MetadataTools` to confirm untracked sources are embedded.
3. Building twice and comparing output binaries to verify determinism.

No breaking changes. These properties affect compiler and PDB output only — they do not change the runtime behavior or public API surface of any assembly.

## Types of changes

## Checklist

## License

### DIFF
--- a/buildtools/build.proj
+++ b/buildtools/build.proj
@@ -419,53 +419,45 @@
     <Target Name="package-extensions" DependsOnTargets="package-netcore-extensions;package-crt-extensions;package-cloudfront-extensions;package-ec2-extensions;package-meai-extensions;package-cborprotocol-extensions;package-logging-extensions" />
 
     <Target Name="build-extensions" >
-        <!-- depends: restore-nuget;run-generator -->
         <Exec Command="dotnet restore -f  AWSSDK.Extensions.sln" WorkingDirectory="../extensions"/>
         <Exec Command="dotnet clean AWSSDK.Extensions.sln" WorkingDirectory="../extensions"/>
         <Exec Command="dotnet build -c $(Configuration) AWSSDK.Extensions.sln /p:AWSKeyFile=$(CustomSnkFileLocation);SignAssembly=true;RunAnalyzersDuringBuild=$(RunAnalyzersDuringBuild);RuleSetFileForBuild=$(RuleSetFileForBuild)"  WorkingDirectory="../extensions"/>
-
-        <Message Text="Build and Test .NET Core Setup Extension"/>
-        <Exec Command="dotnet test -c $(Configuration) --no-build NETCore.SetupTests.csproj" WorkingDirectory="../extensions/test/NETCore.SetupTests"/>
-
-        <Message Text="Build and Test Crt Integration Extension"/>
-        <Exec Command="dotnet test -c $(Configuration) --no-build CrtIntegrationTests.csproj" WorkingDirectory="../extensions/test/CrtIntegrationTests"/>
     </Target>
 
     <Target Name="package-netcore-extensions" DependsOnTargets="$(package-extensions-DependsOnTargets)">
         <MakeDir Directories="../Deployment/nuget" />
-        <Exec Command="..\..\..\sdk\.nuget\NuGet.exe pack ./AWSSDK.Extensions.NETCore.Setup.nuspec -BasePath . -OutputDirectory ../../../Deployment/nuget" WorkingDirectory="../extensions/src/AWSSDK.Extensions.NETCore.Setup" />
+        <Exec Command="..\..\..\sdk\.nuget\NuGet.exe pack ./AWSSDK.Extensions.NETCore.Setup.nuspec -BasePath . -Symbols -SymbolPackageFormat snupkg -OutputDirectory ../../../Deployment/nuget" WorkingDirectory="../extensions/src/AWSSDK.Extensions.NETCore.Setup" />
     </Target>
 
     <Target Name="package-crt-extensions" DependsOnTargets="$(package-extensions-DependsOnTargets)">
         <MakeDir Directories="../Deployment/nuget" />
-        <Exec Command="..\..\..\sdk\.nuget\NuGet.exe pack ./AWSSDK.Extensions.CrtIntegration.nuspec -BasePath . -OutputDirectory ../../../Deployment/nuget" WorkingDirectory="../extensions/src/AWSSDK.Extensions.CrtIntegration" />
+        <Exec Command="..\..\..\sdk\.nuget\NuGet.exe pack ./AWSSDK.Extensions.CrtIntegration.nuspec -BasePath . -Symbols -SymbolPackageFormat snupkg -OutputDirectory ../../../Deployment/nuget" WorkingDirectory="../extensions/src/AWSSDK.Extensions.CrtIntegration" />
     </Target>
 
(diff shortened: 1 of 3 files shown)


dscpinheiro:
This is somewhat related to https://github.com/aws/aws-sdk-net/issues/4110, right? I did give it a try when that issue was created, but even after applying Martin's suggestions `dotnet-validate` still complained after I rebuilt the SDK in our build system...

I only got those warnings to go away after I manually changed the SDK not to use the `.nuspec` file (moving the properties to the `csproj` files) and used `dotnet pack` (instead of what we do today, which is using https://github.com/aws/aws-sdk-net/tree/main/sdk/.nuget). So I think we need to make some other changes on our end before this is resolved.

slang25:
@dscpinheiro Yes it's basically #4110 

I won't go into too much detail about what I'm trying to do, but I writing some meta-programing tools, and it's helpful if packages are "complete" and contain links to source, or embed untracked source files.

You'll see what's missing if you go here:
https://nuget.info/packages/AWSSDK.Core/4.0.3.15

As I understand, this should still work with the nuspec approach, because it's about the symbols/dlls and you are already getting the sourcelink document embedded there, it's just that the symbols don't include the untracked files, and the paths are deterministic (`/_/` etc...).

So it should be a case of tweaking the build step. Are the actual release build scripts part of the repo or are these private?

dscpinheiro:
Yeah, most of our release infrastructure is private, but at a very high level we:
- Run the generator (`MSBuild .\buildtools\build.proj /t:run-generator`)
- Build the solutions with all services (`cd .\sdk\; dotnet build .\AWSSDK.NetFramework.sln /p:SignAssembly=true`)
- Create the NuGet packages (`cd buildtools; ./create-nuget-packages.ps1`)

Maybe I was missing the symbols then (because I did have `IncludeSymbols` in my `.csproj` but not at first when I was trying with the nuspec approach).

dscpinheiro:
Just an update here: I updated our build script to create symbols (`NuGet.exe pack $nuspec -verbosity detailed -Symbols -SymbolPackageFormat snupkg`), ran your PR through our build system (with the `CI` environment variable set), and `dotnet-validate` still complained 🫠:
```
PS> dotnet validate package local .\AWSSDK.ARCZonalShift.4.0.2.3.nupkg
Validating C:\Users\dspin\Downloads\dotnet4\nuget\AWSSDK.ARCZonalShift.4.0.2.3.nupkg
• Source Link: ⚠️ Contains untracked sources (obj)
    Contains untracked sources:
To Fix:
    <EmbedUntrackedSources>true</EmbedUntrackedSources>
    Also, use 3.1.300 SDK to build or
workaround in: https://github.com/dotnet/sourcelink/issues/572
    Assembly: lib\net472\AWSSDK.ARCZonalShift.dll
    Assembly: lib\net8.0\AWSSDK.ARCZonalShift.dll
    Assembly: lib\netcoreapp3.1\AWSSDK.ARCZonalShift.dll
    Assembly: lib\netstandard2.0\AWSSDK.ARCZonalShift.dll

• Deterministic (dll/exe): ⚠️ Contains untracked sources (obj)

• Compiler Flags: ✅ Valid
```

^ This was odd since the `EmbedUntrackedSources` is there and our build system uses an SDK newer than `3.1.300` (the workaround on that issue did not help either). Unless I'm missing something obvious I think we'll need to migrate away from the `nuspec` and use `dotnet pack` (which does require more changes unfortunately).

dscpinheiro:
Never mind my previous comment, I _was_ missing something obvious: My local build system uses CodeCommit for one of the origins so that's why SourceLink was not working properly. 

I just pushed a new commit with some changes (https://github.com/aws/aws-sdk-net/pull/4337/commits/6993b42208fb3bf36d716b6eda9a8c4571229d70), but running the build commands now on the repo when the remote is actually GitHub does work!
```
PS> dotnet validate package local .\AWSSDK.AccessAnalyzer.4.0.4.12.nupkg
Validating C:\Projects\aws-sdk-net-v4\Deployment\nuget\AWSSDK.AccessAnalyzer.4.0.4.12.nupkg
• Source Link: ✅ Valid with Symbol Server

• Deterministic (dll/exe): ✅ Valid

• Compiler Flags: ✅ Valid

PS> dotnet validate package local .\AWSSDK.Extensions.Bedrock.MEAI.4.0.5.7.nupkg
Validating C:\Projects\aws-sdk-net-v4\Deployment\nuget\AWSSDK.Extensions.Bedrock.MEAI.4.0.5.7.nupkg
• Source Link: ✅ Valid with Symbol Server

• Deterministic (dll/exe): ✅ Valid

• Compiler Flags: ✅ Valid
```

I'll chase the approvals so we can finally get this resolved (for our team reviewers, dry-run was `DRY_RUN-7c80816e-20f4-45d2-86cb-878da9f3e3ff`).

buildtools/build.proj
```diff
@@ -419,53 +419,45 @@
     <Target Name="package-extensions" DependsOnTargets="package-netcore-extensions;package-crt-extensions;package-cloudfront-extensions;package-ec2-extensions;package-meai-extensions;package-cborprotocol-extensions;package-logging-extensions" />
 
     <Target Name="build-extensions" >
-        <!-- depends: restore-nuget;run-generator -->
         <Exec Command="dotnet restore -f  AWSSDK.Extensions.sln" WorkingDirectory="../extensions"/>
         <Exec Command="dotnet clean AWSSDK.Extensions.sln" WorkingDirectory="../extensions"/>
         <Exec Command="dotnet build -c $(Configuration) AWSSDK.Extensions.sln /p:AWSKeyFile=$(CustomSnkFileLocation);SignAssembly=true;RunAnalyzersDuringBuild=$(RunAnalyzersDuringBuild);RuleSetFileForBuild=$(RuleSetFileForBuild)"  WorkingDirectory="../extensions"/>
-
-        <Message Text="Build and Test .NET Core Setup Extension"/>
```
dscpinheiro, line 427: I removed this since our build system tests the extensions separately. (And it was missing all the new V4 extensions too)

normj (approved):

dscpinheiro:
This was included in today's release: https://github.com/aws/aws-sdk-net/releases/tag/4.0.206.0

Thanks again for the contribution! (and bearing with me while I was figuring out the details on our end)

PR #4353, opened 2026-03-13
fix: Render <important> and <note> XML doc tags in generated API reference

AlexDaines:
## Description

Add handling for `<important>` and `<note>` XML doc tags in the doc generator's HTML transformation pipeline. These tags appear across 2,583+ occurrences in the SDK's XML documentation corpus but were previously passed through as unrecognized elements. They now render as styled `<div class="noteblock">` blocks with an appropriate header (`Important:` or `Note:`).

## Motivation and Context

Service teams use `<important>` and `<note>` tags in their API model documentation to highlight critical information. The doc generator was silently passing these through as raw `<important>`/`<note>` HTML elements, which browsers ignore as unknown tags — the content rendered as unstyled inline text with no visual distinction.

### Dry-run - Powershell5
- **Dry-run (build) ID:** DRY_RUN-74b028b4-df80-443f-8f11-5c997bcf9c45
- **Status:**
  - [ ] Pending
  - [x] Completed successfully
  - [ ] Failed
- **Failed bypass reason:**

### Dry-run - Dotnet4
- **Dry-run (build) ID:** DRY_RUN-7db66e91-4d87-448b-bfa5-382f29243673
- **Status:**
  - [ ] Pending
  - [x] Completed successfully
  - [ ] Failed
- **Failed bypass reason:**

5 unit tests added in `NDocUtilitiesHtmlTests.cs`, all using exact output assertions (`Assert.Equal`):

- `ImportantTag_RendersAsNoteblock` — standard `<important>` with inner `<para>`
- `NoteTag_RendersAsNoteblock` — standard `<note>` with inner `<para>`
- `ParaTag_StillRendersAsParagraph` — regression test for existing element mapping
- `SelfClosingImportantTag_ProducesEmptyNoteblock` — self-closing `<important />` edge case
- `ImportantTagWithInnerMarkup_PreservesNestedElements` — `<c>` tags nested inside callout, matching real corpus patterns

All 10 tests in `SDKDocGenerator.UnitTests` pass.

image

## Types of changes

## Checklist

## License

### DIFF
--- a/docgenerator/SDKDocGeneratorLib/NDocUtilities.cs
+++ b/docgenerator/SDKDocGeneratorLib/NDocUtilities.cs
@@ -41,14 +41,15 @@ namespace SDKDocGenerator
             { "summary", "p" },
             { "para", "p" },
             { "see", "a" },
-            { "paramref", "code" }
+            { "paramref", "code" },
+            { "important", "div" },
+            { "note", "div" }
         };
 
-        // Callout elements that render as styled blocks with a header
-        private static readonly Dictionary<string, string> CalloutElementMapping = new Dictionary<string, string>(StringComparer.Ordinal)
+        private static readonly Dictionary<string, string> NdocToHtmlClassMapping = new Dictionary<string, string>(StringComparer.Ordinal)
         {
-            { "important", "Important:" },
-            { "note", "Note:" }
+            { "important", "noteblock noteblock-warning" },
+            { "note", "noteblock" }
         };
 
         #region manage ndoc instances
@@ -615,17 +616,8 @@ namespace SDKDocGenerator
 
                                 // element name substitution, if necessary
                                 string elementName;
-                                string calloutHeader;
-                                bool isCallout = CalloutElementMapping.TryGetValue(originalLocalName, out calloutHeader);
-
-                                if (isCallout)
-                                {
-                                    elementName = "div";
-                                }
-                                else if (!NdocToHtmlElementMapping.TryGetValue(originalLocalName, out elementName))
-                                {
+                                if (!NdocToHtmlElementMapping.TryGetValue(originalLocalName, out elementName))
                                     elementName = originalLocalName;
-                                }
 
                                 // some elements can't be empty, use this variable for that
                                 string emptyElementContents = null;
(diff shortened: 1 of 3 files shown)


dscpinheiro:
Screenshots on this PR would be helpful to understand what the pages will look like (you can get the artifacts from the doc updater step).

AlexDaines:
> Screenshots on this PR would be helpful to understand what the pages will look like (you can get the artifacts from the doc updater step).

Good idea. Added an example screenshot

muhammad-othman:
I thought we will use `<strong>` tag or something to convey the importance of these paragraphs, adding `Important:` seems overly verbose.

peterrsongg:
I do kind of agree with Muhammad's general comment that the <important> tag should wrap the text in bold using <strong> or something like that. Don't feel too strongly about keeping or removing "Important" but i think the text should be bolded at least

AlexDaines:
> I thought we will use <strong> tag or something to convey the importance of these paragraphs, adding Important: seems overly verbose.

> I do kind of agree with Muhammad's general comment that the tag should wrap the text in bold using ** or something like that. Don't feel too strongly about keeping or removing "Important" but i think the text should be bolded at least**

I agree with this assessment. I went ahead and revised to use css changes to emphasize <important> and <note> sections and updated screenshots with the new implementation.

docgenerator/SDKDocGeneratorLib/NDocUtilities.cs
```diff
@@ -604,18 +612,26 @@ private static string DocBlobToHTML(XElement rootNode, AbstractTypeProvider type
...
+                                    elementName = originalLocalName;
 
                                 // some elements can't be empty, use this variable for that
                                 string emptyElementContents = null;
 
                                 // start element
                                 writer.WriteStartElement(elementName);
 
+                                // Add CSS class if the original element has a class mapping
+                                string cssClass;
```
muhammad-othman, line 629: nit: not need to define the class beforehand as it isn't used anywhere outside the if condition, you can do this instead

```
                                if (NdocToHtmlClassMapping.TryGetValue(originalLocalName, out var cssClass))
                                {
                                    writer.WriteAttributeString("class", cssClass);
                                }
```

docgenerator/SDKDocGeneratorLib/NDocUtilities.cs
```diff
@@ -41,7 +41,15 @@ public static class NDocUtilities
             { "summary", "p" },
             { "para", "p" },
             { "see", "a" },
-            { "paramref", "code" }
+            { "paramref", "code" },
+            { "important", "div" },
```
muhammad-othman, line 45: I would prefer if we use another element for `important` other than `div` for better accessibility.

docgenerator/SDKDocGeneratorLib/NDocUtilities.cs
```diff
@@ -41,7 +41,15 @@ public static class NDocUtilities
...
+            { "note", "div" }
```
muhammad-othman, line 46: maybe we can even use `small` for `note` and if we did so we wont need `NdocToHtmlClassMapping` as `important`&`note` can be identified with html tags and no need for css classes.

muhammad-othman (approved):
Added few other comments but they are suggestions and not blocking so I'm approving the PR anyway.

PR #4359, opened 2026-03-17
Added input sanitization for deprecatedMessage values from service models before they are emitted into generated C# code.

GarrettBeatty:
## Description

Added input sanitization for `deprecatedMessage` values from service models before they are emitted into generated C# code. The `deprecatedMessage` field was being used raw in `[Obsolete("...")]` attributes, which could produce broken generated code if the message contained newlines (`\r`, `\n`), double quotes (`"`), or HTML tags (`<p>`, `</p>`, `<code>`, `</code>`).

A new `SanitizeDeprecationMessage()` extension method was added to the `StringExtensions` class in `GeneratorHelpers.cs` and is now called from the `DeprecationMessage` property getters in `Member.cs`, `Operation.cs`, and `Shape.cs`.

## Motivation and Context

Other fields in the generator (e.g., documentation strings) are sanitized via `Replace("\n", string.Empty)` and similar transformations before being emitted into generated code. However, `deprecatedMessage` was not receiving any sanitization. Since deprecation messages are placed directly into `[Obsolete("...")]` string literals in generated C# code, unsanitized content such as newlines or quotes could break compilation of the generated SDK code.

## Testing

- Added 3 new unit tests in `ServiceClientGeneratorTests/UnitTests.cs` that verify sanitization of deprecation messages for operations, shapes, and members.
- Created a test model (`TestModelDeprecated.json`) containing deprecated items with newlines, `\r`, HTML tags (`<p>`, `</p>`, `<code>`, `</code>`), and double quotes in their `deprecatedMessage` values.
- All 3 new tests pass, and all existing tests continue to pass.

### Dry-run
- **Dry-run ID:** b87882a7-07cc-4fb6-9f71-88c01f4fdb2b
- **Status:**
  - [x] Pending
  - [ ] Completed successfully
  - [ ] Failed
- **Failed bypass reason:**

No breaking changes. This is a strictly additive sanitization step applied to generated code output. The sanitization only removes/replaces characters that would have caused issues in generated code (`\r`, `\n`, `"`, HTML tags). Existing valid deprecation messages that don't contain these characters are unaffected.

N/A

## Types of changes

## Checklist

## License

### DIFF
--- a/generator/ServiceClientGeneratorTests/Content/TestModelDeprecated.json
+++ b/generator/ServiceClientGeneratorTests/Content/TestModelDeprecated.json
@@ -1,0 +1,43 @@
+{
+  "version": "1.0",
+  "metadata": {
+    "apiVersion": "2020-6-25",
+    "serviceAbbreviation": "TestService",
+    "serviceFullName": "TestService",
+    "signatureVersion": "v4",
+    "serviceId": "TestService"
+  },
+  "operations": {
+    "DeprecatedOperation": {
+      "name": "DeprecatedOperation",
+      "input": { "shape": "DeprecatedOperationInput" },
+      "output": { "shape": "DeprecatedOperationOutput" },
+      "documentation": "a deprecated operation",
+      "deprecated": true,
+      "deprecatedMessage": "This operation is deprecated.\nUse <code>NewOperation</code> instead.\r\nSee \"docs\" for <p>details</p>."
+    }
+  },
+  "shapes": {
+    "DeprecatedOperationInput": {
+      "type": "structure",
+      "members": {
+        "OldField": {
+          "shape": "OldFieldShape",
+          "deprecated": true,
+          "deprecatedMessage": "This field is deprecated.\nUse <code>NewField</code> instead.\r\nSee \"docs\" for <p>details</p>."
+        }
+      }
+    },
+    "DeprecatedOperationOutput": {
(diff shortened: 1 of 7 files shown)


generator/ServiceClientGeneratorTests/Content/TestModelDeprecated.json
```diff
@@ -0,0 +1,43 @@
+{
+  "version": "1.0",
```
peterrsongg, line 2: I wonder if this approach using a json file can be simplified by just testing that the method works as intended by testing it with a string. Is there a reason a json file was chosen like this?

dscpinheiro: Right now the only way to verify the generator works is by running against the real models (these unit tests don't run anywhere in our build system, and I think they were broken for quite some time).

So for now I'd say let's just included the `generator/ServiceClientGeneratorLib/` changes in the PR. Also, I assume we'd need to do the same for V3?

GarrettBeatty: i did run locally by manually updating one of the service json files, i just havent updated the PR description yet. and yeah i will also need to do same for v3

dscpinheiro:
@GarrettBeatty I'm going to close this PR, we can re-open if the issue shows up again.

PR #4375, opened 2026-03-31
chore: add additional text to CONTRIBUTING.md

kellertk:
## License


CONTRIBUTING.md
```diff
@@ -213,6 +213,25 @@ If you're contributing from outside AWS, you have several options for adding Dev
...
+## Automated Tools
+
+The use of AI tooling for assisted development work is accepted and encouraged
+in this repository, but due to the volume of submissions we ask that you observe
+the following rules:
+
+- All issue and pull request submissions to this repository that are sourced by
+  AI must first be reviewed by a human before submitting to the repository.
+  Items reviewed in this way must include a statement like "generated by AI
+  tools, and reviewed by <person>"
```
dscpinheiro: When I look at the rich diff, this `<person>` doesn't render:

Image

Also, side question: Would it make sense to put this at the top? I'm thinking people may not scroll down to notice this.

CONTRIBUTING.md
```diff
@@ -213,6 +213,25 @@ If you're contributing from outside AWS, you have several options for adding Dev
...
 
+## Automated Tools
+
+The use of AI tooling for assisted development work is accepted and encouraged
+in this repository, but due to the volume of submissions we ask that you observe
+the following rules:
+
+- All issue and pull request submissions to this repository that are sourced by
+  AI must first be reviewed by a human before submitting to the repository.
+  Items reviewed in this way must include a statement like "generated by AI
```
AlexDaines, line 224: PR is good as it stands, though I'm curious if you have given consideration to more formalized AI disclosure requirements? Having data on specifically which models were used for each contribution could be _interesting_ at the very least, if not useful- especially as model choices become more and more stratified.

     See: https://github.com/joaomlourenco/aidisclose

kellertk: This needs further discussion internally outside of the scope of this change

CONTRIBUTING.md
kellertk: Fixed the person thing. You're the only one that noticed that :)
As for placement, that's up to you if you want this to look differently. Every team has their own spin on the Contributing guide

dscpinheiro (approved):

AlexDaines (approved):

PR #4379, opened 2026-04-01
Fix Credential resolution storm caused by high parallelization

peterrsongg:
## Description

Fixed concurrent credential resolution failure in `DefaultAWSCredentialsIdentityResolver` by replacing the `ReaderWriterLockSlim` with a `SemaphoreSlim(1,1)` single-flight pattern. The key changes in `DefaultAWSCredentialsIdentityResolver.cs`:

1. **Replaced `ReaderWriterLockSlim` with `SemaphoreSlim(1,1)`** — `SemaphoreSlim` supports both synchronous `Wait()` and asynchronous `WaitAsync()`, preventing thread pool starvation when 64+ async tasks contend simultaneously.

2. **Made `_cachedCredentials` volatile** (this is key) — enables a lock-free fast path where concurrent callers can read cached credentials without acquiring the semaphore. This allows the _cachedCredentials to be updated by any thread and then when read by all other threads it forces the new cached credentials to be read.

3. **Added truly async `ResolveIdentityAsync`/`InternalGetCredentialsAsync`** — the previous implementation wrapped the sync method with `Task.FromResult`, causing all concurrent async callers to block thread pool threads. The new async path uses `SemaphoreSlim.WaitAsync()` so waiting tasks yield their threads back to the pool.

4. **Extracted `ResolveCredentialChain()` and `TryGetCachedCredentials()`** — shared logic between sync and async paths with a double-check pattern: first check cache without lock, then acquire semaphore and re-check before walking the chain.

5. **Added `ConcurrentCredentialResolutionTests.cs`** with 3 tests: sync 64-concurrency (verifies single chain walk), async 64-concurrency (verifies async path), and failure-then-retry (verifies failures aren't cached). All 3 new tests plus all 5 existing `DefaultCredentialsResolverTests` pass.
## Motivation and Context

**Root cause:** The credential resolution in `DefaultAWSCredentialsIdentityResolver` (`sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs`) is invoked per-request in the pipeline via `BaseAuthResolverHandler.PreInvokeAsync()` (`sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/BaseAuthResolverHandler.cs`). When many concurrent requests are made before credentials have been cached, each request independently walks the credential chain and attempts to call `DefaultInstanceProfileAWSCredentials.Instance.GetCredentials()` (`sdk/src/Core/Amazon.Runtime/Credentials/DefaultInstanceProfileAWSCredentials.cs`).

While `DefaultInstanceProfileAWSCredentials` uses a `SemaphoreSlim(1,1)` to serialize IMDS access, the issue is that `DefaultAWSCredentialsIdentityResolver.InternalGetCredentials()` (line ~216) wraps the IMDS call in a try/catch and converts failures to `InvalidOperationException("Failed to connect to EC2 instance metadata to retrieve credentials")`. Under high concurrency, the IMDS endpoint (169.254.169.254) becomes overwhelmed with simultaneous HTTP requests from the credential chain probing (environment variables → web identity → profile → IMDS), causing timeouts that propagate as credential resolution failures.
## Testing

Added Unit Tests
### Dry-runs

- **DotNet Dry-run ID:** 2797b819-c62a-457a-9abe-113ec4cbf2bd
  - [ ] Pending
  - [x] Completed successfully
  - [ ] Failed
- **PowerShell Dry-run ID:**8ed357ee-c4af-492a-98ab-08aefec2f1d0
  - [x] Pending
  - [ ] Completed successfully
  - [ ] Failed

NEW REVISION

- **DotNet Dry-run ID:** 12f8fb02-e7a3-4c2f-8cef-95121759f82
  - [ ] Pending
  - [x] Completed successfully
  - [ ] Failed
- **PowerShell Dry-run ID:**f0310498-2077-4630-8e4c-a0ce2807a16f
  - [] Pending
  - [x] Completed successfully
  - [ ] Failed

DefaultAWSCredentialsIdentityResolver.cs implements IDisposable now, because it utilizes a semaphoreslim, which must be disposed of. This is an acceptable breaking change because most users are just using the static `GetCredentials / GetCredentialsAsync`

    * What functionality was changed? Now only one thread will walk the credential chain while the other threads wait. Once credentials are resolved, the other threads can read the cached credentials. This will eliminate the possibility of overwhelming IMDS on the initial call with 64+ threads.
    * How will this impact customers? Allows for higher concurrency and less threadpool starvation in multi-threaded environments.
    * 

## Types of changes

## Checklist

## License

### DIFF
--- a/generator/.DevConfigs/60e32337-c0b9-457f-8b6c-b4e018ce73d0.json
+++ b/generator/.DevConfigs/60e32337-c0b9-457f-8b6c-b4e018ce73d0.json
@@ -3,7 +3,8 @@
     "updateMinimum": true,
     "type": "patch",
     "changeLogMessages": [
-      "Fix credential resolution issue under high concurrency where multiple threads simultaneously walked the credential chain causing initial service calls to fail."
+      "Fix credential resolution issue under high concurrency where multiple threads simultaneously walked the credential chain causing initial service calls to fail.",
+      "[Breaking Change] DefaultAWSCredentialsIdentityResolver implements IDisposable now. Most users will be unaffected since the static GetCredentials() or GetCredentialsAsync() methods are being called."
     ]
   }
 }
--- a/generator/.DevConfigs/60e32337-c0b9-457f-8b6c-b4e018ce73d0.json
+++ b/generator/.DevConfigs/60e32337-c0b9-457f-8b6c-b4e018ce73d0.json
@@ -1,0 +1,9 @@
+{
+  "core": {
+    "updateMinimum": true,
+    "type": "patch",
+    "changeLogMessages": [
+      "Fix credential resolution issue under high concurrency where multiple threads simultaneously walked the credential chain causing initial service calls to fail."
+    ]
+  }
+}
--- a/sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
+++ b/sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
@@ -390,9 +390,9 @@ namespace Amazon.Runtime.Credentials
             if (_disposed) return;
             if (disposing)
             {
-                _disposed = true;
                 _credentialResolutionLock.Dispose();
             }
+            _disposed = true;
         }
 
         /// <summary>
--- a/sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
+++ b/sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
@@ -32,31 +32,20 @@ namespace Amazon.Runtime.Credentials
     /// The default search order used is described in the <a href="https://docs.aws.amazon.com/sdk-for-net/v4/developer-guide/creds-assign.html">
     /// developer guide</a>, but it can be overwritten by setting the <see cref="AWSConfigs.AWSCredentialsGenerators"/> property.
     /// </summary>
-    [System.Diagnostics.CodeAnalysis.SuppressMessage("Usage", "CA1001:Types that own disposable fields should be disposable",
-        Justification = "This resolver is designed to be long-lived (singleton via DefaultIdentityResolverConfiguration). Disposing the semaphore would break concurrent callers.")]
-    public class DefaultAWSCredentialsIdentityResolver : IIdentityResolver<AWSCredentials>
+    public class DefaultAWSCredentialsIdentityResolver : IIdentityResolver<AWSCredentials>, IDisposable
     {
         private const string AWS_PROFILE_ENVIRONMENT_VARIABLE = "AWS_PROFILE";
         private const string DEFAULT_PROFILE_NAME = "default";
 
-        /// <summary>
-        /// A method that should either return valid <see cref="AWSCredentials"/> or throw an exception (so
-        /// that the SDK can move on and attempt the next generator in the credential chain).
-        /// </summary>
         public delegate AWSCredentials CredentialsGenerator();
 
-        /// <summary>
-        /// SemaphoreSlim supports both synchronous Wait() and asynchronous WaitAsync(), preventing
-        /// thread pool starvation when many concurrent async requests need to resolve credentials
-        /// simultaneously. Only one caller walks the credential chain; all others wait and reuse
-        /// the cached result.
-        /// </summary>
         private readonly SemaphoreSlim _credentialResolutionLock = new SemaphoreSlim(1, 1);
         private volatile AWSCredentials _cachedCredentials;
         private readonly List<CredentialsGenerator> _credentialsGenerators;
         private readonly CredentialProfileStoreChain _credentialProfileChain = new();
         private readonly EnvironmentState _lastKnownEnvironmentState = new();
         private static readonly Lazy<DefaultAWSCredentialsIdentityResolver> _defaultInstance = new();
+        private bool _disposed;
 
         public DefaultAWSCredentialsIdentityResolver()
         {
@@ -387,6 +376,22 @@ namespace Amazon.Runtime.Credentials
                 throw new InvalidOperationException(string.Format(CultureInfo.InvariantCulture,
                     "Failed to connect to EC2 instance metadata to retrieve credentials: {0}.",
                     e.Message), e);
+            }
+        }
+
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if (_disposed) return;
+            if (disposing)
+            {
+                _disposed = true;
+                _credentialResolutionLock.Dispose();
             }
         }
 
--- a/sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
+++ b/sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
@@ -218,11 +218,11 @@ namespace Amazon.Runtime.Credentials
 
             if (resolvedCredentials != null)
             {
-                // Update the environment snapshot before publishing the credentials via the
-                // volatile write. This avoids a race where a concurrent reader in
-                // TryGetCachedCredentials sees non-null _cachedCredentials but stale
-                // environment state, causing a false "environment changed" detection and
-                // an unnecessary re-resolution attempt.
+                // Clear the cache first so that any concurrent lock-free reader in
+                // TryGetCachedCredentials sees a cache miss and waits for the lock
+                // rather than returning stale credentials while the environment
+                // snapshot is being updated.
+                _cachedCredentials = null;
                 _lastKnownEnvironmentState.UpdateEnvironment();
                 _cachedCredentials = resolvedCredentials;
                 return resolvedCredentials;
--- a/sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
+++ b/sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
@@ -32,6 +32,8 @@ namespace Amazon.Runtime.Credentials
     /// The default search order used is described in the <a href="https://docs.aws.amazon.com/sdk-for-net/v4/developer-guide/creds-assign.html">
     /// developer guide</a>, but it can be overwritten by setting the <see cref="AWSConfigs.AWSCredentialsGenerators"/> property.
     /// </summary>
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Usage", "CA1001:Types that own disposable fields should be disposable",
+        Justification = "This resolver is designed to be long-lived (singleton via DefaultIdentityResolverConfiguration). Disposing the semaphore would break concurrent callers.")]
     public class DefaultAWSCredentialsIdentityResolver : IIdentityResolver<AWSCredentials>
     {
         private const string AWS_PROFILE_ENVIRONMENT_VARIABLE = "AWS_PROFILE";
@@ -43,8 +45,14 @@ namespace Amazon.Runtime.Credentials
         /// </summary>
         public delegate AWSCredentials CredentialsGenerator();
 
-        private static readonly ReaderWriterLockSlim _cachedCredentialsLock = new();
-        private AWSCredentials _cachedCredentials;
+        /// <summary>
+        /// SemaphoreSlim supports both synchronous Wait() and asynchronous WaitAsync(), preventing
+        /// thread pool starvation when many concurrent async requests need to resolve credentials
+        /// simultaneously. Only one caller walks the credential chain; all others wait and reuse
+        /// the cached result.
+        /// </summary>
+        private readonly SemaphoreSlim _credentialResolutionLock = new SemaphoreSlim(1, 1);
+        private volatile AWSCredentials _cachedCredentials;
         private readonly List<CredentialsGenerator> _credentialsGenerators;
         private readonly CredentialProfileStoreChain _credentialProfileChain = new();
         private readonly EnvironmentState _lastKnownEnvironmentState = new();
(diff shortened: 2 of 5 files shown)


sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
```diff
@@ -116,108 +124,183 @@ public AWSCredentials ResolveIdentity(IClientConfig clientConfig)
...
-                        break;
-                    }
+                    break;
                 }
+            }
+
+            if (resolvedCredentials != null)
+            {
+                _cachedCredentials = resolvedCredentials;
+                _lastKnownEnvironmentState.UpdateEnvironment();
```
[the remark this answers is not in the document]
peterrsongg: good suggestion, i'll aply it

sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
```diff
@@ -116,108 +124,183 @@ public AWSCredentials ResolveIdentity(IClientConfig clientConfig)
...
+            // Slow path: acquire the lock synchronously so only one thread walks the credential chain.
+            _credentialResolutionLock.Wait();
+            try
+            {
+                // Double-check: another thread may have resolved credentials while we waited.
+                if (_cachedCredentials != null && !hasEnvironmentChanged)
                 {
-                    _lastKnownEnvironmentState.UpdateEnvironment();
                     return _cachedCredentials;
                 }
```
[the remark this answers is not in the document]
peterrsongg: that is true, it could've changed. will fix

sdk/test/UnitTests/Custom/Runtime/Credentials/ConcurrentCredentialResolutionTests.cs
```diff
@@ -0,0 +1,223 @@
...
+            AWSConfigs.AWSCredentialsGenerators = new List<DefaultAWSCredentialsIdentityResolver.CredentialsGenerator>
+            {
+                () =>
+                {
+                    int count = Interlocked.Increment(ref generatorCallCount);
+                    if (shouldFail)
+                    {
+                        throw new InvalidOperationException("Simulated IMDS failure");
+                    }
+                    return new BasicAWSCredentials("test-access-key", "test-secret-key");
```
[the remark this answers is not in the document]
peterrsongg: good catch, will remove use of count

sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
```diff
@@ -116,108 +124,194 @@ public AWSCredentials ResolveIdentity(IClientConfig clientConfig)
...
+            }
+
+            if (resolvedCredentials != null)
+            {
+                // Update the environment snapshot before publishing the credentials via the
+                // volatile write. This avoids a race where a concurrent reader in
+                // TryGetCachedCredentials sees non-null _cachedCredentials but stale
+                // environment state, causing a false "environment changed" detection and
+                // an unnecessary re-resolution attempt.
```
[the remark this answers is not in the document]
peterrsongg: I disagree with this assesement, if new credentials have been resolved, it could be triggered from a background refresh of the credentials, the "stale" credentials you are referring to would still be valid since the pre-empty expiry window kicks in and the credential is never completely stale. Returning them for one more request while a refresh is in-flight is completely harmless. The credentials themselves haven't been revoked. 

Also adding an atomic update handler adds signifcant complexity for a scenario that would still lead a successful request.s 

Thirdly, the original logic did it this way too, `UpdateEnvironment` and the credential assignment weren't atomic from the reader's perspective either.

So, i will not make this change

sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
```diff
@@ -116,108 +124,194 @@ public AWSCredentials ResolveIdentity(IClientConfig clientConfig)
...
+                var source = new CredentialProfileStoreChain(profile.Location);
+                if (source.TryGetProfile(profile.Name, out CredentialProfile storedProfile))
+                {
+                    return storedProfile.GetAWSCredentials(source, true);
+                }
 
-        public Task<AWSCredentials> ResolveIdentityAsync(IClientConfig clientConfig, CancellationToken cancellationToken = default) =>
-            Task.FromResult(ResolveIdentity(clientConfig));
+                throw new AmazonClientException($"Unable to find the \"{profile.Name}\" profile specified in the client configuration.");
+            }
```
boblodgett: This looks like common logic from ResolveIdentity. So we don't have to update this in two places it could be put in  a method such as `private AWSCredentials TryResolveFromProfile(IClientConfig clientConfig)`

sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
```diff
@@ -116,108 +124,194 @@ public AWSCredentials ResolveIdentity(IClientConfig clientConfig)
...
+        /// </summary>
+        private AWSCredentials TryGetCachedCredentials(out bool hasEnvironmentChanged)
+        {
+            hasEnvironmentChanged = false;
+            var cached = _cachedCredentials;
+            if (cached != null)
             {
-                _cachedCredentialsLock.EnterReadLock();
-                if (_cachedCredentials != null)
+                hasEnvironmentChanged = _lastKnownEnvironmentState.HasEnvironmentChanged();
```
boblodgett: This doesn't look safe anymore because HasEnvironmentChanged can be being called at the same time as ResolveCredentialChain which also lost the locks.  I think AI also had a comment about this below.

sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
```diff
@@ -116,108 +124,194 @@ public AWSCredentials ResolveIdentity(IClientConfig clientConfig)
...
+            return await ResolveIdentityAsync(clientConfig, cancellationToken).ConfigureAwait(false);
+        }
 
-            try
+        /// <summary>
+        /// Checks whether valid cached credentials are available without acquiring the lock.
+        /// Returns the cached credentials if they exist and the environment hasn't changed,
+        /// otherwise returns null to indicate that credentials need to be resolved.
+        /// </summary>
+        private AWSCredentials TryGetCachedCredentials(out bool hasEnvironmentChanged)
```
boblodgett: what is the purpose of the out param when it isn't used anywhere? All everything does is use `out _` so why is it needed?

sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
```diff
@@ -116,108 +124,194 @@ public AWSCredentials ResolveIdentity(IClientConfig clientConfig)
         }
 
         Task<BaseIdentity> IIdentityResolver.ResolveIdentityAsync(IClientConfig clientConfig, CancellationToken cancellationToken) =>
-            Task.FromResult<BaseIdentity>(ResolveIdentity(clientConfig));
+            InternalResolveIdentityAsync(clientConfig, cancellationToken);
```
boblodgett: Why do we need the extra internal wrapper method? You should be able to do something like:

```
async Task<BaseIdentity> IIdentityResolver.ResolveIdentityAsync(IClientConfig clientConfig, CancellationToken cancellationToken)
    => await ResolveIdentityAsync(clientConfig, cancellationToken).ConfigureAwait(false);
```

sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
```diff
@@ -116,108 +124,194 @@ public AWSCredentials ResolveIdentity(IClientConfig clientConfig)
...
+            // Fast path: return cached credentials without acquiring the lock.
+            var cached = TryGetCachedCredentials(out _);
+            if (cached != null)
+            {
+                return cached;
+            }
 
-                if (_cachedCredentials != null)
+            // Slow path: acquire the lock synchronously so only one thread walks the credential chain.
+            _credentialResolutionLock.Wait();
```
boblodgett: In what instance could this wait forever because there is no timeouts or cancellation tokens supplied: https://learn.microsoft.com/en-us/dotnet/api/system.threading.semaphoreslim.wait?view=net-10.0

Maybe if IMDS times out or something hangs this will wait forever.

sdk/test/UnitTests/Custom/Runtime/Credentials/ConcurrentCredentialResolutionTests.cs
boblodgett, line 35: There should be a test that tests out the use of CancellationTokens with these code changes. Something like: this (Not tested):

```
[TestMethod]
public async Task AsyncResolution_CancellationWhileWaiting_ThrowsOperationCanceled()
{
    var blockGenerator = new ManualResetEventSlim(false);

    AWSConfigs.AWSCredentialsGenerators = new List<DefaultAWSCredentialsIdentityResolver.CredentialsGenerator>
    {
        () =>
        {
            blockGenerator.Wait(); // Hold the semaphore indefinitely
            return new BasicAWSCredentials("test-access-key", "test-secret-key");
        }
    };

    var resolver = new DefaultAWSCredentialsIdentityResolver();
    using var cts = new CancellationTokenSource();

    // First call grabs the semaphore and blocks inside the generator
    var holdingTask = Task.Run(() => resolver.ResolveIdentity(clientConfig: null));

    // Give it a moment to acquire the semaphore
    await Task.Delay(50).ConfigureAwait(false);

    // Second call waits on WaitAsync with our cancellation token
    var waitingTask = resolver.ResolveIdentityAsync(clientConfig: null, cancellationToken: cts.Token);

    // Cancel while waiting on the semaphore
    cts.Cancel();

    await Assert.ThrowsExceptionAsync<OperationCanceledException>(() => waitingTask).ConfigureAwait(false);

    // Cleanup: unblock the first call
    blockGenerator.Set();
    await holdingTask.ConfigureAwait(false);
}
```

sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
peterrsongg: yeah you're right, the internal wrapper isn't necessary

sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
peterrsongg: looks like it was intended to be used at some point but never was, i will remove it

sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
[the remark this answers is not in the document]
peterrsongg: On second glance, this assessment is valid, the fix to clear out the cachedCredentials is valid

sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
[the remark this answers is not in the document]
peterrsongg: This has been fixed. Since hasEnvironmentChanged was removed as an out parameter, TryGetCachedCredentials now recomputes HasEnvironmentChanged() inline every time it's called. The slow path calls TryGetCachedCredentials() again after acquiring the lock, so it always gets a fresh environment check. There's no pre-lock value being reused across the wait.

sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
peterrsongg: added the cancellation token as an overload as to not break binary compatibility

sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
peterrsongg: fixed this issue by clearing out the cache

sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
```diff
@@ -32,6 +32,8 @@ namespace Amazon.Runtime.Credentials
     /// The default search order used is described in the <a href="https://docs.aws.amazon.com/sdk-for-net/v4/developer-guide/creds-assign.html">
     /// developer guide</a>, but it can be overwritten by setting the <see cref="AWSConfigs.AWSCredentialsGenerators"/> property.
     /// </summary>
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Usage", "CA1001:Types that own disposable fields should be disposable",
```
normj: There is a default singleton but users could create their own instance of this class. You should implement the `IDisposable` interface and yeah our default singleton will never use it but if anybody decides to create a bunch of their own instances it can be disposed.

Calling out in the change log adding the `IDisposable` as a breaking change but point out most users are unaffected because they are calling the static `GetCredentials()` or `GetCredentialsAsync()`.

sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
```diff
@@ -98,126 +106,212 @@ public static Task<AWSCredentials> GetCredentialsAsync(IClientConfig clientConfi
...
+        /// </summary>
+        private static AWSCredentials TryGetProfileCredentials(IClientConfig clientConfig)
         {
-            var hasEnvironmentChanged = false;
+            var profile = clientConfig?.Profile;
+            if (string.IsNullOrEmpty(profile?.Name)) return null;
 
-            try
+            var source = new CredentialProfileStoreChain(profile.Location);
+            if (source.TryGetProfile(profile.Name, out CredentialProfile storedProfile))
```
normj, line 144: I think this is an issue with the existing code but looking at this it looks like we are rereading the credentials file every time we are resolving credentials if there is a profile set. I don't want to expand your PR too much so if I'm right and we are doing disk IO for every request can you add a backlog item to fix this up.

peterrsongg: yup, you're right. I'll create a backlog for this.

sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
peterrsongg: implemented IDispoable and called it out in the changelog

sdk/test/UnitTests/Custom/Runtime/Credentials/ConcurrentCredentialResolutionTests.cs
peterrsongg: I've added a suite of tests for cancellation token behavior

normj (dismissed):

sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
```diff
@@ -98,126 +95,212 @@ public static Task<AWSCredentials> GetCredentialsAsync(IClientConfig clientConfi
...
-        {
-            var profile = clientConfig?.Profile;
-            if (!string.IsNullOrEmpty(profile?.Name))
-            {
-                var source = new CredentialProfileStoreChain(profile.Location);
-                if (source.TryGetProfile(profile.Name, out CredentialProfile storedProfile))
-                {
-                    return storedProfile.GetAWSCredentials(source, true);
-                }
+            => ResolveIdentity(clientConfig, CancellationToken.None);
```
boblodgett: Why would we purposely use None to allow an operation that can hang instead of picking a default timeout ourselves?

generator/.DevConfigs/60e32337-c0b9-457f-8b6c-b4e018ce73d0.json
```diff
@@ -0,0 +1,10 @@
+{
+  "core": {
+    "updateMinimum": true,
+    "type": "patch",
+    "changeLogMessages": [
+      "Fix credential resolution issue under high concurrency where multiple threads simultaneously walked the credential chain causing initial service calls to fail.",
+      "[Breaking Change] DefaultAWSCredentialsIdentityResolver implements IDisposable now. Most users will be unaffected since the static GetCredentials() or GetCredentialsAsync() methods are being called."
```
boblodgett, line 7: Summary needs the breaking change assessment. Also if this is a real breaking change why isn't there n override for it because it wouldn't pass the build system checks if it was.

sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
```diff
@@ -296,6 +379,22 @@ private static AWSCredentials ContainerEC2CredentialsWrapper()
...
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if (_disposed) return;
+            if (disposing)
+            {
+                _disposed = true;
```
boblodgett: Although it doesn't hurt for the current code, this should be on the outside of the `if (disposing)` if check.   Not matching the pattern.

https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose#the-disposebool-method-overload

sdk/test/UnitTests/Custom/Runtime/Credentials/CancellationTokenCredentialResolutionTests.cs
boblodgett: These `new DefaultAWSCredentialsIdentityResolver ` should be in using blocks since IDisposable is implemented.

generator/.DevConfigs/60e32337-c0b9-457f-8b6c-b4e018ce73d0.json
peterrsongg: Adding `IDisposable` is a semantic / behavior breaking change, but not a binary or source breaking change. In some cases when `ref structs` are used, it can be binary / source but since that's not the case here, the backwards compatibility check didn't flag it. Consumers won't have to recompile because of this.

sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
peterrsongg: this is a good callout, we should pick a default that makes sense here.

boblodgett (approved):

normj (approved):

PR #4383, opened 2026-04-07
Document customization hooks added during s3 model gen

peterrsongg:
## Description

Documents all customization hooks added during s3 model generation, with examples.

## Motivation and Context

Allows other devs or ai to utilize these customization hooks.
## Testing

N/A 
### Dry-runs

- **DotNet Dry-run ID:**21e8d4c4-ac8c-4659-84ad-db1e489d76b4
  - [ ] Pending
  - [x] Completed successfully
  - [ ] Failed
- **PowerShell Dry-run ID:**324212e9-ed1d-4e54-91cb-90b3e40e3145
  - [ ] Pending
  - [x] Completed successfully
  - [ ] Failed

## Types of changes

## Checklist

## License

### DIFF
--- a/generator/ServiceClientGeneratorLib/Customizations.cs
+++ b/generator/ServiceClientGeneratorLib/Customizations.cs
@@ -1052,7 +1052,6 @@ namespace ServiceClientGenerator
             public const string DeprecatedMessageKey = "deprecatedMessage";
             public const string ExcludeFromMarshallingKey = "excludeFromMarshalling";
             public const string SkipXmlTestExpressionKey = "skipXmlTestExpression";
-            public const string NewObjectIfNullKey = "newObjectIfNull";
             public const string ShapeDocumentationKey = "shapeDocumentation";
             public const string ShapeModifierXmlNamespaceKey = "xmlNamespace";
             public const string OriginalMemberIsOutsideContainingShapeKey = "originalMemberIsOutsideContainingShape";
@@ -1066,7 +1065,6 @@ namespace ServiceClientGenerator
             private readonly Dictionary<string, JsonData> _injectedProperties;
             private readonly HashSet<string> _excludedMarshallingProperties;
             private readonly HashSet<string> _skipXmlTestExpressionProperties;
-            private readonly HashSet<string> _newObjectIfNullProperties;
             private readonly HashSet<string> _shapeDocumentation;
             private readonly string _shapeModifierXmlNamespace;
             private readonly Dictionary<string, JsonData> _predicateListUnmarshallers;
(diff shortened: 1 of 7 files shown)


generator/ServiceClientGeneratorLib/Customizations.cs
```diff
@@ -1052,7 +1052,6 @@ public class ShapeModifier
             public const string DeprecatedMessageKey = "deprecatedMessage";
             public const string ExcludeFromMarshallingKey = "excludeFromMarshalling";
             public const string SkipXmlTestExpressionKey = "skipXmlTestExpression";
-            public const string NewObjectIfNullKey = "newObjectIfNull";
```
peterrsongg, line 1055: this was only being used once and could be replaced with a different customization hook so i removed it.

generator/customization-hooks.md
```diff
@@ -0,0 +1,488 @@
+## Description
+This file describes customization hooks that can be used to generate custom code.
+
+### Name: excludeFromMarshalling 
+### Type: string[]
```
dscpinheiro: Nit, but could you use different header levels (i.e. `###` vs `####`) for the type. Then the rendered doc outline will not have everything at the same level: 

Image

generator/customization-hooks.md
```diff
@@ -0,0 +1,488 @@
...
+                }
+            }
+        ]
+    }
+}
+```
+
+### Name: skipSetter
+### Type: bool
+- Description: This customization hook is used to completely skip generation of the setter for a given member. If this is set to true, no setter will be generated.
```
dscpinheiro: I see this is only used for a couple of operations, is it also a _rare customization_? If yes, maybe it would be good to explain when it would be used.

^ By that I mean I see other entries with details such as _This is useful when the property gets its value from another property or custom logic rather than storing its own value in a private field._

generator/customization-hooks.md
```diff
@@ -0,0 +1,488 @@
...
+        "modify":[
+            {
+                "InputStream":{"additionalDocumentation":["<para>Input stream for the request; content for the request will be read from the stream.</para>"]}
+            }
+        ]
+    }
+}
+```
+
+### Name: unwrapXmlOutput
```
dscpinheiro, line 372: This is equivalent to Smithy's https://smithy.io/2.0/aws/customizations/s3-customizations.html#aws-customizations-s3unwrappedxmloutput-trait, right?

generator/customization-hooks.md
```diff
@@ -0,0 +1,488 @@
...
+    "mergedEnums": {
+        "MFADeleteStatus": {
+            "mergeInto": "VersionStatus"
+        }
+    }
+```
+
+### Name: skipChecksumDuringMarshalling
+### Type: bool
+- Description: Use this customization hook to skip setting the checksum during marshalling. If an operation has custom logic for setting checksums, this customization hook can be used. It can also be used to completely skip setting the checksum, although that should rarely be the case.
```
dscpinheiro: I'd call out this as rare too, it shouldn't be applied to new operations.

dscpinheiro (approved):

muhammad-othman (approved):

PR #4391, opened 2026-04-16
Implements the updated Retry Behavior behind a feature flag

muhammad-othman:
## Description
This PR implements the Retry Behavior 2.1 SEP specification updates for the AWS SDK for .NET, gated behind a feature flag (`AWS_NEW_RETRIES_2026` environment variable). When enabled, the SDK uses updated retry parameters including revised backoff timing, updated retry quota costs, service-specific configurations, long-polling backoff behavior, and `x-amz-retry-after` header support.

All changes are backward-compatible, the feature flag defaults to `false`, preserving existing behavior. When the flag is removed at end of 2026, the new behavior will become the default.

## Motivation and Context
`DOTNET-8570`

## Testing
- Added `RetryHandlerStandardMode21Tests.cs` with 18 test cases covering all SEP 2.1 standard mode scenarios.
- All existing retry tests (standard mode, adaptive mode, capacity manager) continue to pass.

### Dry-runs

- **DotNet Dry-run ID:** 1666b78d-9279-4431-9ead-6214171098c8
  - [ ] Pending
  - [x] Completed successfully
  - [ ] Failed
- **PowerShell Dry-run ID:** c559ee8d-4c35-49d6-a66e-43a1424d3bc6
  - [ ] Pending
  - [x] Completed successfully
  - [ ] Failed

### 1. Identify all breaking changes

This PR modifies several APIs in the `Amazon.Runtime.Internal` namespace. While technically public (due to C# accessibility), these are internal SDK implementation details not intended for external consumption.

#### Change 1: `CapacityManager` — Old constructors removed

* **What functionality was changed?** The 3-parameter `CapacityManager(throttleRetryCount, throttleRetryCost, throttleCost)` and 4-parameter `CapacityManager(..., timeoutRetryCost)` constructors were replaced with a single 5-parameter constructor `CapacityManager(initialRetryTokens, retryCost, noRetryIncrement, timeoutRetryCost, throttlingRetryCost)`.
* **How will this impact customers?** No impact. `CapacityManager` is in the `Amazon.Runtime.Internal` namespace. No service DLL or external code constructs `CapacityManager` directly, it is only instantiated by `StandardRetryPolicy` and `DefaultRetryPolicy` within the Core assembly.
* **Why does this need to be a breaking change?** The old constructors computed `initialRetryTokens` from `throttleRetryCount * throttleRetryCost`, which was confusing and hid the actual token count. The new constructor takes the total token count directly, matching the SEP 2.1 specification terminology.
* **Are best practices being followed?** Yes. The `Internal` namespace convention signals these are not part of the public API contract.
* **How have you tested this?** 
  - All 47 unit tests pass (18 SEP 2.1 + 13 standard + 12 adaptive + 4 capacity manager)
  - All service DLLs (EC2, S3, DynamoDB, STS) compile successfully against the new Core
  - **Binary compatibility test**: Built old S3/STS service DLLs from `main` branch, then ran them with the new Core DLL, real AWS calls (S3 ListBuckets, STS GetCallerIdentity) succeeded without any `MissingMethodException` or `TypeLoadException`

#### Change 2: `StandardRetryPolicy.CalculateRetryDelay` — Signature changed

* **What functionality was changed?** `CalculateRetryDelay(int retries, int maxBackoffInMilliseconds)` (protected static) was replaced with `CalculateRetryDelay(IExecutionContext, int maxBackoffInMilliseconds)` (internal static) to support service-aware backoff and x-amz-retry-after header.
* **How will this impact customers?** No impact. No service DLL or known external code calls this method. DynamoDB has its own private `CalculateRetryDelay(int)` with a different signature.
* **Why does this need to be a breaking change?** The new formula requires `IExecutionContext` to determine the service type (DynamoDB vs others), error type (throttling vs transient), and x-amz-retry-after header value.
* **How have you tested this?** Same as above, binary compatibility test confirmed no runtime errors.

#### Change 3: `StandardRetryPolicy.WaitBeforeRetry(int, int)` — Removed

* **What functionality was changed?** The public static helper `WaitBeforeRetry(int retries, int maxBackoffInMilliseconds)` was removed. The virtual `WaitBeforeRetry(IExecutionContext)` override remains unchanged.
* **How will this impact customers?** No impact. All service retry policies use the virtual `WaitBeforeRetry(IExecutionContext)` method, not the static helper.
* **How have you tested this?** Binary compatibility test with old service DLLs confirmed no `MissingMethodException`.

#### Change 4: DynamoDB/DynamoDB Streams `MaxErrorRetry` default

* **What functionality was changed?** The `MaxErrorRetry = 10` assignment was removed from the generated `AmazonDynamoDBConfig`/`AmazonDynamoDBStreamsConfig` constructors. The retry policy now sets the default internally: 3 retries when `UseNewRetries2026` is enabled, 10 retries when disabled.
* **How will this impact customers?** When `UseNewRetries2026` is disabled (default), DynamoDB still defaults to 10 retries, no behavioral change. When enabled, it uses 3 retries per SEP 2.1.
* **How have you tested this?** Unit test `DynamoDBBaseBackoffAndIncreasedRetries` verifies 4 max attempts (3 retries) under the new flag.

## Types of changes

## Checklist

## License

### DIFF
--- a/generator/.DevConfigs/94c889f3-af3a-4d8f-940a-0205846afe42.json
+++ b/generator/.DevConfigs/94c889f3-af3a-4d8f-940a-0205846afe42.json
@@ -1,0 +1,26 @@
+{
+  "core": {
+    "changeLogMessages": [
+      "Implements the updated Retry Behavior behind a feature flag (AWS_NEW_RETRIES_2026)."
+    ],
+    "type": "minor",
+    "updateMinimum": true,
+    "backwardIncompatibilitiesToIgnore": []
+  },
+  "services": [
+    {
+      "serviceName": "DynamoDBv2",
+      "type": "patch",
+      "changeLogMessages": [
+        "Remove max-retries from DynamoDB service and depend on values applied in the `AWSSDK.Core` package."
+      ]
+    },
+    {
+      "serviceName": "DynamoDBStreams",
+      "type": "patch",
+      "changeLogMessages": [
+        "Remove max-retries from DynamoDB Streams service and depend on values applied in the `AWSSDK.Core` package."
+      ]
+    }
+  ]
+}
--- a/sdk/src/Core/Amazon.Runtime/CapacityManager.cs
+++ b/sdk/src/Core/Amazon.Runtime/CapacityManager.cs
@@ -41,7 +41,12 @@ namespace Amazon.Runtime.Internal
             /// <summary>
             /// The timeout capacity type uses the timeout capacity amount.
             /// </summary>
-            Timeout
+            Timeout,
+            /// <summary>
+            /// The throttling capacity type uses the throttling retry cost amount.
+            /// Used when the new retry behavior (SEP 2.1) is enabled.
+            /// </summary>
+            Throttling
         }
 
 
@@ -63,17 +68,46 @@ namespace Amazon.Runtime.Internal
             }
         }
 
-        public CapacityManager(int throttleRetryCount, int throttleRetryCost, int throttleCost) 
-            : this(throttleRetryCount, throttleRetryCost, throttleCost, throttleRetryCost)
-        {    
+        /// <summary>
+        /// Constructor for CapacityManager.
+        /// </summary>
+        /// <param name="throttleRetryCount">The initial and maximum number of retry tokens.</param>
+        /// <param name="throttleRetryCost">The cost of a retry.</param>
+        /// <param name="throttleCost">The capacity to add on a successful non-retry request.</param>
+        [Obsolete("Use the constructor with explicit parameter names (initialRetryTokens, retryCost, noRetryIncrement, timeoutRetryCost, throttlingRetryCost) instead.")]
+        public CapacityManager(int throttleRetryCount, int throttleRetryCost, int throttleCost)
+            : this(throttleRetryCount, throttleRetryCost, throttleCost, throttleRetryCost, 0)
+        {
         }
 
+        /// <summary>
+        /// Constructor for CapacityManager.
+        /// </summary>
+        /// <param name="throttleRetryCount">The initial and maximum number of retry tokens.</param>
+        /// <param name="throttleRetryCost">The cost of a retry.</param>
+        /// <param name="throttleCost">The capacity to add on a successful non-retry request.</param>
+        /// <param name="timeoutRetryCost">The cost of a timeout retry.</param>
+        [Obsolete("Use the constructor with explicit parameter names (initialRetryTokens, retryCost, noRetryIncrement, timeoutRetryCost, throttlingRetryCost) instead.")]
         public CapacityManager(int throttleRetryCount, int throttleRetryCost, int throttleCost, int timeoutRetryCost)
+            : this(throttleRetryCount, throttleRetryCost, throttleCost, timeoutRetryCost, 0)
         {
-            retryCost = throttleRetryCost;
-            initialRetryTokens = throttleRetryCount;
-            noRetryIncrement = throttleCost;
+        }
+
+        /// <summary>
+        /// Constructor for CapacityManager.
+        /// </summary>
+        /// <param name="initialRetryTokens">The initial and maximum number of retry tokens.</param>
+        /// <param name="retryCost">The cost of a non-throttling retry.</param>
+        /// <param name="noRetryIncrement">The capacity to add on a successful non-retry request.</param>
+        /// <param name="timeoutRetryCost">The cost of a timeout retry.</param>
+        /// <param name="throttlingRetryCost">The cost of a throttling retry (0 if not applicable).</param>
+        public CapacityManager(int initialRetryTokens, int retryCost, int noRetryIncrement, int timeoutRetryCost, int throttlingRetryCost)
+        {
+            this.retryCost = retryCost;
+            this.initialRetryTokens = initialRetryTokens;
+            this.noRetryIncrement = noRetryIncrement;
             this.timeoutRetryCost = timeoutRetryCost;
+            this.throttlingRetryCost = throttlingRetryCost;
         }
 
         /// <summary>
@@ -92,7 +126,19 @@ namespace Amazon.Runtime.Internal
         /// <param name="capacityType">Specifies what capacity type cost to use for obtaining capacity</param>
         public bool TryAcquireCapacity(RetryCapacity retryCapacity, CapacityType capacityType)
         {
-            var capacityCost = capacityType == CapacityType.Timeout ? timeoutRetryCost : retryCost;
+            int capacityCost;
+            switch (capacityType)
+            {
+                case CapacityType.Timeout:
+                    capacityCost = timeoutRetryCost;
+                    break;
+                case CapacityType.Throttling:
+                    capacityCost = throttlingRetryCost;
+                    break;
+                default:
+                    capacityCost = retryCost;
+                    break;
+            }
             if (capacityCost < 0)
             {
                 return false;
@@ -126,6 +172,9 @@ namespace Amazon.Runtime.Internal
                     break;
                 case CapacityType.Timeout:
                     ReleaseCapacity(timeoutRetryCost, retryCapacity);
+                    break;
+                case CapacityType.Throttling:
+                    ReleaseCapacity(throttlingRetryCost, retryCapacity);
                     break;
                 case CapacityType.Increment:
                     ReleaseCapacity(noRetryIncrement, retryCapacity);
@@ -163,6 +212,10 @@ namespace Amazon.Runtime.Internal
         // legacy retry modes and 10 for all other retry modes.
         private readonly int timeoutRetryCost;
 
+        // This parameter sets the cost of making a retry call when the error is a throttling error.
+        // Used when the new retry behavior (SEP 2.1) is enabled. The default value is 5.
+        private readonly int throttlingRetryCost;
+
         // Maximum capacity in a bucket set to 100 for legacy retry mode and 500 for all other retry modes.
         private readonly int initialRetryTokens;
 
@@ -199,7 +252,7 @@ namespace Amazon.Runtime.Internal
                     _rwlock.EnterWriteLock();
                     try
                     {
-                        retryCapacity = new RetryCapacity(retryCost * initialRetryTokens);
+                        retryCapacity = new RetryCapacity(initialRetryTokens);
                         _serviceUrlToCapacityMap.Add(serviceURL, retryCapacity);
                         return retryCapacity;
                     }
--- a/sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryHandler.cs
+++ b/sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryHandler.cs
@@ -121,6 +121,21 @@ namespace Amazon.Runtime.Internal
                     shouldRetry = this.RetryPolicy.Retry(executionContext, exception);
                     if (!shouldRetry)
                     {
+                        // SEP 2.1: Long-polling operations must always back off when the error is
+                        // retryable and retry quota is exhausted (but NOT when max attempts is reached).
+                        // The RetryLimitReached check distinguishes quota exhaustion from max attempts.
+                        if (IsLongPollingOperation(executionContext)
+                            && requestContext.IsLastExceptionRetryable
+                            && !this.RetryPolicy.RetryLimitReached(executionContext))
+                        {
+                            // Temporarily increment Retries so the backoff formula computes the
+                            // delay for the correct attempt index (i=1 for first failure), then
+                            // restore the original value so LogForError reports accurately.
+                            requestContext.Retries++;
+                            this.RetryPolicy.WaitBeforeRetry(executionContext);
+                            requestContext.Retries--;
+                        }
+
                         LogForError(requestContext, exception);
                         throw;
                     }
@@ -215,6 +230,21 @@ namespace Amazon.Runtime.Internal
                     shouldRetry = await this.RetryPolicy.RetryAsync(executionContext, capturedException.SourceException).ConfigureAwait(false);
                     if (!shouldRetry)
                     {
+                        // SEP 2.1: Long-polling operations must always back off when the error is
+                        // retryable and retry quota is exhausted (but NOT when max attempts is reached).
+                        // The RetryLimitReached check distinguishes quota exhaustion from max attempts.
+                        if (IsLongPollingOperation(executionContext)
+                            && requestContext.IsLastExceptionRetryable
+                            && !this.RetryPolicy.RetryLimitReached(executionContext))
+                        {
+                            // Temporarily increment Retries so the backoff formula computes the
+                            // delay for the correct attempt index (i=1 for first failure), then
+                            // restore the original value so LogForError reports accurately.
+                            requestContext.Retries++;
+                            await RetryPolicy.WaitBeforeRetryAsync(executionContext).ConfigureAwait(false);
+                            requestContext.Retries--;
+                        }
+
                         LogForError(requestContext, capturedException.SourceException);
                         capturedException.Throw();
                     }
@@ -346,6 +376,35 @@ namespace Amazon.Runtime.Internal
                 return true;
             }
             
+            return false;
+        }
+
+        /// <summary>
+        /// Determines if the current operation is a long-polling operation that should always back off
+        /// when retryable, even if retry quota is exhausted.
+        /// </summary>
+        private static bool IsLongPollingOperation(IExecutionContext executionContext)
+        {
+            if (!RetryPolicy.UseNewRetries2026) return false;
+
+            // TODO: Check longPoll trait from model when available in C2J models.
+            // Until the trait is available, use hard-coded service/operation combinations.
+            var serviceId = executionContext.RequestContext.ClientConfig?.ServiceId;
+            var operationName = AWSSDKUtils.ExtractOperationName(executionContext.RequestContext.RequestName);
+
+            if (string.Equals(serviceId, "SQS", StringComparison.OrdinalIgnoreCase) &&
+                string.Equals(operationName, "ReceiveMessage", StringComparison.OrdinalIgnoreCase))
+                return true;
+
+            if (string.Equals(serviceId, "SFN", StringComparison.OrdinalIgnoreCase) &&
+                string.Equals(operationName, "GetActivityTask", StringComparison.OrdinalIgnoreCase))
+                return true;
+
+            if (string.Equals(serviceId, "SWF", StringComparison.OrdinalIgnoreCase) &&
+                (string.Equals(operationName, "PollForActivityTask", StringComparison.OrdinalIgnoreCase) ||
+                 string.Equals(operationName, "PollForDecisionTask", StringComparison.OrdinalIgnoreCase)))
+                return true;
+
             return false;
         }
 
--- a/sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs
+++ b/sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs
@@ -35,6 +35,16 @@ namespace Amazon.Runtime
     public abstract partial class RetryPolicy
     {
         /// <summary>
+        /// Temporary feature flag for updated retry behavior improvements including
+        /// revised backoff timing, updated retry quota costs, and other enhancements.
+        /// Enabled by setting the AWS_NEW_RETRIES_2026 environment variable to "true".
+        /// Defaults to false. This flag will be removed at end of 2026 when the new
+        /// behavior becomes the default.
+        /// </summary>
+        internal static bool UseNewRetries2026 { get; set; } =
+            string.Equals(Environment.GetEnvironmentVariable("AWS_NEW_RETRIES_2026"), "true", StringComparison.OrdinalIgnoreCase);
+
+        /// <summary>
         /// Maximum number of retries to be performed.
         /// This does not count the initial request.
         /// </summary>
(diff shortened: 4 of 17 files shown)


sdk/src/Core/Amazon.Runtime/CapacityManager.cs
```diff
@@ -63,17 +68,21 @@ protected virtual void Dispose(bool disposing)
...
+        /// <param name="timeoutRetryCost">The cost of a timeout retry.</param>
+        /// <param name="throttlingRetryCost">The cost of a throttling retry (0 if not applicable).</param>
+        public CapacityManager(int initialRetryTokens, int retryCost, int noRetryIncrement, int timeoutRetryCost, int throttlingRetryCost)
         {
-            retryCost = throttleRetryCost;
-            initialRetryTokens = throttleRetryCount;
-            noRetryIncrement = throttleCost;
+            this.retryCost = retryCost;
+            this.initialRetryTokens = initialRetryTokens;
+            this.noRetryIncrement = noRetryIncrement;
```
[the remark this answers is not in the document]
muhammad-othman: Details can be found in the breaking change assessment

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/StandardRetryPolicy.cs
```diff
@@ -217,34 +232,86 @@ public override bool RetryLimitReached(IExecutionContext executionContext)
...
-        {
-            AWSSDKUtils.Sleep(CalculateRetryDelay(retries, maxBackoffInMilliseconds));
-        }        
-
-        protected static int CalculateRetryDelay(int retries, int maxBackoffInMilliseconds)
+        /// <param name="executionContext">The execution context for the current request.</param>
+        /// <param name="maxBackoffInMilliseconds">The maximum backoff in milliseconds.</param>
+        /// <returns>The delay in milliseconds before the next retry.</returns>
+        internal static int CalculateRetryDelay(IExecutionContext executionContext, int maxBackoffInMilliseconds)
         {
```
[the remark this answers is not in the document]
muhammad-othman: Details can be found in the breaking change assessment

sdk/src/Services/DynamoDBv2/Generated/AmazonDynamoDBConfig.cs
```diff
@@ -57,7 +57,6 @@ public AmazonDynamoDBConfig()
         {
             base.ServiceId = "DynamoDB";
             this.AuthenticationServiceName = "dynamodb";
```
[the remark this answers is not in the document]
muhammad-othman: Details can be found in the breaking change assessment

sdk/src/Services/DynamoDBStreams/Generated/AmazonDynamoDBStreamsConfig.cs
```diff
@@ -57,7 +57,6 @@ public AmazonDynamoDBStreamsConfig()
         {
             base.ServiceId = "DynamoDB Streams";
```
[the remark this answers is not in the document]
muhammad-othman: Details can be found in the breaking change assessment

generator/.DevConfigs/94c889f3-af3a-4d8f-940a-0205846afe42.json
```diff
@@ -0,0 +1,10 @@
+
+{
+  "core": {
+    "changeLogMessages": [
+      "Implements the updated Retry Behavior behind a feature flag (AWS_NEW_RETRIES_2026)."
+    ],
+    "type": "patch",
+    "updateMinimum": true
```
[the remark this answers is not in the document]
muhammad-othman: updated.

sdk/test/UnitTests/Custom/Runtime/RetryHandlerStandardMode21Tests.cs
```diff
@@ -0,0 +1,983 @@
+using Amazon;
+using Amazon.Runtime;
+using Amazon.Runtime.Internal;
+using Amazon.Runtime.Internal.Auth;
+using Amazon.Runtime.Internal.Transform;
```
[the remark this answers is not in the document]
muhammad-othman: updated.

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryHandler.cs
```diff
@@ -303,7 +329,7 @@ private void LogForError(IRequestContext requestContext, Exception exception)
                           exception.GetType().Name,
                           requestContext.RequestName,
                           requestContext.Request.Endpoint.ToString(),
-                          requestContext.Retries + 1);
+                          requestContext.Retries);
```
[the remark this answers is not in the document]
muhammad-othman: fixed

sdk/test/UnitTests/Custom/Runtime/RetryHandlerStandardMode21Tests.cs
```diff
@@ -0,0 +1,983 @@
...
+    }
+
+    public class Mock21StandardRetryPolicy : StandardRetryPolicy
+    {
+        private static CapacityManager _originalCapacityManager;
+        private static Random _originalRandomJitter;
+
+        private static readonly FieldInfo RandomJitterField = typeof(StandardRetryPolicy)
+            .GetField("_randomJitter", BindingFlags.NonPublic | BindingFlags.Static);
+
```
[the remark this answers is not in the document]
muhammad-othman: updated.

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs
```diff
@@ -146,8 +156,17 @@ public bool Retry(IExecutionContext executionContext, Exception exception)
...
+                        executionContext.RequestContext.LastCapacityType = IsThrottlingError(exception) ?
+                            CapacityManager.CapacityType.Throttling : CapacityManager.CapacityType.Retry;
+                        StoreRetryAfterHeader(executionContext, exception);
+                    }
+                    else
+                    {
+                        executionContext.RequestContext.LastCapacityType = IsServiceTimeoutError(exception) ?
+                            CapacityManager.CapacityType.Timeout : CapacityManager.CapacityType.Retry;
+                    }
                     return OnRetry(executionContext, isClockSkewError,  IsThrottlingError(exception));
```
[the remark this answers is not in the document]
muhammad-othman: This is by design for SEP 2.1, the spec only has `RETRY_COST` and `THROTTLING_RETRY_COST`, no timeout category, will keep open for others to review.

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/_async/RetryPolicy.cs
```diff
@@ -50,6 +50,12 @@ public async Task<bool> RetryAsync(IExecutionContext executionContext, Exception
                     {
                         return false;
                     }
+                    if (UseNewRetries2026)
+                    {
+                        executionContext.RequestContext.LastCapacityType = IsThrottlingError(exception) ?
+                            CapacityManager.CapacityType.Throttling : CapacityManager.CapacityType.Retry;
+                        StoreRetryAfterHeader(executionContext, exception);
```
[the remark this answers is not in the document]
muhammad-othman: The missing `else` block in the async `RetryAsync()` is a pre-existing bug, not part of SEP 2.1. The sync `Retry()` method already had the `LastCapacityType = IsServiceTimeoutError ? Timeout : Retry` assignment, but the async path was missing it entirely. As This is a pre-existing gap I think it should be addressed separately to keep this PR without any changes for customers that don't enable `UseNewRetries2026`, will keep this comment open for others to review.

generator/.DevConfigs/94c889f3-af3a-4d8f-940a-0205846afe42.json
```diff
@@ -0,0 +1,22 @@
+{
+  "core": {
+    "changeLogMessages": [
+      "Implements the updated Retry Behavior behind a feature flag (AWS_NEW_RETRIES_2026)."
```
dscpinheiro, line 4: I know it's probably not available yet, but would be nice to link to the SDK guide (https://docs.aws.amazon.com/sdkref/latest/guide/feature-retry-behavior.html). Looking at this entry how would I know what the "updated retry behavior" is?

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryHandler.cs
```diff
@@ -121,6 +121,19 @@ public override void InvokeSync(IExecutionContext executionContext)
...
+                        // SEP 2.1: Long-polling operations must always back off when the error is
+                        // retryable, even if retry quota is exhausted.
+                        if (IsLongPollingOperation(executionContext)
+                            && requestContext.IsLastExceptionRetryable)
+                        {
+                            // Temporarily increment Retries so the backoff formula computes the
+                            // delay for the correct attempt index (i=1 for first failure), then
+                            // restore the original value so LogForError reports accurately.
+                            requestContext.Retries++;
+                            this.RetryPolicy.WaitBeforeRetry(executionContext);
```
dscpinheiro, line 135: Can `this.RetryPolicy.WaitBeforeRetry` fail? If yes, should this be in a `try / finally` block to make sure the context retries are correct for `LogForError`? (same comment applies to the async method)

sdk/test/UnitTests/Custom/Runtime/RetryHandlerStandardMode21Tests.cs
dscpinheiro, line 1: With the new spec tests, could we get rid of https://github.com/aws/aws-sdk-net/blob/main/sdk/test/Services/S3/IntegrationTests/CapacityManagerTests.cs? This test is really annoying because it takes ~1 minute to run.

sdk/src/Core/Amazon.Runtime/CapacityManager.cs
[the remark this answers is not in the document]
dscpinheiro: I get the reasoning but this class is not in the `Internal` namespace... I don't think customers would be using these constructors but perhaps the `Obsolete` makes sense here.

sdk/test/UnitTests/Custom/Runtime/RetryHandlerStandardMode21Tests.cs
```diff
@@ -0,0 +1,982 @@
...
+    /// Used to eliminate jitter in tests.
+    /// </summary>
+    internal class FixedRandom : Random
+    {
+        private readonly double _value;
+        public FixedRandom(double value) { _value = value; }
+        public override double NextDouble() => _value;
+    }
+
+    public class Mock21StandardRetryPolicy : StandardRetryPolicy
```
dscpinheiro, line 1057: This PR also changes the adaptive retry policy, right? Are there no tests for it in the SEP?

(Also this is a personal nit but I really dislike the reflection to set the jitter - since you're changing `Core` anyway wouldn't it be possible to at least set the field to internal?)

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/_async/RetryPolicy.cs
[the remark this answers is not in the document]
dscpinheiro: Agree it makes sense to fix this on a separate PR.

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs
```diff
@@ -34,6 +34,16 @@ namespace Amazon.Runtime
...
     public abstract partial class RetryPolicy
     {
+        /// <summary>
+        /// Temporary feature flag for updated retry behavior improvements including
+        /// revised backoff timing, updated retry quota costs, and other enhancements.
+        /// Enabled by setting the AWS_NEW_RETRIES_2026 environment variable to "true".
+        /// Defaults to false. This flag will be removed at end of 2026 when the new
+        /// behavior becomes the default.
+        /// </summary>
+        internal static bool UseNewRetries2026 { get; set; } =
```
dscpinheiro, line 44: I don't see it in this PR (or the SEP...), but would it make sense to update the user agent to include whether the new retry option was set? (Otherwise how would we validate customers are using it?)

generator/.DevConfigs/94c889f3-af3a-4d8f-940a-0205846afe42.json
```diff
@@ -0,0 +1,22 @@
...
+      "Amazon.Runtime.Internal.CapacityManager/MethodRemoved",
+      "Amazon.Runtime.Internal.StandardRetryPolicy/MethodRemoved"
+    ]
+  },
+  "services": [
+    {
+      "serviceName": "DynamoDBv2",
+      "type": "patch",
+      "changeLogMessages": [
+        "Remove max-retries from DynamoDB service and depend on values applied in the CORE package."
```
dscpinheiro:
```suggestion
        "Remove max-retries from DynamoDB service and depend on values applied in the `AWSSDK.Core` package."
```

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryHandler.cs
```diff
@@ -215,6 +228,19 @@ public override async System.Threading.Tasks.Task<T> InvokeAsync<T>(IExecutionCo
                     shouldRetry = await this.RetryPolicy.RetryAsync(executionContext, capturedException.SourceException).ConfigureAwait(false);
                     if (!shouldRetry)
                     {
+                        // SEP 2.1: Long-polling operations must always back off when the error is
+                        // retryable, even if retry quota is exhausted.
+                        if (IsLongPollingOperation(executionContext)
```
dscpinheiro, line 236: Kiro found this which seems worth checking:
```
Long-polling backoff incorrectly triggers on max-attempts-exceeded path (RetryHandler.cs, both sync and async)                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                     
  The long-polling guard checks !shouldRetry && IsLongPollingOperation && IsLastExceptionRetryable, but shouldRetry is false for both quota-exhausted and max-attempts-exceeded cases. The spec pseudocode only sleeps on the quota-exhausted path:                                                                  
                                                                                                                                                                                                                                                                                                                     
  if not HasRetryQuota(response)                                                                                                                                                                                                                                                                                     
      if IsLongPollingOperation():                                                                                                                                                                                                                                                                                   
          sleep(ExponentialBackoff(attempts - 1))                                                                                                                                                                                                                                                                    
      return response                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                     
  When max attempts is reached, the spec returns immediately with no sleep. But the current code adds an unnecessary backoff delay before throwing. The fix needs to distinguish why shouldRetry is false — quota exhaustion vs. max attempts vs. non-retryable error.
```

sdk/src/Core/Amazon.Runtime/CapacityManager.cs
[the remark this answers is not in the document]
muhammad-othman: But if customers aren't using it, why do we need to keep an additional obsolete constructor when we can remove it?

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryHandler.cs
muhammad-othman: I don't think so, it is just calling `AWSSDKUtils.Sleep` which has this line `System.Threading.Thread.Sleep(ms);`.

sdk/test/UnitTests/Custom/Runtime/RetryHandlerStandardMode21Tests.cs
muhammad-othman: It isn't really related, the numbers might change but if we needed this test before we still need it now, unless we didn't need it before when we should have removed it anyway.

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs
muhammad-othman: This needs to be added to the feature ids list first, maybe they were planning to do so when removing the feature flag, but let me check and get back to you.

sdk/test/UnitTests/Custom/Runtime/RetryHandlerStandardMode21Tests.cs
muhammad-othman: The changes are only applied to the `CapacityManager` and the `StandardRetryPolicy`, Adaptive retry is affect by these changes as it uses both, so if we added adaptive retry tests we will be testing the same cases.

sdk/src/Core/Amazon.Runtime/CapacityManager.cs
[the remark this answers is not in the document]
dscpinheiro: I didn't say customers aren't using it, just that it's unlikely. 

We've definitely seen other libraries take dependencies on components of the SDK that were supposed to be internal, and breaking them is not a good customer experience at all.

sdk/test/UnitTests/Custom/Runtime/RetryHandlerStandardMode21Tests.cs
muhammad-othman: For using the reflection to set the jitter, I think about it the other way, that we shouldn't change the code just for testing, and for this one keeping it private will make the code easier to reason about and the reader will know that the jitter wont change outside this class, but I don't feel very strongly about it.

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryHandler.cs
muhammad-othman: Didn't consider this one, will need to fix it.

generator/.DevConfigs/94c889f3-af3a-4d8f-940a-0205846afe42.json
muhammad-othman: I don't think the changes will be available if we are planning to release it this week or the week after that.

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryHandler.cs
muhammad-othman: Updated to account for `max-attempts-exceeded` when `IsLongPollingOperation` is true.

sdk/test/UnitTests/Custom/Runtime/RetryHandlerStandardMode21Tests.cs
```diff
@@ -0,0 +1,1140 @@
...
+                {
+                    RuntimePipeline.InvokeSync(executionContext);
+                });
+
+                Assert.AreEqual("Mocked throttling", exception.Message);
+                Assert.AreEqual(0, executionContext.RequestContext.Retries);
+                Assert.AreEqual(1, Tester.CallCount);
+
+                // Should still back off even with empty quota — uses 1000ms throttling base
+                retryPolicy.AssertDelaysMatch(new int[] { 1000 });
```
dscpinheiro, line 689: The test in the SEP for this scenario (`Long-Polling Backoff After Throttling Error When Token Bucket Empty`) has `delay: 0.05` but you have 1000 here. Is that intentional?

muhammad-othman: Reached out to update the SEP, `1000ms` should be the correct delay.

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryHandler.cs
```diff
@@ -349,6 +379,35 @@ private bool ShouldRetryForStaleConnection(
...
 
+        /// <summary>
+        /// Determines if the current operation is a long-polling operation that should always back off
+        /// when retryable, even if retry quota is exhausted.
+        /// </summary>
+        private static bool IsLongPollingOperation(IExecutionContext executionContext)
+        {
+            if (!RetryPolicy.UseNewRetries2026) return false;
+
+            // TODO: Check longPoll trait from model when available in C2J models.
```
boblodgett, line 390: TODO needs to be resolved but a better method than leaving a TODO would be to create a backlog task that we will checkup on. We won't remember to come check TODOs in the future.

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/_async/RetryPolicy.cs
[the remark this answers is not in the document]
boblodgett: I am ok with a different backlog item on this/new PR.

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryPolicy.cs
boblodgett: Did you get an answer for this?

sdk/src/Core/Amazon.Runtime/CapacityManager.cs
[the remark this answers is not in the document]
boblodgett: The issue is we do not know in all cases how customers are using the SDK. Despite being in internal my default thinking is to avoid the possibility of breaking someone for an unknown reason by simply overloading the constructor and mark the other one as obsolete. You also remove the requirement for a senior+ engineer on the PR.

Unless of course you can prove nobody is using those constructors. Basically we should avoid using the breaking changes override unless necessary. If this is a safe change it should be added to the tool so it doesn't flag it.

sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/RetryHandler.cs
muhammad-othman: We should do this when we are removing the `UseNewRetries2026` flag which we will have to remember.

boblodgett: Ok, fair point. This is ok then.

boblodgett (approved):

dscpinheiro (approved):

Issue #3468, opened 2024-09-08
v4 HttpPipeline ReadOnlyMemory


danielmarbach:
### Describe the feature

As of today, the HTTP pipeline supports either Content as byte[] or a stream when the ContentStream property is set. I would appreciate it if the http pipeline could handle `ReadOnlyMemory<byte>` to support better options for buffering. 

Technically you could support ROM already by wrapping the ROM in a ReadOnlyStream that uses the ROM behind the scenes but that would require the Content property to be deprecated and phased out. If the Content property is switched to ROM it probably would also require quite a few changes to the SDK to properly adjust the generator to handle it as well as introducing necessary overloads on the methods that are used to pass in the byte array content. All in all that is too much work to handle as a community contribution and that is why I wanted to open up the discussion whether the team has already ideas how and when they will tackle it. 

Another option would be to introduce new properties on the request and response interfaces and slowly transition over

### Use Case

More efficient memory handling in the Core of the SDK

### Acknowledgements

- [X] I may be able to implement this feature request
- [X] This feature might incur a breaking change

### AWS .NET SDK and/or Package version used

v4

### Targeted .NET Platform

All

### Operating System and version

All

normj:
Closing since PR https://github.com/aws/aws-sdk-net/pull/4409 switching to ReadOnlyMemory.

normj:
As a note this PR is also coming to reduce allocation. https://github.com/aws/aws-sdk-net/pull/4424

Issue #1602, opened 2020-05-06
Terrible SQS SendMessage performance


billpoole-mi:
## Expected Behavior
An application should be able to send significantly more than 10MB/s to SQS per vCPU. Ideally, an application should be able to send well over 100MB/s per vCPU.

## Current Behavior
We are finding that sending about 130MB/s of messages to SQS is consuming about 15 vCPUs. We are finding a lot of CPU time being spent by the GC because we are finding that there are over 5GB of allocations for each 100MB of messages sent. 

We also find that this issue is proportional to total/aggregate payload size, not number of messages. That is, if we send much less data spread over many more messages, the CPU load is significantly less.

## Possible Solution
Simplify/streamline the .NET SQS client so it is performance-optimised. Minimise allocations, reducing GC pressure.

## Steps to Reproduce (for bugs)
Just create a simple application that uses the SQS client to concurrently send a large number of large messages, such that over 100MB/s of messages are being sent. It uses about 15 vCPUs on .NET Core 3.1 on Linux. The performance is even worse on Windows.

## Context
Our application produces a massive amount of data that needs to be sent through SQS, and at 15 vCPUs per 100MB, we find that a lot of our compute costs are coming from the .NET SQS client.

## Your Environment
* AWSSDK.SQS version used: 3.3.102.104
* Operating System and version: Ubuntu 18.04
* Visual Studio version: Visual Studio 2019 (16.5)
* Targeted .NET platform: .NET Core 3.1

## .NET Core Info
* .NET Core version used for development: .NET Core 3.1
* .NET Core version installed in the environment where application runs: .NET Core 3.1

billpoole-mi:
I just did a quick calculation, and with the above reported performance, it seems like there is a currently 25% cost overhead to using the SQS client provided by the AWS .NET SDK; where that cost is for the EC2 CPU-seconds needed by the SQS client to send each million messages, assuming each message is 64kB.

billpoole-mi:
Hi, just wondering if there’s an update on this? Does AWS recognise there is a problem? If so, does AWS plan on a fix/update? If so, is the fix scheduled? If so, what’s the time horizon?

billpoole-mi:
Is the SQS .NET client still being maintained by AWS?

normj:
[...]

PR #4409, opened 2026-05-15
Eliminate byte[] copy in JSON marshalling pipeline by keeping pooled buffer in request

muhammad-othman:
## Description
Eliminates unnecessary `byte[]` allocation and copy in the JSON request marshalling pipeline for non .NET Framework targets. Previously, the marshaller serialized JSON into a pooled `ArrayPoolBufferWriter<byte>`, then immediately called `.ToArray()` to copy the data into `request.Content`. This copy existed because downstream pipeline stages (SigV4 signer, HTTP handler) only supported `byte[]`.

In benchmarks against real SQS, this reduces per-request allocations by **37–46%** for large payloads (64KB) and **8–22%** for medium payloads (4KB), with no latency regression.

This change introduces `PooledContentStream` a seekable, read-only `Stream` backed by an `ArrayPoolBufferWriter<byte>`. The marshaller writes JSON into the stream's buffer and assigns it to `request.ContentStream`. This uses the existing `ContentStream` infrastructure that the pipeline already supports (seeking for retries, Content-Length from `stream.Length`, etc.).

Pipeline consumers that need zero-copy access cast `ContentStream` to `PooledContentStream` to read directly from the in-memory buffer:

- **Signer**: `ComputeContentStreamHash()` detects `PooledContentStream` and hashes directly from `Content.Span` using `SHA256.HashData` (.NET 8) or `ComputeSHA256Hash(byte[], offset, count)` via `MemoryMarshal.TryGetArray` (netcoreapp3.1/netstandard2.0). No stream seek overhead.
- **HttpHandler**: Writes the body using `ReadOnlyMemoryContent` (.NET 8) or zero-copy `ByteArrayContent(array, offset, count)` (netcoreapp3.1/netstandard2.0). Sets `Content-Length` from `stream.Length`.
- **CompressionHandler**: Reads from `PooledContentStream.Content`, disposes the pooled stream, and sets `request.Content` to the compressed bytes.
- **ChecksumHandler**: Computes checksums from the pooled array segment directly.

The pooled buffer is returned to the `ArrayPool` when `IRequest.Dispose()` is called in the Marshaller handler's finally block, after the entire inner pipeline (including retries) completes.

The `PooledContentStream` infrastructure is protocol-agnostic. This PR only addresses JSON marshallers (`JsonRPCRequestMarshaller.tt`), but the same pattern may be extended to other protocols if possible.

## Motivation and Context
`DOTNET-8585`
#3468 

## Testing
- Added `PooledContentWriterTests.cs`.
- Measured using `SendMessageBenchmark` against real SQS (includes network I/O).

  Here are the results before this change:

| Method           | PayloadSize | Mean     | Error    | StdDev   | Median   | P50      | P90      | Gen0    | Gen1    | Gen2    | Allocated |
|----------------- |------------ |---------:|---------:|---------:|---------:|---------:|---------:|--------:|--------:|--------:|----------:|
| SendMessage      | large       | 32.93 ms | 1.394 ms | 3.953 ms | 31.49 ms | 31.49 ms | 39.50 ms |       - |       - |       - | 173.86 KB |
| SendMessageBatch | large       | 50.53 ms | 3.090 ms | 8.816 ms | 48.05 ms | 48.05 ms | 63.89 ms | 83.3333 | 83.3333 | 83.3333 |  690.5 KB |
| SendMessage      | medium      | 31.46 ms | 1.396 ms | 4.094 ms | 30.97 ms | 30.97 ms | 37.49 ms |       - |       - |       - |  52.58 KB |
| SendMessageBatch | medium      | 34.29 ms | 1.512 ms | 4.409 ms | 33.59 ms | 33.59 ms | 41.10 ms |       - |       - |       - |  90.19 KB |
| SendMessage      | small       | 31.38 ms | 0.905 ms | 2.567 ms | 31.19 ms | 31.19 ms | 35.02 ms |       - |       - |       - |  45.61 KB |
| SendMessageBatch | small       | 35.01 ms | 1.588 ms | 4.531 ms | 33.82 ms | 33.82 ms | 41.74 ms |       - |       - |       - |  52.68 KB |

And after this change:

| Method           | PayloadSize | Mean     | Error    | StdDev   | Median   | P50      | P90      | Allocated |
|----------------- |------------ |---------:|---------:|---------:|---------:|---------:|---------:|----------:|
| SendMessage      | large       | 33.63 ms | 1.567 ms | 4.394 ms | 31.95 ms | 31.95 ms | 41.18 ms | 108.52 KB |
| SendMessageBatch | large       | 51.80 ms | 2.606 ms | 7.392 ms | 50.77 ms | 50.77 ms | 63.17 ms | 369.93 KB |
| SendMessage      | medium      | 29.35 ms | 1.053 ms | 3.038 ms | 28.16 ms | 28.16 ms | 33.52 ms |   48.5 KB |
| SendMessageBatch | medium      | 33.75 ms | 1.295 ms | 3.695 ms | 32.31 ms | 32.31 ms | 39.87 ms |  69.88 KB |
| SendMessage      | small       | 34.17 ms | 1.725 ms | 5.005 ms | 33.10 ms | 33.10 ms | 40.99 ms |  44.79 KB |
| SendMessageBatch | small       | 37.76 ms | 2.033 ms | 5.962 ms | 36.25 ms | 36.25 ms | 46.32 ms |  51.27 KB |

Savings scale with payload size, the optimization eliminates a `byte[]` copy proportional to the serialized request body. Large payloads (64KB) see the biggest reduction. Small payloads (256 bytes) see minimal improvement since the copy was already cheap, but no regression.

### Dry-runs

- **DotNet Dry-run ID:** 9690e644-331a-4c92-b3ab-8d79b242638f
  - [ ] Pending
  - [x] Completed successfully
  - [ ] Failed
- **PowerShell Dry-run ID:** d2e80f7a-87e1-4b86-afa4-c3d29fadf698
  - [ ] Pending
  - [x] Completed successfully
  - [ ] Failed

    No public API breaking changes.
       * `IRequest`: Already had `[AWSIsBackwardsCompatible]` adding `IDisposable` is safe.
       * `ICryptoUtil`: Already had `[AWSIsBackwardsCompatible]` so adding `ComputeSHA256Hash(byte[], int, int)` overload is safe.
       * `ICompressionAlgorithm`: Added `[AWSIsBackwardsCompatible]` attribute in this PR. This interface is only implemented internally (by `GZipCompression`), so adding `Compress(ReadOnlyMemory<byte>)` is safe. The attribute was missing previously but the interface was never intended for external implementation, this should be reviewed to confirm no external implementors exist.
       * `IHttpRequest<TRequestContent>`: Added `[AWSIsBackwardsCompatible]` attribute in this PR too similar to  `ICompressionAlgorithm`.
      No customer impact. The `ICompressionAlgorithm` interface is internal to the SDK (under `Amazon.Runtime.Internal.Compression` namespace) and only implemented by `GZipCompression` within the SDK. No customer code implements this interface. The `[AWSIsBackwardsCompatible]` attribute we added formalizes this existing reality.

        This is not a customer-facing breaking change. The alternative to adding `Compress(ReadOnlyMemory<byte>)` to the interface would be:
        - Casting to the concrete `GZipCompression` type in the `CompressionHandler` — avoids interface change but couples the handler to a specific implementation.
        - Adding a separate `ICompressionAlgorithmV2` interface — unnecessary complexity for an internal-only interface.
        
        Adding the method directly is the cleanest approach given this interface is never implemented externally.

      Yes. The `[AWSIsBackwardsCompatible]` attribute follows the same pattern as `IRequest` and `ICryptoUtil`, interfaces that are public for testability/visibility but only implemented within the SDK. Adding the attribute makes this contract explicit for future contributors.

      Built `AWSSDK.Core` after this change and `SQS` before this change, and tested `SQS` operations.
Pending review

## Types of changes

## Checklist

## License

### DIFF
--- a/generator/.DevConfigs/44066c9b-8ced-46a4-8b3b-d79bbc52c65d.json
+++ b/generator/.DevConfigs/44066c9b-8ced-46a4-8b3b-d79bbc52c65d.json
@@ -1,0 +1,16 @@
+
+{
+  "core": {
+    "changeLogMessages": [
+      "Eliminate byte[] copy in JSON marshalling pipeline by keeping pooled buffer in request"
+    ],
+    "type": "minor",
+    "updateMinimum": true,
+    "backwardIncompatibilitiesToIgnore": [ 
+      "Amazon.Runtime.Internal.Compression.GZipCompression/MethodOverloadAdded",
+      "Amazon.Runtime.IHttpRequest/MethodAbstractMethodAdded",
+      "Amazon.Runtime.HttpWebRequestMessage/MethodOverloadAdded",
+      "Amazon.Util.ICryptoUtil/MethodAbstractMethodAdded"
+    ]
+  }
+}
--- a/generator/ServiceClientGeneratorLib/Generators/Marshallers/JsonRPCRequestMarshaller.tt
+++ b/generator/ServiceClientGeneratorLib/Generators/Marshallers/JsonRPCRequestMarshaller.tt
@@ -128,18 +128,18 @@ namespace <#=this.Config.Namespace #>.Model.Internal.MarshallTransformations
 	{
 #>
 #if !NETFRAMEWORK
-			using ArrayPoolBufferWriter<byte> arrayPoolBufferWriter = new ArrayPoolBufferWriter<byte>();
-			using Utf8JsonWriter writer = new Utf8JsonWriter(arrayPoolBufferWriter);
+            request.ContentStream = new PooledContentStream();
+            using Utf8JsonWriter writer = new Utf8JsonWriter(((PooledContentStream)request.ContentStream).BufferWriter);
 #else
             using var memoryStream = new MemoryStream();
-			using Utf8JsonWriter writer = new Utf8JsonWriter(memoryStream);
+            using Utf8JsonWriter writer = new Utf8JsonWriter(memoryStream);
 #endif
-<#		
+<#
 		if (shouldMarshallPayload)
 		{
 #>
 			var context = new JsonMarshallerContext(request, writer);
-<#		
+<#
 			ProcessStructure(0, "publicRequest." + payload.PropertyName, payload.Shape, isSparse: requestStructure.IsSparse);
 #>
 			writer.Flush();
@@ -150,7 +150,7 @@ namespace <#=this.Config.Namespace #>.Model.Internal.MarshallTransformations
 #>
 			writer.WriteStartObject();
 			var context = new JsonMarshallerContext(request, writer);
-<#		
+<#
 			ProcessMembers(0, "publicRequest", this.Operation.RequestBodyMembers);
 #>
 			writer.WriteEndObject();
@@ -158,10 +158,7 @@ namespace <#=this.Config.Namespace #>.Model.Internal.MarshallTransformations
 <#
 		}
 #>
-			// ToArray() must be called here because aspects of sigv4 signing require a byte array
-#if !NETFRAMEWORK
-			request.Content = arrayPoolBufferWriter.WrittenMemory.ToArray();
-#else
+#if NETFRAMEWORK
 			request.Content = memoryStream.ToArray();
 #endif
 <#
--- a/generator/ServiceClientGeneratorLib/Generators/Marshallers/JsonRPCRequestMarshaller.tt
+++ b/generator/ServiceClientGeneratorLib/Generators/Marshallers/JsonRPCRequestMarshaller.tt
@@ -128,18 +128,18 @@ namespace <#=this.Config.Namespace #>.Model.Internal.MarshallTransformations
 	{
 #>
 #if !NETFRAMEWORK
-			using ArrayPoolBufferWriter<byte> arrayPoolBufferWriter = new ArrayPoolBufferWriter<byte>();
-			using Utf8JsonWriter writer = new Utf8JsonWriter(arrayPoolBufferWriter);
+			request.PooledContentWriter = new ArrayPoolBufferWriter<byte>();
+			using Utf8JsonWriter writer = new Utf8JsonWriter(request.PooledContentWriter);
 #else
             using var memoryStream = new MemoryStream();
 			using Utf8JsonWriter writer = new Utf8JsonWriter(memoryStream);
 #endif
-<#		
+<#
 		if (shouldMarshallPayload)
 		{
 #>
 			var context = new JsonMarshallerContext(request, writer);
-<#		
+<#
 			ProcessStructure(0, "publicRequest." + payload.PropertyName, payload.Shape, isSparse: requestStructure.IsSparse);
 #>
 			writer.Flush();
@@ -150,7 +150,7 @@ namespace <#=this.Config.Namespace #>.Model.Internal.MarshallTransformations
 #>
 			writer.WriteStartObject();
 			var context = new JsonMarshallerContext(request, writer);
-<#		
+<#
 			ProcessMembers(0, "publicRequest", this.Operation.RequestBodyMembers);
 #>
 			writer.WriteEndObject();
@@ -158,10 +158,7 @@ namespace <#=this.Config.Namespace #>.Model.Internal.MarshallTransformations
 <#
 		}
 #>
-			// ToArray() must be called here because aspects of sigv4 signing require a byte array
-#if !NETFRAMEWORK
-			request.Content = arrayPoolBufferWriter.WrittenMemory.ToArray();
-#else
+#if NETFRAMEWORK
 			request.Content = memoryStream.ToArray();
 #endif
 <#
--- a/sdk/src/Core/Amazon.Runtime/Internal/Compression/ICompressionAlgorithm.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Compression/ICompressionAlgorithm.cs
@@ -13,6 +13,7 @@
  * permissions and limitations under the License.
  */
 
+using System;
 using System.IO;
 
 namespace Amazon.Runtime.Internal.Compression
@@ -20,6 +21,7 @@ namespace Amazon.Runtime.Internal.Compression
     /// <summary>
     /// Interface for an implementation of compression algorithms
     /// </summary>
+    [AWSIsBackwardsCompatible]
     public interface ICompressionAlgorithm
     {
         /// <summary>
@@ -34,6 +36,15 @@ namespace Amazon.Runtime.Internal.Compression
         /// <param name="content"> Request payload content represented in byte array </param>
         /// <returns> Compressed data in a byte array format </returns>
         byte[] Compress(byte[] content);
+
+#if !NETFRAMEWORK
+        /// <summary>
+        /// Compresses request payload content from a read-only memory
+        /// </summary>
+        /// <param name="content"> Request payload content </param>
+        /// <returns> Compressed data in a byte array format </returns>
+        byte[] Compress(ReadOnlyMemory<byte> content);
+#endif
 
         /// <summary>
         /// Wraps <paramref name="inputStream"/> with an instance object of the compression stream algorithm
--- a/sdk/src/Core/Amazon.Runtime/Internal/DefaultRequest.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/DefaultRequest.cs
@@ -358,13 +358,37 @@ namespace Amazon.Runtime.Internal
 
             if (this.contentStreamHash == null)
             {
-                var seekableStream = WrapperStream.SearchWrappedStream(this.contentStream, s => s.CanSeek);
-                if (seekableStream != null)
+#if !NETFRAMEWORK
+                // Fast path: hash directly from the pooled buffer without stream seek overhead.
+                if (this.contentStream is PooledContentStream pooledStream)
                 {
-                    var position = seekableStream.Position;
-                    byte[] payloadHashBytes = CryptoUtilFactory.CryptoInstance.ComputeSHA256Hash(seekableStream);
+                    var memory = pooledStream.Content;
+#if NET8_0_OR_GREATER
+                    byte[] payloadHashBytes = System.Security.Cryptography.SHA256.HashData(memory.Span);
+#else
+                    byte[] payloadHashBytes;
+                    if (System.Runtime.InteropServices.MemoryMarshal.TryGetArray(memory, out var segment))
+                    {
+                        payloadHashBytes = CryptoUtilFactory.CryptoInstance.ComputeSHA256Hash(segment.Array, segment.Offset, segment.Count);
+                    }
+                    else
+                    {
+                        payloadHashBytes = CryptoUtilFactory.CryptoInstance.ComputeSHA256Hash(memory.ToArray());
+                    }
+#endif
                     this.contentStreamHash = AWSSDKUtils.ToHex(payloadHashBytes, true);
-                    seekableStream.Seek(position, SeekOrigin.Begin);
+                }
+                else
+#endif
+                {
+                    var seekableStream = WrapperStream.SearchWrappedStream(this.contentStream, s => s.CanSeek);
+                    if (seekableStream != null)
+                    {
+                        var position = seekableStream.Position;
+                        byte[] payloadHashBytes = CryptoUtilFactory.CryptoInstance.ComputeSHA256Hash(seekableStream);
+                        this.contentStreamHash = AWSSDKUtils.ToHex(payloadHashBytes, true);
+                        seekableStream.Seek(position, SeekOrigin.Begin);
+                    }
                 }
             }
 
@@ -589,5 +613,28 @@ namespace Amazon.Runtime.Internal
         /// Auth scheme chosen for the current request.
         /// </summary>
         public IAuthSchemeOption ChosenAuthScheme { get; set; }
+
+        /// <summary>
+        /// Disposes pooled resources held by this request.
+        /// </summary>
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        /// <summary>
+        /// Disposes managed resources.
+        /// </summary>
+        protected virtual void Dispose(bool disposing)
+        {
+#if !NETFRAMEWORK
+            if (disposing && contentStream is PooledContentStream)
+            {
+                contentStream.Dispose();
+                contentStream = null;
+            }
+#endif
+        }
     }
 }
--- a/sdk/src/Core/Amazon.Runtime/Internal/DefaultRequest.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/DefaultRequest.cs
@@ -22,6 +22,9 @@ using Amazon.Runtime.Internal.Util;
 using Amazon.Runtime.Internal.Auth;
 using Amazon.Util;
 using Amazon.Runtime.EventStreams;
+#if !NETFRAMEWORK
+using ThirdParty.RuntimeBackports;
+#endif
 
 namespace Amazon.Runtime.Internal
 {
@@ -301,10 +304,31 @@ namespace Amazon.Runtime.Internal
         {
             get
             {
+#if !NETFRAMEWORK
+                // Lazy copy: only allocates if external code accesses Content directly.
+                // Internal pipeline code should read PooledContentWriter.WrittenMemory instead.
+                if (content == null && PooledContentWriter != null)
+                {
+                    content = PooledContentWriter.WrittenMemory.ToArray();
+
+                    // Once we materialize the byte[] copy, release the pooled buffer immediately
+                    // to avoid holding two copies of the payload in memory.
+                    PooledContentWriter.Dispose();
+                    PooledContentWriter = null;
+                }
+#endif
                 return this.content;
             }
             set
             {
+#if !NETFRAMEWORK
+                // Return the pooled buffer before replacing the content.
+                if (PooledContentWriter != null)
+                {
+                    PooledContentWriter.Dispose();
+                    PooledContentWriter = null;
+                }
+#endif
                 this.content = value;
             }
         }
@@ -589,5 +613,33 @@ namespace Amazon.Runtime.Internal
         /// Auth scheme chosen for the current request.
         /// </summary>
         public IAuthSchemeOption ChosenAuthScheme { get; set; }
+
+#if !NETFRAMEWORK
+        /// <inheritdoc/>
+        public ArrayPoolBufferWriter<byte> PooledContentWriter { get; set; }
+#endif
+
+        /// <summary>
+        /// Disposes pooled resources held by this request.
+        /// </summary>
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        /// <summary>
+        /// Disposes managed resources.
+        /// </summary>
+        protected virtual void Dispose(bool disposing)
+        {
+            if (disposing)
+            {
+#if !NETFRAMEWORK
+                PooledContentWriter?.Dispose();
+                PooledContentWriter = null;
+#endif
+            }
+        }
     }
 }
--- a/sdk/src/Core/Amazon.Runtime/Internal/PooledContentStream.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/PooledContentStream.cs
@@ -1,0 +1,138 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+#if !NETFRAMEWORK
+using System;
+using System.IO;
+using ThirdParty.RuntimeBackports;
+
+namespace Amazon.Runtime.Internal
+{
+    /// <summary>
+    /// A seekable, read-only stream backed by an ArrayPoolBufferWriter.
+    /// Used as request.ContentStream to avoid byte[] copies in the marshalling pipeline.
+    /// Disposing returns the pooled buffer.
+    /// </summary>
+    public sealed class PooledContentStream : Stream
+    {
+        private ArrayPoolBufferWriter<byte> _writer;
+        private int _position;
+
+        /// <summary>
+        /// Creates a new PooledContentStream with its own ArrayPoolBufferWriter.
+        /// Use <see cref="BufferWriter"/> to write content before setting as ContentStream.
+        /// </summary>
+        public PooledContentStream()
+        {
+            _writer = new ArrayPoolBufferWriter<byte>();
+        }
+
+        /// <summary>
+        /// The underlying buffer writer. Use this to write content (e.g. via Utf8JsonWriter or XmlWriter).
+        /// Must not be accessed after the stream is set as ContentStream and pipeline processing begins.
+        /// </summary>
+        public ArrayPoolBufferWriter<byte> BufferWriter => _writer;
+
+        /// <summary>
+        /// The written content as a contiguous memory region.
+        /// Pipeline consumers use this for zero-copy access to the request body.
+        /// </summary>
+        public ReadOnlyMemory<byte> Content => _writer?.WrittenMemory ?? ReadOnlyMemory<byte>.Empty;
+
+        public override bool CanRead => true;
+        public override bool CanSeek => true;
+        public override bool CanWrite => false;
+        public override long Length => _writer?.WrittenCount ?? 0;
+
+        public override long Position
+        {
+            get => _position;
+            set
+            {
+                if (value < 0 || value > Length)
+                    throw new ArgumentOutOfRangeException(nameof(value));
+                _position = (int)value;
+            }
+        }
+
+        public override int Read(byte[] buffer, int offset, int count)
+        {
+            if (_writer == null)
+                throw new ObjectDisposedException(nameof(PooledContentStream));
+
+            var source = _writer.WrittenMemory;
+            int available = source.Length - _position;
+            if (available <= 0)
+                return 0;
+
+            int toCopy = Math.Min(count, available);
+            source.Slice(_position, toCopy).Span.CopyTo(buffer.AsSpan(offset, toCopy));
+            _position += toCopy;
+            return toCopy;
+        }
+
+#if NET8_0_OR_GREATER
+        public override int Read(Span<byte> buffer)
+        {
+            if (_writer == null)
+                throw new ObjectDisposedException(nameof(PooledContentStream));
+
+            var source = _writer.WrittenMemory;
+            int available = source.Length - _position;
+            if (available <= 0)
+                return 0;
+
+            int toCopy = Math.Min(buffer.Length, available);
+            source.Slice(_position, toCopy).Span.CopyTo(buffer.Slice(0, toCopy));
+            _position += toCopy;
+            return toCopy;
+        }
+#endif
+
+        public override long Seek(long offset, SeekOrigin origin)
+        {
+            if (_writer == null)
+                throw new ObjectDisposedException(nameof(PooledContentStream));
+
+            long newPosition = origin switch
+            {
+                SeekOrigin.Begin => offset,
+                SeekOrigin.Current => _position + offset,
+                SeekOrigin.End => _writer.WrittenCount + offset,
+                _ => throw new ArgumentOutOfRangeException(nameof(origin))
+            };
+            if (newPosition < 0 || newPosition > _writer.WrittenCount)
+                throw new ArgumentOutOfRangeException(nameof(offset));
+
+            _position = (int)newPosition;
+            return _position;
+        }
+
+        public override void Flush() { }
+        public override void SetLength(long value) => throw new NotSupportedException();
+        public override void Write(byte[] buffer, int offset, int count) => throw new NotSupportedException();
+
+        protected override void Dispose(bool disposing)
+        {
+            if (disposing)
+            {
+                _writer?.Dispose();
+                _writer = null;
+                _position = 0;
+            }
+            base.Dispose(disposing);
+        }
+    }
+}
+#endif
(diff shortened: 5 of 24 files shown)


muhammad-othman:
I also ran some Benchmark to test SQS throughput considering #1602, and the change shows consistent improvements especailly with large payloads
- Before

| Method                     | Concurrency | MessageSizeBytes | TotalMessages | Mean        | Error     | StdDev     | Median      | P50         | P90         | Allocated |
|--------------------------- |------------ |----------------- |-------------- |------------:|----------:|-----------:|------------:|------------:|------------:|----------:|
| ConcurrentSendMessage      | 4           | 4096             | 100           | 1,662.78 ms | 33.147 ms |  72.058 ms | 1,653.63 ms | 1,653.63 ms | 1,759.60 ms |   5.42 MB |
| ConcurrentSendMessageBatch | 4           | 4096             | 100           |   210.22 ms |  4.371 ms |  12.472 ms |   205.38 ms |   205.38 ms |   227.79 ms |   1.33 MB |
| ConcurrentSendMessage      | 4           | 65536            | 100           | 1,787.90 ms | 40.930 ms | 119.395 ms | 1,769.12 ms | 1,769.12 ms | 1,938.61 ms |  16.88 MB |
| ConcurrentSendMessageBatch | 4           | 65536            | 100           |   336.73 ms | 16.933 ms |  47.203 ms |   316.83 ms |   316.83 ms |   425.57 ms |  13.05 MB |
| ConcurrentSendMessage      | 16          | 4096             | 100           |   434.03 ms |  6.546 ms |   9.997 ms |   430.96 ms |   430.96 ms |   449.83 ms |   5.13 MB |
| ConcurrentSendMessageBatch | 16          | 4096             | 100           |    81.21 ms |  2.583 ms |   7.157 ms |    80.50 ms |    80.50 ms |    90.21 ms |   1.33 MB |
| ConcurrentSendMessage      | 16          | 65536            | 100           |   480.31 ms | 11.754 ms |  32.571 ms |   467.10 ms |   467.10 ms |   523.37 ms |  16.86 MB |
| ConcurrentSendMessageBatch | 16          | 65536            | 100           |   188.27 ms | 12.509 ms |  35.890 ms |   175.75 ms |   175.75 ms |   255.01 ms |  13.05 MB |

- After

| Method                     | Concurrency | MessageSizeBytes | TotalMessages | Mean        | Error     | StdDev     | Median      | P50         | P90         | Allocated   |
|--------------------------- |------------ |----------------- |-------------- |------------:|----------:|-----------:|------------:|------------:|------------:|------------:|
| ConcurrentSendMessage      | 4           | 4096             | 100           | 1,609.75 ms | 32.079 ms |  67.666 ms | 1,587.97 ms | 1,587.97 ms | 1,706.44 ms |  4844.78 KB |
| ConcurrentSendMessageBatch | 4           | 4096             | 100           |   224.31 ms |  7.607 ms |  22.068 ms |   219.51 ms |   219.51 ms |   256.09 ms |   953.54 KB |
| ConcurrentSendMessage      | 4           | 65536            | 100           | 1,762.76 ms | 43.596 ms | 127.860 ms | 1,731.12 ms | 1,731.12 ms | 1,950.82 ms | 10899.02 KB |
| ConcurrentSendMessageBatch | 4           | 65536            | 100           |   336.17 ms | 15.469 ms |  42.864 ms |   318.27 ms |   318.27 ms |   404.49 ms |   6955.6 KB |
| ConcurrentSendMessage      | 16          | 4096             | 100           |   435.31 ms |  4.859 ms |   5.785 ms |   433.10 ms |   433.10 ms |   441.53 ms |  4908.46 KB |
| ConcurrentSendMessageBatch | 16          | 4096             | 100           |    79.97 ms |  2.236 ms |   6.159 ms |    78.71 ms |    78.71 ms |    87.80 ms |   953.51 KB |
| ConcurrentSendMessage      | 16          | 65536            | 100           |   473.45 ms |  9.462 ms |  20.569 ms |   467.43 ms |   467.43 ms |   501.73 ms | 10846.49 KB |
| ConcurrentSendMessageBatch | 16          | 65536            | 100           |   188.85 ms |  9.637 ms |  26.218 ms |   181.20 ms |   181.20 ms |   230.21 ms |  6958.52 KB |



<details>
<summary>Benchmark code</summary>

```csharp
using Amazon;
using Amazon.SQS;
using Amazon.SQS.Model;
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Columns;
using BenchmarkDotNet.Configs;

namespace SQSBenchmarks;

/// <summary>
/// Validates the fix for https://github.com/aws/aws-sdk-net/issues/1602
/// The issue reports excessive allocations (~5GB per 100MB sent) and high CPU usage
/// when sending large volumes of SQS messages. This benchmark measures throughput (MB/s)
/// and GC pressure under sustained concurrent load.
/// </summary>
[MemoryDiagnoser]
[GcServer(true)]
[Config(typeof(Config))]
public class ThroughputBenchmark
{
    private class Config : ManualConfig
    {
        public Config()
        {
            AddColumn(StatisticColumn.P50);
            AddColumn(StatisticColumn.P90);
        }
    }

    private IAmazonSQS _client;
    private string _queueUrl;
    private string _messageBody;

    [Params(4, 16)]
    public int Concurrency { get; set; }

    [Params(4096, 65536)]
    public int MessageSizeBytes { get; set; }

    /// <summary>
    /// Total number of messages to send per benchmark iteration.
    /// </summary>
    [Params(100)]
    public int TotalMessages { get; set; }

    [GlobalSetup]
    public async Task Setup()
    {
        _client = new AmazonSQSClient();

        var createResponse = await _client.CreateQueueAsync(new CreateQueueRequest
        {
            QueueName = $"throughput-bench-{Guid.NewGuid():N}"
        });
        _queueUrl = createResponse.QueueUrl;

        _messageBody = GeneratePayload(MessageSizeBytes);
    }

    [GlobalCleanup]
    public async Task Cleanup()
    {
        if (_queueUrl != null)
        {
            await _client.DeleteQueueAsync(new DeleteQueueRequest { QueueUrl = _queueUrl });
        }
        _client?.Dispose();
    }

    /// <summary>
    /// Sends TotalMessages concurrently using SendMessageAsync with the specified concurrency level.
    /// Measures total allocation and GC pressure for the entire batch.
    /// </summary>
    [Benchmark(Description = "ConcurrentSendMessage")]
    public async Task ConcurrentSendMessage()
    {
        using var semaphore = new SemaphoreSlim(Concurrency);
        var tasks = new List<Task>(TotalMessages);

        for (int i = 0; i < TotalMessages; i++)
        {
            await semaphore.WaitAsync();
            tasks.Add(Task.Run(async () =>
            {
                try
                {
                    await _client.SendMessageAsync(new SendMessageRequest
                    {
                        QueueUrl = _queueUrl,
                        MessageBody = _messageBody
                    });
                }
                finally
                {
                    semaphore.Release();
                }
            }));
        }

        await Task.WhenAll(tasks);
    }

    [Benchmark(Description = "ConcurrentSendMessageBatch")]
    public async Task ConcurrentSendMessageBatch()
    {
        const int batchSize = 10;
        int totalBatches = (TotalMessages + batchSize - 1) / batchSize;

        using var semaphore = new SemaphoreSlim(Concurrency);
        var tasks = new List<Task>(totalBatches);

        for (int b = 0; b < totalBatches; b++)
        {
            await semaphore.WaitAsync();
            int currentBatchSize = Math.Min(batchSize, TotalMessages - b * batchSize);
            tasks.Add(Task.Run(async () =>
            {
                try
                {
                    var entries = new List<SendMessageBatchRequestEntry>(currentBatchSize);
                    for (int i = 0; i < currentBatchSize; i++)
                    {
                        entries.Add(new SendMessageBatchRequestEntry
                        {
                            Id = i.ToString(),
                            MessageBody = _messageBody
                        });
                    }

                    await _client.SendMessageBatchAsync(new SendMessageBatchRequest
                    {
                        QueueUrl = _queueUrl,
                        Entries = entries
                    });
                }
                finally
                {
                    semaphore.Release();
                }
            }));
        }

        await Task.WhenAll(tasks);
    }

    private static string GeneratePayload(int length)
    {
        var chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
        var random = new Random(42);
        var buffer = new char[length];
        for (int i = 0; i < length; i++)
        {
            buffer[i] = chars[random.Next(chars.Length)];
        }
        return new string(buffer);
    }
}

</details> ```

File: generator/ServiceClientGeneratorLib/Generators/Marshallers/JsonRPCRequestMarshaller.tt [hunk 1]
```diff
@@ -150,17 +150,16 @@ namespace <#=this.Config.Namespace #>.Model.Internal.MarshallTransformations
...
 #>
 			writer.WriteEndObject();
 			writer.Flush();
 <#
 		}
 #>
-			// ToArray() must be called here because aspects of sigv4 signing require a byte array
 #if !NETFRAMEWORK
-			request.Content = arrayPoolBufferWriter.WrittenMemory.ToArray();
+			request.PooledContentWriter = arrayPoolBufferWriter;
```
normj: If there is an exception in the marshaller the `arrayPoolBufferWriter` should be disposed.

sdk/src/Core/Amazon.Runtime/Internal/DefaultRequest.cs
```diff
@@ -301,10 +304,31 @@ public byte[] Content
         {
             get
             {
+#if !NETFRAMEWORK
+                // Lazy copy: only allocates if external code accesses Content directly.
+                // Internal pipeline code should read PooledContentWriter.WrittenMemory instead.
+                if (content == null && PooledContentWriter != null)
```
normj: This makes me nervous because all somebody has to do later is make an accidentally call to this property before checking `PooledContentWriter` and then the the performance gain is lost but tests will still show everything is successful. 

We need some compile mechanism to let us know when we mess this up.

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/Marshaller.cs
```diff
@@ -56,7 +67,24 @@ public override void InvokeSync(IExecutionContext executionContext)
         public override System.Threading.Tasks.Task<T> InvokeAsync<T>(IExecutionContext executionContext)
         {
             PreInvoke(executionContext);
-            return base.InvokeAsync<T>(executionContext);
+            return DisposeRequestAfterAsync<T>(executionContext, base.InvokeAsync<T>(executionContext));
+        }
+
+        // Separated to avoid creating an async state machine in InvokeAsync on every call.
```
normj: This separation doesn't gain you anything. Yes you moved the state machine allocation out of InvokeAsync but it is still created in DisposeRequestAfterAsync so the net effect is the same amount of state machine allocations.

sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/HttpHandler.cs
```diff
@@ -424,37 +432,50 @@ protected virtual IHttpRequest<TRequestContent> CreateWebRequest(IRequestContext
...
 
+#if !NETFRAMEWORK
+            if (request.PooledContentWriter != null)
+            {
+                request.Headers[HeaderKeys.ContentLengthHeader] =
+                    request.PooledContentWriter.WrittenCount.ToString(CultureInfo.InvariantCulture);
+                return httpRequest;
+            }
+#endif
+            if (request.Content != null)
```
normj: If there is a `PooledContentWriter` doesn't that mean ` if (request.Content != null)` will also evaluate to true and set content length again after it wen in the `if (request.PooledContentWriter != null)` block?

generator/ServiceClientGeneratorLib/Generators/Marshallers/JsonRPCRequestMarshaller.tt
muhammad-othman: This happens when `Request?.Dispose()` is called from the `finally` clause at the `Marshaller` pipeline handler 
https://github.com/aws/aws-sdk-net/blob/muhamoth/DOTNET-8585-perf-support-ReadOnlyMemory/sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/Marshaller.cs#L86

Which calls `PooledContentWriter?.Dispose()` 
https://github.com/aws/aws-sdk-net/blob/muhamoth/DOTNET-8585-perf-support-ReadOnlyMemory/sdk/src/Core/Amazon.Runtime/Internal/DefaultRequest.cs#L639

sdk/src/Core/Amazon.Runtime/Internal/DefaultRequest.cs
muhammad-othman: I thought about making Content obsolete for `!NETFRAMEWORK`, but it's still used by other protocols that and using `PooledContentWriter` for all protocols might be tricky. For xml it should be straightforward, CBOR is a bit tricky as we don't control the `CborWriter` internals, though it may be possible if we were able to rework it, Query protocol is tricky too and not sure if it is possible.

For now, we can add a test that runs the JSON pipeline and verifies that `PooledContentWriter` is still not null before the unmarshaller Dispose is called. This won't catch every possible path, but it will break if someone adds a Content access in the main pipeline flow.

Do you think there is a way that is more foolproof?

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/Marshaller.cs
muhammad-othman: I agree, will remove it.

sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/HttpHandler.cs
muhammad-othman: Not really, as we have `return httpRequest;` at the end of the if statement, so we shouldn't get to ` if (request.Content != null)` when `PooledContentWriter` isn't null.
This proofs your earlier point that a simple `request.Content` access here or there even to check if it is null will materialize the `PooledContentWriter` and will make us lose the optimizations, still I didn't want to be aggressive and do something that would break the customers if the `request.Content` was accessed by mistake as I would rather lose the optimization to avoid making the code very fragile.

generator/ServiceClientGeneratorLib/Generators/Marshallers/JsonRPCRequestMarshaller.tt
normj: I'm talking about an exception that happens after the `arrayPoolBufferWriter` is create but before `request.PooledContentWriter` is set. If `request.PooledContentWriter` is not set then the dispose in the `Marshaller` pipeline won't have an instance to dispose.

sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/HttpHandler.cs
normj: Ah missed the `return` in the block.

From a coding style point of view I would recommend having the `if (request.PooledContentWriter != null)` do an immediate return while all the other `if` / `else if` blocks setting content length do not do an immediate return. It is asking for a subtle bug is somebody adds some new logic before the return on line 479 thinking that after all branches for setting the content length are resolved code flow will go there.

muhammad-othman: It took me a while to understand this comment  😅, if you mean `I would "not" recommend having the ...`  then this sounds reasonable and I have updated it.

sdk/src/Core/Amazon.Runtime/Internal/DefaultRequest.cs
muhammad-othman: I've added this test https://github.com/aws/aws-sdk-net/blob/be6d463119fb68ea6ac09ee6f957b65bf0703a0a/sdk/test/UnitTests/Custom/Runtime/PooledContentWriterTests.cs#L110
Will add similar tests to cover other protocols when they support `PooledContentWriter` too.

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/Marshaller.cs
muhammad-othman: Removed it, now I'm using simple `await` and didn't see performance differences between them.

generator/ServiceClientGeneratorLib/Generators/Marshallers/JsonRPCRequestMarshaller.tt
muhammad-othman: Make sense, I totally missed that, the temporary `arrayPoolBufferWriter` variable isn't needed anyway so I've updated it to set `request.PooledContentWriter = new ArrayPoolBufferWriter<byte>();` directly, so we always have a reference to the `ArrayPoolBufferWriter` when creating it which we will dispose in the finally clause.

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/Marshaller.cs
```diff
@@ -34,29 +34,51 @@ namespace Amazon.Runtime.Internal
...
+                return await base.InvokeAsync<T>(executionContext).ConfigureAwait(false);
+            }
+            finally
+            {
+                // Returns pooled buffers (e.g. ArrayPoolBufferWriter) to the pool after the entire inner
+                // pipeline completes, including all retries. Safe to call even if already disposed.
+                // This must not happen before the HttpHandler finishes reading the WrittenMemory,
+                // since HttpRequestMessageFactory wraps it in ReadOnlyMemoryContent/ByteArrayContent.
+                executionContext.RequestContext.Request?.Dispose();
+            }
```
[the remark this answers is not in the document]
muhammad-othman: Discussed here https://github.com/aws/aws-sdk-net/pull/4409#discussion_r3251355840

sdk/src/Core/Amazon.Runtime/Internal/DefaultRequest.cs
normj: What if instead of creating the new property we create our own wrapper custom stream around the `IBufferedWriter<byte>`. In the marshaller we set the `ContentStream` property to our `BufferWriterStream(pooledContentWriter)`. Then in case where we need to be aware that it is the `IBufferedWriter<byte>` like in the HttpHandler we cast the `ContentStream` to our custom stream that exposes what we need from the `IBufferedWriter<byte>`. That way we avoid side effect behavior by just accessing a property like `ContentStream` before checking another property.

sdk/src/Core/Amazon.Runtime/Internal/DefaultRequest.cs
```diff
@@ -589,5 +613,26 @@ public string GetHeaderValue(string headerName)
...
+        }
+
+        /// <summary>
+        /// Disposes managed resources.
+        /// </summary>
+        protected virtual void Dispose(bool disposing)
+        {
+            if (disposing)
+            {
+                contentStream?.Dispose();
```
normj: I think you need to do this only if `contentStream` is the `PooledContentStream`. The `contentStream` could also be an input stream from S3 which has it's own user controlled choice on whether to dispose the stream via the `AutoCloseStream` property on the `PutObjectRequest`.

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/CompressionHandler.cs
```diff
@@ -93,13 +93,26 @@ protected virtual void PreInvoke(IExecutionContext executionContext)
...
+#endif
                 if (input.Length >= minCompressionSize)
                 {
                     executionContext.RequestContext.Metrics.AddProperty(Metric.UncompressedRequestSize, input.Length);
                     using (TracingUtilities.CreateSpan(executionContext.RequestContext, TelemetryConstants.RequestCompressionSpanName))
                     using (executionContext.RequestContext.Metrics.StartEvent(Metric.RequestCompressionTime))
                     {
+#if !NETFRAMEWORK
+                        // Dispose the pooled stream before replacing with compressed content
+                        request.ContentStream?.Dispose();
```
normj: Is this a problem that we got `input` from `ContentStream` but then `ContentStream` is disposed and after that we use `input`. That doesn't cause anything with the memory behind `input` to be possibly reused before compress does it?

sdk/test/UnitTests/Custom/Runtime/PooledContentStreamTests.cs
```diff
@@ -0,0 +1,185 @@
...
+
+        private class TestRequest : AmazonWebServiceRequest
+        {
+        }
+
+        private class NoOpHandler : PipelineHandler
+        {
+            public override void InvokeSync(IExecutionContext executionContext) { }
+            public override System.Threading.Tasks.Task<T> InvokeAsync<T>(IExecutionContext executionContext)
+                => System.Threading.Tasks.Task.FromResult(default(T));
```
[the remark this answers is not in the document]
muhammad-othman: This isn't accurate as this class inherits from `PipelineHandler` not implement `IPipelineHandler`.

sdk/src/Core/Amazon.Runtime/Internal/Compression/ICompressionAlgorithm.cs
```diff
@@ -13,13 +13,15 @@
...
 
+using System;
 using System.IO;
 
 namespace Amazon.Runtime.Internal.Compression
 {
     /// <summary>
     /// Interface for an implementation of compression algorithms
     /// </summary>
+    [AWSIsBackwardsCompatible]
```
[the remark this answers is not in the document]
muhammad-othman: This isn't accurate, AWSIsBackwardsCompatible is resolved correctly

sdk/src/Core/Amazon.Runtime/Internal/PooledContentStream.cs
```diff
@@ -0,0 +1,137 @@
...
+        {
+            _writer = new ArrayPoolBufferWriter<byte>();
+        }
+
+        /// <summary>
+        /// The underlying buffer writer. Use this to write content (e.g. via Utf8JsonWriter or XmlWriter).
+        /// Must not be accessed after the stream is set as ContentStream and pipeline processing begins.
+        /// </summary>
+        public ArrayPoolBufferWriter<byte> BufferWriter => _writer;
+
```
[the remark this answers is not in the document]
muhammad-othman: We don't access the writer after we dispose the PooledContentStream but fixed it for correctness.

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/CompressionHandler.cs
```diff
@@ -83,6 +83,20 @@ protected virtual void PreInvoke(IExecutionContext executionContext)
             }
 
             var compressionAlgorithm = CompressionFactory.GetCompressionAlgorithm(request.CompressionAlgorithm);
+#if !NETFRAMEWORK
+            // PooledContentStream holds all data in memory and compress as a whole buffer
+            // rather than wrapping in CompressionWrapperStream with chunked encoding.
+            if (request.ContentStream is PooledContentStream pooledStream)
+            {
+                if (CompressBufferContent(executionContext, request, pooledStream.Content, compressionAlgorithm, minCompressionSize))
+                {
```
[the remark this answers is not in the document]
muhammad-othman: PooledContentStream Content always represents the full request body and Position is not relevant for the zero-copy path.

sdk/src/Core/Amazon.Runtime/Internal/Util/ChecksumUtils.cs
```diff
@@ -207,6 +207,23 @@ public static void SetRequestChecksumMD5(IRequest request)
...
+                {
+                    var checksumBytes = algorithm.ComputeHash(segment.Array, segment.Offset, segment.Count);
+                    return Convert.ToBase64String(checksumBytes);
+                }
+                else
+                {
+                    var checksumBytes = algorithm.ComputeHash(memory.ToArray());
+                    return Convert.ToBase64String(checksumBytes);
+                }
+            }
```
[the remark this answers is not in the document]
muhammad-othman: `PooledContentStream.Content` always represents the full request body and Position is not relevant for the zero-copy path.

sdk/src/Core/Amazon.Runtime/Internal/DefaultRequest.cs
```diff
@@ -358,13 +358,37 @@ public string ComputeContentStreamHash()
...
-                if (seekableStream != null)
+#if !NETFRAMEWORK
+                // Fast path: hash directly from the pooled buffer without stream seek overhead.
+                if (this.contentStream is PooledContentStream pooledStream)
                 {
-                    var position = seekableStream.Position;
-                    byte[] payloadHashBytes = CryptoUtilFactory.CryptoInstance.ComputeSHA256Hash(seekableStream);
+                    var memory = pooledStream.Content;
+#if NET8_0_OR_GREATER
+                    byte[] payloadHashBytes = System.Security.Cryptography.SHA256.HashData(memory.Span);
```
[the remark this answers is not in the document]
muhammad-othman: `PooledContentStream.Content` always represents the full request body and Position is not relevant for the zero-copy path.

sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/HttpHandler.cs
```diff
@@ -292,6 +292,13 @@ private void WriteContentToRequestBody(TRequestContent requestContent,
             // This code path ends up using a ByteArrayContent for System.Net.HttpClient used by .NET Core.
             // HttpClient can't seem to handle ByteArrayContent with 0 length so in that case use
             // the StreamContent code path.
+#if !NETFRAMEWORK
+            else if (wrappedRequest.ContentStream is PooledContentStream pooledStream && pooledStream.Content.Length > 0)
+            {
+                requestContext.Metrics.AddProperty(Metric.RequestSize, pooledStream.Content.Length);
+                httpRequest.WriteToRequestBody(requestContent, pooledStream.Content, requestContext.Request.Headers);
+            }
```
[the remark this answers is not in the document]
muhammad-othman: `PooledContentStream.Content` always represents the full request body and Position is not relevant for the zero-copy path, the fast path intentionally bypasses stream position semantics.

sdk/src/Core/Amazon.Runtime/Internal/PooledContentStream.cs
```diff
@@ -0,0 +1,137 @@
...
+        public override void Flush() { }
+        public override void SetLength(long value) => throw new NotSupportedException();
+        public override void Write(byte[] buffer, int offset, int count) => throw new NotSupportedException();
+
+        protected override void Dispose(bool disposing)
+        {
+            if (disposing)
+            {
+                _writer?.Dispose();
+                _writer = null;
```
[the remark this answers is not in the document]
muhammad-othman: We don't really access the position after dispose but updated it for correctness.

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/CompressionHandler.cs
```diff
@@ -83,6 +83,20 @@ protected virtual void PreInvoke(IExecutionContext executionContext)
...
+                if (CompressBufferContent(executionContext, request, pooledStream.Content, compressionAlgorithm, minCompressionSize))
+                {
+                    // Dispose after Compress has fully consumed the pooled buffer
+                    request.ContentStream.Dispose();
+                    request.ContentStream = null;
+                }
+            }
+            else
+#endif
             if (request.ContentStream != null)
```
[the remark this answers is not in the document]
muhammad-othman: updated.

sdk/src/Core/Amazon.Runtime/Pipeline/Handlers/CompressionHandler.cs
muhammad-othman: updated.

sdk/src/Core/Amazon.Runtime/Internal/DefaultRequest.cs
muhammad-othman: Updated it to only dispose if the contentStream is PooledContentStream

normj (dismissed):

generator/.DevConfigs/44066c9b-8ced-46a4-8b3b-d79bbc52c65d.json
```diff
@@ -0,0 +1,16 @@
+
+{
+  "core": {
+    "changeLogMessages": [
+      "Eliminate byte[] copy in JSON marshalling pipeline by keeping pooled buffer in request"
```
peterrsongg, line 5: Should we call out that any external code that mocks `IRequest` will break because it implements `IDisposable`? That seems like one of those silent breaking changes, even though it is in the internal namespace and fair game. idk what do you think @normj ?

generator/ServiceClientGeneratorLib/Generators/Marshallers/JsonRPCRequestMarshaller.tt
```diff
@@ -128,18 +128,18 @@ namespace <#=this.Config.Namespace #>.Model.Internal.MarshallTransformations
 	{
 #>
 #if !NETFRAMEWORK
-			using ArrayPoolBufferWriter<byte> arrayPoolBufferWriter = new ArrayPoolBufferWriter<byte>();
-			using Utf8JsonWriter writer = new Utf8JsonWriter(arrayPoolBufferWriter);
+            request.ContentStream = new PooledContentStream();
```
peterrsongg, line 131: Great idea to wrap the writer in a class so that we can just use the existing `request.ContentStream` property. When I originally migrated the json marshalling code to System.Text.Json I was trying to decide how we would remove the byte array allocation and I was only thinking from the perspective of adding a new property (which would've been the writer). And that would've had its own issues, especially needing to check the `PooledContentWriter` first which would then silently create a byte[] copy. This wrapper approach removes that completely

peterrsongg (dismissed):
Approved, just a minor comment about the changelog message regarding the addition of `IDisposable` to `IRequest`

generator/.DevConfigs/44066c9b-8ced-46a4-8b3b-d79bbc52c65d.json
muhammad-othman: It even has the `AWSIsBackwardsCompatible` attribute already.

muhammad-othman:
Reverted `ObjectDisposedException` when accessing `BufferWriter` as it causes `CA1065` which says properties shouldn't throw exceptions .
I didn't want to overcomplicate the code to fix `CA1065` as since the doc already says "Must not be accessed after disposal." The property is only called by the marshaller during construction, never after pipeline processing. The `ObjectDisposedException` guard was over-engineering for a code path that don't happen in practice anyway.

generator/.DevConfigs/44066c9b-8ced-46a4-8b3b-d79bbc52c65d.json
normj: I don't want to start the precedence of saying we need to add change log entries for changes to "internal" code.

peterrsongg (approved):

normj (approved):

boblodgett (approved):

PR #4410, opened 2026-05-18
Allow configuring of HttpClientFactory through DI factories

### LINKED ISSUE #4092: Configure HttpClientFactory in ClientConfig instance using dependency injection factories
### Describe the feature

Allow configuring `clientConfig.HttpClientFactory` through dependency injection factories.

### Use Case

I'm faced with the problem that it's not possible to configure `clientConfig.HttpClientFactory` through `AddAWSService<>`; there's no overload that allows you to configure that instance using IServiceProvider.

I know that many `AWSOptions` options can be configured from appsettings.json for example, but that `AWSOptions` only has a subset of ClientConfig fields and only primitive type fields, not for those that require, for example, an instance of `HttpClientFactory`:

https://github.com/aws/aws-sdk-net/blob/dc51341e3fc91d9b09e6546eb834efc1d425b6df/extensions/src/AWSSDK.Extensions.NETCore.Setup/ClientFactory.cs#L221

The `DefaultClientConfig` class does not allow to configure HttpClientFactory:

https://github.com/aws/aws-sdk-net/blob/dc51341e3fc91d9b09e6546eb834efc1d425b6df/extensions/src/AWSSDK.Extensions.NETCore.Setup/DefaultClientConfig.cs#L24

In my case, I would like to build it with the default factory but be able to configure it later in another step. My intention is to use `IAmazonS3` but with: https://learn.microsoft.com/es-es/dotnet/core/extensions/httpclient-factory

Currently, the only way (without using DI, in my case it's not possible as I'm using IHttpClientFactory from Microsoft) is to set `AWSConfigs.HttpClientFactory`, but this is not ideal as it is static:

https://github.com/aws/aws-sdk-net/blob/dc51341e3fc91d9b09e6546eb834efc1d425b6df/sdk/src/Core/Amazon.Runtime/_netstandard/ClientConfig.cs#L123

### Proposed Solution

I suppose it could be solved with a minor change in one of the many PRs created by @chase-miller: https://github.com/aws/aws-sdk-net/pull/3715#issuecomment-2802557547

I hope they get unblocked somehow because he has done a tremendous job.

### Other Information

_No response_

### Acknowledgements

AWSSDK.Extensions.NETCore.Setup -> 4.0.3.9
AWSSDK.S3 -> 4.0.9.1

### Targeted .NET Platform

.NET 9

Windows 11

peterrsongg:
## Description

Adds an overload to `AddAWSService` to provide a way to configure `clientConfig.HttpClientFactory` directly at the point of registration. This allows users to use the `IHttpClientFactory` provided by microsoft which is ONLY available through DI (or any custom factory), with an AWS service.

Allows something like:
```
  services.AddHttpClient("aws-s3");

  services.AddAWSService<IAmazonS3>((config, sp) =>
  {
      var httpFactory = sp.GetRequiredService<IHttpClientFactory>();
      config.HttpClientFactory = new AwsHttpClientFactory(httpFactory, "aws-s3");
  });
```

The user still has to create a class that bridges the `IHttpClientFactory` with the SDK's HttpClientFactory like so (just an example):
```
  public class AwsHttpClientFactory : Amazon.Runtime.HttpClientFactory
  {
      private readonly IHttpClientFactory _httpClientFactory;
      private readonly string _name;

      public AwsHttpClientFactory(IHttpClientFactory httpClientFactory, string name)
      {
          _httpClientFactory = httpClientFactory;
          _name = name;
      }

      public override HttpClient CreateHttpClient(IClientConfig clientConfig)
      {
          return _httpClientFactory.CreateClient(_name);
      }

      public override bool UseSDKHttpClientCaching(IClientConfig clientConfig)
      {
          // Microsoft's IHttpClientFactory manages the handler lifecycle,
          // so don't let the SDK cache on top of it.
          return false;
      }

      public override bool DisposeHttpClientsAfterUse(IClientConfig clientConfig)
      {
          // IHttpClientFactory-created clients are safe to dispose (it only
          // disposes the HttpClient wrapper, not the underlying handler).
          return true;
      }
  }
```

## Motivation and Context

Fixes #4092 
## Testing

UnitTests added, dry run passed
### Dry-runs

- **DotNet Dry-run ID:** ea61611a-8e2c-4872-8bbd-f5c3664228fa
  - [ ] Pending
  - [x] Completed successfully
  - [ ] Failed
- **PowerShell Dry-run ID:** 8f291e19-3267-4b79-8231-2c31c9d4ad5c
  - [ ] Pending
  - [x] Completed successfully
  - [ ] Failed

## Types of changes

## Checklist

## License

### DIFF
--- a/extensions/src/AWSSDK.Extensions.NETCore.Setup/ServiceCollectionExtensions.cs
+++ b/extensions/src/AWSSDK.Extensions.NETCore.Setup/ServiceCollectionExtensions.cs
@@ -106,7 +106,7 @@ namespace Microsoft.Extensions.DependencyInjection
         /// <returns>Returns back the IServiceCollection to continue the fluent system of IServiceCollection.</returns>
         public static IServiceCollection AddAWSService<T>(this IServiceCollection collection, ServiceLifetime lifetime = ServiceLifetime.Singleton) where T : IAmazonService
         {
-            return AddAWSService<T>(collection, null, lifetime);
+            return AddAWSService<T>(collection, (AWSOptions)null, lifetime);
         }
 
         /// <summary>
@@ -130,6 +130,42 @@ namespace Microsoft.Extensions.DependencyInjection
         }
 
         /// <summary>
+        /// Adds the Amazon service client to the dependency injection framework. The Amazon service client is not
+        /// created until it is requested. If the ServiceLifetime property is set to Singleton, the default, then the same
+        /// instance will be reused for the lifetime of the process and the object should not be disposed.
+        /// </summary>
+        /// <typeparam name="T">The AWS service interface, like IAmazonS3.</typeparam>
+        /// <param name="collection"></param>
+        /// <param name="configAction">An action to configure the ClientConfig using services from the IServiceProvider.</param>
+        /// <param name="lifetime">The lifetime of the service client created. The default is Singleton.</param>
+        /// <returns>Returns back the IServiceCollection to continue the fluent system of IServiceCollection.</returns>
+        public static IServiceCollection AddAWSService<T>(this IServiceCollection collection, Action<ClientConfig, IServiceProvider> configAction, ServiceLifetime lifetime = ServiceLifetime.Singleton) where T : IAmazonService
+        {
+            return AddAWSService<T>(collection, null, configAction, lifetime);
+        }
+
+        /// <summary>
+        /// Adds the Amazon service client to the dependency injection framework. The Amazon service client is not
+        /// created until it is requested. If the ServiceLifetime property is set to Singleton, the default, then the same
+        /// instance will be reused for the lifetime of the process and the object should not be disposed.
+        /// </summary>
+        /// <typeparam name="T">The AWS service interface, like IAmazonS3.</typeparam>
+        /// <param name="collection"></param>
+        /// <param name="options">The AWS options used to create the service client overriding the default AWS options added using AddDefaultAWSOptions.</param>
+        /// <param name="configAction">An action to configure the ClientConfig using services from the IServiceProvider.</param>
+        /// <param name="lifetime">The lifetime of the service client created. The default is Singleton.</param>
+        /// <returns>Returns back the IServiceCollection to continue the fluent system of IServiceCollection.</returns>
+        public static IServiceCollection AddAWSService<T>(this IServiceCollection collection, AWSOptions options, Action<ClientConfig, IServiceProvider> configAction, ServiceLifetime lifetime = ServiceLifetime.Singleton) where T : IAmazonService
+        {
+            Func<IServiceProvider, object> factory =
+                new ClientFactory<T>(options, configAction).CreateServiceClient;
+
+            var descriptor = new ServiceDescriptor(typeof(T), factory, lifetime);
+            collection.Add(descriptor);
+            return collection;
+        }
+
+        /// <summary>
         /// Adds the Amazon service client to the dependency injection framework if the service type hasn't already been registered.
         /// The Amazon service client is not created until it is requested. If the ServiceLifetime property is set to Singleton,
         /// the default, then the same instance will be reused for the lifetime of the process and the object should not be disposed.
@@ -140,7 +176,7 @@ namespace Microsoft.Extensions.DependencyInjection
         /// <returns>Returns back the IServiceCollection to continue the fluent system of IServiceCollection.</returns>
         public static IServiceCollection TryAddAWSService<T>(this IServiceCollection collection, ServiceLifetime lifetime = ServiceLifetime.Singleton) where T : IAmazonService
         {
-            return TryAddAWSService<T>(collection, null, lifetime);
+            return TryAddAWSService<T>(collection, (AWSOptions)null, lifetime);
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


extensions/src/AWSSDK.Extensions.NETCore.Setup/ServiceCollectionExtensions.cs
```diff
@@ -120,9 +120,40 @@ public static IServiceCollection AddAWSService<T>(this IServiceCollection collec
...
+        /// Adds the Amazon service client to the dependency injection framework. The Amazon service client is not
+        /// created until it is requested. If the ServiceLifetime property is set to Singleton, the default, then the same
+        /// instance will be reused for the lifetime of the process and the object should not be disposed.
+        /// </summary>
+        /// <typeparam name="T">The AWS service interface, like IAmazonS3.</typeparam>
+        /// <param name="collection"></param>
+        /// <param name="configAction">An action to configure the ClientConfig using services from the IServiceProvider.</param>
+        /// <param name="lifetime">The lifetime of the service client created. The default is Singleton.</param>
+        /// <returns>Returns back the IServiceCollection to continue the fluent system of IServiceCollection.</returns>
+        public static IServiceCollection AddAWSService<T>(this IServiceCollection collection, Action<ClientConfig, IServiceProvider> configAction, ServiceLifetime lifetime = ServiceLifetime.Singleton) where T : IAmazonService
```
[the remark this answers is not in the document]
peterrsongg: While this is technically valid, it is practically irrelevant. A user could use the parameterless overload without the need to pass in `null` for AWSOptions. In practice this doesn't make sense. However, since this is a possibility I think calling it out in the changelog should suffice unless ppl think otherwise.

sdk/src/Core/Amazon.Runtime/_netstandard/ClientConfig.cs
```diff
@@ -115,10 +115,6 @@ public int? MaxConnectionsPerServer
         /// If null, an HttpClient will be created by the SDK.
         /// Note that IClientConfig members such as ProxyHost, ProxyPort, GetWebProxy, and AllowAutoRedirect
         /// will have no effect unless they're used explicitly by the HttpClientFactory implementation.
-        ///
-        /// See https://docs.microsoft.com/en-us/xamarin/cross-platform/macios/http-stack?context=xamarin/ios and
```
dscpinheiro, line 119: I assume you haven't done a dry-run with the latest commit because changing `Core` does require an entry in the DevConfig (it can be `updateMinimum=false` since it's doc only).

peterrsongg: gotcha, i'll do that thanks

generator/.DevConfigs/37c0ad85-9929-448a-aaeb-7732bcf8ec27.json
```diff
@@ -0,0 +1,13 @@
+{
+  "extensions": [
+    {
+      "extensionName": "Extensions.NETCore.Setup",
+      "type": "patch",
+      "changeLogMessages": [
+        "Allow users to configure HttpClientFactory through DI Factories, by providing an overload of AddAWSService.",
```
normj: You made a generic solution with adding the callback but then the change log is specific to HttpClientFactory which might make users think there is some specific HttpClientFactory like properties being added. What something like

```
Added new callback for the `AddAWSService` and similar methods allowing customization of the service client config before being used to create the service client. For example the callback can be used to configure the `HttpClientFactory` with an implementation coming from the DI using the `IHttpClientFactory` provided by microsoft.
```

generator/.DevConfigs/37c0ad85-9929-448a-aaeb-7732bcf8ec27.json
```diff
@@ -0,0 +1,13 @@
+{
+  "extensions": [
+    {
+      "extensionName": "Extensions.NETCore.Setup",
+      "type": "patch",
```
normj: Since this is a feature bump this to minor.

generator/.DevConfigs/37c0ad85-9929-448a-aaeb-7732bcf8ec27.json
peterrsongg: I was struggling with a good changelog message. I like this one better, will update 👍

normj (dismissed):

generator/.DevConfigs/37c0ad85-9929-448a-aaeb-7732bcf8ec27.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extensions": [
+    {
+      "extensionName": "Extensions.NETCore.Setup",
+      "type": "minor",
+      "changeLogMessages": [
+        "Added new callback for the `AddAWSService` and similar methods allowing customization of the service client config before being used to create the service client. For example the callback can be used to configure the `HttpClientFactory` with an implementation coming from the DI using the `IHttpClientFactory` provided by microsoft.",
```
dscpinheiro:
```suggestion
        "Added new callback for the `AddAWSService` and similar methods allowing customization of the service client config before being used to create the service client. For example the callback can be used to configure the `HttpClientFactory` with an implementation coming from the DI using the `IHttpClientFactory` provided by Microsoft.",
```

dscpinheiro (approved):

normj (approved):

PR #4417, opened 2026-06-01
Fix issue where region was being overriden if AWS_ENDPOINT_URL (or service specific variants) were set.

peterrsongg:
## Description

While working on something else I realized that when `AWS_ENDPOINT_URL` is set and `AWS_REGION` is also set, the `AWS_REGION` is set to null by the getter in `ServiceURL`. This is because the `serviceURL `getter was incorrectly writing to `this.ServiceUrl` instead of the backing field `this.serviceURL`, which was triggering the logic to nullify the regionEndpoint. 

The setter's behavior is unchanged, explicitly setting `config.ServiceURL` will still nullify the regionEndpoint. However, we should consider if we want to change this "mutually exclusive" behavior and be smarter about the behavior when both are set. 

The null region means it was just using the default value of us-east-1.

## Motivation and Context

The `ServiceURL` property getter auto-resolves from environment variables (`AWS_ENDPOINT_URL`, `AWS_ENDPOINT_URL_{SERVICE}`) and profile configuration. When it finds a value, it previously called `this.ServiceURL = value` (the property setter), which unconditionally sets `regionEndpoint = null` and `probeForRegionEndpoint = false` due to the "mutually exclusive" design between RegionEndpoint and ServiceURL.

This meant that any code path reading `config.ServiceURL` after the region was already resolved (e.g.,
EndpointDiscoveryResolver constructor, signers, endpoint resolvers) would permanently destroy the region — even though
the user explicitly configured both via environment variables.

The fix is to write to the backing field and create a new `NormalizeServiceURL` static helper which bypasses the setter altogether while preserving the previous behavior.

### Dry-runs

- **DotNet Dry-run ID:** e72953ee-befa-40ef-b1dd-a2beacfc35ce
  - [ ] Pending
  - [x] Completed successfully
  - [ ] Failed
- **PowerShell Dry-run ID:** e55ceb36-f326-46c4-9900-300b0e625bc8
  - [ ] Pending
  - [x] Completed successfully
  - [ ] Failed

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/ClientConfig.cs
+++ b/sdk/src/Core/Amazon.Runtime/ClientConfig.cs
@@ -339,22 +339,22 @@ namespace Amazon.Runtime
         /// </summary>
         public string ServiceURL
         {
-            get 
+            get
             {
                 if (!didProcessServiceURL && this.serviceURL == null && IgnoreConfiguredEndpointUrls == false && ServiceId != null)
                 {
-                    
+
                     string serviceSpecificTransformedEnvironmentVariable = TransformServiceId.TransformServiceIdToEnvVariable(ServiceId);
                     string transformedConfigServiceId = TransformServiceId.TransformServiceIdToConfigVariable(ServiceId);
 
                     if (Environment.GetEnvironmentVariable(serviceSpecificTransformedEnvironmentVariable) != null)
                     {
                         Logger.GetLogger(GetType()).InfoFormat($"ServiceURL configured from service specific environment variable: {serviceSpecificTransformedEnvironmentVariable}.");
-                        this.ServiceURL = Environment.GetEnvironmentVariable(serviceSpecificTransformedEnvironmentVariable);                    
+                        this.serviceURL = NormalizeServiceURL(Environment.GetEnvironmentVariable(serviceSpecificTransformedEnvironmentVariable));
                     }
                     else if (Environment.GetEnvironmentVariable(EnvironmentVariables.GLOBAL_ENDPOINT_ENVIRONMENT_VARIABLE) != null)
                     {
-                        this.ServiceURL = Environment.GetEnvironmentVariable(EnvironmentVariables.GLOBAL_ENDPOINT_ENVIRONMENT_VARIABLE);
+                        this.serviceURL = NormalizeServiceURL(Environment.GetEnvironmentVariable(EnvironmentVariables.GLOBAL_ENDPOINT_ENVIRONMENT_VARIABLE));
                         Logger.GetLogger(GetType()).InfoFormat($"ServiceURL configured from global environment variable: {EnvironmentVariables.GLOBAL_ENDPOINT_ENVIRONMENT_VARIABLE}.");
                     }
                     else
@@ -378,7 +378,7 @@ namespace Amazon.Runtime
                                 {
                                     Logger.GetLogger(GetType()).InfoFormat($"ServiceURL configured from service specific endpoint url in " +
                                     $"profile {profile.Name} from key {transformedConfigServiceId}.");
-                                    this.ServiceURL = endpointUrlValue;
+                                    this.serviceURL = NormalizeServiceURL(endpointUrlValue);
                                 }
 
                             }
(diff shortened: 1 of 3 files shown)


sdk/src/Core/Amazon.Runtime/ClientConfig.cs
```diff
@@ -403,32 +402,33 @@ public string ServiceURL
...
+            }
+        }
+
+        private static string NormalizeServiceURL(string value)
+        {
+            if(!string.IsNullOrEmpty(value))
+            {
+                try
                 {
-                    // If the URL passed in only has a host name make sure there is an ending "/" to avoid signature mismatch issues.
```
dscpinheiro, line 409: Is it worth keeping this comment? (I noticed sometimes AI tends to remove these but the explanation here seems important to me)

sdk/test/UnitTests/Custom/Runtime/ClientConfigTests.cs
```diff
@@ -114,6 +114,72 @@ public void TestDoesNotThrowForCustomServiceUrl()
...
+            var savedRegion = Environment.GetEnvironmentVariable("AWS_REGION");
+            var savedEndpoint = Environment.GetEnvironmentVariable("AWS_ENDPOINT_URL_S3");
+            try
+            {
+                Environment.SetEnvironmentVariable("AWS_REGION", "eu-west-1");
+                Environment.SetEnvironmentVariable("AWS_ENDPOINT_URL_S3", "http://localhost:9000");
+                FallbackRegionFactory.Reset();
+
+                var config = new TestClientConfig();
+                config.ServiceId = "S3";
```
dscpinheiro, line 155: Curious, is the `ServiceId` needed for the test to pass?

sdk/src/Core/Amazon.Runtime/ClientConfig.cs
peterrsongg: yeah I think this is useful context, I'll add it back!

sdk/test/UnitTests/Custom/Runtime/ClientConfigTests.cs
peterrsongg: nope it isn't needed for this test to pass, I'll remove it

peterrsongg: huh.. i thought it wouldn't but it does... that's weird

peterrsongg: oh it's because in the ServiceURL getter one of the checks is `ServiceId != null`. In a real service, this value would always be populated. Looks like since I can't override it in the TestConfig I have to set it here.

dscpinheiro (dismissed):

sdk/src/Core/Amazon.Runtime/ClientConfig.cs
```diff
@@ -339,22 +339,21 @@ public abstract string RegionEndpointServiceName
...
                 {
-                    
                     string serviceSpecificTransformedEnvironmentVariable = TransformServiceId.TransformServiceIdToEnvVariable(ServiceId);
                     string transformedConfigServiceId = TransformServiceId.TransformServiceIdToConfigVariable(ServiceId);
 
                     if (Environment.GetEnvironmentVariable(serviceSpecificTransformedEnvironmentVariable) != null)
                     {
                         Logger.GetLogger(GetType()).InfoFormat($"ServiceURL configured from service specific environment variable: {serviceSpecificTransformedEnvironmentVariable}.");
-                        this.ServiceURL = Environment.GetEnvironmentVariable(serviceSpecificTransformedEnvironmentVariable);                    
+                        this.serviceURL = NormalizeServiceURL(Environment.GetEnvironmentVariable(serviceSpecificTransformedEnvironmentVariable));
```
normj, line 352: I would recommend adding a comment the purposeful use of the field instead of the property to avoid the region logic.

peterrsongg: good idea, done

dscpinheiro (approved):

normj (approved):

PR #4431, opened 2026-06-12
Fix regression with AmazonUnmarshallingException.LastKnownLocation

dscpinheiro:
After #4422 was merged, Muhammad noticed there was a regression where the `LastKnownLocation` property of the `AmazonUnmarshallingException` (which is thrown if the JSON parsing fails) did not contain the same value as before. I added a test, but the code in `main` today would only return the delimiters (i.e. "////////").

I have two commits on this branch:
- First one is the initial attempt to restore the previous behaviour. It worked (the dry-run succeeded), but it erased the latency wins from the PR (which we're still interested in)
- Second one leverages the fact that the `Utf8JsonReader` already has the property names as UTF-8 bytes, and it uses a single `byte[]` instead of a StringBuilder

## Testing
Ran the relevant benchmarks on my personal laptop (a Mac but ran before and after so that we could compare apples-to-apples):
| Benchmark | pre-PR #4422 | main | proposed fix |
|---|---:|---:|---:|
| GetItemOutput_S | 1,714.9 ns / 7.34 KB | 1,414.5 ns / 6.57 KB | 1,390.4 ns / 6.55 KB |
| GetItemOutput_M | 5,461.3 ns / 13.76 KB | 4,464.8 ns / 11.23 KB | 4,382.0 ns / 11.21 KB |
| GetItemOutput_L | 12,951.4 ns / 25.85 KB | 10,200.5 ns / 19.80 KB | 10,113.3 ns / 19.79 KB |

### Dry-runs
- **DotNet Dry-run ID:** `289cd0f7-dfb5-4297-9727-dc35ad67409f`
  - [X] Completed
- **PowerShell Dry-run ID:** `5fa7fcdb-38de-4fe2-a3e6-4d3c8abff6ef`
  - [X] Completed

N/A

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Internal/Transform/JsonUnmarshallerContext.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Transform/JsonUnmarshallerContext.cs
@@ -46,7 +46,6 @@ namespace Amazon.Runtime.Internal.Transform
     /// </summary>
     public class JsonUnmarshallerContext : UnmarshallerContext
     {
-        private const string DELIMITER = "/";
         #region Private members
 
         private Stream baseStream;
@@ -57,20 +56,17 @@ namespace Amazon.Runtime.Internal.Transform
         private bool disposed = false;
         private bool wasPeeked = false;
 
-        // Property names are no longer pushed onto the path stack (that was the per-property string
-        // allocation we removed). But AWSSDK.Core must stay compatible with service packages that were
-        // generated before reader-based matching existed: those still call the legacy path-based
-        // TestExpression(string, int) overload, which matches against CurrentPath. To keep that working
-        // without re-introducing an allocation on the hot path, we capture the current property name as
-        // raw UTF-8 bytes into a reused, pooled buffer on each Read, and only materialize a string lazily
-        // when CurrentPath is actually read (i.e. by old packages or by exception diagnostics).
+        // Property names are pushed onto the path stack's StringBuilder on each PropertyName token
+        // via a pooled char[] (no string allocation on the hot path). The path string is only
+        // materialized (via StringBuilder.ToString) when CurrentPath is actually read — typically only
+        // on exception for diagnostics, or by the legacy path-based TestExpression(string, int) overload
+        // used by older service packages. We also capture the current property name into a reused pooled
+        // buffer for reader-based TestExpression matching (ValueTextEquals).
         private byte[] currentPropertyNameBuffer = null;
         private int currentPropertyNameLength = 0;
         // Set only when a property name contains JSON escapes, where a raw-byte copy can't be compared
         // against the unescaped expression the legacy overload expects.
         private string currentPropertyNameEscaped = null;
-        private bool onPropertyName = false;
-        private string cachedCurrentPath = null;
         #endregion
 
         #region Constructors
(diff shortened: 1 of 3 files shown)


sdk/src/Core/Amazon.Runtime/Internal/Transform/JsonUnmarshallerContext.cs
```diff
@@ -546,14 +465,20 @@ private enum PathSegmentType
...
-            private Stack<PathSegment> stack = new Stack<PathSegment>();
+            private readonly Stack<PathSegment> stack = new Stack<PathSegment>();
             int currentDepth = 0;
-            private StringBuilder stackStringBuilder = new StringBuilder(128);
+            // The current path as UTF-8 bytes. Delimiters and property names are appended as they
+            // are read and trimmed back off (by adjusting pathLength) as their scopes close. The
+            // path string is only materialized when CurrentPath is read — typically on exception
+            // for diagnostics, or by the legacy path-based TestExpression(string, int) overload
+            // used by service packages generated before reader-based matching existed.
+            private byte[] pathBytes = new byte[256];
```
normj: You could get rent the `byte[]` from the array pool and then return it when the `JsonUnmarshallerContext` is disposed. Essentially make the private class `JsonPathStack` implement `IDisposable` and dispose from `JsonUnmarshallerContext`'s dispose method.

sdk/test/UnitTests/Custom/Marshalling/ErrorUnmarshallingTests.cs
```diff
@@ -172,6 +174,29 @@ public void UnmarshalJsonErrorHtmlProxyResponseDoesNotInfiniteLoop()
             Assert.AreEqual(ErrorType.Unknown, response.Type);
         }
 
+        [TestMethod]
+        [TestCategory("UnitTest")]
+        public void UnmarshalJsonExceptionIncludesDeeplyNestedPropertyInLastKnownLocation()
```
normj: Seems like you also need a test case that includes a JSON escaped character in the property name to test the logic `if (name.IndexOf((byte)'\\') >= 0)` on line 408.

normj (changes requested):

sdk/test/UnitTests/Custom/Marshalling/ErrorUnmarshallingTests.cs
dscpinheiro: Added a new test case in https://github.com/aws/aws-sdk-net/pull/4431/changes/fdd29de3456305726e195af5bd97755470505035

sdk/src/Core/Amazon.Runtime/Internal/Transform/JsonUnmarshallerContext.cs
dscpinheiro: Updated the context to implement `IDisposable` in https://github.com/aws/aws-sdk-net/pull/4431/changes/fdd29de3456305726e195af5bd97755470505035

normj (approved):

muhammad-othman (approved):

dscpinheiro:
New dry-run (`DRY_RUN-2327d6c6-d1e8-4a66-98f9-395d9c5fc447`) hasn't completed yet, so will wait before merging the PR (likely after today's release completes)

PR #4436, opened 2026-06-15
Re-resolve default-chain credentials when the shared credentials file changes

sankettangade:
## Description

When credentials are resolved through the default credential chain, the resolver caches them and only re-checks them when an environment variable changes. It never noticed when the shared credentials file (`~/.aws/credentials`) was changed on disk. So, if an external tool rotated the credentials in that file, the process kept using the old credentials until it was restarted.

This PR change makes the default chain watch the credentials/config file and re-read it when it changes. It reuses the file watcher that already exists for the client-config profile path (`CachedProfileCredentialResolver`), so the check is push-based and adds no per-request file access.

The cache is invalidated on any change to the backing credentials/config file, **regardless of the resolved credential type**. Invalidation is intentionally not gated on the credential type: a profile edit can switch a profile between credential types (e.g. from static keys to `source_profile`/assume-role) or repoint `source_profile` at a different role — changes a self-refreshing credential object (SSO, assume-role, web identity, container, instance profile) cannot detect on its own, because it keeps using the role/values it was constructed with. Re-resolving on any file change is cheap and rare (the check is an in-memory change-token comparison; the actual re-read only happens when the file truly changed), and avoids serving credentials from a stale profile definition.

**What changed**
- Extracted `SharedFileWatcher` into its own internal class (under `Amazon.Util.Internal`) so both the client-config profile path and the default chain can share it.
- `DefaultAWSCredentialsIdentityResolver` now tracks the backing credentials/config files and invalidates its cache on file change, for any credential type resolved via the default chain.
- Added unit tests covering default-chain file rotation (sync + async), no change when the file is unchanged, repeated rotations, and that environment variables still take precedence.

### Performance

Measured on net8.0 Release, cache-hit path (static `[default]` profile, file unchanged), 2,000,000 iterations per loop, best-of-3 timing with warmup. The "before" build is the pre-fix resolver; "after" is this change.

| Flow | Metric | Before | After | Delta |
|---|---|---|---|---|
| Default (`clientConfig = null`) | Time | 474.5 ns/call | 481.8 ns/call | +7.3 ns (~1.5%) |
| Default (`clientConfig = null`) | Memory | 144 B/call | 144 B/call | 0 B |
| Non-default (`clientConfig.Profile` set) | Time | 120.4 ns/call | 107.8 ns/call | no measurable change (noise) |
| Non-default (`clientConfig.Profile` set) | Memory | 144 B/call | 144 B/call | 0 B |

**Notes**
- **Zero added allocation.** The file-change check compares in-memory `long` change tokens and allocates nothing. The 144 B/call appears identically in both builds and is the pre-existing `CancellationTokenSource` allocated by `ResolveIdentity` (35s timeout wrapper) — unrelated to this change.
- **Default flow: +~7 ns/call** from three in-memory token comparisons in `HasCredentialsFileChanged()`. No per-request syscall on the hot path.

 
There is a PowerShell PR that has to be merged along with this PR to fix a failing test case - https://github.com/aws/PRIVATE-aws-tools-for-powershell-staging/pull/679

## Motivation and Context

https://github.com/aws/aws-tools-for-powershell/issues/415 - Though this issue was reported for PS session variable caching, .NET SDK bug was found out while testing.

## Testing

- New unit tests in `DefaultCredentialsResolverTests` for the default-chain
  (`clientConfig: null`) rotation cases.
- Verified in .NET with a standalone .NET 8 console app that references the built `AWSSDK.Core` and calls the static `DefaultAWSCredentialsIdentityResolver.GetCredentials()` — the process-wide singleton
  path used by real applications and service clients. 
  - After Change: credentials file rotated mid-process is picked up (AKID_V1 -> AKID_V2), 3/3 PASS.
  - Before Change: returns stale credentials (AKID_V1 -> AKID_V1), 3/3 FAIL.
- Verified end-to-end through the PowerShell module: credentials rotated mid session in credentials file are being refreshed and cached now.

### Dry-runs

- **DotNet Dry-run ID:** `d0851977-048a-4d83-8bd5-3bf7b24a5f8f`
  - [ ] Pending
  - [x] Completed successfully
  - [ ] Failed
- **PowerShell Dry-run ID:** `88107504-fa2d-40da-bbf2-0f8e6d5803e2`
  - [ ] Pending
  - [x] Completed successfully
  - [ ] Failed

Not a breaking change. Behavior changes only for static, file-backed credentials
resolved via the default chain, and only when the file is actually changed on disk. All other credential types and code paths behave exactly as before. 

## Types of changes

## Checklist

## License

### DIFF
--- a/generator/.DevConfigs/4fc0120c-965d-42cc-a1e2-1ab9427e6525.json
+++ b/generator/.DevConfigs/4fc0120c-965d-42cc-a1e2-1ab9427e6525.json
@@ -3,7 +3,7 @@
     "updateMinimum": true,
     "type": "patch",
     "changeLogMessages": [
-      "Invalidate the DefaultAWSCredentialsIdentityResolver default-chain credential cache when the shared credentials file is modified, so static credentials (BasicAWSCredentials/SessionAWSCredentials) rotated by an external tool are picked up mid-process without requiring a restart. "
+      "Invalidate the DefaultAWSCredentialsIdentityResolver default-chain credential cache when the shared credentials file is modified, so credentials rotated by an external tool are picked up mid-process without requiring a restart."
     ]
   }
 }
--- a/generator/.DevConfigs/4fc0120c-965d-42cc-a1e2-1ab9427e6525.json
+++ b/generator/.DevConfigs/4fc0120c-965d-42cc-a1e2-1ab9427e6525.json
@@ -1,0 +1,9 @@
+{
+  "core": {
+    "updateMinimum": true,
+    "type": "patch",
+    "changeLogMessages": [
+      "Invalidate the DefaultAWSCredentialsIdentityResolver default-chain credential cache when the shared credentials file is modified, so static credentials (BasicAWSCredentials/SessionAWSCredentials) rotated by an external tool are picked up mid-process without requiring a restart. "
+    ]
+  }
+}
--- a/sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
+++ b/sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
@@ -24,6 +24,7 @@ using System.Threading.Tasks;
 using Amazon.Runtime.CredentialManagement;
 using Amazon.Runtime.Identity;
 using Amazon.Runtime.Internal.Util;
+using Amazon.Util.Internal;
 
 namespace Amazon.Runtime.Credentials
 {
@@ -55,7 +56,6 @@ namespace Amazon.Runtime.Credentials
         private long _lastSeenCredentialsFileToken;
         private long _lastSeenConfigFileToken;
         private long _lastSeenNetSdkFileToken;
-        private volatile bool _cachedCredentialsAreFileBacked;
 
         public DefaultAWSCredentialsIdentityResolver()
         {
@@ -160,19 +160,19 @@ namespace Amazon.Runtime.Credentials
         }
 
         /// <summary>
-        /// Returns true if the cached credentials were resolved from a file (a static
-        /// BasicAWSCredentials / SessionAWSCredentials profile) and any of the backing files have
+        /// Returns true if any of the credentials/config files backing the cached credentials have
         /// changed since they were resolved. Purely an in-memory change-token comparison, no I/O on
         /// this path, so it is safe to call on every credential resolution.
         /// <para />
-        /// Returns false for credential types that refresh themselves (SSO, assume-role, web identity,
-        /// container, instance profile), so a file touch never forces a needless re-resolution of those.
+        /// Invalidation is intentionally not gated on the resolved credential type. Profile edits can
+        /// switch a profile between credential types (e.g., from static keys to <c>source_profile</c>/
+        /// assume-role, or repoint <c>source_profile</c> at a different role) — changes that a
+        /// self-refreshing credential object cannot detect on its own because it keeps using the
+        /// role/values it was constructed with. Re-resolving on any file change is cheap and rare and
+        /// avoids serving credentials from a stale profile definition.
         /// </summary>
         private bool HasCredentialsFileChanged()
         {
-            if (!_cachedCredentialsAreFileBacked)
-                return false;
-
             if (_credentialsFileWatcher != null && _credentialsFileWatcher.ChangeToken != _lastSeenCredentialsFileToken)
                 return true;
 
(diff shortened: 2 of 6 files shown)


sdk/src/Core/Amazon.Util/Internal/SharedFileWatcher.cs
```diff
@@ -0,0 +1,93 @@
...
+    /// When a change is detected, a monotonic <see cref="ChangeToken"/> is incremented. Each
+    /// consumer stores the token value it last observed; staleness is determined by comparing the
+    /// stored token to the current token rather than resetting shared state. This avoids cross-entry
+    /// interference when multiple consumers depend on the same file.
+    /// <para />
+    /// This is used both by <see cref="CachedProfileCredentialResolver"/> (for credentials resolved
+    /// from a profile set on the client config) and by <see cref="DefaultAWSCredentialsIdentityResolver"/>
+    /// (for static credentials resolved from a file via the default credential chain).
+    /// </summary>
+    internal sealed class SharedFileWatcher : IDisposable
```
peterrsongg, line 36: Wondering if a better place to put this is in https://github.com/aws/aws-sdk-net/tree/main/sdk/src/Core/Amazon.Util/Internal

Even though you made the class internal, it is still in the `Amazon.Runtime.Credentials` namespace along with the other credential providers. This class isn't really a credential provider or a credential POCO, and could technically be used for anything.

generator/.DevConfigs/4fc0120c-965d-42cc-a1e2-1ab9427e6525.json
```diff
@@ -0,0 +1,9 @@
+{
```
peterrsongg, line 1: can you provide some performance numbers from before and after of the default code path? And one of the non-default too just in case

sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
```diff
@@ -142,14 +151,90 @@ private AWSCredentials TryGetCachedCredentials()
...
+        /// type manages its own refresh, so file-change invalidation is intentionally skipped for them.
+        /// <para />
+        /// Must only be called while holding <see cref="_credentialResolutionLock"/>, before
+        /// <see cref="_cachedCredentials"/> is published, so that a concurrent lock-free reader that
+        /// observes the new credentials also observes the tracking state captured here.
+        /// </summary>
+        private void UpdateFileChangeTracking(AWSCredentials resolvedCredentials)
+        {
+            var type = resolvedCredentials.GetType();
+            var isFileBacked = type == typeof(BasicAWSCredentials) || type == typeof(SessionAWSCredentials);
```
peterrsongg: Do we need this type-checked logic at all? If a user were to do something like
```
[default]
source_profile = a
[profile a]
role_arn = "foo"
region="us-east-1"
```

then mid session, the profile is updated for whatever reason:
```
[default]
source_profile=b
[profile b]
role_arn = "bar"
region="us-west-2"
```
wouldn't the credentials from profile a beused with this guard when it should've picked up credentials from profile b?

If so, can we just remove the volatile bool and this logic here? At the worst case the file has changed and a file will be re-read, but that seems like a better alternative than picking up the wrong credentials.

peterrsongg (commented):
I have one concern about type-checking the type of credentials as it seems like it would miss some valid scenarios that I listed in the comments. I feel like we should just compare tokens unconditionally. If the file has changed then we must invalidate the cache regardless of the type of credential that was resolved. 

Here is the test case:
```
        [TestMethod]
        public void DefaultChain_AssumeRoleProfile_FileChanged_ReResolvesCredentials()
        {
            ClearChainEnvironmentVariables();

            var initialContents =
                "[default]\n" +
                "role_arn = arn:aws:iam::111111111111:role/RoleA\n" +
                "source_profile = source_a\n" +
                "[source_a]\n" +
                "aws_access_key_id = AKID_SOURCE_A\n" +
                "aws_secret_access_key = SECRET_SOURCE_A\n";

            var updatedContents =
                "[default]\n" +
                "role_arn = arn:aws:iam::222222222222:role/RoleB\n" +
                "source_profile = source_b\n" +
                "[source_b]\n" +
                "aws_access_key_id = AKID_SOURCE_B\n" +
                "aws_secret_access_key = SECRET_SOURCE_B\n";

            using (var fixture = new SharedCredentialsFileTestFixture(initialContents))
            using (var resolver = new DefaultAWSCredentialsIdentityResolver())
            {
                var previousLocation = RedirectDefaultChainToFixture(fixture);
                try
                {
                    var original = resolver.ResolveIdentity(clientConfig: null);
                    Assert.IsInstanceOfType(original, typeof(AssumeRoleAWSCredentials),
                        "Default profile with role_arn + source_profile should resolve to AssumeRoleAWSCredentials.");

                    var originalAssumeRole = (AssumeRoleAWSCredentials)original;
                    Assert.AreEqual("arn:aws:iam::111111111111:role/RoleA", originalAssumeRole.RoleArn);

                    fixture.UpdateCredentialsFile(updatedContents);

                    var updated = resolver.ResolveIdentity(clientConfig: null);
                    Assert.IsInstanceOfType(updated, typeof(AssumeRoleAWSCredentials),
                        "Updated profile should still resolve to AssumeRoleAWSCredentials.");
                    Assert.AreNotSame(original, updated,
                        "Cache should have been invalidated after file change, even for non-Basic/non-Session credential types.");

                    var updatedAssumeRole = (AssumeRoleAWSCredentials)updated;
                    Assert.AreEqual("arn:aws:iam::222222222222:role/RoleB", updatedAssumeRole.RoleArn,
                        "Should have picked up the new role_arn from the updated file.");
                }
                finally
                {
                    AWSConfigs.AWSProfilesLocation = previousLocation;
                }
            }
        }
```

sdk/src/Core/Amazon.Util/Internal/SharedFileWatcher.cs
sankettangade: Moved it to `Amazon.Util/Internal`

sdk/src/Core/Amazon.Runtime/Credentials/DefaultAWSCredentialsIdentityResolver.cs
sankettangade: Agree. It is safer to not have this type-check. Added the changes in the latest commit.

generator/.DevConfigs/4fc0120c-965d-42cc-a1e2-1ab9427e6525.json
sankettangade: I have added the performance numbers of before and after for both default and non-default flows in PR description.

peterrsongg (approved):

muhammad-othman (approved):

dscpinheiro:
@sankettangade Can this be merged or are you waiting for the PR on the PS repo to be approved first?

sankettangade:
> @sankettangade Can this be merged or are you waiting for the PR on the PS repo to be approved first?

Merged it. Was waiting on the PS PR that had to go with this.

Issue #1605, opened 2020-05-07
DynamoDbContext.SaveAsync Should Support Conditional Update Expressions


jopincar:
## Expected Behavior

Change the QueryFilter element of DynamoDBOperationConfig to just Filter or add a SaveFilter and allow specification of conditional update expressions.  The fact that QueryFilter works with ConditionalOperator (why no Query in front of that?) is odd and I had the expectation of being able to supply an update expression.

## Current Behavior
Poorly named members promote confusion about the ability to perform conditional updates.

## Context
I am unable to implement conditional update logic using DynamoDbContext.SaveAsync which is a great inconvenience.  It prevents me from implementing this check without doing an extra read and introducing a hole in the check due to latency and timing.  My alternative is to introduce the lower level APIs but then, what's the point of the higher level.  It's also inconsistent that query supports expressions but save does not.

I was wanting to use an existing date field as the "version" to implement optimistic locking.

## .NET Core Info
* .NET Core version used for development:  2.1
* .NET Core version installed in the environment where application runs: 2.1

[logging]: http://docs.aws.amazon.com/sdk-for-net/v3/developer-guide/net-dg-config-other.html

jgarciaa00:
I have the same issue, query supports expressions but save does not. Can do it?

Dreamescaper:
Please reopen this issue, it is still actual.

ashishdhingra:
Mode information about [Condition Expressions](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.ConditionExpressions.html). 

@jopincar Are your using Document model or Object Persistence model? Following [Example: CRUD Operations Using the AWS SDK for .NET Document Model](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/ItemCRUDDotNetDocumentAPI.html), document model appears to support conditional update (refer `UpdateItemOperationConfig`). Could you please share sample code with the current and expected behavior?

mncoldsmoke:
Please reopen this issue. It has NOT been resolved. It is still a bug where in the object model QueryFilters do not work on saves. This has been clear as to what was asked for the ENTIRE thread, and asking people what model they are using is an irresponsible way to do support when it was clearly stated to be an issue with DynamoDbContext.

mncoldsmoke:
Thank you Alex Shovlin
________________________________

Hecedu:
Any updates on this? This would be an extremely useful feature.

[...]

PR #4439, opened 2026-06-18
DynamoDbContext.SaveAsync Should Support Conditional Update Expressions

aanton-git:
## Description
Change the QueryFilter element of DynamoDBOperationConfig to just Filter or add a SaveFilter and allow specification of conditional update expressions. The fact that QueryFilter works with ConditionalOperator (why no Query in front of that?) is odd and I had the expectation of being able to supply an update expression.

## Motivation and Context
I am unable to implement conditional update logic using DynamoDbContext.SaveAsync which is a great inconvenience. It prevents me from implementing this check without doing an extra read and introducing a hole in the check due to latency and timing. My alternative is to introduce the lower level APIs but then, what's the point of the higher level. It's also inconsistent that query supports expressions but save does not.

I was wanting to use an existing date field as the "version" to implement optimistic locking.

Initial issue : https://github.com/aws/aws-sdk-net/issues/1605

## Types of changes

## Checklist

## License

### DIFF
--- a/generator/.DevConfigs/b2626677-dca2-4ade-9565-4ae38b4d4eff.json
+++ b/generator/.DevConfigs/b2626677-dca2-4ade-9565-4ae38b4d4eff.json
@@ -1,0 +1,11 @@
+{
+  "services": [
+    {
+      "serviceName": "DynamoDBv2",
+      "type": "patch",
+      "changeLogMessages": [
+        "DynamoDbContext.SaveAsync Should Support Conditional Update Expressions"
+      ]
+    }
+  ]
+}
--- a/generator/.DevConfigs/b2626677-dca2-4ade-9565-4ae38b4d4eff.json
+++ b/generator/.DevConfigs/b2626677-dca2-4ade-9565-4ae38b4d4eff.json
@@ -2,9 +2,9 @@
   "services": [
     {
       "serviceName": "DynamoDBv2",
-      "type": "patch",
+      "type": "minor",
       "changeLogMessages": [
-        "DynamoDbContext.SaveAsync Should Support Conditional Update Expressions"
+        "Added `SaveConfig.ConditionalExpression` to support conditional expressions on `DynamoDBContext.SaveAsync` and fixed expression evaluation for nullable-typed constants (e.g. comparisons against int?) in `DynamoDBContext` filter/condition expressions"
       ]
     }
   ]
--- a/sdk/src/Services/DynamoDBv2/Custom/DataModel/Context.cs
+++ b/sdk/src/Services/DynamoDBv2/Custom/DataModel/Context.cs
@@ -412,18 +412,36 @@ namespace Amazon.DynamoDBv2.DataModel
             ApplyPostUpdate(storage, value, flatConfig, updateDocument, updateConfig);
         }
 
-        private Expression ComposeConditionalExpression(Expression inputConditionalExpression, Expression versionExpression)
+        private Expression ComposeConditionalExpression(ItemStorage storage, DynamoDBFlatConfig flatConfig, Table table)
         {
-            Expression conditionalExpression = new Expression();
+            Expression expression = null;
 
-            if (versionExpression.IsSet && !inputConditionalExpression.IsSet)
-                return versionExpression;
-            if(inputConditionalExpression.IsSet && !versionExpression.IsSet)
-                return inputConditionalExpression;
+            bool hasVersionCheck = !(flatConfig.SkipVersionCheck.HasValue && flatConfig.SkipVersionCheck.Value) && storage.Config.HasVersion;
+            if (flatConfig.FilterExpression is not { Filter: null })
+            {
+                if (hasVersionCheck)
+                {
+                    throw new InvalidOperationException("QueryFilter is not supported with version check. Use SkipVersionCheck to skip version check and use QueryFilter");
+                }
 
+                if (flatConfig.QueryFilter != null && flatConfig.QueryFilter.Count != 0)
+                {
+                    throw new InvalidOperationException("QueryFilter is not supported with filter expression. Use either QueryFilter or filter expression, but not both.");
+                }
 
+                expression = ComposeExpression(flatConfig.FilterExpression.Filter, storage.Config, flatConfig);
+            }
+            else
+            {
+                if (hasVersionCheck)
+                {
+                    var conversionConfig = new DynamoDBEntry.AttributeConversionConfig(table.Conversion, table.IsEmptyStringValueEnabled);
+                    expression = CreateConditionExpressionForVersion(storage, conversionConfig);
+                    SetNewVersion(storage);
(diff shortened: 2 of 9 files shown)


aanton-git:
@dscpinheiro Copilot comments were fixed

sdk/test/Services/DynamoDBv2/UnitTests/Custom/DataModel/ContextInternalTests.cs
dscpinheiro, line 1702: This test fails on the .NET Framework for me:
```
 Save_WithConditionalExpression_ThrowsIfVersionCheckPresent
   Source: ContextInternalTests.cs line 1702
   Duration: 174 ms

  Message: 
Assert.IsTrue failed. 

  Stack Trace: 
ContextInternalTests.Save_WithConditionalExpression_ThrowsIfVersionCheckPresent() line 1728
```

---

Another thing on this test is the verifiable setup seems unnecessary? The code will throw before `UpdateItem`, so could we verify it's never called?

sdk/src/Services/DynamoDBv2/Custom/DataModel/SaveConfig.cs
```diff
@@ -35,12 +35,22 @@ public class SaveConfig : BaseOperationConfig
...
         public bool? IgnoreNullValues { get; set; }
 
+        /// <summary>
+        /// Represents a conditional expression applied to the underlying Save (UpdateItem) request.
+        /// </summary>
+        /// <remarks>
+        /// Conditional expressions may target key properties and non-key properties.
+        /// When used with versioned entities, <see cref="SkipVersionCheck"/> must be enabled.
+        /// </remarks>
+        public ContextExpression FilterExpression { get; set; }
```
dscpinheiro: Since this is a new property, would it make sense to use `ConditionalExpression` instead? (the original feature request did mention naming confusion as an issue)

generator/.DevConfigs/b2626677-dca2-4ade-9565-4ae38b4d4eff.json
```diff
@@ -0,0 +1,11 @@
+{
+  "services": [
+    {
+      "serviceName": "DynamoDBv2",
+      "type": "patch",
+      "changeLogMessages": [
+        "DynamoDbContext.SaveAsync Should Support Conditional Update Expressions"
```
dscpinheiro: What do you think of this as a changelog message?
```
Added `SaveConfig.ConditionalExpression` to support conditional expressions on `DynamoDBContext.SaveAsync`
```

Also, I think this is worth a `minor` version bump.

sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextExpression.cs
```diff
@@ -150,7 +150,10 @@ private static object EvaluateExpression(Expression expr)
 
                 case UnaryExpression u when u.NodeType == ExpressionType.Convert:
                     var operand = EvaluateExpression(u.Operand);
-                    return Convert.ChangeType(operand, u.Type);
+                    // Convert.ChangeType does not support Nullable<T>; use the underlying type
+                    // to correctly handle expressions such as p => p.Version == (int?)3.
+                    var type = Nullable.GetUnderlyingType(u.Type) ?? u.Type;
```
dscpinheiro, line 155: Was this an existing bug? If yes, worth calling out in the changelog too. For example:
```
Fixed expression evaluation for nullable-typed constants (e.g. comparisons against int?) in `DynamoDBContext` filter/condition expressions
```

sdk/src/Services/DynamoDBv2/Custom/DataModel/SaveConfig.cs
aanton-git: Renamed it to ConditionalExpression

generator/.DevConfigs/b2626677-dca2-4ade-9565-4ae38b4d4eff.json
aanton-git: I've updated the changelog message and type, also included expression evaluation for nullable-typed constants in message

sdk/src/Services/DynamoDBv2/Custom/DataModel/ContextExpression.cs
aanton-git: Updated

sdk/test/Services/DynamoDBv2/UnitTests/Custom/DataModel/ContextInternalTests.cs
aanton-git: I've updated the test to check that UpdateItem is never called. It should pass now

dscpinheiro:
I pushed a small commit (removing some hard-coded values I've seen cause flaky tests) and queued a dry-run (`e584846f-d880-4c76-964a-c838f942324e`)

dscpinheiro (dismissed):

sdk/src/Services/DynamoDBv2/Custom/DataModel/Context.cs
```diff
@@ -412,12 +412,44 @@ private async Task SaveHelperAsync([DynamicallyAccessedMembers(InternalConstants
...
+        private Expression ComposeConditionalExpression(ItemStorage storage, DynamoDBFlatConfig flatConfig, Table table)
+        {
+            Expression expression = null;
+
+            bool hasVersionCheck = !(flatConfig.SkipVersionCheck.HasValue && flatConfig.SkipVersionCheck.Value) && storage.Config.HasVersion;
+            if (flatConfig.FilterExpression is not { Filter: null })
+            {
+                if (hasVersionCheck)
+                {
+                    throw new InvalidOperationException("FilterExpression is not supported with version check. Set SkipVersionCheck to true to use FilterExpression with versioned entities.");
```
muhammad-othman: nit: Should we use `ConditionalExpression` in the exception message rather than `FilterExpression` ? I think this exception will be thrown when users set `ConditionalExpression` on save then get an exception complaining about `FilterExpression` which might be confusing, Or Am I missing something?

muhammad-othman (dismissed):

sdk/src/Services/DynamoDBv2/Custom/DataModel/Context.cs
dscpinheiro: I think this was a miss from the rename, and since the method is new it'd only be called with a `ConditionalExpression` (so should be updated - same for the test that asserts the exception message)

aanton-git: @dscpinheiro  @muhammad-othman Thank you. I've updated the exception message and unit test

dscpinheiro (approved):

AlexDaines (approved):

Issue #4425, opened 2026-06-09
[BedrockChatClient] `ResponseFormat` cannot be combined with `Tools` (and is unsupported for streaming) — both are avoidable now that Bedrock has native structured outputs


anton-prepared:
**Package:** `AWSSDK.Extensions.Bedrock.MEAI`
**Type:** Bug / Feature request (adopt native structured outputs)

> Versions / environment:
> - `AWSSDK.Extensions.Bedrock.MEAI` version: `4.0.5.7`
> - `AWSSDK.BedrockRuntime` version: `4.0.16`
> - `Microsoft.Extensions.AI` version: `10.4.1`
> - Target framework: `net10.0`
> - Model tested: `us.anthropic.claude-sonnet-4-5-20250929-v1:0`

## Summary

`BedrockChatClient` implements `ChatOptions.ResponseFormat` (structured output) by injecting a synthetic forced tool. Because Bedrock's `ToolChoice` can only force one tool at a time, the adapter throws when a caller supplies both `ResponseFormat` **and** their own `Tools`, and it also refuses `ResponseFormat` entirely for streaming.

Both restrictions were correct before February 2026, when forcing a tool was the only way to get schema-constrained output through the Converse API. They are no longer necessary. Bedrock shipped **native structured outputs** (GA 2026-02-04) via a dedicated `outputConfig.textFormat` request field that is *orthogonal* to `toolConfig`, and the corresponding types already exist in the generated `AWSSDK.BedrockRuntime` model. The adapter just hasn't adopted them yet.

## Current behaviour

In `AddResponseFormatTool` (`BedrockChatClient.cs`), `ResponseFormat` is mapped to a synthetic tool named `generate_response`, and `ToolConfiguration.ToolChoice` is set to force that tool. When the user has also provided tools, the method throws:

```
ArgumentException:
"ResponseFormat cannot be used with Tools in Amazon Bedrock.
 ResponseFormat uses Bedrock's tool mechanism for structured output,
 which conflicts with user-provided tools."
```

The accompanying comment states *"This is a Bedrock API constraint, not an SDK limitation."* That was accurate when written, but as of the native structured outputs release it is no longer a platform constraint — it is now specific to the synthetic-tool strategy this adapter uses.

[...]

PR #4441, opened 2026-06-19
Add native structured outputs support to Bedrock MEAI BedrockChatClient

anton-prepared:
## Description

Adopts Bedrock **native structured outputs** (`outputConfig.textFormat`) for `ChatOptions.ResponseFormat` in the `BedrockChatClient` MEAI adapter, addressing #4425.

Previously, `ResponseFormat` was implemented by injecting a synthetic forced tool (`generate_response`) and forcing it via `toolConfig.toolChoice`. Because Bedrock's `ToolChoice` can only force one tool at a time, the adapter:

- threw `ArgumentException` when a caller supplied both `ResponseFormat` **and** their own `Tools`, and
- threw `NotSupportedException` for `ResponseFormat` on the streaming path.

Both restrictions were correct before native structured outputs shipped. Now that Bedrock exposes `outputConfig.textFormat` as a top-level request field that is orthogonal to `toolConfig` (and the generated `Amazon.BedrockRuntime.Model` types already exist), the adapter can map `ResponseFormat` directly and lift both restrictions.

## What changed

- **New `BedrockStructuredOutputMode` enum** (`Native` = default, `SyntheticTool`). Native uses `outputConfig.textFormat`; SyntheticTool keeps the legacy forced-tool strategy for models without native support (Claude 3.x, Titan, Llama, Cohere, AI21, ...).
- **New `AsIChatClient` overload** that accepts the mode. The existing `AsIChatClient(runtime, defaultModelId)` signature is **left unchanged** (the new overload's mode parameter has no default), so this is **source- and binary-compatible** — existing callers are unaffected.
- **Request building:** in `Native` mode, `ResponseFormat` maps to `OutputConfig.TextFormat` (`Type = json_schema`, `JsonSchema.Schema` = the caller's schema serialized to a JSON string, `Name`/`Description` from `SchemaName`/`SchemaDescription`, with the schema's own `title`/`name` used as a name fallback). The synthetic tool, the tool-output extraction, and the streaming guard are all gated to `SyntheticTool` mode only. A caller-supplied `OutputConfig` (via `RawRepresentationFactory`) is respected.
- **Behavior:** `Native` mode composes with user-provided `Tools` and supports streaming — the constrained JSON returns as ordinary text content (and as text deltas while streaming).

## Notes / design

- **Model coverage.** Native structured outputs is supported on newer models (e.g. Claude 4.5+ and a number of open-weight models) and not on older ones. Rather than break older models, the legacy synthetic-tool path is retained and selectable via `BedrockStructuredOutputMode.SyntheticTool`.
- **Schema validation.** The caller's schema is forwarded verbatim; schemas outside Bedrock's supported JSON Schema subset produce a `ValidationException` from the service, surfaced to the caller (matches other SDKs).
- `ToolSpecification.Strict` (enforcing tool *input* schemas) is intentionally out of scope here and can be a separate follow-up.

## Tests

- Existing `ResponseFormat_Json_*` tests repointed to `SyntheticTool` mode (their behavior is unchanged in that mode).
- New native-mode tests: `OutputConfig` shape (not a synthetic tool), `Tools` + `ResponseFormat` composing without throwing, text-response parsing, and streaming without throwing.
- Full `BedrockMEAITests` suite passes on net8.0 (232) and net472 (170); the project builds warning-clean (`TreatWarningsAsErrors`).

Relates to #4425

### DIFF
--- a/extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
+++ b/extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
@@ -268,7 +268,7 @@ internal sealed partial class BedrockChatClient : IChatClient
     /// <remarks>
     /// <para>
     /// <see cref="ChatOptions.ResponseFormat"/> is supported while streaming only in
-    /// <see cref="BedrockStructuredOutputMode.Native"/> mode (the default), where the constrained JSON
+    /// <see cref="BedrockStructuredOutputMode.Native"/> mode, where the constrained JSON
     /// streams back as ordinary text deltas. In <see cref="BedrockStructuredOutputMode.SyntheticTool"/>
     /// mode a JSON <see cref="ChatOptions.ResponseFormat"/> throws <see cref="NotSupportedException"/>;
     /// use <see cref="GetResponseAsync"/> for that strategy.
@@ -298,7 +298,7 @@ internal sealed partial class BedrockChatClient : IChatClient
             throw new NotSupportedException(
                 "ResponseFormat is not supported for streaming responses when using " +
                 "BedrockStructuredOutputMode.SyntheticTool. Use BedrockStructuredOutputMode.Native " +
-                "(the default) for streaming structured output, or GetResponseAsync for the synthetic-tool strategy.");
+                "for streaming structured output, or GetResponseAsync for the synthetic-tool strategy.");
         }
 
         ConverseStreamRequest request = options?.RawRepresentationFactory?.Invoke(this) as ConverseStreamRequest ?? new();
--- a/extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
+++ b/extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
@@ -40,6 +40,12 @@ internal sealed partial class BedrockChatClient : IChatClient
     private const string ResponseFormatToolName = "generate_response";
     /// <summary>The description used for the synthetic tool that enforces response format.</summary>
     private const string ResponseFormatToolDescription = "Generate response in specified format";
+    /// <summary>
+    /// Permissive schema used for native structured outputs when the caller requests JSON output without
+    /// supplying a schema (<c>ChatResponseFormat.Json</c>). Mirrors the synthetic-tool path's behavior so
+    /// that schema-less JSON mode keeps working, and satisfies the required <c>jsonSchema.schema</c> field.
+    /// </summary>
+    private const string DefaultJsonSchema = "{\"type\":\"object\"}";
 
     /// <summary>The wrapped <see cref="IAmazonBedrockRuntime"/> instance.</summary>
     private readonly IAmazonBedrockRuntime _runtime;
@@ -57,7 +63,7 @@ internal sealed partial class BedrockChatClient : IChatClient
     /// <param name="defaultModelId">Model ID to use as the default when no model ID is specified in a request.</param>
     /// <param name="structuredOutputMode">How <see cref="ChatOptions.ResponseFormat"/> is realized against the Converse API.</param>
     public BedrockChatClient(IAmazonBedrockRuntime runtime, string? defaultModelId,
-        BedrockStructuredOutputMode structuredOutputMode = BedrockStructuredOutputMode.Native)
+        BedrockStructuredOutputMode structuredOutputMode = BedrockStructuredOutputMode.SyntheticTool)
     {
         Debug.Assert(runtime is not null);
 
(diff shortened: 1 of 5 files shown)


extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
```diff
@@ -53,12 +55,15 @@ internal sealed partial class BedrockChatClient : IChatClient
     /// </summary>
     /// <param name="runtime">The <see cref="IAmazonBedrockRuntime"/> instance to wrap.</param>
     /// <param name="defaultModelId">Model ID to use as the default when no model ID is specified in a request.</param>
-    public BedrockChatClient(IAmazonBedrockRuntime runtime, string? defaultModelId)
+    /// <param name="structuredOutputMode">How <see cref="ChatOptions.ResponseFormat"/> is realized against the Converse API.</param>
+    public BedrockChatClient(IAmazonBedrockRuntime runtime, string? defaultModelId,
+        BedrockStructuredOutputMode structuredOutputMode = BedrockStructuredOutputMode.Native)
```
dscpinheiro: Could `= BedrockStructuredOutputMode.Native` cause a customer using an older (but still supported) model to fail? (since there'd be an `OutputConfig` in the request to Bedrock)

If yes, I don't think it can be the default because of the runtime break (i.e. you'd have to opt-in to `Native` but this value would need to be `SyntheticTool`).

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
```diff
@@ -920,6 +955,75 @@ private static Document ToDocument(JsonElement json)
...
+        {
+            TextFormat = new OutputFormat
+            {
+                Type = OutputFormatType.Json_schema,
+                Structure = new OutputFormatStructure
+                {
+                    JsonSchema = new JsonSchemaDefinition
+                    {
+                        // JsonElement -> raw JSON string; the field is a string on the wire.
+                        Schema = jsonFormat.Schema?.GetRawText(),
```
dscpinheiro: If the JSON doesn't have a schema should this be `{"type":"object"}`? That's what the previous behavior did: https://github.com/aws/aws-sdk-net/blob/07815c241bb0024ebca00512e7b985b3d58608ba/extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs#L1014-L1021

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
```diff
@@ -920,6 +955,75 @@ private static Document ToDocument(JsonElement json)
...
+                        Name = jsonFormat.SchemaName ?? TryGetSchemaTitle(jsonFormat.Schema) ?? "output_schema",
+                        Description = jsonFormat.SchemaDescription,
+                    }
```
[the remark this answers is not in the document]
dscpinheiro: I think this is the same as my comment above.

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
anton-prepared: Good call — you're right that defaulting to `Native` is a runtime break, since it adds `outputConfig` to every `ResponseFormat` request and older models (Claude 3.x, Titan, Llama, etc.) would start rejecting them. I've flipped the default to `SyntheticTool` so existing callers keep the exact same request shape, and `Native` is now strictly opt-in. I also reordered the enum so `SyntheticTool = 0` (the default value) to keep `default(BedrockStructuredOutputMode)` consistent with the constructor default. Done in 399dc0771c70.

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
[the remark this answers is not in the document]
anton-prepared: Agreed — with `ChatResponseFormat.Json` (no schema), `jsonFormat.Schema` is null and `jsonSchema.schema` is required by the model, so the request would fail. Native mode now falls back to a permissive `{type:object}` schema in that case, matching what the synthetic-tool path did. Added a regression test (`ResponseFormat_Json_Native_NoSchema_UsesDefaultObjectSchema`). Fixed in 399dc0771c70.

dscpinheiro (dismissed):
Internal build (`7e2ae24a-6e23-44d8-bbc6-3ff5286a9089`) succeeded.

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
```diff
@@ -239,6 +267,13 @@ public async Task<ChatResponse> GetResponseAsync(
     /// <inheritdoc />
     /// <remarks>
     /// <para>
+    /// <see cref="ChatOptions.ResponseFormat"/> is supported while streaming only in
+    /// <see cref="BedrockStructuredOutputMode.Native"/> mode (the default), where the constrained JSON
```
peterrsongg: this comment is out of date now since `Native` isn't the default anymore. Can you update the docs so users aren't confused?

peterrsongg (commented):
@anton-prepared  thanks for the PR. looks good but just one comment is out of date can you update that and I'll hit approve?

dscpinheiro (approved):

extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
dscpinheiro: I updated it myself in https://github.com/aws/aws-sdk-net/pull/4441/commits/327b79adc38091c11c63f3536a04a3264d650e3f

peterrsongg (approved):

Issue #3722, opened 2025-03-25
AWSSDK.DynamoDBv2 does not detect self-references in PutItem request


OleksandrKrutykh:
### Describe the bug

If the item passed into `Amazon.DynamoDBv2.AmazonDynamoDBClient.PutItemAsync` method contains an `AttributeValue` with a self-reference, the method call results in a StackOverflowException. Since this exception cannot be caught, I assume it will crash whatever runtime environment the code is running in. 

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

The call results in AmazonDynamoDBException or one of its inheritors.

### Current Behavior

A StackOverflowException is thrown. Please, see the stack trace in `ExceptionStack.txt` attached. Note that the stack trace has been shortened for readability, because it contained a lot of identical `AttributeValueMarshaller.Marshall` calls. 

[ExceptionStack.txt](https://github.com/user-attachments/files/19446117/ExceptionStack.txt)

### Reproduction Steps

Assuming that we have a DynamoDB table with a numeric hash key attribute called `Id`:

```csharp
using Amazon.DynamoDBv2;
using Amazon.DynamoDBv2.Model;

using var dynamoDbClient = new AmazonDynamoDBClient();

// An attribute value with a self-reference
var badItem = new Dictionary<string, AttributeValue>();
badItem["C"] = new AttributeValue { M = badItem };

var request = new PutItemRequest
{
    TableName = "<Table name here>",
    Item = new Dictionary<string, AttributeValue>
    {
        ["Id"] = new AttributeValue { N = "1" },
        ["Data"] = new AttributeValue { M = badItem }
    }
};

await dynamoDbClient.PutItemAsync(request);
```

### AWS .NET SDK and/or Package version used

AWSSDK.DynamoDBv2 3.7.406.7

### Targeted .NET Platform

.NET 8

### Operating System and version

Windows 11

ashishdhingra:
@OleksandrKrutykh Good afternoon. Thanks for opening the issue. The issue is reproducible with the below stack trace:
```C#
>	AWSSDK.Core.dll!ThirdParty.Json.LitJson.JsonWriter.Put(string str) Line 200	C#
 	AWSSDK.Core.dll!ThirdParty.Json.LitJson.JsonWriter.PutString(string str) Line 225	C#
 	AWSSDK.Core.dll!ThirdParty.Json.LitJson.JsonWriter.WritePropertyName(string property_name) Line 515	C#
[...]

PR #4456, opened 2026-06-30
Add JsonMaxDepth ClientConfig option for deeply nested json documents

### LINKED ISSUE #4455: AmazonUnmarshallingException from GetQueryRuntimeStatistics when query plan tree exceeds System.Text.Json default MaxDepth of 64
### Describe the feature

Expose a configurable `MaxDepth` on `AmazonAthenaConfig` (or `ClientConfig`) that is
  threaded through to the SDK's internal `StreamingUtf8JsonReader` / `Utf8JsonReader`
  options used during response unmarshalling.

  Currently the SDK v4 deserialises all service responses with `System.Text.Json` at the
  default `MaxDepth = 64`. `GetQueryRuntimeStatistics` returns a recursive
  `QueryStagePlanNode.Children` plan tree that can exceed 64 levels for complex Athena
  queries (many CTEs / JOINs), causing the SDK to throw `AmazonUnmarshallingException`
  even though the HTTP response is `200 OK` and the query itself succeeded.

### Use Case

When calling GetQueryRuntimeStatisticsAsync for an Athena query whose execution plan
  nests QueryStagePlanNode.Children deeper than 64 levels, the SDK throws:

    Amazon.Runtime.AmazonUnmarshallingException: Error unmarshalling response back from AWS.
      Last Parsed Path: …Children, HTTP Status Code: 200 OK
     ---> System.Text.Json.JsonReaderException:
          The maximum configured depth of 64 has been exceeded.

  The Athena query completed
  successfully in ~3 seconds; the failure is purely at deserialisation time.

  Only complex structural mappings (many CTEs / JOINs) reproduce the issue, so it surfaces silently in production
  rather than in testing.

### Proposed Solution

Add a MaxDepth property to AmazonAthenaConfig (or the ClientConfig base class) and
  thread it through to the JsonReaderOptions passed to the internal StreamingUtf8JsonReader.
  This would be analogous to the existing StreamingUtf8JsonReaderBufferSize knob already
  exposed on ClientConfig.

  Alternatively, document a supported extensibility hook (custom response handler,
  AWSConfigs switch, or pipeline interceptor) if one already exists, as none is visible
  in the v4 public API.

  A per-request override on GetQueryRuntimeStatisticsRequest would also be acceptable,
  though a global config property is more ergonomic.

### Other Information

Alternatives we investigated and rejected:

  1. StreamingUtf8JsonReaderBufferSize (already exposed on ClientConfig) — controls
     buffer size, not recursion depth. Does not help.

  2. Subclassing AmazonAthenaClient / overriding the unmarshaller — the internal
     StreamingUtf8JsonReader instantiation is sealed inside the generated unmarshaller;
     no override point is accessible.

  3. Raw HTTP + manual SigV4 + custom JSON parse — functional but brittle and
     maintenance-heavy.

  4. SELECT COUNT(*) over the aggregated Glue table — over-counts for multi-mapping
     domains (the aggregated table is per-domain, shared across mappings).

  Applied workaround: catch AmazonUnmarshallingException specifically and fall back to
  counting rows via GetQueryResultsAsync paging (skipping the header row on the first
  page). This is functional but turns an O(1) statistics call into O(N/1000) paging
  calls for affected queries, and requires callers to work around an SDK limitation
  that should be configurable.

### Acknowledgements

 AWSSDK.Athena 4.0.5.4
  AWSSDK.Core 4.0.9

### Targeted .NET Platform

.NET 8

Amazon Linux 2 (AWS Lambda)

peterrsongg:
## Description

Adds `JsonMaxDepth` option for unmarshalling deeply nested JSON documents. Also adds it to the Extensions package DefaultClientConfig. I was gonna add a test that it throws an exception if the nesting is more than the allowed amount but due to accessibility level of JsonReaderException, I was unable to. I did a manual test to make sure that if the depth of the document is 65 and the allowed level is 64 it still throws.

Fixes
## Motivation and Context

Fixes #4455 . Adds `JsonMaxDepth` option for unmarshalling deeply nested JSON documents.

## Testing

Added unit tests.
### Dry-runs

- **DotNet Dry-run ID:**90f3707b-d961-4c76-b095-7c9bebab6e8e
  - [ ] Pending
  - [x] Completed successfully
  - [ ] Failed
- **PowerShell Dry-run ID:**54a7a27b-47d6-4688-b61e-32ae75a1d22f
  - [ ] Pending
  - [x] Completed successfully
  - [ ] Failed

## Types of changes

## Checklist

## License


sdk/src/Core/Amazon.Runtime/IClientConfig.cs
```diff
@@ -236,11 +236,18 @@ public partial interface IClientConfig
...
         /// out to the request.
         /// </summary>
         int BufferSize { get; }
 
+        /// <summary>
+        /// Gets the maximum depth allowed when reading JSON responses.
+        /// The default is 64. Increase this value if you encounter
+        /// AmazonUnmarshallingException due to deeply nested response structures.
+        /// </summary>
+        int JsonMaxDepth { get; }
```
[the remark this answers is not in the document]
dscpinheiro: `IClientConfig` has the backwards compatible attribute because its concrete implementation is in `Core` itself (so this comment is not applicable here).

sdk/test/UnitTests/Custom/Marshalling/StreamingUtf8JsonReaderTests.cs
```diff
@@ -210,5 +210,35 @@ public void StreamingUtf8JsonReaderHandlesBufferSizeMinusOne()
...
+                var reader = new StreamingUtf8JsonReader(stream, 4096, 128);
+                int maxDepthSeen = 0;
+                while (reader.Read())
+                {
+                    if (reader.Reader.CurrentDepth > maxDepthSeen)
+                        maxDepthSeen = (int)reader.Reader.CurrentDepth;
+                }
+                Assert.IsTrue(maxDepthSeen >= depth);
+            }
+        }
```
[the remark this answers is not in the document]
peterrsongg: I already said in the description why I didn't add this case.

peterrsongg:
Both dry runs passed, but I just merged in some stuff from copilot so will run another:
netv4: a53558e5-d2c2-449a-a0a4-32c06812df86 (PASSED)
psv5: a466d5c9-d0c0-453a-8aaf-8a893e687c1d (PASSED)

extensions/src/AWSSDK.Extensions.NETCore.Setup/ConfigurationExtensions.cs
```diff
@@ -235,15 +235,19 @@ public static AWSOptions GetAWSOptions(this IConfiguration config, string config
 
                     options.DefaultClientConfig.IgnoreConfiguredEndpointUrls = ignoreConfiguredEndpointUrls;
                 }
-                else if (string.Equals(element.Key, nameof(DefaultClientConfig.JsonMaxDepth), StringComparison.OrdinalIgnoreCase))
-                {
-                    if (!int.TryParse(element.Value, out var jsonMaxDepth))
-                    {
-                        throw new ArgumentException($"Invalid integer value for {nameof(DefaultClientConfig.JsonMaxDepth)}.");
-                    }
+else if (string.Equals(element.Key, nameof(DefaultClientConfig.JsonMaxDepth), StringComparison.OrdinalIgnoreCase))
```
dscpinheiro: The indentation on this commit looks odd... I think it's Copilot, I've seen it on my PRs too when I accept the automatic suggestion.

peterrsongg: ah thanks for pointing that out. fixed.

sdk/src/Core/Amazon.Runtime/Internal/Transform/JsonUnmarshallerContext.cs
```diff
@@ -54,6 +54,9 @@ public class JsonUnmarshallerContext : UnmarshallerContext
         private JsonTokenType? currentToken = null;
         private bool disposed = false;
         private bool wasPeeked = false;
+        // a default value of 64 is used here to mirror the default on system.text.json as specified here:
+        // https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonreaderoptions.maxdepth?view=net-8.0#system-text-json-jsonreaderoptions-maxdepth
+        private readonly int _jsonMaxDepth = 64;
```
normj: I understand `64` is MSFT default but what is the harm if we increase the default? I only want users to have think about this config as a very last option.

sdk/src/Core/Amazon.Runtime/Internal/Transform/JsonUnmarshallerContext.cs
```diff
@@ -136,6 +139,8 @@ public JsonUnmarshallerContext(
                 baseStream = this.CrcStream;
             else
                 baseStream = responseStream;
+
+            _jsonMaxDepth = requestContext?.ClientConfig?.JsonMaxDepth ?? 64;
```
normj: Don't put the magic number 64 in 2 places. I suspect you should just not set `_jsonMaxDepth` which is already set to 64 if the ClientConfig property is not set.

sdk/src/Core/Amazon.Runtime/ClientConfig.cs
```diff
@@ -645,6 +645,25 @@ public int BufferSize
             set { this.bufferSize = value; }
         }
 
+        private int jsonMaxDepth = 64;
```
normj: Third reference of the number 64. Please use a constant so we if want to change the depth we only have one place to change.

sdk/src/Core/Amazon.Runtime/Internal/Util/StreamingUtf8JsonReader.cs
```diff
@@ -79,7 +75,20 @@ public StreamingUtf8JsonReader(Stream stream) : this(stream, AWSConfigs.Streamin
         /// <param name="stream">the stream containing the data</param>
         /// <param name="bufferSize">the size of the buffer when reading data from the stream</param>
         /// <exception cref="ArgumentException"></exception>
-        public StreamingUtf8JsonReader(Stream stream, int bufferSize)
+        public StreamingUtf8JsonReader(Stream stream, int bufferSize) : this(stream, bufferSize, 64)
```
normj: Use constant for the magic number `64`.

normj (changes requested):

generator/ServiceClientGeneratorLib/Generators/Marshallers/JsonRPCResponseUnmarshaller.tt
dscpinheiro: Should this constructor (and the one on line 181) also be updated to use the new signature?

dscpinheiro, line 1: https://github.com/aws/aws-sdk-net/blob/main/generator/ServiceClientGeneratorLib/Generators/SourceFiles/EventStreamOutputGenerator.tt also uses the `new StreamingUtf8JsonReader(context.Stream)` constructor, should it be updated as well?

dscpinheiro (commented):
Nit but you could've included only a couple of generated files to make the review easier (no need to change now).

sdk/src/Core/Amazon.Runtime/Internal/Transform/JsonUnmarshallerContext.cs
dscpinheiro: The one scenario I could think of is this DDB issue (but that was 1000 levels deep and the `Utf8JsonWriter` instead of the reader): https://github.com/aws/aws-sdk-net/issues/3722

generator/ServiceClientGeneratorLib/Generators/Marshallers/JsonRPCResponseUnmarshaller.tt
peterrsongg: this one doesn't need to be updated since the error will never be nested 64+ levels deep. That is why i left it out

peterrsongg: same goes for an event stream output, its never going to be nested that deep so the default is fine here as well.

sdk/src/Core/Amazon.Runtime/Internal/Transform/JsonUnmarshallerContext.cs
peterrsongg: Not really, I just did 64 because that was MSFT default. I can increase it to 256. It's just supposed to protect against malicious JSON to avoid infinite recursion

generator/ServiceClientGeneratorLib/Generators/Marshallers/JsonRPCResponseUnmarshaller.tt
peterrsongg: done. just added it here and in the stream copy version in the exception unmarshalling

peterrsongg: just updated.

sdk/src/Core/Amazon.Runtime/Internal/Transform/JsonUnmarshallerContext.cs
peterrsongg: updated to 256

peterrsongg:
new dry run id dotnetv4:966aa6a3-5345-4e8b-93c6-523b60bbd5f9
psv5: 636ddc69-0e7c-4921-8ac2-435db260e4c4

dscpinheiro (dismissed):

sdk/src/Core/Amazon.Runtime/Internal/Transform/JsonUnmarshallerContext.cs
```diff
@@ -333,6 +338,15 @@ public Stream Stream
             get { return baseStream; }
         }
 
+        /// <summary>
+        /// Gets the maximum depth allowed when reading JSON responses.
+        /// The default is 64.
```
normj: Update this since you change the default

normj (changes requested):

normj (approved):

AlexDaines (approved):

dscpinheiro (approved):

Issue #4433, opened 2026-06-12
Upgrading sdk packages causes "28000: PAM authentication failed for user" error


mrizzini:
### Describe the bug

Hi there,
 
My team has been investigating an error occurring when connecting to our our postgres DB instance, from our EKS container, and have not been able to get to the root cause. The error we are seeing is "28000: PAM authentication failed for user {ourUser}"
 
We have recently started upgrading our AWSSDK packages in our EKS container, and that is when we started getting these errors. Before the errors began, we were stable, and we were on the following packages:
 
    <PackageReference Include="AWSSDK.Core" Version="4.0.1.3" />
    <PackageReference Include="AWSSDK.EventBridge" Version="4.0.5.2" />
    <PackageReference Include="AWSSDK.RDS" Version="4.0.10.7" />
    <PackageReference Include="AWSSDK.S3" Version="4.0.9.1" />
    <PackageReference Include="AWSSDK.SecretsManager" Version="4.0.2" />
    <PackageReference Include="AWSSDK.SecurityToken" Version="4.0.3" />
    <PackageReference Include="AWSSDK.SQS" Version="4.0.2" />
    <PackageReference Include="AWSSDK.SSO" Version="4.0.2" />
    <PackageReference Include="AWSSDK.SSOOIDC" Version="4.0.3.1" />
    <PackageReference Include="AWSSDK.StepFunctions" Version="4.0.1" />
    <PackageReference Include="AWSSDK.Transfer" Version="4.0.4.3" />
 
Sometime in April or May, we upgraded to the following:
 
    <PackageReference Include="AWSSDK.Core" Version="4.0.3.31" />
    <PackageReference Include="AWSSDK.EventBridge" Version="4.0.5.2" />
    <PackageReference Include="AWSSDK.RDS" Version="4.0.10.7" />
    <PackageReference Include="AWSSDK.S3" Version="4.0.9.1" />
    <PackageReference Include="AWSSDK.SecretsManager" Version="4.0.2" />
    <PackageReference Include="AWSSDK.SecurityToken" Version="4.0.5.20" />
    <PackageReference Include="AWSSDK.SQS" Version="4.0.2" />
    <PackageReference Include="AWSSDK.SSO" Version="4.0.2" />
    <PackageReference Include="AWSSDK.SSOOIDC" Version="4.0.3.1" />
    <PackageReference Include="AWSSDK.StepFunctions" Version="4.0.1" />
[...]

PR #4457, opened 2026-07-06
Update RDS and DSQL token generators to force credentials refresh when needed

dscpinheiro:
Fix for #4433: A `ForceRefresh` method was added to the `RefreshingAWSCredentials` provider, and it's called from the RDS / DSQL token generators to guarantee the credentials used for signing can outlive the token (in most scenarios, see DSQL comments for limitations with longer expirations).

### Dry-runs
- **DotNet Dry-run ID:** `DRY_RUN-c34e8386-fedd-42aa-ad0c-a32f186b5cd1`
- **PowerShell Dry-run ID:** `DRY_RUN-92251ca3-000c-4c1d-a423-d15bb8a958fc`

There's a small behavior change when the refresh fails, I called it out in the DevConfig: `Token generation may now throw if the credentials refresh fails; previously a token could be returned signed with credentials nearing expiration, which the database could reject.`

## License

### DIFF
--- a/generator/.DevConfigs/6c5704cb-97e5-47ce-b930-8733977e2a94.json
+++ b/generator/.DevConfigs/6c5704cb-97e5-47ce-b930-8733977e2a94.json
@@ -1,7 +1,7 @@
 {
   "core": {
     "changeLogMessages": [
-      "Added `RefreshingAWSCredentials.ForceRefresh(TimeSpan)` and `ForceRefreshAsync(TimeSpan)` methods that regenerate credentials when the current credentials would expire within the requested minimum lifetime, and return the existing credentials otherwise."
+      "Added `RefreshingAWSCredentials.GetCredentials(TimeSpan)` and `GetCredentialsAsync(TimeSpan)` methods that regenerate credentials when the current credentials would expire within the requested minimum lifetime, and return the existing credentials otherwise."
     ],
     "type": "patch",
     "updateMinimum": true
--- a/generator/.DevConfigs/6c5704cb-97e5-47ce-b930-8733977e2a94.json
+++ b/generator/.DevConfigs/6c5704cb-97e5-47ce-b930-8733977e2a94.json
@@ -1,0 +1,25 @@
+{
+  "core": {
+    "changeLogMessages": [
+      "Added `RefreshingAWSCredentials.ForceRefresh(TimeSpan)` and `ForceRefreshAsync(TimeSpan)` methods that regenerate credentials when the current credentials would expire within the requested minimum lifetime, and return the existing credentials otherwise."
+    ],
+    "type": "patch",
+    "updateMinimum": true
+  },
+  "services": [
+    {
+      "serviceName": "RDS",
+      "type": "minor",
+      "changeLogMessages": [
+        "`RDSAuthTokenGenerator` now forces a refresh when the credentials would expire before the generated token, so the token remains valid for its full lifetime. This applies to credential providers deriving from `RefreshingAWSCredentials` (such as `AssumeRoleAWSCredentials` and `AssumeRoleWithWebIdentityCredentials`); EC2 instance profile credentials and static credentials are unaffected. `GenerateAuthToken` may now throw if the credentials refresh fails; previously a token could be returned signed with credentials nearing expiration, which the database could reject."
+      ]
+    },
+    {
+      "serviceName": "DSQL",
+      "type": "minor",
+      "changeLogMessages": [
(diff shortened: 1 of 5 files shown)


generator/.DevConfigs/6c5704cb-97e5-47ce-b930-8733977e2a94.json
```diff
@@ -0,0 +1,25 @@
...
+    ],
+    "type": "patch",
+    "updateMinimum": true
+  },
+  "services": [
+    {
+      "serviceName": "RDS",
+      "type": "minor",
+      "changeLogMessages": [
+        "`RDSAuthTokenGenerator` now forces a refresh when the credentials would expire before the generated token, so the token remains valid for its full lifetime. This applies to credential providers deriving from `RefreshingAWSCredentials` (such as `AssumeRoleAWSCredentials` and `AssumeRoleWithWebIdentityCredentials`); EC2 instance profile credentials and static credentials are unaffected. `GenerateAuthToken` may now throw if the credentials refresh fails; previously a token could be returned signed with credentials nearing expiration, which the database could reject."
```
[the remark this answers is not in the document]
dscpinheiro: `InstanceProfileAWSCredentials` is not part of the default credential chain, _EC2 instance profile credentials_ here refers to https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Core/Amazon.Runtime/Credentials/DefaultInstanceProfileAWSCredentials.cs

(Copilot is right if a customer was invoking the token generator using an explicit `InstanceProfileAWSCredentials` it'd also get a refresh, but I don't believe that's a common use case at all)

sdk/src/Core/Amazon.Runtime/Credentials/RefreshingAWSCredentials.cs
```diff
@@ -459,6 +459,100 @@ public virtual void ClearCredentials()
...
 
+        /// <summary>
+        /// Returns credentials that remain valid for at least <paramref name="minimumLifetime"/> when the credential
+        /// provider can supply them, regenerating the current credentials if they would expire sooner. This lets callers
+        /// such as the RDS and DSQL auth-token generators ensure the generated token stays valid for its full lifetime.
+        /// If the regenerated credentials still expire within <paramref name="minimumLifetime"/> (for example, the
+        /// requested lifetime exceeds the provider's maximum session duration), they are returned anyway and an
+        /// informational message is logged. Exceptions thrown while regenerating the credentials propagate to the caller.
+        /// </summary>
+        public ImmutableCredentials ForceRefresh(TimeSpan minimumLifetime)
```
normj: I'm thinking we should name this as `GetCredentials` and just treat it as an overload of the normal method. I think `ForceRefresh` isn't right because depending on `minimumLiftetime` and the current credentials expiration it won't actually force a refresh.

sdk/src/Core/Amazon.Runtime/Credentials/RefreshingAWSCredentials.cs
```diff
@@ -459,6 +459,100 @@ public virtual void ClearCredentials()
...
+                return newState.Credentials;
+            }
+            finally
+            {
+                _updateGeneratedCredentialsSemaphore.Release();
+            }
+        }
+
+        /// <inheritdoc cref="ForceRefresh(TimeSpan)"/>
+        public async Task<ImmutableCredentials> ForceRefreshAsync(TimeSpan minimumLifetime)
```
normj: Same comment as above and call this `GetCredentialsAsync` and treat as a overload.

dscpinheiro: Updated in https://github.com/aws/aws-sdk-net/pull/4457/commits/087a8c92536e266d885590b8559433b5cb0369a2

sdk/src/Core/Amazon.Runtime/Credentials/RefreshingAWSCredentials.cs
dscpinheiro: Updated in https://github.com/aws/aws-sdk-net/pull/4457/commits/087a8c92536e266d885590b8559433b5cb0369a2

normj (approved):

sdk/src/Core/Amazon.Runtime/Credentials/RefreshingAWSCredentials.cs
```diff
@@ -459,6 +459,100 @@ public virtual void ClearCredentials()
...
+            {
+                _updateGeneratedCredentialsSemaphore.Release();
+            }
+        }
+
+        /// <inheritdoc cref="GetCredentials(TimeSpan)"/>
+        public async Task<ImmutableCredentials> GetCredentialsAsync(TimeSpan minimumLifetime)
+        {
+            if (minimumLifetime < TimeSpan.Zero)
+                throw new ArgumentOutOfRangeException("minimumLifetime", "minimumLifetime cannot be negative");
```
muhammad-othman, line 509: nit: I think using `nameof` is better for all `ArgumentOutOfRangeException` so we don't forget about the string incase the parameter name changed for any reason.

muhammad-othman (approved):

PR #4459, opened 2026-07-06
feat: support strict tool use in Bedrock MEAI BedrockChatClient

### LINKED ISSUE #4458: Surface strict tool use (ToolSpecification.Strict) through BedrockChatClient in AWSSDK.Extensions.Bedrock.MEAI
### Describe the feature

BedrockChatClient (in AWSSDK.Extensions.Bedrock.MEAI) currently lacks the ability to specify strict tool use. The underlying Amazon.BedrockRuntime.Model.ToolSpecification already has a Strict property that marshals to "strict" on the wire, but the MEAI bridge never sets it: AddUserTools builds each ToolSpecification with only Name/Description/InputSchema, and rebuilds the input schema with only type/properties/required — dropping additionalProperties, which strict mode requires. 

Request: The proposed addition would allow callers to opt a tool into strict mode through the MEAI AIFunction abstraction, and have the bridge emit both ToolSpecification.Strict = true and additionalProperties: false.

### Use Case

We run an agentic app (Claude on Bedrock, through AWSSDK.Extensions.Bedrock.MEAI). Occasionally the model emits tool inputs that violate the declared JSON schema - missing required fields, wrong types, invalid enum values - which breaks downstream tool execution. Bedrock's strict tool use eliminates that class of error by constraining tool-call generation to the schema, but it's currently unreachable through MEAI without dropping to the raw IAmazonBedrockRuntime client.

### Proposed Solution

Per-tool opt-in via the AIFunction's AdditionalProperties, mirroring the existing CachePoint pattern. In AddUserTools:

```c#
// Opt-in strict tool use: callers set a boolean "strict" on the AIFunction's AdditionalProperties.
bool strict =
    f.AdditionalProperties?.TryGetValue("strict", out object? maybeStrict) == true &&
    maybeStrict is true;

// ...when building the schema dictionary:
if (strict)
{
    schemaDictionary["additionalProperties"] = new Document(false);
}

// ...on the ToolSpecification:
Strict = strict ? true : null,
```

Behaviour is unchanged when the flag is absent (Strict stays null, no additionalProperties), so it's fully backward-compatible.

### Other Information

This approach is similar to the prompt-caching support added by reading AdditionalProperties (#3930).

### Acknowledgements

- [x] I may be able to implement this feature request

AWSSDK.Extensions.Bedrock.MEAI 4.0.101.2, AWSSDK.BedrockRuntime 4.0.100.2

### Targeted .NET Platform

.NET 8, .NET 10

Windows 11, AmazonLinux

dimo-coremont:
## Description

Adds opt-in support for **strict tool use** to `BedrockChatClient` in `AWSSDK.Extensions.Bedrock.MEAI`.

`Amazon.BedrockRuntime.Model.ToolSpecification` already exposes a `Strict` property that marshals to `"strict"` on the wire, but `BedrockChatClient.AddUserTools` never sets it - every tool is built with only `Name`/`Description`/`InputSchema`. The method also rebuilds the input schema from scratch using only `type`/`properties`/`required`, dropping any `additionalProperties` the caller's schema declared, which strict mode requires.

This PR lets a caller opt a tool into strict mode via a boolean `"strict"` entry on the `AIFunction`'s `AdditionalProperties` (mirroring the existing `CachePoint` pattern used elsewhere in this file). When set:
- `schemaDictionary["additionalProperties"]` is added as `false`.
- `ToolSpecification.Strict` is set to `true`.

When the flag is absent (the default), behavior is unchanged: `Strict` stays `null` (the generated marshaller omits a null `Strict`, so no new field appears on the wire) and no `additionalProperties` is added to the schema.

## Motivation and Context

Fixes #4458 

## Testing

Added two unit tests to `extensions/test/BedrockMEAITests/BedrockChatClientTests.cs`, following the existing patterns in that file for constructing the client and asserting on the built `ConverseRequest`/`ToolConfiguration`:

- `Tool_WithoutStrictOptIn_LeavesStrictNull_AndNoAdditionalProperties` — confirms unchanged behavior for tools that don't opt in (`Strict == null`, no `additionalProperties` in the emitted schema).
- `Tool_WithStrictOptIn_SetsStrictTrue_AndAdditionalPropertiesFalse` — confirms `Strict == true` and `additionalProperties: false` when the tool's `AdditionalProperties["strict"]` is `true`.

Ran the full `BedrockMEAITests` project locally (`dotnet test`)

No breaking changes

## Types of changes

## Checklist

## License

### DIFF
--- a/extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
+++ b/extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
@@ -1061,6 +1061,12 @@ internal sealed partial class BedrockChatClient : IChatClient
                 }
             }
 
+            // Opt-in strict tool use: callers request it per-tool via a boolean "strict"
+            // entry on the AIFunction's AdditionalProperties (mirrors the CachePoint pattern).
+            bool strict =
+                f.AdditionalProperties?.TryGetValue("strict", out object? maybeStrict) == true &&
+                maybeStrict is true;
+
             Dictionary<string, Document> schemaDictionary = new()
             {
                 ["type"] = new Document("object"),
@@ -1076,6 +1082,12 @@ internal sealed partial class BedrockChatClient : IChatClient
                 schemaDictionary["required"] = new Document(required);
             }
 
+            if (strict)
+            {
+                // Bedrock strict tool use requires the input schema to forbid extra properties.
+                schemaDictionary["additionalProperties"] = new Document(false);
+            }
+
             toolConfig ??= new();
             toolConfig.Tools ??= [];
             toolConfig.Tools.Add(new()
@@ -1088,6 +1100,7 @@ internal sealed partial class BedrockChatClient : IChatClient
                     {
                         Json = new(schemaDictionary)
                     },
+                    Strict = strict ? true : null,
                 },
             });
         }
--- a/extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
+++ b/extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
@@ -1061,10 +1061,8 @@ internal sealed partial class BedrockChatClient : IChatClient
                 }
             }
 
-            // Opt-in strict tool use: callers request it per-tool via a boolean "strict"
-            // entry on the AIFunction's AdditionalProperties (mirrors the CachePoint pattern).
             bool strict =
-                f.AdditionalProperties?.TryGetValue("strict", out object? maybeStrict) == true &&
+                f.AdditionalProperties?.TryGetValue(nameof(ToolSpecification.Strict), out object? maybeStrict) == true &&
                 maybeStrict is true;
 
             Dictionary<string, Document> schemaDictionary = new()
--- a/extensions/test/BedrockMEAITests/BedrockChatClientTests.cs
+++ b/extensions/test/BedrockMEAITests/BedrockChatClientTests.cs
@@ -667,7 +667,7 @@ public class BedrockChatClientTests
             "get_weather",
             "Get the weather",
             schema,
-            new Dictionary<string, object> { ["strict"] = true });
+            new Dictionary<string, object> { ["Strict"] = true });
 
         var options = new ChatOptions { Tools = new[] { tool } };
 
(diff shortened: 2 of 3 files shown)


extensions/src/AWSSDK.Extensions.Bedrock.MEAI/BedrockChatClient.cs
```diff
@@ -1061,6 +1061,12 @@ private static ToolConfiguration AddUserTools(ToolConfiguration? toolConfig, ILi
                 }
             }
 
+            // Opt-in strict tool use: callers request it per-tool via a boolean "strict"
+            // entry on the AIFunction's AdditionalProperties (mirrors the CachePoint pattern).
+            bool strict =
+                f.AdditionalProperties?.TryGetValue("strict", out object? maybeStrict) == true &&
```
dscpinheiro: I think this change looks reasonable, one small question is that previous PRs used `nameof` when reading from `AdditionalProperties`, for example:
- `trc.AdditionalProperties?.TryGetValue(nameof(ReasoningContentBlock.RedactedContent), out redactedContent);`
- `content.AdditionalProperties?.TryGetValue(nameof(ContentBlock.CachePoint), out var maybeCachePoint)`

Do you think it'd make sense to use `nameof(ToolSpecification.Strict)` here as well?

dimo-coremont: Thanks. I made the change. I also removed the comment which seemed unnecessary.

dscpinheiro: You'll need to update the test too, because it's still using the lowercase "strict" and fails with the latest commit:
```
[xUnit.net 00:00:00.25]     Amazon.BedrockRuntime.BedrockChatClientTests.Tool_WithStrictOptIn_SetsStrictTrue_AndAdditionalPropertiesFalse [FAIL]
  Failed Amazon.BedrockRuntime.BedrockChatClientTests.Tool_WithStrictOptIn_SetsStrictTrue_AndAdditionalPropertiesFalse [81 ms]
  Error Message:
   Assert.True() Failure
Expected: True
Actual:   False
  Stack Trace:
     at Amazon.BedrockRuntime.BedrockChatClientTests.Tool_WithStrictOptIn_SetsStrictTrue_AndAdditionalPropertiesFalse() in /_/extensions/test/BedrockMEAITests/BedrockChatClientTests.cs:line 680
```

dimo-coremont: Apologies, this is done now.

peterrsongg:
Looks good, kicked off an internal build:
.NET :  5721a468-f8c4-4d4a-9489-57d621590679,
Powershell: a660fbe6-b667-479f-9f5a-36127532285a

will approve if it passes.

peterrsongg (approved):
builds succeeded. Approving.

dscpinheiro (approved):

dscpinheiro:
@dimo-coremont Thanks for the contribution, this will be included in the next SDK release.

PR #4461, opened 2026-07-09
Feature/s3 download directory file request event

breadswonders:
## Description
Adds `TransferUtilityDownloadDirectoryRequest.DownloadDirectoryFileRequestEvent`, raised for each file before it is downloaded, allowing subscribers to inspect and modify the individual `TransferUtilityDownloadRequest` before it executes. This mirrors the existing `TransferUtilityUploadDirectoryRequest.UploadDirectoryFileRequestEvent` so the download-directory API has parity with the upload-directory API.

## Motivation and Context
`UploadDirectory` already lets callers customize each per-file request via `UploadDirectoryFileRequestEvent`, but `DownloadDirectory` had no equivalent hook. This adds the symmetric capability for downloads (e.g. setting `RequestPayer`, custom headers, or SSE-C settings per file). My personal motivation for this change is being able to handle S3 to Windows file path conversions when uploading (already supported) and downloading directories. Without this event, we have no way of "fixing" an invalid filepath before `TransferUtility` throws an exception at us.

## Testing
Added 4 unit tests to `DownloadDirectoryCommandTests` covering: the event firing once per file, the args carrying a fully-populated request, handler mutations propagating through to the underlying `GetObjectRequest`, and the no-subscriber path. All 38 tests in the class pass (`net8.0`). No integration tests were run (they require AWS credentials).

### Dry-runs

- **DotNet Dry-run ID:**
  - [x] Pending
  - [ ] Completed successfully
  - [ ] Failed
- **PowerShell Dry-run ID:**
  - [x] Pending
  - [ ] Completed successfully
  - [ ] Failed

No breaking changes. The change is purely additive (a new public event and a new `DownloadDirectoryFileRequestArgs : EventArgs` type). No existing signatures or behavior change.

## Types of changes

## Checklist

  - I am unsure of this.
  - Could not run integration tests.

## License

## AI Disclosure
Generated by AI tools, and reviewed by breadswonders. I drove, Claude helped.

### DIFF
--- a/sdk/src/Services/S3/Custom/Transfer/Internal/DownloadDirectoryCommand.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/Internal/DownloadDirectoryCommand.cs
@@ -154,8 +154,6 @@ namespace Amazon.S3.Transfer.Internal
 
             downloadRequest.WriteObjectProgressEvent += downloadedProgressEventCallback;
 
-            _request.RaiseDownloadDirectoryFileRequestEvent(downloadRequest);
-
             return downloadRequest;
         }
 
--- a/sdk/src/Services/S3/Custom/Transfer/Internal/DownloadDirectoryCommand.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/Internal/DownloadDirectoryCommand.cs
@@ -154,6 +154,8 @@ namespace Amazon.S3.Transfer.Internal
 
             downloadRequest.WriteObjectProgressEvent += downloadedProgressEventCallback;
 
+            _request.RaiseDownloadDirectoryFileRequestEvent(downloadRequest);
+
             return downloadRequest;
         }
 
--- a/generator/.DevConfigs/d3c19aa1-9757-422f-9bc9-31eaaeb8ed05.json
+++ b/generator/.DevConfigs/d3c19aa1-9757-422f-9bc9-31eaaeb8ed05.json
@@ -1,0 +1,11 @@
+{
+  "services": [
+    {
+      "serviceName": "S3",
+      "type": "minor",
+      "changeLogMessages": [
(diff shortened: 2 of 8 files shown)


sdk/src/Services/S3/Custom/Transfer/Internal/DownloadDirectoryCommand.cs
```diff
@@ -154,6 +154,8 @@ internal TransferUtilityDownloadRequest ConstructTransferUtilityDownloadRequest(
 
             downloadRequest.WriteObjectProgressEvent += downloadedProgressEventCallback;
 
+            _request.RaiseDownloadDirectoryFileRequestEvent(downloadRequest);
```
GarrettBeatty: i think this should be moved outside the method to https://github.com/breadswonders/aws-sdk-net/blob/main/sdk/src/Services/S3/Custom/Transfer/Internal/_bcl%2Bnetstandard/DownloadDirectoryCommand.cs#L312 what do you think? that way we keep the logic of `ConstructTransferUtilityDownloadRequest` to only build the request

GarrettBeatty (changes requested):
see comment

sdk/src/Services/S3/Custom/Transfer/Internal/DownloadDirectoryCommand.cs
breadswonders: @GarrettBeatty That makes sense to me. I was following the pattern that was used for the `UploadDirectoryCommand` [here](https://github.com/breadswonders/aws-sdk-net/blob/feature/s3-download-directory-file-request-event/sdk/src/Services/S3/Custom/Transfer/Internal/UploadDirectoryCommand.cs#L140), so we should probably update both at the same time, sound good?

GarrettBeatty: yes that makes sense to me!

breadswonders: @GarrettBeatty Done.

GarrettBeatty (approved):

GarrettBeatty:
code looks good. i started a dry run 8793aa67-1dbd-48eb-b3db-119460864b67 internally to verify all tests pass. will merge once that is done

dscpinheiro (approved):

PR #4462, opened 2026-07-10
PoC: RMS prototype

JanHyka:
# DO NOT MERGE, THIS IS PROOF OF CONCEPT

## Description
Proof of concept for usage of high performance RecyclableMemoryStream

## Motivation and Context
As benchmarks reveal, current baseline is still very cpu and memory intensive, where esp. cpu load doesn't grow in linear fashion. Profiling revealed MemoryStream being main culprit (each resize allocates new memory on heap and copies content of previous one - which is especially significant for blocks above LoH threshold (default 85kB)).
For testing purposes, very minimalistic change was applied and inludes dependency on latest RecyclableMemoryStream version (one reason being it natively inherits from MemoryStream so works as an easy inplace change).  Where applied, the change reveals full costs of simple MemoryStream approach.

Both base and rms benchmarks are included in this branch for detailed observation.

RecyclableMemoryStream Setup: default + zeroing enabled, rented memory return on finalizer (no using, no dispose, worst case scenario)

Request/suggestion based on the data collected in this branch will be added in Issues for further processing.

### Benchmark Comparison: `base` vs `rms` full Serde suite

Environment for both runs: BenchmarkDotNet v0.15.2, Windows 11, Intel Core Ultra 9 285K, .NET 8.0.28, InProcess toolchain.

Noise threshold: changes below roughly 2x the reported StdDev% (typically <2-3%) are treated as run-to-run noise and not called out as regressions/improvements.

#### Major Improvements (well above noise)

| Benchmark | base Mean | rms Mean | Δ Mean | base Alloc | rms Alloc | Notes |
|---|---:|---:|---:|---:|---:|---|
| RestXml `GetObject_M` | 53,785.9 ns | 17,347.8 ns | **-67.7%** | 127.85 KB | 27.85 KB | Gen2 collections eliminated (19.9→0) |
| RestXml `GetObject_L` | 140,891.4 ns | 71,394.2 ns | **-49.3%** | 1,052.75 KB | 28.57 KB | Gen0/1/2 collections eliminated, allocation cut ~97% |
| RestJson1 `GetObject_L` | 132,312.3 ns | 71,731.0 ns | **-45.8%** | 1,050.66 KB | 97.5 KB | Gen2 collections eliminated, allocation cut ~90% |
| RestJson1 `GetObject_M` | 35,317.1 ns | 17,579.7 ns | **-50.2%** | 125.71 KB | 25.71 KB | Gen2 collections eliminated |

These four are the standout wins — large streaming/object-read paths in RestJson1 and RestXml now avoid large-object-heap / Gen2 allocations entirely, driving both latency and memory down dramatically.

#### Moderate Improvements

| Benchmark | base Mean | rms Mean | Δ Mean |
|---|---:|---:|---:|
| AwsJson10 `PutItem_M` | 11,220.4 ns | 10,820.1 ns | -3.6% |
| AwsJson10 `PutItem_L` | 13,140.7 ns | 12,688.7 ns | -3.4% |
| AwsJson10 `GetItem_L` | 17,913.9 ns | 17,032.9 ns | -4.9% |
| RestJson1 `PutObject_M` | 33,836.7 ns | 32,577.5 ns | -3.7% |
| RestJson1 `GetObject_S` | 10,000.7 ns | 9,660.0 ns | -3.4% |
| RestJson1 `PutMetricData_S` | 10,319.6 ns | 9,838.0 ns | -4.7% |
| RestJson1 `PutMetricData_M` | 15,277.0 ns | 14,754.7 ns | -3.4% |
| RestJson1 `GetMetricData_S` | 11,326.4 ns | 11,001.1 ns | -2.9% |
| RestJson1 `GetMetricData_M` | 17,599.2 ns | 17,143.8 ns | -2.6% |

#### Regressions (above noise)

| Benchmark | base Mean | rms Mean | Δ Mean |
|---|---:|---:|---:|
| AwsQuery `PutMetricData_S` | 11,767.4 ns | 12,112.7 ns | +2.9% |
| AwsQuery `PutMetricData_M` | 17,565.8 ns | 18,113.0 ns | +3.1% |

These two AwsQuery `PutMetricData` cases are small but consistent regressions (low StdDev in both runs, ~0.6-1.3 ns% error), so they are likely real rather than noise, though the magnitude is modest.

#### Within Noise (no significant change)

- AwsJson10: `Healthcheck`, `PutItem_S`, `GetItem_S`, `GetItem_M`
- AwsQuery: `Healthcheck`, `GetMetricData_S`, `GetMetricData_M`
- RestJson1: `CopyObject_Baseline`, `CopyObject_M`, `PutObject_S`, `PutObject_L`, `PutMetricData_L`, `GetMetricData_L`
- RestXml: `CopyObject`, `PutObject_S`, `PutObject_M`, `PutObject_L`, `GetObject_S`, `PutMetricData_S`, `PutMetricData_M`, `GetMetricData_S`, `GetMetricData_M`
- RpcV2Cbor: all methods (`PutItem_S/M/L`, `GetItem_S/M/L`) — differences are all under ~1.5%, consistent with measurement noise.

#### Summary

- **Biggest wins**: RestJson1 and RestXml `GetObject_M`/`GetObject_L` show 45-68% latency reductions driven by eliminating Gen2 garbage collections and cutting allocations by 90%+ — this is the headline result of the `rms` changes.
- **Secondary wins**: Several AwsJson10 and RestJson1 `PutItem`/`PutObject`/`PutMetricData`/`GetMetricData` cases show consistent 2.5-5% latency improvements with unchanged allocation, suggesting general CPU-path micro-optimizations.
- **Regressions**: Two AwsQuery `PutMetricData` (S/M) benchmarks regressed by ~3%, worth investigating but not alarming.
- **Unaffected**: AwsJson10/AwsQuery/RestJson1/RestXml healthcheck and several baseline paths, and the entire RpcV2Cbor suite, show no measurable change.

## License


normj:
Apologize @JanHyka for not getting back to you sooner. We have been debating the `Microsoft.IO.RecyclableMemoryStream` dependency. We take a pretty hard stance about taking any dependencies on but the performance improvements you were seeing with `Microsoft.IO.RecyclableMemoryStream` was hard to ignore.

Unfortunately we need to stick with our rule of not taking dependencies. Particular any dependencies outside of the `System.*` namespace. The challenge with `Microsoft.IO.RecyclableMemoryStream` and dependencies in general with the AWS SDK is our SDK is generally used at a pretty low level part of the application stack. Any dependency we take can cause collisions with our parts of the application stack that reference a different version of `Microsoft.IO.RecyclableMemoryStream`. For example with `Microsoft.IO.RecyclableMemoryStream` if we were to take on version 3.0.1 but the application itself takes a explicit dependency on 2.3.2 the application starts to get runtime missing method exceptions.

Preview 4.0.0 of `Microsoft.IO.RecyclableMemoryStream` was just released and we would expect to get similar issues between v3 and v4.

Areas we can make exceptions to our dependency rule is when the feature is in a separate opt-in package like `AWSSDK.Extensions.Logging.ILoggerAdaptor` where pulling associated `Microsoft.*` dependencies is an opt-in for an optional feature. But putting a dependency on `AWSSDK.Core` has an extremely high bar.

That being said we should look to see if we can use `Microsoft.IO.RecyclableMemoryStream` as inspiration to what we can do inside `AWSSDK.Core`. We have the internal [PooledContentStream](https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Core/Amazon.Runtime/Internal/PooledContentStream.cs) recently added that does somethings similar but in a limited use case. We can look to expand the feature set of that implementation with `Microsoft.IO.RecyclableMemoryStream` as an example.

PR #4465, opened 2026-07-11
Consolidate configuration for build targets

normj:
## Description

Consolidates the configuration of the SDK's .NET build targets so that the set of target
frameworks is defined in **one place** — the `Sdk*Targets` MSBuild properties in
`sdk/Directory.Build.props` — instead of being restated across the code generator, templates, and
`_manifest.json`.

`sdk/Directory.Build.props` now owns the target framework set:

| Property | Value | Used by |
|---|---|---|
| `SdkNetFrameworkTargets` | `net472` | .NET Framework source/projects |
| `SdkNetTargets` | `netstandard2.0;netcoreapp3.1;net8.0` | NetStandard source/projects |
| `SdkTestTargets` | `net472;net8.0` | xunit.v3 integration tests |
| `SdkMSTestAllTestTargets` | `net472;netcoreapp3.1;net8.0` | MSTest unit test projects (per-service + NetFramework uber) |
| `SdkMSTestNetTestTargets` | `netcoreapp3.1;net8.0` | NetStandard uber unit test project |
| `SdkPerformanceTargets` | `net8.0` | performance/benchmark runners |

### What changed

- **Generated project files** now emit `<TargetFrameworks>$(Sdk*Targets)</TargetFrameworks>` rather
  than literal TFM lists. Each project type maps to its property via
  `TargetFrameworkProperties.ForProjectType(...)`; a missing mapping is a hard generation error
  rather than a silent fallback.
- **Per-target-framework conditionals no longer hardcode TFM names.** `AWS_ASYNC_ENUMERABLES_API`,
  `IsTrimmable`, and `GenerateBindingRedirectsOutputType` use MSBuild framework-identifier/version
  comparisons (e.g. `$([MSBuild]::GetTargetFrameworkIdentifier('$(TargetFramework)')) != '.NETFramework'`),
  so they adapt to any current or future target automatically.
- **`SuppressTfmSupportBuildWarnings` / `GenerateBindingRedirectsOutputType`** were centralized into
  `Directory.Build.props` and removed from the generated csproj and the redundant hand-maintained
  copies.
- **Replaced the custom `NETSTANDARD20` define** with the built-in `NETSTANDARD2_0` constant, and
  consolidated `AWS_ASYNC_ENUMERABLES_API` from three per-TFM lines into a single identifier-based line.
- **Removed the `targetFrameworks` lists from `_manifest.json`** (both the `projects[]` items and the
  `unittestprojects` object form). Unit-test project generation now derives project-type
  references/excludes from framework-identifier conditions instead of a per-TFM map, which also
  corrected the test projects so netcoreapp3.1/netstandard2.0 are only targeted where appropriate.
- **The `AssemblyInfo` / `CoreAssemblyInfo` `#if/#elif` `AssemblyDescription` chain is now generated
  from the `Sdk*Targets` values** in `Directory.Build.props`. Previously each target needed a
  hand-added `#elif` branch (with an `#else #error` that broke the build for any new target); the
  branches and their labels are now produced automatically.

### Forward-looking code fixes (.NET 10 / 11 readiness)

While validating the consolidation by test-building with **.NET 10 and .NET 11** target frameworks, a
few source-level issues surfaced. These fixes are **not required by the current target set** (net472,
netstandard2.0, netcoreapp3.1, net8.0) but are included here so the SDK is ready when those targets
are added:

- **`sdk/src/Core/Amazon.Runtime/Pipeline/HttpHandler/AmazonSecurityProtocolManager.cs`** — suppress
  `SYSLIB0014` where the obsolete `SecurityProtocolType` API is used for older-framework compatibility
  (on newer .NET this code path is a no-op).
- **`sdk/src/Core/Amazon.Util/AWSSDKUtils.cs`** — add a `[RequiresUnreferencedCode]` trimming
  annotation (under `NET8_0_OR_GREATER`) on the reflection-based `IsSet` helper.
- **`sdk/src/Services/SimpleNotificationService/Custom/Util/Message.cs`** — on `NET10_0_OR_GREATER`,
  load the signing certificate via `X509CertificateLoader.LoadCertificate(...)` instead of the
  `new X509Certificate2(...)` constructor, which is obsolete in .NET 10.

### Remaining work (not in this PR)

There is still more consolidation to do — most notably the **NuGet `.nuspec` generation**
(`CoreNuspec.tt` / `Nuspec.tt`), which still hardcodes the per-TFM `lib/<tfm>` file mappings and
dependency groups. That is a separate follow-up.

**Important:** with the current `Directory.Build.props` settings, the exact same set of assemblies is
built as is shipped today (net472, netstandard2.0, netcoreapp3.1, net8.0). So the currently generated
nuspec files remain correct and packaging is unaffected by this PR — the nuspec consolidation only
becomes necessary when the target set actually changes.

## Motivation and Context

Changing the SDK's .NET target frameworks previously required editing many places (the generator
code, T4 templates, and `_manifest.json`), which was error-prone and easy to get out of sync. The
goal of this work is that adding or removing a .NET target requires editing only
`sdk/Directory.Build.props`, with the generator adapting automatically.

## Testing

- Rebuilt the code generator and regenerated all project/source files; verified the working-tree diff
  is limited to the intended target-framework/define changes.
- Built representative projects across all target frameworks and project types (Core and a sample
  service, both NetStandard and NetFramework variants; per-service and uber unit-test projects) — all
  compile with 0 errors, and net472 selects the correct `#if NET472` branch rather than the
  `#else #error`.
- **Add-a-target simulation:** temporarily added `net9.0` to `SdkNetTargets`, regenerated, and
  confirmed the generated projects targeted it and a new `#elif NET9_0` AssemblyDescription branch was
  produced — with no generator or template edits — then reverted.

### Dry-runs

- **DotNet Dry-run ID:**
  - [ ] Pending
  - [x] Completed successfully
  - [ ] Failed
- **PowerShell Dry-run ID:**
  - [ ] Pending
  - [x] Completed successfully
  - [ ] Failed

This is a build-infrastructure/code-generation change. The set of shipped assemblies and target
frameworks is unchanged from today, so there is no customer-facing runtime or packaging impact. One
minor generated-metadata change: the net472 `AssemblyDescription` text now reads
`.NET Framework 4.7.2` instead of `4.7.2`.

## Types of changes

## Checklist

## License


normj:
Put in draft while I push through dry runs.

normj:
Taking out of draft now that dry run is successful: DRY_RUN-4f74514a-9e06-42aa-9dfc-7a367a22826b

sdk/Directory.Build.props
```diff
@@ -8,5 +8,45 @@
    <!-- NuGet audit vulnerabilities are handled by a separate CI pipeline, not the build -->
    <NuGetAuditMode>direct</NuGetAuditMode>
    <WarningsNotAsErrors>$(WarningsNotAsErrors);NU1901;NU1902;NU1903;NU1904</WarningsNotAsErrors>
+   
+   <LangVersion>12.0</LangVersion>
```
dscpinheiro, line 12: Is there a reason you chose `12.0` here? Previously it was either `9.0` / `11.0` (and the latter is still hard-coded in some places such as `sdk/test/UnitTests/AWSSDK.UnitTests.Core.csproj` and `sdk/test/Common/AWSSDK.CommonTest.csproj`)

generator/ServiceClientGeneratorLib/Generators/ProjectFiles/VS2017ProjectFile.tt
dscpinheiro, line 1: This is a nit, but I ran the generator locally and a bunch of extra (empty) lines were added to every project file:

image

Can we remove them?

dscpinheiro (commented):
I knew of one place where the TFM list / conditions were hard-coded (`buildtools/ZipArtifactGenerator/aws-sdk-zip-artifact.csproj`), but after running the generator locally I found these others:
- `sdk/src/Core/AWSSDK.Core.*.csproj` are still using `net472` / `netstandard2.0;netcoreapp3.1;net8.0` and the existing `NETSTANDARD20` constant
- `CloudFront.SignersTests` also has `netcoreapp3.1` conditions (but no comment as the one added in `AWSSDK.Extensions.NETCore.Setup`)
- `sdk/test/Common/AWSSDK.CommonTest.csproj` has `Condition="'$(TargetFramework)' != 'net472'"`
- All the integration test projects (which are hand-written) have both `Condition="'$(TargetFramework)' == 'net472'"` and `Condition="'$(TargetFramework)' != 'net472'"`

I assume we'd need to update them too?

sdk/test/UnitTests/AWSSDK.UnitTests.Core.csproj
```diff
@@ -8,11 +8,12 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
    <RunAnalyzersDuringBuild Condition="'$(RunAnalyzersDuringBuild)'==''">true</RunAnalyzersDuringBuild>
-    <TargetFrameworks>net472;netstandard2.0;netcoreapp3.1;net8.0</TargetFrameworks>
+    <TargetFrameworks>$(SdkTestTargets)</TargetFrameworks>
```
AlexDaines: This one is an MSTest project (MSTest.TestAdapter/TestFramework below) but got `$(SdkTestTargets)`, the xunit.v3 integration-test set — evaluated TargetFrameworks go from `net472;netstandard2.0;netcoreapp3.1;net8.0` at the merge base to `net472;net8.0` here, so the core unit tests stop running on netcoreapp3.1. CommonTest and UnitTestUtilities kept the full set, and the props comment says MSTest projects keep netcoreapp3.1 — so I'm guessing this wanted `$(SdkMSTestAllTestTargets)`, but maybe dropping 3.1 for the core unit tests was the intent?

sdk/Directory.Build.props
```diff
@@ -8,5 +8,45 @@
...
+   <SdkTestXunitv2NetTargets>netcoreapp3.1;net8.0</SdkTestXunitv2NetTargets>
+
+   <!--
+     MSTest-based unit test targets. Unlike the xunit.v3 integration tests (SdkTestTargets), MSTest
+     supports netcoreapp3.1, so the unit tests continue to cover it. SdkMSTestAllTestTargets is the full
+     set (used by the per-service and .NET Framework unit test projects); SdkMSTestNetTestTargets is the
+     non-.NET-Framework slice used by the NetStandard uber unit test project.
+   -->
+   <SdkMSTestAllTestTargets>net472;netcoreapp3.1;net8.0</SdkMSTestAllTestTargets>
+   <SdkMSTestNetTestTargets>netstandard2.0;netcoreapp3.1;net8.0</SdkMSTestNetTestTargets>
```
AlexDaines, line 48: The value here (`netstandard2.0;netcoreapp3.1;net8.0`) and its docs disagree: the PR description's table and the two new code comments (ProjectFileConfiguration.cs:25-27, UnitTestProjectFileCreator.cs:196-198) all say `netcoreapp3.1;net8.0`, with the last noting netstandard2.0 is excluded because tests can't run on it. Which one is the intent? The value matches the committed AWSSDK.UnitTests.NetStandard.csproj (and ProtocolTests.NetStandard compile-targets netstandard2.0 the same way), so I'd lean toward keeping the value and updating the comments — but either way, one of them should win before the next regen decides silently.

AlexDaines (changes requested):
For the "I assume we'd need to update them too?" question, I ran a script over `git ls-files` at HEAD matching literal TFM lists, literal `$(TargetFramework)` comparisons, and constants gated on them. The updated template emits a generated-file marker and 0 files at HEAD carry it, so the bulk regeneration isn't committed. Every generated csproj picks up the new form on the next `generate` run and is not a straggler.

Beyond the ones already listed (Core csproj pair, CommonTest, UnitTestUtilities, the integration-test net472 conditions, extensions, ZipArtifactGenerator), the search turned up one more: `sdk/test/Services/Polly/IntegrationTests/AWSSDK.IntegrationTests.Polly.csproj`. It still has the literal `net472;net8.0` (line 3) and `GenerateBindingRedirectsOutputType` (line 19) that the other 13 integration-test csproj were converted away from, and it's hand-written so regeneration won't pick it up. I'm guessing it just slipped through, but flagging in case it was deliberate. (Analyzer csproj are intentionally single-target netstandard2.0 for Roslyn and don't count.)

I also ran MSBuild evaluation per TFM against both this head and the merge base. `AWS_ASYNC_ENUMERABLES_API`, `IsTrimmable`, and the NETCore.Setup conditions all evaluate identically to before. No hand-written code is stranded on the old `NETSTANDARD20` constant. The only remaining consumers are generated AssemblyInfo.cs files, which regenerate.

Two inline comments below: a property mapping that drops netcoreapp3.1 from the core unit tests (question there about whether that was intended), and a value-vs-comment mismatch on `SdkMSTestNetTestTargets`. I'm marking request-changes for those two plus Polly. Everything else I checked came back clean.

sdk/Directory.Build.props
AlexDaines: I was curious about this too, so I ran the evaluation to see what the 12.0 does today — turns out almost nothing yet, and the interesting part is the next regeneration:

- UnitTests.Core kept an unconditional `<LangVersion>11.0</LangVersion>`, which shadows the central 12.0. Collapsing the old net8.0=11.0 / else=9.0 pair also moved net472 from C#9 to C#11 (base net472=9.0, head net472=11.0). Same leftover in CommonTest, UnitTestUtilities, and the uber test projects.
- The committed generated csproj still carry the old 9.0/11.0 pairs (S3.NetStandard evaluates to 9.0 at head), and the template no longer emits LangVersion — so the next full regen takes every generated project from C#9/11 to C#12 in one step.
- The new extensions Directory.Build.props import also flows 12.0 into every extensions project without its own override.

Nothing breaks today, but the 12.0 is shadowed where the leftovers exist and takes effect repo-wide on regen. Might be worth either sweeping the per-project LangVersion leftovers here or holding the central LangVersion for a follow-up, so the two halves land together — though if the C#12 jump on regen is the plan, this all works as-is.

generator/ServiceClientGeneratorLib/Generators/ProjectFiles/VS2017ProjectFile.tt
AlexDaines: I chased down where these come from — it's inherent T4 rendering, so .tt content edits alone can't get rid of them. Literal blank lines in the template (lines 26, 37, 46, 53) render verbatim, and each standalone `<#` / `#>` / `}` control line leaves its terminating newline, so a block that emits nothing still yields a blank output line. GeneratorDriver.WriteFile only trims edge whitespace (line 1338), never internal. One option for a single-place fix: collapse runs of 3+ newlines at the csproj write site (ProjectFileCreator.cs, GenerateVS2017ProjectFile), e.g. `Regex.Replace(content, "(\\r?\\n){3,}", "$1$1")`, scoped to csproj output so .cs generation keeps its blank lines.

sdk/test/UnitTests/AWSSDK.UnitTests.Core.csproj
dscpinheiro: This is a good catch. This specific project only runs during preview builds so the dry-run didn't catch it.

sdk/Directory.Build.props
normj: I set it to `12.0` because when I did a test build adding .NET 10 and 11 I got some compiler errors because the RegEx source generator that we use will use some C# 12 features when generating source for those targets. So setting it to `12.0` was getting ready for new .NET targets.

sdk/test/UnitTests/AWSSDK.UnitTests.Core.csproj
normj: Fixed

sdk/Directory.Build.props
normj: Added a comment that `netstandard2.0` is here to maintain the behavior before this work. In the future we need to rethink how we test the `netstandard2.0` target.

normj:
> I knew of one place where the TFM list / conditions were hard-coded (`buildtools/ZipArtifactGenerator/aws-sdk-zip-artifact.csproj`), but after running the generator locally I found these others:
> 
> * I was thinking the core csproj files were still generated so I didn't check them in. I fixed that.
> * `CloudFront.SignersTests` also has `netcoreapp3.1` conditions (but no comment as the one added in `AWSSDK.Extensions.NETCore.Setup`)
> * `sdk/test/Common/AWSSDK.CommonTest.csproj` has `Condition="'$(TargetFramework)' != 'net472'"`
> * All the integration test projects (which are hand-written) have both `Condition="'$(TargetFramework)' == 'net472'"` and `Condition="'$(TargetFramework)' != 'net472'"`
> 
> I assume we'd need to update them too?

* For the `AWSSDK.Core.*.csproj` files were you looking at them before running the generator? Core project file is generated and I didn't check in the generated project files.
* Added a comment that the `ItemGroup` with the conditional `netcoreapp3.1` can be dropped once we take out support for 3.1. If we do forget to take it out when we drop 3.1 it won't cause any problems.
* Fixed `AWS.CommonTest.csproj` to use `$([MSBuild]::GetTargetFrameworkIdentifier('$(TargetFramework)')) != '.NETFramework'` to identify it is .NET Framework. 
* Fixed up the handcoded test projects that had conditional checks for `net472` to use the same .NET Framework check.

Fixed the extra lines in the generated csproj file. I think Claude preferred `\n` over windows `\r\n` and that messed up the tt to cs conversion tool.

Fixed the `sdk/test/Services/Polly/IntegrationTests/AWSSDK.IntegrationTests.Polly.csproj` project that was for some reason not updated to use the new MSBuild property.

normj:
New dry run was successful: DRY_RUN-cfcd8d29-36e9-4ec5-ab0e-634e6820b4c5

dscpinheiro:
> For the AWSSDK.Core.*.csproj files were you looking at them before running the generator?

I thought so! But probably just missed it. There's one more file from my original comment with hard-coded TFMs still: https://github.com/aws/aws-sdk-net/blob/main/buildtools/ZipArtifactGenerator/aws-sdk-zip-artifact.csproj#L7 (and a few other places checking for `net472`, `netcoreapp3.1`, etc...)

It's not in the `sdk` folder, but could we use an import like the extensions do?
```xml
<Project>
  <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../../sdk'))" />
</Project>
```

generator/ServiceClientGeneratorLib/ProjectFileConfiguration.cs
```diff
@@ -8,6 +8,41 @@
...
+    public static class TargetFrameworkProperties
+    {
+        public const string NetFrameworkProperty = "$(SdkNetFrameworkTargets)";
+        public const string NetStandardProperty = "$(SdkNetTargets)";
+
+        /// <summary>
+        /// MSTest-based unit test target frameworks. Unlike the xunit.v3 integration tests (which use
+        /// $(SdkTestTargets)), MSTest supports netcoreapp3.1. MSTestAllProperty is the full set
+        /// (net472;netcoreapp3.1;net8.0) used by the per-service and .NET Framework unit test projects;
+        /// MSTestNetProperty is the non-.NET-Framework slice (netcoreapp3.1;net8.0) used by the NetStandard
```
AlexDaines: Comment still says (netcoreapp3.1;net8.0), but the property keeps netstandard2.0 now. Same wording in UnitTestProjectFileCreator.cs:197.

sdk/Directory.Build.props
```diff
@@ -8,5 +8,51 @@
...
+   <!--
+     MSTest-based unit test targets. Unlike the xunit.v3 integration tests (SdkTestTargets), MSTest
+     supports netcoreapp3.1, so the unit tests continue to cover it. SdkMSTestAllTestTargets is the full
+     set (used by the per-service and .NET Framework unit test projects); SdkMSTestNetTestTargets is the
+     non-.NET-Framework slice used by the NetStandard uber unit test project.
+   -->
+   <SdkMSTestAllTestTargets>net472;netcoreapp3.1;net8.0</SdkMSTestAllTestTargets>
+   
+   <!--
+   In this framework list for test projects netstandard2.0 is maintain the behavior before target consolidation.
```
AlexDaines: "netstandard2.0 is maintain the behavior" reads garbled. Maybe "is kept to maintain".

AlexDaines (dismissed):
All three addressed, dry run green after the commit. Approving.

sdk/Directory.Build.props
normj: Done

generator/ServiceClientGeneratorLib/ProjectFileConfiguration.cs
normj: I removed the whole block comment because we the source of truth should be `Directory.build.props` and this comment is going to just go out of date. Actually the code will hopefully just go away once we move over to the new Smithy generator.

AlexDaines (approved):

dscpinheiro (approved):

normj:
Latest dry run was successful: DRY_RUN-13544da5-d4e0-422a-af16-8f56d85c1a82

PR #4466, opened 2026-07-11
CPU and memory improvements on XmlUnmarshallerContext

JanHyka:
## Description
Refactor XmlUnmarshallerDevConfig, build path stack as ancestor path and avoid repeated string rematerializations.

## Motivation and Context
Lower cpu and gc pressure on xml unmarshaller.

### Benchmark
Before:

| Method        | Shape | ItemCount | Mean      | Gen0    | Gen1   | Allocated |
|-------------- |------ |---------- |----------:|--------:|-------:|----------:|
| **ParseDocument** | **empty** | **10**        |  **8.597 μs** |  **2.1362** | **0.1221** |  **40.29 KB** |
| **ParseDocument** | **empty** | **100**       | **71.417 μs** | **13.6719** | **0.4883** | **252.64 KB** |
| **ParseDocument** | **text**  | **10**        | **11.156 μs** |  **2.6855** | **0.1221** |  **49.98 KB** |
| **ParseDocument** | **text**  | **100**       | **97.764 μs** | **18.5547** | **0.9766** | **349.51 KB** |

After:

| Method        | Shape | ItemCount | Mean      | Gen0   | Gen1   | Allocated |
|-------------- |------ |---------- |----------:|-------:|-------:|----------:|
| **ParseDocument** | **empty** | **10**        |  **3.274 μs** | **0.9766** |      **-** |  **18.76 KB** |
| **ParseDocument** | **empty** | **100**       | **18.776 μs** | **2.1973** |      **-** |  **44.85 KB** |
| **ParseDocument** | **text**  | **10**        |  **4.707 μs** | **1.3428** | **0.0610** |  **24.85 KB** |
| **ParseDocument** | **text**  | **100**       | **36.416 μs** | **5.3711** |      **-** | **104.94 KB** |

After the modification the pass is, based on scenario 2.3-3.8x faster while consuming less than 50% memory.

Benchmark code
<details>

```

public class XmlUnmarshallerContextBenchmarks
{
    private byte[] _payload = null!;

    [Params("text", "empty")]
    public string Shape { get; set; } = null!;

    /// <summary>Number of repeated <c>member</c> entries in the list payload.</summary>
    [Params(10, 100)]
    public int ItemCount { get; set; }

    [GlobalSetup]
    public void Setup()
    {
        var sb = new StringBuilder();
        sb.Append("<?xml version=\"1.0\" encoding=\"UTF-8\"?>");
[...]

PR #4467, opened 2026-07-13
Eliminates per node path allocation in XML and AWSQuery response unmarshalling.

muhammad-othman:
## Description

  Eliminates the per-node path-string allocation in the **XML and AWSQuery response unmarshalling** pipeline by replacing the full path-string rebuild on every `Read()` with an incrementally-maintained, pooled character buffer and allocation-free span matching.

  In unmarshalling benchmarks this reduces **time by up to ~66% and allocations by up to ~95%** on large responses (e.g. `awsQuery_GetMetricDataResponse_L`: **304 MB → 14.6 MB allocated, 171.9 ms → 57.6 ms**), and the win survives end-to-end: a full `GetMetricData` round-trip is **53-62% faster and allocates 72–75% less**. See the full before/after tables in [Testing](#testing).

  Previously `XmlUnmarshallerContext` tracked the current element path as a `Stack<string>` plus a `stackString`, and on **every** `Read()` it rebuilt the entire slash-joined path string (`StackToPath` is a `stack.ToArray()` + `string.Format` cascade over every segment). Generated unmarshallers call `context.TestExpression("path/to/element")` once per node, and each call compared against that freshly-rebuilt string. For a document with N nodes at depth D, this allocated a fresh path string per node, producing super-linear (O(N·D)) time and allocation, the source of the ~304 MB / 172 ms on the large response.

This PR mirrors the approach the JSON context already uses. The XML context now maintains a single reusable `char[]` rented from `ArrayPool<char>.Shared`: `Push` appends `/segment` and `Pop` trims it, with the full path string materialized **lazily and cached** only when actually read. The boolean results are **identical** to the previous string path validated by `XmlUnmarshallerContextPathTests` . The buffer returns to the `ArrayPool` when the context is disposed.

## Motivation and Context
 `DOTNET-8588`

## Testing

  - Added `XmlUnmarshallerContextPathTests`.
  
  - Measured using the AWS Query and RestXml serde/service benchmarks. The results below compare the branch **without** this change (`before`) against the branch **with** it (`after`):

  #### Serde benchmarks - AWS Query

  **Before**

  | Method                                 | Mean             | Error           | StdDev          | Max              | P50              | P90              | P95              | Gen0       | Gen1      | Allocated    |
  |--------------------------------------- |-----------------:|----------------:|----------------:|-----------------:|-----------------:|-----------------:|-----------------:|-----------:|----------:|-------------:|
  | awsQuery_PutMetricDataRequest_Baseline |         385.5 ns |         6.19 ns |         3.24 ns |         390.3 ns |         384.8 ns |         389.1 ns |         389.7 ns |     0.1073 |         - |      1.32 KB |
  | awsQuery_PutMetricDataRequest_L        |     344,353.8 ns |     5,460.08 ns |       844.95 ns |     345,085.8 ns |     344,570.4 ns |     345,014.5 ns |     345,050.2 ns |    20.9961 |    6.3477 |    262.84 KB |
  | awsQuery_GetMetricDataRequest_L        |     258,520.2 ns |     2,590.37 ns |       400.86 ns |     259,106.3 ns |     258,366.5 ns |     258,909.1 ns |     259,007.7 ns |    19.5313 |    5.8594 |    240.43 KB |
  | awsQuery_GetMetricDataResponse_S       |      24,647.0 ns |     1,817.96 ns |     2,093.56 ns |      28,844.2 ns |      24,235.8 ns |      26,979.9 ns |      28,672.8 ns |     4.5166 |    0.2441 |     55.72 KB |
  | awsQuery_GetMetricDataResponse_M       |   1,702,363.8 ns |    34,025.61 ns |     5,265.50 ns |   1,707,707.6 ns |   1,702,728.4 ns |   1,707,100.5 ns |   1,707,404.1 ns |   261.7188 |   31.2500 |   3210.47 KB |
  | awsQuery_GetMetricDataResponse_L       | 171,898,520.4 ns | 3,407,717.75 ns | 3,646,219.82 ns | 179,693,366.7 ns | 172,442,816.7 ns | 175,911,450.0 ns | 177,257,493.3 ns | 25333.3333 | 1000.0000 | 311795.04 KB |

  **After**

  | Method                                 | Mean            | Error           | StdDev        | Max             | P50             | P90             | P95             | Gen0      | Gen1     | Allocated   |
  |--------------------------------------- |----------------:|----------------:|--------------:|----------------:|----------------:|----------------:|----------------:|----------:|---------:|------------:|
  | awsQuery_PutMetricDataRequest_Baseline |        395.7 ns |        19.03 ns |      21.15 ns |        443.9 ns |        390.5 ns |        427.1 ns |        437.4 ns |    0.1073 |        - |     1.32 KB |
  | awsQuery_PutMetricDataRequest_L        |    359,965.8 ns |     8,186.80 ns |   9,427.94 ns |    378,610.1 ns |    355,839.0 ns |    375,768.6 ns |    377,840.2 ns |   20.9961 |   6.3477 |   262.84 KB |
  | awsQuery_GetMetricDataRequest_L        |    262,248.3 ns |     5,151.35 ns |   3,407.30 ns |    270,116.2 ns |    261,416.6 ns |    264,993.3 ns |    267,554.7 ns |   19.5313 |   5.8594 |   240.43 KB |
  | awsQuery_GetMetricDataResponse_S       |      9,027.7 ns |       417.32 ns |     480.59 ns |     10,136.9 ns |      8,820.9 ns |      9,711.0 ns |      9,777.5 ns |    1.4343 |   0.0763 |    17.76 KB |
  | awsQuery_GetMetricDataResponse_M       |    576,731.7 ns |     9,729.86 ns |   3,469.76 ns |    580,381.7 ns |    577,537.0 ns |    579,920.3 ns |    580,151.0 ns |   13.6719 |   0.9766 |   178.04 KB |
  | awsQuery_GetMetricDataResponse_L       | 57,597,468.5 ns | 1,086,302.98 ns | 387,386.11 ns | 58,270,355.6 ns | 57,510,400.0 ns | 58,026,127.8 ns | 58,148,241.7 ns | 1222.2222 | 777.7778 | 14998.38 KB |

  **Response unmarshalling delta (AWS Query):**

  | Method                            | Time before      | Time after      | Δ time  | Alloc before  | Alloc after | Δ alloc         |
  |---------------------------------- |-----------------:|----------------:|--------:|--------------:|------------:|----------------:|
  | awsQuery_GetMetricDataResponse_S  |      24,647 ns    |      9,028 ns   | −63.4%  |    55.72 KB   |   17.76 KB  | −68.1%          |
  | awsQuery_GetMetricDataResponse_M  |   1,702,364 ns    |    576,732 ns   | −66.1%  |  3,210.47 KB  |  178.04 KB  | −94.5% (18.0×)  |
  | awsQuery_GetMetricDataResponse_L  | 171,898,520 ns    | 57,597,469 ns   | −66.5%  | 311,795 KB    | 14,998 KB   | −95.2% (20.8×)  |

  The `…Request_*` marshalling rows are unchanged (byte-identical allocations, timing within noise), confirming this PR touches only the response unmarshalling path. The allocation win grows with response size (18× → 21× as M → L).

  ### Serde benchmarks - RestXml

  **Before**

  | Method                             | Mean         | Error       | StdDev       | Max          | P50          | P90          | P95          | Gen0     | Gen1     | Gen2     | Allocated  |
  |----------------------------------- |-------------:|------------:|-------------:|-------------:|-------------:|-------------:|-------------:|---------:|---------:|---------:|-----------:|
  | restXml_CopyObjectRequest_M        |   1,030.8 ns |    34.67 ns |     38.53 ns |   1,103.6 ns |   1,025.5 ns |   1,092.8 ns |   1,095.0 ns |   0.2632 |   0.0019 |        - |    3.24 KB |
  | restXml_CopyObjectOutput_Baseline  |   1,103.1 ns |    41.85 ns |     46.51 ns |   1,199.9 ns |   1,081.2 ns |   1,191.8 ns |   1,198.9 ns |   1.1368 |   0.0534 |        - |   13.93 KB |
  | restXml_CopyObjectOutput_M         |   3,438.6 ns |   440.29 ns |    507.04 ns |   4,444.4 ns |   3,255.0 ns |   4,289.6 ns |   4,345.5 ns |   1.3237 |   0.0687 |        - |   16.24 KB |
  | restXml_GetObject_M                |  10,637.3 ns |   437.47 ns |    468.09 ns |  11,580.1 ns |  10,560.1 ns |  11,312.5 ns |  11,435.2 ns |  18.4326 |  18.0969 |  18.0969 |  103.95 KB |
  | restXml_GetObject_L                | 144,962.9 ns | 9,359.17 ns | 10,014.20 ns | 167,947.0 ns | 143,444.9 ns | 156,568.6 ns | 158,326.2 ns | 166.9922 | 166.5039 | 166.5039 | 1028.36 KB |

  **After**

  | Method                             | Mean         | Error       | StdDev      | Max          | P50          | P90          | P95          | Gen0     | Gen1     | Gen2     | Allocated |
  |----------------------------------- |-------------:|------------:|------------:|-------------:|-------------:|-------------:|-------------:|---------:|---------:|---------:|----------:|
  | restXml_CopyObjectRequest_M        |   1,059.4 ns |    50.02 ns |    55.59 ns |   1,187.5 ns |   1,042.5 ns |   1,146.0 ns |   1,157.4 ns |   0.2632 |   0.0019 |        - |   3.24 KB |
  | restXml_CopyObjectOutput_Baseline  |   1,072.1 ns |    21.38 ns |    21.96 ns |   1,112.6 ns |   1,069.1 ns |   1,094.5 ns |   1,099.3 ns |   1.1292 |   0.0381 |        - |  13.84 KB |
  | restXml_CopyObjectOutput_M         |   2,340.4 ns |    42.47 ns |    35.47 ns |   2,410.4 ns |   2,334.7 ns |   2,381.5 ns |   2,395.4 ns |   1.2054 |   0.0191 |        - |  14.79 KB |
  | restXml_GetObject_M                |  11,641.1 ns |   605.53 ns |   621.84 ns |  12,911.6 ns |  11,719.8 ns |  12,341.6 ns |  12,518.7 ns |  19.0277 |  18.6920 |  18.6920 |    104 KB |
  | restXml_GetObject_L                | 153,763.4 ns | 3,064.34 ns | 2,392.43 ns | 159,123.8 ns | 153,211.3 ns | 155,663.1 ns | 157,224.6 ns | 164.5508 | 164.0625 | 164.0625 | 1028.4 KB |

  `CopyObjectOutput_M` is **−31.9% time / −8.9% allocation** (3,439 ns → 2,340 ns, 16.24 KB → 14.79 KB). The `GetObject_*`  body is a raw binary blob with essentially no XML element structure, so the cost is the stream copy, not path tracking and they are flat within noise (`GetObject_L` StdDev ≈ 10 µs before).

  #### End-to-end benchmarks - AWS Query

  **Before**

  | Method                       | Mean         | Error      | StdDev     | Max          | P50          | P90          | P95          | Gen0    | Gen1   | Allocated |
  |----------------------------- |-------------:|-----------:|-----------:|-------------:|-------------:|-------------:|-------------:|--------:|-------:|----------:|
  | awsQuery_e2e_Healthcheck     |  17,294.0 ns |   252.0 ns |   223.4 ns |  17,646.0 ns |  17,294.7 ns |  17,562.1 ns |  17,602.4 ns |  3.2043 | 1.0681 |  39.34 KB |
  | awsQuery_e2e_PutMetricData_S |  19,479.1 ns |   342.1 ns |   320.0 ns |  20,123.4 ns |  19,419.0 ns |  19,818.3 ns |  19,950.5 ns |  3.3264 | 0.9155 |  41.04 KB |
  | awsQuery_e2e_PutMetricData_M |  30,679.4 ns |   391.5 ns |   326.9 ns |  31,228.7 ns |  30,721.7 ns |  30,977.8 ns |  31,082.4 ns |  4.0283 | 1.3428 |  49.54 KB |
  | awsQuery_e2e_GetMetricData_S |  31,624.0 ns |   535.4 ns |   657.6 ns |  32,794.2 ns |  31,507.4 ns |  32,288.0 ns |  32,605.8 ns |  5.1880 | 1.7090 |   64.2 KB |
  | awsQuery_e2e_GetMetricData_M | 111,192.0 ns | 1,809.3 ns | 1,510.8 ns | 114,892.5 ns | 111,072.7 ns | 112,395.5 ns | 113,436.0 ns | 16.9678 | 3.1738 | 208.57 KB |

  **After**

  | Method                       | Mean        | Error    | StdDev     | Median      | Max         | P50         | P90         | P95         | Gen0   | Gen1   | Allocated |
  |----------------------------- |------------:|---------:|-----------:|------------:|------------:|------------:|------------:|------------:|-------:|-------:|----------:|
  | awsQuery_e2e_Healthcheck     | 17,005.7 ns | 325.4 ns |   348.2 ns | 16,888.6 ns | 17,866.6 ns | 16,888.6 ns | 17,496.9 ns | 17,582.1 ns | 3.1128 | 1.0376 |  38.47 KB |
  | awsQuery_e2e_PutMetricData_S | 19,127.1 ns | 377.3 ns |   660.8 ns | 18,844.1 ns | 21,094.3 ns | 18,844.1 ns | 19,971.0 ns | 20,691.7 ns | 3.2654 | 1.0681 |  40.08 KB |
  | awsQuery_e2e_PutMetricData_M | 32,380.1 ns | 646.2 ns | 1,484.8 ns | 32,054.1 ns | 36,135.0 ns | 32,054.1 ns | 34,259.4 ns | 35,746.6 ns | 3.9368 | 1.3123 |  48.59 KB |
  | awsQuery_e2e_GetMetricData_S | 24,596.0 ns | 609.1 ns | 1,767.1 ns | 24,027.1 ns | 29,120.9 ns | 24,027.1 ns | 27,330.3 ns | 28,149.4 ns | 3.4180 | 1.1292 |  41.97 KB |
  | awsQuery_e2e_GetMetricData_M | 52,403.2 ns | 789.6 ns |   738.6 ns | 52,345.4 ns | 53,825.7 ns | 52,345.4 ns | 53,186.1 ns | 53,411.4 ns | 4.2725 | 1.4038 |  53.08 KB |

The response-heavy operations improve substantially: `GetMetricData_S` 31.6 µs → 24.6 µs (−22%), 64.2 KB → 42.0 KB (−35%); `GetMetricData_M` 111.2 µs → 52.4 µs (−53%), 208.6 KB → 53.1 KB (−75%). The marshal-only operations (`Healthcheck`, `PutMetricData_*`) are flat within noise.

  ### End-to-end benchmarks - RestXml

  **Before**

  | Method                      | Mean           | Error       | StdDev      | Max            | P50            | P90            | P95            | Gen0     | Gen1     | Gen2     | Allocated  |
  |---------------------------- |---------------:|------------:|------------:|---------------:|---------------:|---------------:|---------------:|---------:|---------:|---------:|-----------:|
  | restXml_e2e_GetMetricData_S |    29,650.2 ns |    409.1 ns |    362.7 ns |    30,268.0 ns |    29,536.1 ns |    30,183.2 ns |    30,248.7 ns |   6.1646 |   1.5259 |        - |   75.84 KB |
  | restXml_e2e_GetMetricData_M |    91,386.2 ns |  1,739.4 ns |  1,861.2 ns |    96,302.1 ns |    91,169.9 ns |    93,034.5 ns |    93,582.3 ns |  17.3340 |   2.8076 |        - |  213.43 KB |
  | restXml_e2e_GetObject_M     |    37,518.9 ns |    737.8 ns |  1,540.0 ns |    41,446.2 ns |    37,318.6 ns |    39,865.7 ns |    40,300.6 ns |  25.3296 |  23.1323 |  23.0103 |  128.31 KB |
  | restXml_e2e_PutObject_M     |   272,210.0 ns |  6,351.1 ns | 17,704.4 ns |   316,904.4 ns |   268,200.4 ns |   297,685.4 ns |   312,109.6 ns |   2.4414 |   0.9766 |        - |   30.23 KB |

  **After**

  | Method                      | Mean           | Error       | StdDev      | Median         | Max            | P50            | P90            | P95            | Gen0     | Gen1     | Gen2     | Allocated  |
  |---------------------------- |---------------:|------------:|------------:|---------------:|---------------:|---------------:|---------------:|---------------:|---------:|---------:|---------:|-----------:|
  | restXml_e2e_GetMetricData_S |    21,169.7 ns |    305.1 ns |    238.2 ns |    21,185.5 ns |    21,510.0 ns |    21,185.5 ns |    21,384.4 ns |    21,442.8 ns |   4.4556 |   1.4648 |        - |   54.62 KB |
  | restXml_e2e_GetMetricData_M |    34,453.1 ns |    456.6 ns |    404.8 ns |    34,309.8 ns |    35,048.9 ns |    34,309.8 ns |    34,994.0 ns |    35,029.4 ns |   4.7607 |   1.5869 |        - |   58.96 KB |
  | restXml_e2e_GetObject_M     |    37,582.3 ns |    742.7 ns |  1,358.1 ns |    37,411.7 ns |    41,337.9 ns |    37,411.7 ns |    39,623.6 ns |    39,913.3 ns |  26.3672 |  24.1089 |  23.9868 |  128.89 KB |
  | restXml_e2e_PutObject_M     |   253,140.7 ns |  4,827.5 ns |  6,105.2 ns |   251,656.5 ns |   268,957.5 ns |   251,656.5 ns |   260,418.9 ns |   264,477.1 ns |   2.4414 |   0.9766 |        - |    30.8 KB |

  `GetMetricData_S` 29.7 µs → 21.2 µs (−29%), 75.8 KB → 54.6 KB (−28%); `GetMetricData_M` 91.4 µs → 34.5 µs (−62%), 213.4 KB → 59.0 KB (−72%). The binary-body operations (`GetObject_M`, `PutObject_M`) are flat within noise.

### Dry-runs

- **DotNet Dry-run ID:** af54336a-46cf-434e-a9af-43036d471048
  - [ ] Pending
  - [x] Completed successfully
  - [ ] Failed
- **PowerShell Dry-run ID:** 1e6192be-574c-4f6a-9a35-2452898153ad
  - [ ] Pending
  - [x] Completed successfully
  - [ ] Failed

## Types of changes

## Checklist

## License

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Internal/Transform/XmlUnmarshallerContext.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Transform/XmlUnmarshallerContext.cs
@@ -1,5 +1,6 @@
 ﻿using Amazon.Runtime.Internal.Util;
 using System;
+using System.Buffers;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
@@ -46,8 +47,8 @@ namespace Amazon.Runtime.Internal.Transform
 
         private StreamReader streamReader;
         private XmlTextReader _xmlTextReader;
-        private Stack<string> stack = new Stack<string>();
-        private string stackString = "";
+        private readonly XmlPathBuffer pathBuffer = new XmlPathBuffer();
+        private string currentAttributeName;
         private Dictionary<string, string> attributeValues;
         private List<string> attributeNames;
         private IEnumerator<string> attributeEnumerator;
@@ -158,7 +159,12 @@ namespace Amazon.Runtime.Internal.Transform
         /// </summary>
         public override string CurrentPath
         {
-            get { return this.stackString; }
+            get
+            {
+                return this.currentAttributeName == null
+                    ? this.pathBuffer.CurrentPath
+                    : this.pathBuffer.BuildAttributePath(this.currentAttributeName);
+            }
         }
 
         /// <summary>
@@ -167,7 +173,16 @@ namespace Amazon.Runtime.Internal.Transform
         /// </summary>
         public override int CurrentDepth
         {
-            get { return stack.Count; }
+            get { return pathBuffer.Count; }
+        }
+
+        /// <summary>
+        /// The local name of the element the parser is currently positioned on (the last path
+        /// segment), or an empty string at the document root.
+        /// </summary>
+        public string CurrentElementName
+        {
+            get { return this.pathBuffer.TopSegment; }
         }
 
         /// <summary>
@@ -192,6 +207,57 @@ namespace Amazon.Runtime.Internal.Transform
         public override bool IsStartOfDocument
         {
             get { return XmlReader.ReadState == ReadState.Initial; }
+        }
+
+        /// <summary>
+        /// Matches an expression against the current path suffix without materializing the path string.
+        /// Result-identical to the base string-based <see cref="UnmarshallerContext.TestExpression(string)"/>.
+        /// </summary>
+        public override bool TestExpression(string expression)
+        {
+            if (expression.Equals("."))
+                return true;
+
+            // Build the comparison span. For an attribute node the effective path is
+            // "{elementPath}/@{attr}", which we never materialize on the element buffer; fall back
+            // to the string path in that (rare) case to keep behavior identical.
+            if (currentAttributeName != null)
+                return CurrentPath.EndsWith(expression, StringComparison.OrdinalIgnoreCase);
+
+            return (pathBuffer.PathSpan.Length == 0 ? "/".AsSpan() : pathBuffer.PathSpan)
+                 .EndsWith(expression.AsSpan(), StringComparison.OrdinalIgnoreCase);
+        }
+
+        /// <summary>
+        /// Matches an expression at a given starting stack depth without materializing the path string.
+        /// Result-identical to the base string-based <see cref="UnmarshallerContext.TestExpression(string, int)"/>.
+        /// </summary>
+        public override bool TestExpression(string expression, int startingStackDepth)
+        {
+            if (expression.Equals("."))
+                return true;
+
+            // Count '/' separators to derive the required depth, matching the base algorithm exactly.
+            int requiredDepth = startingStackDepth;
+            for (int i = expression.IndexOf('/'); i > -1; i = expression.IndexOf('/', i + 1))
+            {
+                // Don't consider attributes a new depth level
+                if (expression[0] != '@')
+                    requiredDepth++;
+            }
+
+            if (requiredDepth != CurrentDepth)
+                return false;
+
+            var path = pathBuffer.PathSpan;
+            if (currentAttributeName != null)
+            {
+                path = CurrentPath.AsSpan();
+            }
+
+            return path.Length > expression.Length
+                && path[path.Length - expression.Length - 1] == '/'
+                && path.EndsWith(expression.AsSpan(), StringComparison.OrdinalIgnoreCase);
         }
 
         #endregion
@@ -237,10 +303,14 @@ namespace Amazon.Runtime.Internal.Transform
             if (attributeEnumerator != null && attributeEnumerator.MoveNext())
             {
                 this.nodeType = XmlNodeType.Attribute;
-                stackString = string.Format(CultureInfo.InvariantCulture, "{0}/@{1}", StackToPath(stack), attributeEnumerator.Current);
+                // Positioned on an attribute: the element path is unchanged; CurrentPath/matching
+                // append "/@{name}" on demand via currentAttributeName.
+                this.currentAttributeName = attributeEnumerator.Current;
             }
             else
             {
+                this.currentAttributeName = null;
+
                 // Skip some nodes
                 if (nodesToSkip.Contains(XmlReader.NodeType))
                     XmlReader.Read();
@@ -248,8 +318,7 @@ namespace Amazon.Runtime.Internal.Transform
                 if (currentlyProcessingEmptyElement)
                 {
                     nodeType = XmlNodeType.EndElement;
-                    stack.Pop();
-                    stackString = StackToPath(stack);
+                    pathBuffer.Pop();
                     XmlReader.Read();
                     currentlyProcessingEmptyElement = false;
                 }
@@ -257,8 +326,7 @@ namespace Amazon.Runtime.Internal.Transform
                 {
                     //This is a shorthand form of an empty element <element /> and we want to allow it
                     nodeType = XmlNodeType.Element;
-                    stack.Push(XmlReader.LocalName);
-                    stackString = StackToPath(stack);
+                    pathBuffer.Push(XmlReader.LocalName);
                     currentlyProcessingEmptyElement = true;
                     nodeContent = String.Empty;
 
@@ -270,14 +338,12 @@ namespace Amazon.Runtime.Internal.Transform
                     {
                         case XmlNodeType.EndElement:
                             this.nodeType = XmlNodeType.EndElement;
-                            stack.Pop();
-                            stackString = StackToPath(stack);
+                            pathBuffer.Pop();
                             XmlReader.Read();
                             break;
                         case XmlNodeType.Element:
                             nodeType = XmlNodeType.Element;
-                            stack.Push(XmlReader.LocalName);
-                            stackString = StackToPath(stack);
+                            pathBuffer.Push(XmlReader.LocalName);
                             this.ReadElement();
                             break;
                         default:
@@ -289,7 +355,6 @@ namespace Amazon.Runtime.Internal.Transform
                             // after advancing the underlying XmlReader past an unhandled node type.
                             nodeType = XmlNodeType.None;
                             nodeContent = string.Empty;
-                            stackString = StackToPath(stack);
                             break;
                     }
                 }
@@ -312,16 +377,6 @@ namespace Amazon.Runtime.Internal.Transform
         #endregion
 
         #region Private Methods
-
-        private static string StackToPath(Stack<string> stack)
-        {
-            string path = null;
-            foreach (string s in stack.ToArray())
-            {
-                path = null == path ? s : string.Format(CultureInfo.InvariantCulture, "{0}/{1}", s, path);
-            }
-            return "/" + path;
-        }
 
         // Move to the next element, cache the attributes collection
         // and attempt to cache the inner text of the element if applicable.
@@ -369,10 +424,111 @@ namespace Amazon.Runtime.Internal.Transform
 #endif
                         _xmlTextReader = null;
                     }
+                    pathBuffer.Dispose();
                 }
                 disposed = true;
             }
             base.Dispose(disposing);
+        }
+
+        private sealed class XmlPathBuffer : IDisposable
+        {
+            // The current path as chars, e.g. "/Root/Child/member". Segments are appended on Push and
+            // trimmed (by moving pathLength back) on Pop. The buffer is reused for the whole document.
+            private char[] pathChars = ArrayPool<char>.Shared.Rent(256);
+            private int pathLength;
+            // Length (including the leading '/') of each pushed segment, enabling O(1) Pop.
+            private readonly Stack<int> segmentLengths = new Stack<int>();
+            private string cached;
+
+            /// <summary>Number of element segments on the path (the current depth).</summary>
+            public int Count
+            {
+                get { return segmentLengths.Count; }
+            }
+
+            /// <summary>
+            /// The full path string, e.g. "/Root/Child". An empty stack yields "/" to match the
+            /// historical StackToPath behavior. Materialized lazily and cached until the next mutation.
+            /// </summary>
+            public string CurrentPath
+            {
+                get
+                {
+                    return this.cached ??= pathLength == 0 ? "/" : new string(pathChars, 0, pathLength);
+                }
+            }
+
+            /// <summary>The path chars without the trailing string allocation, for span matching.</summary>
+            public ReadOnlySpan<char> PathSpan
+            {
+                get { return pathChars.AsSpan(0, pathLength); }
+            }
+
+            public string TopSegment
+            {
+                get
+                {
+                    if (segmentLengths.Count == 0)
+                        return string.Empty;
+
+                    int seg = segmentLengths.Peek();
+                    return new string(pathChars, pathLength - seg + 1, seg - 1);
+                }
+            }
+
+            /// <summary>
+            /// Builds the transient attribute path "{CurrentPath}/@{name}" without mutating the buffer.
+            /// </summary>
+            public string BuildAttributePath(string attributeName)
+            {
+                return string.Concat(CurrentPath, "/@", attributeName);
+            }
+
+            public void Push(string localName)
+            {
+                int seg = localName.Length + 1; // leading '/'
+                EnsureCapacity(seg);
+                pathChars[pathLength++] = '/';
+                localName.CopyTo(0, pathChars, pathLength, localName.Length);
+                pathLength += localName.Length;
+                segmentLengths.Push(seg);
+                cached = null;
+            }
+
+            public void Pop()
+            {
+                if (segmentLengths.Count == 0)
+                    return;
+
+                pathLength -= segmentLengths.Pop();
+                cached = null;
+            }
+
+            private void EnsureCapacity(int additionalChars)
+            {
+                if (pathLength + additionalChars <= pathChars.Length)
+                    return;
+
+                int newSize = Math.Max(pathChars.Length * 2, pathLength + additionalChars);
+                char[] newBuffer = ArrayPool<char>.Shared.Rent(newSize);
+                Array.Copy(pathChars, 0, newBuffer, 0, pathLength);
+                ArrayPool<char>.Shared.Return(pathChars);
+                pathChars = newBuffer;
+            }
+
+            public void Dispose()
+            {
+                if (pathChars != null)
+                {
+                    // Materialize the path string before returning the buffer so that a post-dispose
+                    // CurrentPath read (e.g. AmazonUnmarshallingException diagnostics in an error path)
+                    // returns the last known path instead of dereferencing the returned buffer.
+                    cached ??= pathLength == 0 ? "/" : new string(pathChars, 0, pathLength);
+                    ArrayPool<char>.Shared.Return(pathChars);
+                    pathChars = null;
+                }
+            }
         }
     }
 
--- a/generator/.DevConfigs/4e87e50a-2f32-49c7-8fe7-7eab4c1c9f22.json
+++ b/generator/.DevConfigs/4e87e50a-2f32-49c7-8fe7-7eab4c1c9f22.json
@@ -1,0 +1,9 @@
+{
+  "core": {
+    "changeLogMessages": [
+      "Eliminates per node path allocation in XML and AWSQuery response unmarshalling."
+    ],
+    "type": "patch",
+    "updateMinimum": true
+  }
+}
--- a/generator/.DevConfigs/d93f07e6-e27a-4638-95dd-6cf18aa93c86.json
+++ b/generator/.DevConfigs/d93f07e6-e27a-4638-95dd-6cf18aa93c86.json
@@ -1,0 +1,9 @@
+{
+  "core": {
+    "changeLogMessages": [
+      "Reduce memory allocations in XML primitive unmarshalling."
+    ],
+    "type": "patch",
+    "updateMinimum": true
+  }
+}
--- a/sdk/src/Core/Amazon.Runtime/Internal/Transform/SimpleTypeUnmarshaller.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Transform/SimpleTypeUnmarshaller.cs
@@ -25,12 +25,6 @@ namespace Amazon.Runtime.Internal.Transform
 {
     static class SimpleTypeUnmarshaller<T>
     {
-        public static T Unmarshall(XmlUnmarshallerContext context)
-        {
-            string text = context.ReadText();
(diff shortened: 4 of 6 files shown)


sdk/src/Core/Amazon.Runtime/Internal/Transform/XmlUnmarshallerContext.cs
```diff
@@ -369,11 +425,112 @@ protected override void Dispose(bool disposing)
...
+            }
+
+            public void Dispose()
+            {
+                if (pathChars != null)
+                {
+                    // Materialize the path string before returning the buffer so that a post-dispose
+                    // CurrentPath read (e.g. AmazonUnmarshallingException diagnostics in an error path)
+                    // returns the last known path instead of dereferencing the returned buffer.
+                    cached ??= pathLength == 0 ? "/" : new string(pathChars, 0, pathLength);
```
JanHyka, line 527: This leaves the object in undefined state. We should follow classic `bool _isDisposed` approach and throw on attempts to call public methods once the object is disposed. Otherwise you get mixture of working last state (`CurrentPath`, `BuildAttributePath`), something that breaks that state (`Pop`) and something that throws because parts (actual pathChars) are disposed (`Push`) and parts are kept (stack).
At absolutely minimum, methods that get post-disposal defunct should explicitly throw `ObjectDisposedException`. That would ensure predictible behavior for your diagnostics or WeakReference approach.

Also note if the object isn't disposed explicitly, the finalizer won't return rented array.

sdk/src/Core/Amazon.Runtime/Internal/Transform/XmlUnmarshallerContext.cs
```diff
@@ -369,11 +425,112 @@ protected override void Dispose(bool disposing)
...
+                    return new string(pathChars, pathLength - seg + 1, seg - 1);
+                }
+            }
+
+            /// <summary>
+            /// Builds the transient attribute path "{CurrentPath}/@{name}" without mutating the buffer.
+            /// </summary>
+            public string BuildAttributePath(string attributeName)
+            {
+                return string.Concat(CurrentPath, "/@", attributeName);
```
JanHyka, line 485: This can actually cause two string materializations - one in `CurrentPath` getter and second in `Concat`enation. While we already have array buffer, it would be more efficient to avoid forcing `CurrentPath` materialization.

sdk/src/Core/Amazon.Runtime/Internal/Transform/XmlUnmarshallerContext.cs
```diff
@@ -369,11 +425,112 @@ protected override void Dispose(bool disposing)
...
                         _xmlTextReader = null;
                     }
+                    pathBuffer.Dispose();
                 }
                 disposed = true;
             }
             base.Dispose(disposing);
         }
+
+        private sealed class XmlPathBuffer : IDisposable
```
JanHyka, line 434: I believe making this class internal + accompany it with own unit tests would be justified.

sdk/test/UnitTests/Custom/Runtime/XmlUnmarshallerContextPathTests.cs
```diff
@@ -0,0 +1,200 @@
...
+
+namespace AWSSDK.UnitTests
+{
+    [TestClass]
+    [TestCategory("UnitTest")]
+    [TestCategory("Runtime")]
+    public class XmlUnmarshallerContextPathTests
+    {
+        [TestMethod]
+        public void MetricDataResponse_MatcherMatchesHistoricalLogic()
```
JanHyka: Test methods should be explicit in what they test and what assertion expects under what conditions.

`MatchesHistoricalLogic` is essentially a black box that doesn't give any space to read actual rules expectations and what is the foundation for such expectation. This is especially problematic since prior version didn't have any tests at all and all behavioral expectations for current version are symptomatic. We have no means to make sure we are not replicating possible bugs from older code.

Please consider Testing individual rules with clear outcomes.

JanHyka (commented):
As per comment in #4466 I am checking on this implementation proposal, leaving a few comments. Since this proposed version is more efficient than the one in #4466 let's keep this one.

sdk/src/Core/Amazon.Runtime/Internal/Transform/XmlUnmarshallerContext.cs
muhammad-othman: `CurrentPath` is cached, so it shouldn't be calculated again no matter how many times we access it. I'm not sure how can avoid `CurrentPath` materialization, the only way I see is by building from the buffer which is less efficient.
Another thing to consider is that attributes in the SDK models are very rare, so we shouldn't optimize for attributes if it means making everything else less efficient.

sdk/src/Core/Amazon.Runtime/Internal/Transform/XmlUnmarshallerContext.cs
muhammad-othman: I agree that it writing unit tests for it can be justified, but I see it as an implementation detail that we shouldn't care about and only test the `XmlUnmarshallerContext` since it is what the SDK is using, though we don't really do a good job for testing that too :(

sdk/src/Core/Amazon.Runtime/Internal/Transform/XmlUnmarshallerContext.cs
muhammad-othman: This is a very valid concern if this was a public class that can be used by other parts of the SDK, but I think because it is private and only used by `XmlUnmarshallerContext` and it is the one disposing it then we should keep it this way to avoid adding `ObjectDisposedException` checks everywhere.
I'm more worried that maybe in the future we want to consolidate how the unmarshalling exceptions are handled and in this case we can catch them at another pipeline handler after the context is disposed and we can check the `CurrentPath` at this point, this is matching the other protocols patterns which keeps the `CurrentPath` available after the context is disposed.

sdk/test/UnitTests/Custom/Runtime/XmlUnmarshallerContextPathTests.cs
muhammad-othman: Atleast for this PR I wanted to make sure that nothing changes between the current implementation and the previous version which is the point of these tests, even if a bug was found it should be fixed in a different PR.
However I agree that the context behavior isn't well tested and we should add proper unit tests to cover it including the other protocols.

sdk/src/Core/Amazon.Runtime/Internal/Transform/XmlUnmarshallerContext.cs
```diff
@@ -194,6 +209,59 @@ public override bool IsStartOfDocument
...
+                // Don't consider attributes a new depth level
+                if (expression[0] != '@')
+                    requiredDepth++;
+            }
+
+            if (requiredDepth != CurrentDepth)
+                return false;
+
+            var path = pathBuffer.PathSpan;
+
```
dscpinheiro: Nit: extra lines here.

sdk/src/Core/Amazon.Runtime/Internal/Transform/XmlUnmarshallerContext.cs
```diff
@@ -194,6 +209,59 @@ public override bool IsStartOfDocument
...
+        {
+            if (expression.Equals("."))
+                return true;
+
+            // Count '/' separators to derive the required depth, matching the base algorithm exactly.
+            int requiredDepth = startingStackDepth;
+            for (int i = expression.IndexOf('/'); i > -1; i = expression.IndexOf('/', i + 1))
+            {
+                // Don't consider attributes a new depth level
+                if (expression[0] != '@')
```
dscpinheiro, line 245: I know this was existing behavior, but is always checking for `[0]` (inside the loop) intentional?

sdk/test/UnitTests/Custom/Runtime/XmlUnmarshallerContextPathTests.cs
```diff
@@ -0,0 +1,199 @@
...
+
+
+        private static bool HistoricalTestExpression(string expression, string currentPath)
+        {
+            if (expression.Equals("."))
+                return true;
+            return currentPath.EndsWith(expression, StringComparison.OrdinalIgnoreCase);
+        }
+
+        private static bool HistoricalTestExpression(string expression, int startingStackDepth, string currentPath, int currentDepth)
```
dscpinheiro: I somewhat agree with the other comment... I'm not a fan of repeating the previous logic here (and once the PR is merged this code won't be "true" anymore). 

Would also expect most edge cases to be covered by protocol tests (and have this class validate what's missing).

sdk/src/Core/Amazon.Runtime/Internal/Transform/XmlUnmarshallerContext.cs
muhammad-othman: Yes it is, as only none attributes have a depth, attributes are attached to the current element so the depth shouldn't be increased. I guess it could've been rewritten in a better way but I wanted to avoid extra changes if possible.

sdk/test/UnitTests/Custom/Runtime/XmlUnmarshallerContextPathTests.cs
muhammad-othman: I'm not sure what edge cases can be missed by the protocol tests, and if there are shouldn't they be an additional protocol tests to cover other SDKs too?

muhammad-othman: Added new tests at https://github.com/aws/aws-sdk-net/blob/bf296faeee07ab300379527cfb67bcade835ba3b/sdk/test/UnitTests/Custom/Runtime/XmlUnmarshallerContextTests.cs

muhammad-othman: This tests passes on development branch too except `CurrentElementName_IsTopSegment_EmptyAtRootBoundaries` as it uses `CurrentElementName` which didn't exist before.

dscpinheiro (approved):

boblodgett (approved):

PR #4468, opened 2026-07-14
feat(cloudfront-signers): sign with AWS KMS / external signer via ICloudFrontSigner

the-gabe:
Today `AWSSDK.Extensions.CloudFront.Signers` can only sign from a raw private-key PEM held in the process. There is no way to keep the CloudFront signing key in **AWS KMS** (or an HSM) and have the SDK produce the signed URL/cookies — callers must reimplement the policy JSON, url-safe base64, and cookie assembly by hand.

This PR adds a signing seam:

- `ICloudFrontSigner` — `Sign(byte[] policyBytes) -> (byte[] signature, HashAlgorithmName)`.
- `DelegateCloudFrontSigner` — wraps a `Func<byte[], byte[]>` so a caller can plug in `kms:Sign` (or anything) in a couple of lines, with no new dependency added to the package.
- `ICloudFrontSigner` overloads of `SignUrl`, `GetCustomSignedURL`, and `GetCookiesForCustomPolicy`, routing through the existing policy/encoding/cookie builders.

The seam is algorithm-agnostic and works with **both** key types CloudFront accepts, no code change between them:

- **RSA 2048** → KMS `RSASSA_PKCS1_V1_5_SHA_256` (PKCS#1 v1.5 — CloudFront doesn't verify PSS).
- **ECDSA P-256** → KMS `ECDSA_SHA_256`; KMS returns a DER (RFC 3279) signature, exactly what CloudFront expects, so no re-encoding.

A CloudFront policy is well under KMS's 4096-byte `Message` limit, so `MessageType=RAW` is correct. Tests assert the RSA overloads produce byte-identical output to the private-key overloads, and that an ECDSA-signed cookie verifies against the P-256 public key.

No new package dependency (KMS stays in caller code). Existing overloads and downstream logic are unchanged. Includes the required DevConfig (`Extensions.CloudFront.Signers`, minor).

_Generated by AI tools, and reviewed by James Brown._

### DIFF
--- a/extensions/src/AWSSDK.Extensions.CloudFront.Signers/ICloudFrontSigner.cs
+++ b/extensions/src/AWSSDK.Extensions.CloudFront.Signers/ICloudFrontSigner.cs
@@ -1,0 +1,145 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ * 
+ *  http://aws.amazon.com/apache2.0
+ * 
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+using System;
+using System.Security.Cryptography;
+
+namespace Amazon.CloudFront
+{
+    /// <summary>
+    /// Provides a custom mechanism for signing a CloudFront policy document. Implement this to
+    /// compute the signature with whatever key material and cryptographic implementation you
+    /// choose, instead of handing a private key to the SDK.
+    /// <para>
+    /// The common scenario is an <b>AWS KMS</b> asymmetric key, so that the private key never
+    /// leaves KMS, but nothing here assumes that: an implementation is free to use an HSM, an
+    /// external key service, or a key the process holds in memory.
+    /// </para>
+    /// <para>
+    /// The <see cref="AmazonCloudFrontUrlSigner"/> and <see cref="AmazonCloudFrontCookieSigner"/>
+    /// overloads that take an <see cref="ICloudFrontSigner"/> build the policy document and the
+    /// url-safe encoding exactly as the private-key overloads do; only the signature computation
+    /// is delegated here. This keeps the extension free of any hard dependency on KMS while still
+    /// enabling it.
+    /// </para>
+    /// </summary>
+    /// <remarks>
+    /// This is signature-algorithm agnostic, so both key types CloudFront accepts work through the
+    /// same seam. The guidance below names the AWS KMS signing algorithms because that is the common
+    /// case, but the requirement on any implementation is the same, per key type:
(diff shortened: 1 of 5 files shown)


AlexDaines (commented):
.

the-gabe:
Hi, anything needed from my side here?

extensions/src/AWSSDK.Extensions.CloudFront.Signers/ICloudFrontSigner.cs
```diff
@@ -0,0 +1,138 @@
+/*
```
normj, line 1: I recognize this does call out your company but it doesn't scale for us to support/maintain different copyright notices across different files especially as the files evolve over time.

Can you use the copyright from the existing files like the `AmazonCloudFrontCookieSigner.cs`.

extensions/test/CloudFront.SignersTests/SignerDelegationTest.cs
```diff
@@ -0,0 +1,164 @@
+/*
```
normj: Same comment about copyright.

extensions/src/AWSSDK.Extensions.CloudFront.Signers/ICloudFrontSigner.cs
```diff
@@ -0,0 +1,138 @@
...
+ */
+
+using System;
+using System.Security.Cryptography;
+
+namespace Amazon.CloudFront
+{
+    /// <summary>
+    /// Signs a CloudFront policy document. Implement this to sign with a key the process does
+    /// not hold in memory — most notably an <b>AWS KMS</b> asymmetric key (the private key
```
normj: I understand for your KMS example the key is not held in memory but this interface is generic an implementation could have their own mechanism that does involve the key in memory. The docs should say this interface provides a custom mechanism for signing and then mention KMS as the common scenario.

normj (changes requested):

extensions/src/AWSSDK.Extensions.CloudFront.Signers/ICloudFrontSigner.cs
the-gabe: Yeah no problem. It's not like this is a big deal in this context :)

extensions/src/AWSSDK.Extensions.CloudFront.Signers/ICloudFrontSigner.cs
the-gabe: Gotcha, can do

extensions/src/AWSSDK.Extensions.CloudFront.Signers/ICloudFrontSigner.cs
dscpinheiro: @the-gabe Were you planning to make the changes Norm asked about? (If not we'd prefer to close the PR instead of keeping of lingering)

the-gabe: @dscpinheiro changes pushed! Sorry for the delay.

normj (approved):

dscpinheiro (approved):

dscpinheiro:
Thanks for the contribution, this will be included in the next SDK release.

PR #4472, opened 2026-07-16
Add IRequest.PrecomputedContentSha256 honored ahead of the body read

GarrettBeatty:
## Description

Adds `IRequest.PrecomputedContentSha256` so a caller can supply a hex-encoded SHA256 of the request body and have `AWS4Signer.SetRequestBodyHash` use it verbatim as `x-amz-content-sha256` — without the signer reading and hashing the body itself. This is what lets a caller sign large or non-seekable content.

## What's included

- `IRequest.PrecomputedContentSha256` (+ `DefaultRequest` implementation).
- `AWS4Signer.SetRequestBodyHash` honors it **ahead of the body read**, so it survives the `InitializeHeaders`/`CleanHeaders` scrub that clears a stale `x-amz-content-sha256` before (re)signing.

## Design note

The value is routed through a **request field** rather than teaching `CleanHeaders` to preserve the header, specifically so the shared resign/scrub path is left untouched. The standard service pipeline never sets this field, so its resign behavior is unaffected.

## Testing

- New `PrecomputedContentSha256Tests`: verbatim use, survival of a stale-header scrub, and a regression test proving the standard path (field unset) **still** scrubs a stale hash.
- `AWSSDK.Core` builds clean; new tests pass.

## Dry-run

Basic chained dry-run (`catapult-net`, product `dotnetv4`, stage `prod`, source `origin:sigv4-precomputed-content-sha256`):

| Product | Build ID |
|---|---|
| .NET (`dotnetv4`) | `0abed65d-116f-446b-920e-ef2c6c77b7bc` |
| PowerShell (`powershell5`) | `8cf8129f-7b56-4ce2-9830-797a29ece4c4` |

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4Signer.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4Signer.cs
@@ -542,6 +542,14 @@ namespace Amazon.Runtime.Internal.Auth
                 }
             }
 
+            // If the caller supplied a precomputed body hash on the request, use it verbatim.
+            // This is honored ahead of the body read (and survives the InitializeHeaders/CleanHeaders
+            // scrub that clears a stale x-amz-content-sha256 header before signing), letting a caller
+            // sign a body the signer must not read (large or non-seekable). The standard service
+            // pipeline never sets this, so its resign/scrub behavior is unaffected.
+            if (!request.UseChunkEncoding && !string.IsNullOrEmpty(request.PrecomputedContentSha256))
+                return SetPayloadSignatureHeader(request, request.PrecomputedContentSha256);
+
             // if the body hash has been precomputed and already placed in the header, just extract and return it
             string computedContentHash;
             var shaHeaderPresent = request.Headers.TryGetValue(HeaderKeys.XAmzContentSha256Header, out computedContentHash);
--- a/sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4Signer.cs
+++ b/sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4Signer.cs
@@ -547,6 +547,11 @@ namespace Amazon.Runtime.Internal.Auth
             // scrub that clears a stale x-amz-content-sha256 header before signing), letting a caller
             // sign a body the signer must not read (large or non-seekable). The standard service
             // pipeline never sets this, so its resign/scrub behavior is unaffected.
+            //
+            // This branch is only reachable when payload signing is effectively enabled: the
+            // unsigned-payload gate above (DisablePayloadSigning ?? !signPayload) already returned
+            // UNSIGNED-PAYLOAD otherwise, so PrecomputedContentSha256 cannot override that magic string.
+            // It is likewise ignored for chunk-encoded requests, which carry their own streaming hash.
             if (!request.UseChunkEncoding && !string.IsNullOrEmpty(request.PrecomputedContentSha256))
                 return SetPayloadSignatureHeader(request, request.PrecomputedContentSha256);
 
--- a/sdk/test/UnitTests/Custom/Runtime/Signing/PrecomputedContentSha256Tests.cs
+++ b/sdk/test/UnitTests/Custom/Runtime/Signing/PrecomputedContentSha256Tests.cs
@@ -14,6 +14,7 @@
  */
 
 using System;
+using System.IO;
 using System.Text;
 using Amazon.Runtime;
 using Amazon.Runtime.Internal;
@@ -41,14 +42,16 @@ namespace AWSSDK.UnitTests.Signing
         // -----------------------------------------------------------------------
 
         [TestMethod]
-        public void SetRequestBodyHash_WithPrecomputedHash_UsesItVerbatim()
+        public void SetRequestBodyHash_WithPrecomputedHash_UsesItVerbatimWithoutReadingBody()
         {
-            // A precomputed hash must be used verbatim as x-amz-content-sha256, even when a body is present
-            // that would otherwise be read and hashed.
+            // A precomputed hash must be used verbatim as x-amz-content-sha256 AND the signer must not
+            // read the body to produce it. We back the request with a stream that throws on any read: if
+            // the signer honors the precomputed hash ahead of the body read, signing succeeds and the
+            // header carries the supplied value; if it fell through to hashing the body, the read would throw.
             var precomputed = new string('a', 64);
             var config = new MockClientConfig { AuthenticationRegion = Region };
             var request = BuildInternalRequest();
-            request.Content = Encoding.UTF8.GetBytes("some body the signer must not read");
+            request.ContentStream = new ThrowOnReadStream(Encoding.UTF8.GetBytes("some body the signer must not read"));
             request.PrecomputedContentSha256 = precomputed;
 
             new AWS4Signer().SignRequest(request, config, new RequestMetrics(), AccessKey, SecretKey);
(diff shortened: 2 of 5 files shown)


sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4Signer.cs
```diff
@@ -539,6 +539,14 @@ public static string SetRequestBodyHash(IRequest request, bool signPayload, stri
                 }
             }
 
+            // If the caller supplied a precomputed body hash on the request, use it verbatim.
+            // This is honored ahead of the body read (and survives the InitializeHeaders/CleanHeaders
+            // scrub that clears a stale x-amz-content-sha256 header before signing), letting a caller
+            // sign a body the signer must not read (large or non-seekable). The standard service
+            // pipeline never sets this, so its resign/scrub behavior is unaffected.
+            if (!request.UseChunkEncoding && !string.IsNullOrEmpty(request.PrecomputedContentSha256))
+                return SetPayloadSignatureHeader(request, request.PrecomputedContentSha256);
```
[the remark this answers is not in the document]
GarrettBeatty: Copilot said: "Step 2 doesn't check signPayload — add && signPayload so we only use the precomputed hash when signing is on."

  Here's the concrete case that breaks. Imagine an S3 client configured with signPayload = false (its default is unsigned), but for this one upload you want a real signed hash, so you set:

  request.DisablePayloadSigning = false;         // "no, DO sign this one"
  request.PrecomputedContentSha256 = "abc123..."; // "and here's the hash"

  Trace it:

  - Step 1: DisablePayloadSigning is false, so the condition is false → skip. Payload signing is on for this request. ✅
  - Step 2 (current code): hash is present → return "abc123...". ✅ Never reads the 50 GB body. Feature works.
  - Step 2 (Copilot's && signPayload): signPayload is the client's flag = false → skip. ❌ Falls through to Step 3 and reads the entire body — the exact thing you were trying to avoid.


```


i updated the docs to be more clear

File: sdk/src/Core/Amazon.Runtime/Internal/Auth/AWS4Signer.cs [hunk 2]
```diff
@@ -542,6 +542,19 @@ public static string SetRequestBodyHash(IRequest request, bool signPayload, stri
...
+            //
+            // This branch is only reachable when payload signing is effectively enabled: the
+            // unsigned-payload gate above (DisablePayloadSigning ?? !signPayload) already returned
+            // UNSIGNED-PAYLOAD otherwise, so PrecomputedContentSha256 cannot override that magic string.
+            // It is likewise ignored for chunk-encoded requests, which carry their own streaming hash.
+            if (!request.UseChunkEncoding && !string.IsNullOrEmpty(request.PrecomputedContentSha256))
+                return SetPayloadSignatureHeader(request, request.PrecomputedContentSha256);
+
             // if the body hash has been precomputed and already placed in the header, just extract and return it
             string computedContentHash;
```
GarrettBeatty, line 559: this other code with computedcontenthash doesnt work as expected with the signer which is why we need this new property. consider this example

 ## Why setting the header yourself doesn't work

  Say you have a 50 GB non-seekable upload and you already know its SHA-256. You try the "obvious" thing — set the header yourself:

  ```csharp
  var request = new DefaultRequest(uploadRequest, "s3");
  request.Endpoint = new Uri("https://my-bucket.s3.amazonaws.com/huge-file");
  request.HttpMethod = "PUT";
  request.ContentStream = myNonSeekableStream;

  // I already know the hash, so I'll just put it in the header myself:
  request.Headers["X-Amz-Content-SHA256"] = "9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08";

  // Sign it the normal way:
  new AWS4Signer().SignRequest(request, config, metrics, accessKey, secretKey);
  ```

  You'd expect the signer to see your header and use it. Here's what actually executes:

  ### Step 1 — `SignRequest` starts (line 230)

  ```csharp
  signedAt = InitializeHeaders(request.Headers, request.Endpoint, signedAt);
  ```

  ### Step 2 — `InitializeHeaders` calls `CleanHeaders` (line 305 → 338)

  ```csharp
  private static void CleanHeaders(IDictionary<string, string> headers)
  {
      headers.Remove(HeaderKeys.AuthorizationHeader);
      headers.Remove(HeaderKeys.XAmzContentSha256Header);   // 💥 YOUR HEADER IS DELETED HERE
      ...
  }
  ```

  Your `"9f86d0..."` is gone. The headers dictionary no longer contains `X-Amz-Content-SHA256`.

  ### Step 3 — `SetRequestBodyHash` runs (line 254 → 520)

  It reaches the header-lookup branch at line 560:

  ```csharp
  var shaHeaderPresent = request.Headers.TryGetValue(HeaderKeys.XAmzContentSha256Header, out computedContentHash);
  if (shaHeaderPresent && !request.UseChunkEncoding)
      return computedContentHash;
  ```

  `shaHeaderPresent` is now `false` — because `CleanHeaders` removed it two steps ago. So this branch is skipped.

  ### Step 4 — falls through to computing the hash (line 565+)

  With no header and no property, the signer does the only thing left: it reads the entire body to compute the SHA-256 itself.

  ```csharp
  // otherwise continue to calculate the hash and set it in the headers before returning
  ...
  computedContentHash = ComputeHash(request.ContentStream);   // reads all 50 GB
  ```

  **Result:** either it reads 50 GB just to sign (defeating the whole purpose), or — if the stream is non-seekable — it can't rewind afterward to actually send, so the upload throws.

  ## The same code, but with the property

  ```csharp
  request.PrecomputedContentSha256 = "9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08";
  new AWS4Signer().SignRequest(request, config, metrics, accessKey, secretKey);
  ```

  Trace:

  1. `CleanHeaders` runs — removes the *header*, but the **property is untouched** (`CleanHeaders` only touches the headers dictionary).
  2. `SetRequestBodyHash` reaches line 555 — **before** the header lookup:

  ```csharp
  if (!request.UseChunkEncoding && !string.IsNullOrEmpty(request.PrecomputedContentSha256))
      return SetPayloadSignatureHeader(request, request.PrecomputedContentSha256);   // ✅ uses your hash
  ```

  3. Returns your hash verbatim. **Body never read.** Works on non-seekable streams.

sdk/test/UnitTests/Custom/Runtime/Signing/PrecomputedContentSha256Tests.cs
```diff
@@ -0,0 +1,163 @@
...
+using Amazon.Util;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace AWSSDK.UnitTests.Signing
+{
+    [TestClass]
+    [TestCategory("Core")]
+    [TestCategory("Signer")]
+    public class PrecomputedContentSha256Tests
+    {
```
dscpinheiro, line 32: What do you think of this test method (which I don't believe is validated anywhere else?):
```csharp
[DataTestMethod]
[DataRow(true,  true,  false, AWS4Signer.UnsignedPayload)]
[DataRow(false, false, false, AWS4Signer.UnsignedPayload)]
[DataRow(false, true,  true,  AWS4Signer.StreamingBodySha256)]
public void SetRequestBodyHash_WithPrecomputedHash_IgnoredWhenNotEffectivelySigningBody(bool disablePayloadSigning, bool signerSignsPayload, bool useChunkEncoding, string expectedHash)
{
    var config = new MockClientConfig { AuthenticationRegion = Region };
    var request = BuildInternalRequest();
    request.UseChunkEncoding = useChunkEncoding;
    request.PrecomputedContentSha256 = new string('a', 64);

    if (disablePayloadSigning)
    {
        request.DisablePayloadSigning = true;
    }

    new AWS4Signer(signPayload: signerSignsPayload).SignRequest(request, config, new RequestMetrics(), AccessKey, SecretKey);
    Assert.AreEqual(expectedHash, request.Headers[HeaderKeys.XAmzContentSha256Header]);
}
```

GarrettBeatty: i will add this

dscpinheiro (approved):

peterrsongg (approved):

PR #4474, opened 2026-07-20
Add SigV4SigningHandler for transparent HttpClient signing

GarrettBeatty:
## Description

Adds **`SigV4SigningHandler`**, a `System.Net.Http.DelegatingHandler` in the `Amazon.Runtime.Signing` namespace that SigV4-signs each outbound request. Install it on an `HttpClient` — directly or via `IHttpClientFactory` — and every request through that client is signed transparently.

## Why a handler

`AWSSigV4Signer.SignAsync` gives you signed headers to apply yourself. The handler is the transparent, DI-native layer on top: it plugs into the native `HttpMessageHandler` chain and `IHttpClientFactory`, signs every send with no per-call step, and re-signs each retry/redirect attempt with a fresh timestamp. No sibling AWS SDK ships an equivalent because none runs on `System.Net.Http`; the only precedent is the community `AwsSignatureVersion4` package.

## Usage

**`IHttpClientFactory` / DI** (the intended production wiring) — register the client once, every request is signed:

```csharp
services.AddHttpClient("execute-api", c => c.BaseAddress = new Uri("https://abc123.execute-api.us-east-1.amazonaws.com/"))
    .AddHttpMessageHandler(sp => new SigV4SigningHandler(
        DefaultAWSCredentialsIdentityResolver.GetCredentials(), RegionEndpoint.USEast1, "execute-api"))
    // Signing handler must not auto-follow redirects (see caveats), and a retry policy must sit OUTSIDE it.
    .ConfigurePrimaryHttpMessageHandler(() => new HttpClientHandler { AllowAutoRedirect = false });

// elsewhere — plain HttpClient code, signed automatically:
var client = httpClientFactory.CreateClient("execute-api");
var response = await client.GetAsync("items");
```

**Standalone `HttpClient`** — wrap the handler directly:

```csharp
using Amazon.Runtime.Signing;

var handler = new SigV4SigningHandler(
    DefaultAWSCredentialsIdentityResolver.GetCredentials(),
    RegionEndpoint.USEast1,
    service: "execute-api")
{
    InnerHandler = new HttpClientHandler { AllowAutoRedirect = false },
};

var client = new HttpClient(handler);
var response = await client.PostAsync(
    "https://abc123.execute-api.us-east-1.amazonaws.com/items",
    new StringContent(json, Encoding.UTF8, "application/json"));
```

**Per-request override** — one client that signs for more than one service/region:

```csharp
var message = new HttpRequestMessage(HttpMethod.Get, "https://sts.us-west-2.amazonaws.com/?Action=GetCallerIdentity&Version=2011-06-15");
message.Options.Set(new HttpRequestOptionsKey<string>(SigV4SigningHandler.ServiceOptionKey), "sts");
message.Options.Set(new HttpRequestOptionsKey<string>(SigV4SigningHandler.RegionOptionKey), "us-west-2");
var response = await client.SendAsync(message);   // signed for sts / us-west-2, overriding the handler defaults
```

The `(AWSSigningParameters)` constructor is also available when you want to set `SignPayload`/`SignedAt` defaults.

## Behavior

- **Credentials resolve on every send** (not captured at construction), so rotating sources (SSO/assume-role/IMDS) always sign with current credentials.
- **Body buffering** happens only when the signer will hash it, and the content is rebuffered as re-readable so the body survives a resend on retry. `SignPayload = false` and a precomputed `x-amz-content-sha256` header skip the body read entirely (large/streaming uploads).
- **Per-request overrides** for service/region/`SignPayload` via `HttpRequestMessage.Options` (net5.0+) / `Properties` (earlier TFMs), keyed by `ServiceOptionKey` / `RegionOptionKey` / `SignPayloadOptionKey`.
- **Re-sign without duplication** — signing headers are replaced (on both request and content headers), not appended, so a re-signed retry stays clean.
- **Cancellation** is honored throughout, including the body-buffering read.

### Two caveats the handler cannot self-enforce (they depend on how the caller assembles the chain)

**1. `AllowAutoRedirect = false` on the primary handler.** The chain is `HttpClient -> SigV4SigningHandler -> HttpClientHandler`. `HttpClientHandler.AllowAutoRedirect` defaults to `true`, and a 3xx is followed *inside* `HttpClientHandler` — i.e. below the signing handler, which already ran on the way down and never sees the redirect. Because SigV4 signs the target `host` and path into the signature, the redirected hop — which was never re-signed — is not valid for its new destination.

Concretely, say you sign a request to `service-a` and it replies `302 Location: https://service-b.amazonaws.com/other`:

```
1. Handler signs:  GET https://service-a.amazonaws.com/thing
                   Authorization: AWS4-HMAC-SHA256 ... (signed for host=service-a, path=/thing)
2. HttpClientHandler auto-follows the 302 to service-b — beneath the signer, so no re-sign.
```

What service-b receives depends on the redirect:

- **Cross-host (service-a → service-b):** the runtime **strips the `Authorization` header** on a cross-origin redirect (so credentials aren't leaked to another host), so service-b gets an **unsigned** request and rejects it.
- **Same-host (`/thing` → `/other` on service-a):** the `Authorization` header survives, but it was computed for `/thing` while the request is now for `/other`, so the canonical request no longer matches and the service rejects it with `SignatureDoesNotMatch` (and the now-older `X-Amz-Date` can also fall outside the clock-skew window).

The handler can't intercept a follow that happens beneath it, so the fix is to turn auto-redirect off and either let the caller re-issue the request (which re-enters signing and is signed for the new host/path) or place a redirect-following handler *above* the signer — either way each hop is signed for its own destination.

**2. A retry handler must sit *outside* (above) the signing handler.** A `DelegatingHandler` retries by calling `base.SendAsync` again, so whatever is *below* the retry handler is what re-runs each attempt.
- **Correct — `HttpClient -> Retry -> SigV4Signing -> HttpClientHandler`:** a failed attempt loops back through the signer, so each attempt is **re-signed with a fresh `X-Amz-Date`** (and freshly-resolved credentials). Every attempt is independently valid.
- **Wrong — `HttpClient -> SigV4Signing -> Retry -> HttpClientHandler`:** the request is signed once and the retry loop replays that same frozen signature; a backed-off retry can then fall outside SigV4's clock-skew tolerance and be rejected for a stale date.

This re-sign-per-attempt behavior is the handler's headline advantage over signing a message once and resending it — but only if the retry policy is ordered outside the signer. With `Microsoft.Extensions.Http` + Polly, register the retry policy **before** the signing handler:

```csharp
services.AddHttpClient("execute-api")
    .AddPolicyHandler(retryPolicy)                               // outer - retries re-enter signing
    .AddHttpMessageHandler(() => new SigV4SigningHandler(...))   // inner - re-signs each attempt
    .ConfigurePrimaryHttpMessageHandler(() => new HttpClientHandler { AllowAutoRedirect = false });
```

Both are documented on the type rather than enforced: the handler can't see what sits above it in the chain, and only sees its own inner handler.

## Testing

unit and integ tests

## Dry-run

Basic chained dry-run (`catapult-net`, product `dotnetv4`, stage `prod`, source `origin:sigv4-signing-handler`):

| Product | Build ID |
|---|---|
| .NET (`dotnetv4`) | `11c03ce4-6dd1-4366-8ffc-158daefca453` |
| PowerShell (`powershell5`) | `3b4fb208-8d6b-455d-a0aa-4838dcafb9f4` |

### DIFF
--- a/sdk/src/Core/Amazon.Runtime/Signing/SigV4SigningHandler.cs
+++ b/sdk/src/Core/Amazon.Runtime/Signing/SigV4SigningHandler.cs
@@ -1,0 +1,271 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using Amazon.Util;
+
+namespace Amazon.Runtime.Signing
+{
+    /// <summary>
+    /// A <see cref="DelegatingHandler"/> that signs each outbound request with AWS Signature Version 4
+    /// before it is sent. Install it on an <see cref="HttpClient"/> — directly or via
+    /// <c>IHttpClientFactory</c> — to sign transparently, so callers write ordinary <see cref="HttpClient"/>
+    /// code and every request is signed. It is a thin front door over <see cref="AWSSigV4Signer"/> and
+    /// performs no signing computation of its own.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// The signing region, service, and payload-signing mode are handler-level defaults for the common
+    /// one-service-per-client case. A single client that must sign for more than one service or region,
+    /// or vary payload signing per request, can override any of them on an individual message through
+    /// <c>HttpRequestMessage.Options</c> (net5.0+) or <c>HttpRequestMessage.Properties</c> (earlier
+    /// targets), using <see cref="ServiceOptionKey"/>, <see cref="RegionOptionKey"/>, and
+    /// <see cref="SignPayloadOptionKey"/>.
+    /// </para>
+    /// <para>
+    /// Credentials are resolved on every send (not captured at construction), so a rotating credential
+    /// source such as SSO, assume-role, or IMDS always signs with current credentials.
+    /// </para>
+    /// <para>
+    /// <b>Important — disable automatic redirects.</b> The primary/inner handler must set
+    /// <c>AllowAutoRedirect = false</c>. An auto-followed redirect is replayed below this handler, so it is
+    /// not re-signed; because it often targets a different host it would carry a signature computed for the
+    /// original host and be rejected. Handle redirects above this handler so each hop is re-signed.
+    /// </para>
+    /// <para>
+    /// <b>Important — retry ordering.</b> Any retry handler must sit <i>outside</i> (before) this handler in
+    /// the chain so each attempt re-enters signing and is signed with a fresh timestamp. A retry handler
+    /// placed closer to the transport would replay one frozen signature and a delayed retry could fall
+    /// outside the allowed clock-skew window.
+    /// </para>
+    /// <para>
+    /// To hash the body the handler buffers request content into memory (and rebuffers it as re-readable
+    /// content so it survives a resend on retry). For a large or streaming upload where this is undesirable,
+    /// set <see cref="AWSSigningParameters.SignPayload"/> to <c>false</c> (the body is signed as
+    /// UNSIGNED-PAYLOAD and never read; requires HTTPS) or supply a precomputed
+    /// <c>x-amz-content-sha256</c> header (the body is not read).
+    /// </para>
+    /// </remarks>
+    public class SigV4SigningHandler : DelegatingHandler
+    {
+        /// <summary>
+        /// The <c>HttpRequestMessage.Options</c> / <c>Properties</c> key (a <see cref="string"/>) that
+        /// overrides the signing service for a single request.
+        /// </summary>
+        public const string ServiceOptionKey = "Amazon.Runtime.Signing.Service";
+
+        /// <summary>
+        /// The <c>HttpRequestMessage.Options</c> / <c>Properties</c> key (a <see cref="string"/>) that
+        /// overrides the signing region for a single request.
+        /// </summary>
+        public const string RegionOptionKey = "Amazon.Runtime.Signing.Region";
+
+        /// <summary>
+        /// The <c>HttpRequestMessage.Options</c> / <c>Properties</c> key (a <see cref="bool"/>) that
+        /// overrides payload signing for a single request.
+        /// </summary>
+        public const string SignPayloadOptionKey = "Amazon.Runtime.Signing.SignPayload";
+
+        private readonly AWSSigningParameters _parameters;
+
+        /// <summary>
+        /// Creates a handler that signs each request for the given credentials, region, and service.
+        /// </summary>
+        /// <param name="credentials">The credentials to sign with; resolved on every send.</param>
+        /// <param name="region">The default signing region.</param>
+        /// <param name="service">The default signing service name, e.g. "execute-api".</param>
+        public SigV4SigningHandler(AWSCredentials credentials, RegionEndpoint region, string service)
+            : this(BuildParameters(credentials, region, service))
+        {
+        }
+
+        /// <summary>
+        /// Creates a handler from a set of default signing parameters. <see cref="AWSSigningParameters.Credentials"/>
+        /// is resolved on every send. Leave <see cref="AWSSigningParameters.SignedAt"/> unset so each request is
+        /// signed with a current timestamp.
+        /// </summary>
+        /// <param name="parameters">The default signing parameters.</param>
+        public SigV4SigningHandler(AWSSigningParameters parameters)
+        {
+            _parameters = parameters ?? throw new ArgumentNullException(nameof(parameters));
+            if (_parameters.Credentials == null)
+                throw new ArgumentException("Credentials must be set.", nameof(parameters));
+            if (string.IsNullOrEmpty(_parameters.Region))
+                throw new ArgumentException("Region must be set.", nameof(parameters));
+            if (string.IsNullOrEmpty(_parameters.Service))
+                throw new ArgumentException("Service must be set.", nameof(parameters));
+        }
+
+        /// <summary>
+        /// Signs <paramref name="request"/> with SigV4 and forwards it to the inner handler.
+        /// </summary>
+        protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+        {
+            if (request == null)
+                throw new ArgumentNullException(nameof(request));
+            if (request.RequestUri == null || !request.RequestUri.IsAbsoluteUri)
+                throw new InvalidOperationException(
+                    "The request URI must be an absolute URI. Set HttpClient.BaseAddress or use an absolute request URI.");
+
+            var parameters = ResolveParameters(request);
+            var signingRequest = await BuildSigningRequestAsync(request, parameters).ConfigureAwait(false);
+
+            var result = await AWSSigV4Signer.SignAsync(signingRequest, parameters, cancellationToken).ConfigureAwait(false);
+
+            ApplySigningHeaders(request, result);
+
+            return await base.SendAsync(request, cancellationToken).ConfigureAwait(false);
+        }
+
+        private static AWSSigningParameters BuildParameters(AWSCredentials credentials, RegionEndpoint region, string service)
+        {
+            if (region == null)
+                throw new ArgumentNullException(nameof(region));
+
+            return new AWSSigningParameters
+            {
+                Credentials = credentials,
+                Region = region.SystemName,
+                Service = service,
+            };
+        }
+
+        // Produce a per-request copy of the defaults, applying any per-message overrides. A copy is required
+        // because the same handler instance signs concurrent requests that may carry different overrides.
+        private AWSSigningParameters ResolveParameters(HttpRequestMessage request)
+        {
+            var parameters = new AWSSigningParameters
+            {
+                Credentials = _parameters.Credentials,
+                Region = _parameters.Region,
+                Service = _parameters.Service,
+                SignPayload = _parameters.SignPayload,
+                SignedAt = _parameters.SignedAt,
+            };
+
+            if (TryGetOption(request, ServiceOptionKey, out string service) && !string.IsNullOrEmpty(service))
+                parameters.Service = service;
+            if (TryGetOption(request, RegionOptionKey, out string region) && !string.IsNullOrEmpty(region))
+                parameters.Region = region;
+            if (TryGetOption(request, SignPayloadOptionKey, out bool signPayload))
+                parameters.SignPayload = signPayload;
+
+            return parameters;
+        }
+
+        private static async Task<AWSSigningRequest> BuildSigningRequestAsync(HttpRequestMessage request, AWSSigningParameters parameters)
+        {
+            // The body is only needed when the signer will hash it: payload signing is on and the caller
+            // hasn't already supplied a hash. Otherwise (UNSIGNED-PAYLOAD, or a precomputed hash) the body is
+            // never read, which is what keeps a large or streaming upload from being buffered.
+            var hasPrecomputedHash = HasContentSha256Header(request);
+            var mustHashBody = parameters.SignPayload && !hasPrecomputedHash && request.Content != null;
+
+            byte[] body = null;
+            if (mustHashBody)
+            {
+                body = await request.Content.ReadAsByteArrayAsync().ConfigureAwait(false);
+
+                // Replace the (possibly one-shot) content with a re-readable buffer so the body survives a
+                // resend on retry, preserving the original content headers.
+                var buffered = new ByteArrayContent(body);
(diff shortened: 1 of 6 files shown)


sdk/src/Core/Amazon.Runtime/Signing/SigV4SigningHandler.cs
```diff
@@ -0,0 +1,296 @@
...
+            // hasn't already supplied a hash. Otherwise (UNSIGNED-PAYLOAD, or a precomputed hash) the body is
+            // never read, which is what keeps a large or streaming upload from being buffered.
+            var hasPrecomputedHash = HasContentSha256Header(request);
+            var mustHashBody = parameters.SignPayload && !hasPrecomputedHash && request.Content != null;
+
+            byte[] body = null;
+            if (mustHashBody)
+            {
+                cancellationToken.ThrowIfCancellationRequested();
+                // Reading the content here to hash it also buffers it internally: HttpContent loads its body
```
GarrettBeatty, line 244: i tested this with some test throwaway program. originally i had some logic here to rebuffer things but seems like its not needed

sdk/src/Core/Amazon.Runtime/Signing/SigV4SigningHandler.cs
```diff
@@ -0,0 +1,296 @@
...
+    /// <c>HttpRequestMessage.Options</c> (net5.0+) or <c>HttpRequestMessage.Properties</c> (earlier
+    /// targets), using <see cref="ServiceOptionKey"/>, <see cref="RegionOptionKey"/>, and
+    /// <see cref="SignPayloadOptionKey"/>.
+    /// </para>
+    /// <para>
+    /// Credentials are resolved on every send (not captured at construction), so a rotating credential
+    /// source such as SSO, assume-role, or IMDS always signs with current credentials.
+    /// </para>
+    /// <para>
+    /// <b>Important — disable automatic redirects.</b> The primary/inner handler must set
```
GarrettBeatty, line 43: we should document this behavior in the docs too once we make those changes

sdk/src/Core/Amazon.Runtime/Signing/SigV4SigningHandler.cs
```diff
@@ -0,0 +1,296 @@
...
+        // because the same handler instance signs concurrent requests that may carry different overrides.
+        private AWSSigV4Parameters ResolveParameters(HttpRequestMessage request)
+        {
+            var parameters = new AWSSigV4Parameters
+            {
+                Credentials = _parameters.Credentials,
+                Region = _parameters.Region,
+                Service = _parameters.Service,
+                SignPayload = _parameters.SignPayload,
+                SignedAt = _parameters.SignedAt,
```
dscpinheiro, line 198: Is copying the signed time correct for retries?

sdk/test/UnitTests/Custom/Runtime/Signing/SigV4SigningHandlerTests.cs
```diff
@@ -0,0 +1,373 @@
...
+        private static void SetOption<T>(HttpRequestMessage message, string key, T value)
+        {
+#if NET5_0_OR_GREATER
+            message.Options.Set(new HttpRequestOptionsKey<T>(key), value);
+#else
+            message.Properties[key] = value;
+#endif
+        }
+
+        private static void AssertThrows<TException>(Action action) where TException : Exception
```
dscpinheiro: Why did you need to create these two helpers? Can't the `Assert.ThrowsException<T> / ThrowsExceptionAsync<T>` MSTest has be used instead?

(They only catch the exact exception while the `catch (TException)` you have here catches sub-classes too but looks for this test it wouldn't matter)

sdk/test/IntegrationTests/Tests/SigV4SigningHandlerIntegrationTests.cs
```diff
@@ -0,0 +1,103 @@
...
+    /// End-to-end tests for <see cref="SigV4SigningHandler"/>. These install the handler on a real
+    /// <see cref="HttpClient"/> and send requests to STS, proving that a request signed transparently by the
+    /// handler is accepted by a real AWS service.
+    ///
+    /// STS <c>GetCallerIdentity</c> is used because it is callable by any valid identity and needs no
+    /// provisioned resources. It accepts the request either as a GET (empty body) or a POST with the action
+    /// in a form-encoded body, so the same operation exercises both the no-body and body-hashing paths.
+    /// </summary>
+    [Trait("Category", "SigV4Signer")]
+    [Collection("Serial")]
```
dscpinheiro: Same feedback as the previous PR on the serial attribute.

sdk/src/Core/Amazon.Runtime/Signing/SigV4SigningHandler.cs
```diff
@@ -0,0 +1,296 @@
...
+
+            return false;
+        }
+
+        // Apply the computed signing headers, replacing any left over from a previous signing pass so a
+        // re-signed retry does not accumulate duplicate header lines. A signing header may also have been set
+        // by the caller on the content headers (e.g. a precomputed X-Amz-Content-SHA256), so clear it from
+        // there too — otherwise the outgoing request carries the header twice (once on the request, once on
+        // the content), which SigV4 validation can reject.
+        private static void ApplySigningHeaders(HttpRequestMessage request, AWSSigningResult result)
```
dscpinheiro, line 323: Unlikely, but if the credentials change from short-term to long-term the old `x-amz-security-token` would stay in the headers, right? If yes, should that specific header always be removed?

sdk/src/Core/Amazon.Runtime/Signing/SigV4SigningHandler.cs
```diff
@@ -0,0 +1,296 @@
...
+            if (request.Content != null)
+            {
+                foreach (var header in request.Content.Headers)
+                    headers[header.Key] = JoinValues(header.Value);
+            }
+        }
+
+        private static string JoinValues(IEnumerable<string> values)
+        {
+            return string.Join(",", values.Select(v => v == null ? string.Empty : v.Trim()));
```
dscpinheiro, line 301: Also unlikely, but: If a customer specifies the same header on both the request and the content, only the latter will be used in the signature right?

For example:
```csharp
request.Headers.TryAddWithoutValidation("x-custom", "request-value");
request.Content.Headers.TryAddWithoutValidation("x-custom", "content-value");
```

The signer will only consider `x-custom:content-value` but the HttpClient will send both: `x-custom:request-value, content-value`

sdk/src/Core/Amazon.Runtime/Signing/SigV4SigningHandler.cs
GarrettBeatty: i think it would be because:

1. if they provided it - that means they always want this value to be used, even if its a retry
2. If they dont provide it (most common scenario) we auto calculate it behind the scenes and its a fresh value

sdk/test/IntegrationTests/Tests/SigV4SigningHandlerIntegrationTests.cs
GarrettBeatty: will fix

sdk/src/Core/Amazon.Runtime/Signing/SigV4SigningHandler.cs
```diff
@@ -0,0 +1,296 @@
...
+            if (_parameters.Region == null)
+                throw new ArgumentException("Region must be set.", nameof(parameters));
+            if (string.IsNullOrEmpty(_parameters.Service))
+                throw new ArgumentException("Service must be set.", nameof(parameters));
+        }
+
+        /// <summary>
+        /// Signs <paramref name="request"/> with SigV4 and forwards it to the inner handler.
+        /// </summary>
+        protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
```
dscpinheiro, line 131: Noticed this later: Is not overriding the sync `Send` method intentional?

sdk/src/Core/Amazon.Runtime/Signing/SigV4SigningHandler.cs
GarrettBeatty: this is also the reason the user's retry logic has to be outside the sigv4 handler - so a fresh timestamp is created everytime. And if the user provided a `signedAt` time, its assumed that they would be recalculating it on their end for each retry

sdk/src/Core/Amazon.Runtime/Signing/SigV4SigningHandler.cs
GarrettBeatty: updated to add sync

sdk/src/Core/Amazon.Runtime/Signing/SigV4SigningHandler.cs
GarrettBeatty: updated and added an integ test which confirms 200 success when multiple are there

sdk/src/Core/Amazon.Runtime/Signing/SigV4SigningHandler.cs
GarrettBeatty: updated to also remove that header

sdk/src/Core/Amazon.Runtime/Signing/SigV4SigningHandler.cs
```diff
@@ -0,0 +1,361 @@
...
+    /// <remarks>
+    /// <para>
+    /// The signing region, service, and payload-signing mode are handler-level defaults for the common
+    /// one-service-per-client case. A single client that must sign for more than one service or region,
+    /// or vary payload signing per request, can override any of them on an individual message through
+    /// <c>HttpRequestMessage.Options</c> (net5.0+) or <c>HttpRequestMessage.Properties</c> (earlier
+    /// targets), using <see cref="ServiceOptionKey"/>, <see cref="RegionOptionKey"/>, and
+    /// <see cref="SignPayloadOptionKey"/>.
+    /// </para>
+    /// <para>
```
normj, line 42: I think this `para` about credentials might confuse / concern users because they might read it as a bad thing that credentials are being resolved on every request. The fact that this takes in an `AWSCredentials` like any other library we create to me means we don't need to give it extra explanation.

sdk/src/Core/Amazon.Runtime/Signing/SigV4SigningHandler.cs
```diff
@@ -0,0 +1,361 @@
...
+        {
+        }
+
+        /// <summary>
+        /// Creates a handler from a set of default signing parameters. <see cref="AWSSigV4Parameters.Credentials"/>
+        /// is resolved on every send. Leave <see cref="AWSSigV4Parameters.SignedAt"/> unset so each request is
+        /// signed with a current timestamp.
+        /// </summary>
+        /// <param name="parameters">The default signing parameters.</param>
+        public SigV4SigningHandler(AWSSigV4Parameters parameters)
```
normj: To simplify the experience especially since an inner handler is always required I suggest taking it as part of the constructor and defaulting to `SocketsHttpHandler` if the user doesn't specify a handler.

```


        public SigV4SigningHandler(AWSSigV4Parameters parameters, HttpMessageHandler? innerHandler = null)
             : base(innerHandler ?? new SocketsHttpHandler())
        {
            _parameters = parameters ?? throw new ArgumentNullException(nameof(parameters));
            if (_parameters.Credentials == null)
                throw new ArgumentException("Credentials must be set.", nameof(parameters));
            if (_parameters.Region == null)
                throw new ArgumentException("Region must be set.", nameof(parameters));
            if (string.IsNullOrEmpty(_parameters.Service))
                throw new ArgumentException("Service must be set.", nameof(parameters));
        }

```

sdk/src/Core/Amazon.Runtime/Signing/SigV4SigningHandler.cs
```diff
@@ -0,0 +1,361 @@
...
+            };
+        }
+
+        // Produce a per-request copy of the defaults, applying any per-message overrides. A copy is required
+        // because the same handler instance signs concurrent requests that may carry different overrides.
+        private AWSSigV4Parameters ResolveParameters(HttpRequestMessage request)
+        {
+            var parameters = new AWSSigV4Parameters
+            {
+                Credentials = _parameters.Credentials,
```
normj, line 194: This reminded to look at the ValidateArguments in the new stand alone signer. https://github.com/aws/aws-sdk-net/blob/standalone-sigv4-signer/sdk/src/Core/Amazon.Runtime/Signing/AWSSigV4Signer.cs#L367

In there we are throwing an error if `Region` or `Credentials` are not set. But we should only throw if they are null and their values can't be resolved using `FallbackRegionFactory` or `DefaultAWSCredentialsIdentityResolver.GetCredentials()`. That matches the behavior of our other libraries.

normj (changes requested):

sdk/src/Core/Amazon.Runtime/Signing/SigV4SigningHandler.cs
GarrettBeatty: fixed

sdk/src/Core/Amazon.Runtime/Signing/SigV4SigningHandler.cs
GarrettBeatty: fixed

sdk/src/Core/Amazon.Runtime/Signing/SigV4SigningHandler.cs
GarrettBeatty: fixed. SocketsHttpHandler doesnt exist on net472 so i use HttpClientHandler instead

normj (approved):

dscpinheiro (approved):

PR #4477, opened 2026-07-28
fix: Handle zero-byte objects in TransferUtility RANGE multipart download

GarrettBeatty:
## Description

Fixes a bug where `TransferUtility` multipart download in **RANGE** mode fails on zero-byte (empty) objects.

The RANGE strategy's discovery request issues a ranged GET (`bytes=0-{partSize-1}`) as its very first request. For a zero-byte object, S3 returns **416 Range Not Satisfiable** (there is no byte 0 to satisfy), which previously propagated and failed the entire download. The default **PART** strategy is unaffected because it discovers with `partNumber=1` (which succeeds on empty objects).

## Fix

In `MultipartDownloadManager.DiscoverUsingRangeStrategyAsync`, the initial ranged GET is now sent via a helper that:

1. Sends the ranged GET (normal path for objects with data).
2. On **416**, probes with a `partNumber=1` GET to disambiguate.
3. If the probe confirms the object is empty (`ContentLength == 0`), returns it — the response has no `ContentRange`, so the existing single-part branch completes the download as a 0-byte file.
4. If the probe returns a **non-empty** object (the object was replaced server-side between the ranged GET and the probe), throws an `InvalidOperationException` so the caller can retry the whole download. This is intentionally *not* retried internally, consistent with how the IfMatch/ETag guard already fails subsequent parts on mid-download mutation.

This aligns with the AWS SDK for Java v2, whose range-based (presigned-URL) downloader handles the same 416-on-empty-object case explicitly.

## Testing

- **Unit** (`MultipartDownloadManagerTests`):
  - `DiscoverUsingRangeStrategy_EmptyObject_FallsBackToPartNumberGet` — 416 → `partNumber=1` probe → single 0-byte part.
  - `DiscoverUsingRangeStrategy_ObjectReplacedAfter416_Throws` — 416 → probe returns non-empty → throws.
  - Full `MultipartDownloadManager` suite: **109/109 pass**.
- **Integration** (`TransferUtilityDownloadWithResponseTests`):
  - `DownloadWithResponse_RangeStrategy_EmptyObject` — verified against real S3; passes with the fix (fails without it). Fills a coverage gap where the only prior empty-object test used the PART strategy.

## Dry-run builds

Chained Catapult dry-run (prod stage, source `origin:transfer-util-range-empty-object-fallback`):

- **.NET (dotnetv4):** `c2cf323a-f8df-48e3-8ec7-b998763ba3ca`
- **PowerShell (powershell5):** `590100a1-fb60-4d92-9835-88be951865aa`

### DIFF
--- a/generator/.DevConfigs/a6947522-0d0b-4bfb-80cd-67f65f65d9cb.json
+++ b/generator/.DevConfigs/a6947522-0d0b-4bfb-80cd-67f65f65d9cb.json
@@ -1,0 +1,11 @@
+{
+  "services": [
+    {
+      "serviceName": "S3",
+      "type": "patch",
+      "changeLogMessages": [
+        "Fixed `TransferUtility` multipart download in RANGE mode failing on zero-byte objects. The initial ranged GET (`bytes=0-{partSize-1}`) returns `416 Range Not Satisfiable` for an empty object; the RANGE strategy now detects the 416, probes with a `partNumber=1` GET, and completes as an empty file, matching the PART strategy's behavior."
+      ]
+    }
+  ]
+}
--- a/sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
@@ -866,7 +866,7 @@ namespace Amazon.S3.Transfer.Internal
 
                 var partProbeResponse = await _s3Client.GetObjectAsync(partProbeRequest, cancellationToken).ConfigureAwait(false);
 
-                if (partProbeResponse != null && partProbeResponse.ContentLength == 0)
+                if (partProbeResponse != null && partProbeResponse.ContentLength == 0 && partProbeResponse.ContentRange == null)
                 {
                     // Confirmed empty. This response has no ContentRange, so the caller's ContentRange == null
                     // branch treats it as a single empty part and completes with a 0-byte file.
--- a/sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
+++ b/sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
@@ -24,6 +24,7 @@ using System.Buffers;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
+using System.Net;
 using System.Threading;
 using System.Threading.Tasks;
 using Amazon.Runtime;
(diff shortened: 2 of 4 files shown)


generator/.DevConfigs/a6947522-0d0b-4bfb-80cd-67f65f65d9cb.json
```diff
@@ -0,0 +1,11 @@
+{
+  "services": [
+    {
+      "serviceName": "S3",
+      "type": "patch",
+      "changeLogMessages": [
+        "Fixed `TransferUtility` multipart download in RANGE mode failing on zero-byte objects. The initial ranged GET (`bytes=0-{partSize-1}`) returns `416 Range Not Satisfiable` for an empty object; the RANGE strategy now detects the 416, probes with a `partNumber=1` GET, and completes as an empty file, matching the PART strategy's behavior."
```
GarrettBeatty, line 7: this behavior is consistent with java sdk https://github.com/aws/aws-sdk-java-v2/blob/master/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/multipart/PresignedUrlDownloadHelper.java#L74 and i confirmed it with zoe and dimitriy in transfer manager slack channel

sdk/src/Services/S3/Custom/Transfer/Internal/MultipartDownloadManager.cs
```diff
@@ -833,6 +831,61 @@ private async Task<DownloadResult> DiscoverUsingRangeStrategyAsync(CancellationT
...
+            {
+                _logger.DebugFormat("MultipartDownloadManager: [Part 1 Discovery] Received 416 Range Not Satisfiable on the ranged GET; " +
+                    "probing with a partNumber=1 GET to confirm the object is empty.");
+
+                var partProbeRequest = CreateGetObjectRequest();
+                partProbeRequest.PartNumber = 1;
+
+                var partProbeResponse = await _s3Client.GetObjectAsync(partProbeRequest, cancellationToken).ConfigureAwait(false);
+
+                if (partProbeResponse != null && partProbeResponse.ContentLength == 0)
```
dscpinheiro: Do you need to check for `&& partProbeResponse.ContentRange == null` here too? (that's what the - existing - line 769 has, and the comment on line 871 says "this response has no ContentRange")

GarrettBeatty: i dont think it really matters right? either should work i think?

GarrettBeatty: but either way i should update the comment

dscpinheiro (approved):

sankettangade (approved):

PR #4488, opened 2026-08-13
Simplify RDS and DSQL auth token generators onto the public SigV4 facade

GarrettBeatty:
## Description

Task 8 follow-up from the standalone SigV4 signing work: simplify the RDS and DSQL IAM auth-token generators by building their presigned tokens through the public `AWSSigV4Signer.Presign` facade instead of the internal signing types. Also dogfoods the new public signing API.

## What changed

- Both generators now call `AWSSigV4Signer.Presign(...)` over a neutral `AWSSigningRequest`, dropping the hand-built `DefaultRequest`, private request subclass, `RequestMetrics`, `AWS4PreSignedUrlSigner`, `ComposeUrl`, manual query splicing, and `https://`-stripping.
- Output tokens are byte-identical and public method signatures are unchanged (internal refactor only).
- Existing credential-refresh behavior is preserved, including DSQL's 15-minute credential-lifetime floor.
- On net472, `System.Net.Http` is now referenced by the RDS/DSQL projects (needed because `AWSSigningRequest.HttpMethod` is `System.Net.Http.HttpMethod`). Added per-service `reference-dependencies` support to the generator so only these two services get it.

## Testing

- RDS and DSQL unit tests pass on net472, netcoreapp3.1, and net8.0.
- Affected packages and the generator build clean.

Opened as a draft against `development`.

### DIFF
--- a/sdk/src/Services/DSQL/AWSSDK.DSQL.NetFramework.csproj
+++ b/sdk/src/Services/DSQL/AWSSDK.DSQL.NetFramework.csproj
@@ -76,6 +76,7 @@
 
   <ItemGroup>
     <Reference Include="System.Configuration"/>
+    <Reference Include="System.Net.Http"/>
   </ItemGroup>
 
 
--- a/sdk/test/Services/RDS/UnitTests/Custom/RDSAuthTokenGeneratorTest.cs
+++ b/sdk/test/Services/RDS/UnitTests/Custom/RDSAuthTokenGeneratorTest.cs
@@ -213,6 +213,22 @@ namespace AWSSDK.UnitTests.RDS
             }, typeof(ArgumentException));
         }
 
+        [TestMethod]
+        [TestCategory("RDS")]
+        public void GenerateAuthTokenDBUserRequiringEncoding()
+        {
+            // DBUser is a user-controlled query parameter. It must be URL-encoded exactly once so that
+            // decoding the DBUser value out of the token yields the original string. This guards against a
+            // regression where the value is double-encoded (e.g. '@' -> %40 -> %2540).
+            const string specialUser = "db user+admin@corp";
+            var token = RDSAuthTokenGenerator.GenerateAuthToken(BasicCredentials, AWSRegion, DBHost, DBPort, specialUser);
+
+            var match = Regex.Match(token, "[?&]DBUser=([^&]*)");
+            Assert.IsTrue(match.Success, "DBUser parameter not found in token: " + token);
+            Assert.AreEqual(specialUser, Uri.UnescapeDataString(match.Groups[1].Value),
+                "DBUser should be single-encoded so it decodes back to the original value. Token: " + token);
+        }
+
         private void AssertAuthToken(string token, string accessKey, RegionEndpoint region)
         {
             AssertAuthToken(token, accessKey, region, false);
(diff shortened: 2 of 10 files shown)


sdk/src/Services/DSQL/AWSSDK.DSQL.NetFramework.csproj
```diff
@@ -76,6 +76,7 @@
 
   <ItemGroup>
     <Reference Include="System.Configuration"/>
+    <Reference Include="System.Net.Http"/>
```
GarrettBeatty, line 79: is it fine to add this dependency?

dscpinheiro: It's referenced in Core... https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Core/AWSSDK.Core.NetFramework.csproj#L67 (but I don't see it in any service project)

GarrettBeatty: the other option is i add some override to allow just accepting a `string` and then we dont need the dependency here

normj: It is fine to add the reference. It doesn't change anything from the customer point of view because these assemblies are already available to users via the GAC with .NET Framework. The reference just tells the compiler when building this assembly to include it when looking for type references.

GarrettBeatty: oh i didnt realize its not from nuget in this case thanks

GarrettBeatty:
converting to draft while i add some more unit tests

normj (dismissed):

sdk/test/Services/RDS/UnitTests/Custom/RDSAuthTokenGeneratorTest.cs
```diff
@@ -208,6 +208,22 @@ public void GenerateAuthTokenHighPort()
             }, typeof(ArgumentException));
         }
 
+        [TestMethod]
+        [TestCategory("RDS")]
+        public void GenerateAuthTokenDBUserRequiringEncoding()
```
GarrettBeatty, line 218: ran this on development branch and it passes and verified it still passed on new branch

normj (dismissed):

GarrettBeatty:
[catapult-net]   PS   (powershell5): buildId = df668b46-3870-442e-8b42-3624a8c45b20
{
  "buildId": "d62cc97d-0d6e-45b0-8fb4-873c69d3f6e6",
  "buildType": "DRY_RUN",
  "stage": "PROD",
  "overrides": {
    "sources": [
      {
        "sourceId": "origin",
        "sourceReference": "task8-simplify-rds-dsql-token-generators"
      }
    ]
  }
}
{
  "buildId": "df668b46-3870-442e-8b42-3624a8c45b20",
  "buildType": "DRY_RUN",
  "stage": "PROD",
  "overrides": {
    "sources": [
      {
        "sourceId": "dotnetbuildid",
        "sourceReference": "d62cc97d-0d6e-45b0-8fb4-873c69d3f6e6"
      }
    ]
  }
}

dscpinheiro (approved):

normj (approved):

PR #4495, opened 2026-09-03
[4.1] Analyzer warning when targeting deprecated .NET Targets

normj:
## Description
Starting with SDK v4.1 we are dropping the dedicated `netcoreapp3.1` target from the SDK assemblies. When a project still targeting a dropped framework (e.g. .NET Core 3.1) updates to v4.1, NuGet does not fail — it quietly resolves the `netstandard2.0` assets instead. That downgrade is invisible to the user and means they miss the features and optimizations available on supported targets.

To make that fallback visible, this PR adds a Roslyn code analyzer that ships inside every AWS SDK for .NET NuGet package and emits an `AWSSDK3000` warning when the consuming project targets a framework the SDK no longer ships assemblies for.

## How it works
- **Shared analyzer logic** lives in `sdk/code-analysis/SharedAnalysisCode/AbstractDeprecatedTargetFrameworkAnalyzer.cs`. It runs a compilation action that reads the project's target framework (from the `build_property.TargetFramework` global analyzer config, falling back to the `TargetFrameworkAttribute` for older .NET SDKs) and reports `AWSSDK3000` if it matches a deprecated target. A per-compilation claim ensures the warning is reported only once even when multiple SDK packages (e.g. Core + S3) are referenced.
- **The deprecated target list is data-driven.** A new `DeprecatedTargets` MSBuild property in `sdk/Directory.Build.props` (currently `netcoreapp3.1`) is baked by `SharedAnalysisCode.csproj` into the generated `DeprecatedTargetFrameworks.ShortNames` array the analyzer checks against, so future dropped targets only need a props change.
- **Per-service analyzers are code-generated.** The generator now emits a `DeprecatedTargetFrameworkAnalyzer` and a `AWSSDK.<Service>.CodeAnalysis.csproj` for each service (see `DeprecatedTargetFrameworkAnalyzerWriter`, `CodeAnalysisProjectCreator`, and the `.tt` templates), plus the Core analyzer project under `sdk/code-analysis/CoreAnalysis`.
- **Analyzers are bundled into the packages.** The nuspec generators (`CoreNuspec`/`Nuspec` `.tt` + `.cs`) now add the built analyzer DLLs to the `analyzers\dotnet\cs` folder of each package (excluding the Roslyn/framework assemblies), so the warning lights up automatically on install with no extra opt-in.
- **The warning message is not tied to a specific target.** The message format takes the offending framework, the fallback (`netstandard2.0`), and the suppression id as arguments rather than hard-coding `netcoreapp3.1`, so it stays correct as the deprecated set changes.
- Also updates the older generator nuspec templates and the `4.1` framework list (removing `netcoreapp3.1`, adding `net10.0`/`net11.0`) so generated packages match the 4.1 targets.

Users who intentionally stay on a fallback target can suppress the warning with `<NoWarn>AWSSDK3000</NoWarn>`.

## Why ship the analyzer in every service package instead of only Core?
Since every service package depends on `AWSSDK.Core`, it is tempting to place the analyzer only in Core and rely on the dependency to carry it. That d
oes not work: **NuGet analyzers are not applied transitively.** The default `PrivateAssets` for a `PackageReference` is `contentfiles;analyzers;build`
, so a package's `analyzers/dotnet/cs` assets are private to that dependency edge and are only applied to projects that reference the package *directl
y*.

For the common case where a customer references only a service package:
```xml
<PackageReference Include="AWSSDK.S3" ... />
```
`AWSSDK.Core` is pulled in transitively, but Core's analyzer never runs against the customer's compilation, so no `AWSSDK3000` warning would be produc
ed — exactly the customers we want to reach. A Core-only analyzer would fire only for the minority who also add a direct `AWSSDK.Core` reference. Ther
e is no way to fix this from our side either, because asset-flow-to-parent is governed by the consuming project's `PackageReference` settings, which w
e cannot set from a nuspec.

Bundling the analyzer in every service package is therefore required, not redundant. The trade-off is that a project referencing several AWS packages
loads several independent copies of the analyzer. `AbstractDeprecatedTargetFrameworkAnalyzer.TryClaimCompilation` handles this by parking a claim on t
he `AppDomain` keyed by the `Compilation`, so the warning is reported only once no matter how many AWS packages are referenced.

## Testing

Besides the new tests I have built NuGet packages from both the old and new generator and confirmed the warning message showed up when targeting .NET Core 3.1 and not when targeting a modern version of .NET.

Example of the build output when targeting .NET Core 3.1 showing the `AWSSDK3000` warning. The project had both Core and S3 referenced but the warning was only displayed once.
```
> dotnet build
C:\Program Files\dotnet\sdk\11.0.100-preview.7.26381.103\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.EolTargetFrameworks.targets(32,5): warning NETSDK1138: The target framework 'netcoreapp3.1' is out of support and will not receive security updates in the future. Please refer to https://aka.ms/dotnet-core-support for more information about the support policy.
Restore succeeded with 1 warning(s) in 0.1s
    C:\Program Files\dotnet\sdk\11.0.100-preview.7.26381.103\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.EolTargetFrameworks.targets(32,5): warning NETSDK1138: The target framework 'netcoreapp3.1' is out of support and will not receive security updates in the future. Please refer to https://aka.ms/dotnet-core-support for more information about the support policy.
    info NETSDK1057: You are using a preview version of .NET. See: https://aka.ms/dotnet-support-policy
  ConsoleApp17 netcoreapp3.1 succeeded with 4 warning(s) (1.0s) → bin\Debug\netcoreapp3.1\ConsoleApp17.dll
    C:\Users\normj\.nuget\packages\system.text.encodings.web\8.0.0\buildTransitive\netcoreapp2.0\System.Text.Encodings.Web.targets(4,5): warning System.Text.Encodings.Web 8.0.0 doesn't support netcoreapp3.1 and has not been tested with it. Consider upgrading your TargetFramework to net6.0 or later. You may also set <SuppressTfmSupportBuildWarnings>true</SuppressTfmSupportBuildWarnings> in the project file to ignore this warning and attempt to run in this unsupported configuration at your own risk.
    C:\Users\normj\.nuget\packages\system.text.json\8.0.6\buildTransitive\netcoreapp2.0\System.Text.Json.targets(4,5): warning System.Text.Json 8.0.6 doesn't support netcoreapp3.1 and has not been tested with it. Consider upgrading your TargetFramework to net6.0 or later. You may also set <SuppressTfmSupportBuildWarnings>true</SuppressTfmSupportBuildWarnings> in the project file to ignore this warning and attempt to run in this unsupported configuration at your own risk.
    C:\Program Files\dotnet\sdk\11.0.100-preview.7.26381.103\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.EolTargetFrameworks.targets(32,5): warning NETSDK1138: The target framework 'netcoreapp3.1' is out of support and will not receive security updates in the future. Please refer to https://aka.ms/dotnet-core-support for more information about the support policy.
    CSC : warning AWSSDK3000: The target framework 'netcoreapp3.1' is no longer supported by the AWS SDK for .NET. This project has fallen back to the netstandard2.0 assemblies of the SDK, which do not include the features and optimizations available on supported targets. Retarget the project to .NET 8 or later, or suppress this warning with <NoWarn>AWSSDK3000</NoWarn>.

Build succeeded with 6 warning(s) in 1.2s

```


generator/ServiceClientGeneratorLib/Generators/NuGet/CoreNuspec.tt
```diff
@@ -44,12 +46,17 @@
...
     <file src=".\bin\Release\netstandard2.0\AWSSDK.Core.pdb" target="lib\netstandard2.0"/>
 
-    <file src=".\bin\Release\netcoreapp3.1\AWSSDK.Core.dll" target="lib\netcoreapp3.1"/>
-    <file src=".\bin\Release\netcoreapp3.1\AWSSDK.Core.xml" target="lib\netcoreapp3.1"/>
-    <file src=".\bin\Release\netcoreapp3.1\AWSSDK.Core.pdb" target="lib\netcoreapp3.1"/>
-
     <file src=".\bin\Release\net8.0\AWSSDK.Core.dll" target="lib\net8.0"/>
     <file src=".\bin\Release\net8.0\AWSSDK.Core.xml" target="lib\net8.0"/>
     <file src=".\bin\Release\net8.0\AWSSDK.Core.pdb" target="lib\net8.0"/>
+
```
normj, line 52: I made this fix use hardcoding instead of doing the extra work to make it generic because this is the old generator that is in the process of being replaced.

generator/ServiceClientGeneratorLib/Generators/NuGet/Nuspec.tt
```diff
@@ -129,13 +129,17 @@
...
     <file src=".\bin\Release\netstandard2.0\\<#=this.Session["AssemblyName"]#>.pdb" target="lib\netstandard2.0" />
 
-    <file src=".\bin\Release\netcoreapp3.1\\<#=this.Session["AssemblyName"]#>.dll" target="lib\netcoreapp3.1" />
-    <file src=".\bin\Release\netcoreapp3.1\\<#=this.Session["AssemblyName"]#>.xml" target="lib\netcoreapp3.1" />
-    <file src=".\bin\Release\netcoreapp3.1\\<#=this.Session["AssemblyName"]#>.pdb" target="lib\netcoreapp3.1" />
-
     <file src="<#=Path.Combine(".", "bin", "Release", "net8.0", (string)this.Session["AssemblyName"] + ".dll")#>" target="lib\net8.0" />
     <file src="<#=Path.Combine(".", "bin", "Release", "net8.0", (string)this.Session["AssemblyName"] + ".xml")#>" target="lib\net8.0" />
     <file src="<#=Path.Combine(".", "bin", "Release", "net8.0", (string)this.Session["AssemblyName"] + ".pdb")#>" target="lib\net8.0" />
+
```
normj, line 135: I made this fix use hardcoding instead of doing the extra work to make it generic because this is the old generator that is in the process of being replaced.

generator/ServiceClientGeneratorLib/Generators/NuGet/CoreNuspec.tt
```diff
@@ -26,15 +26,17 @@
...
+      <group targetFramework="net10.0">
+      </group>		
+      <group targetFramework="net11.0">
+      </group>		
     </dependencies>
   </metadata> 
   <files>
     <file src="..\..\nuget-content\AWSLogo.png" target="images\" />
-    <file src="..\..\nuget-content\account-management.ps1" target="tools" />
+    <file src="..\..\code-analysis\CoreAnalysis\bin\Release\*.dll" target="analyzers\dotnet\cs" exclude="**\Microsoft.CodeAnalysis.*;**\System.Collections.Immutable.*;**\System.Reflection.Metadata.*;**\System.Composition.*" />
```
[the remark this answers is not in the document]
normj: I removed the `account-management.ps1` script on purpose because it was causing me packaging problems and the script really is very out of date. Instead of ballooning this PR I'm going to create a separate PR for cleaning up the rest of the `account-management.ps1` and `install.ps` scripts.

generator/ServiceClientGeneratorLib/GenerationManifest.cs
```diff
@@ -103,7 +103,7 @@ public string PreviewLabel
         public DefaultConfiguration.DefaultConfigurationModel DefaultConfiguration { get; set; }
 
         //This should be the same version number as SdkVersioning.DefaultAssemblyVersion in BuildTasks
-        private const string DefaultAssemblyVersion = "4.0";
+        private const string DefaultAssemblyVersion = "4.1";
```
dscpinheiro: Does this line need to change too? https://github.com/aws/aws-sdk-net/blob/ed2e2105e0888d1a304c174d6dee8a3d6a6291ff/generator/ServiceModels/_sdk-versions.json#L3546

sdk/code-analysis/SharedAnalysisCode/AbstractDeprecatedTargetFrameworkAnalyzer.cs
```diff
@@ -0,0 +1,212 @@
...
+
+                return Version.TryParse(trimmed.Substring("Version=".Length).TrimStart('v', 'V'), out version);
+            }
+
+            return false;
+        }
+
+        private static bool IsDeprecatedShortName(string targetFramework)
+        {
+            return DeprecatedTargetFrameworks.ShortNames
```
dscpinheiro, line 169: Should we also catch [OS specific TFMs](https://learn.microsoft.com/en-us/dotnet/standard/frameworks#os-specific-tfms) (e.g. `net6.0-browser`)?  If I understand this method correctly it only considers the target as deprecated if the name matches exactly.

dscpinheiro (approved):

sdk/code-analysis/SharedAnalysisCode/AbstractDeprecatedTargetFrameworkAnalyzer.cs
normj: Good catch. Latest commit handles OS specific TFMs.

muhammad-othman (approved):

PR #4497, opened 2026-09-03
fix: preserve sub-second precision for JSON-protocol epoch-seconds timestamps

### LINKED ISSUE #4496: CreateEvent (and all JSON-protocol timestamps) marshal epoch-seconds as truncated Int64, dropping sub-second precision
### Describe the bug

For a Smithy `timestamp` shape that has **no explicit `timestampFormat`** under a JSON protocol (`rest-json` / `json-rpc`), the generated request marshaller serializes the value as a **whole-second `Int64`**, silently discarding any sub-second (millisecond) component.

Per the [restJson1 protocol](https://smithy.io/2.0/aws/protocols/aws-restjson1-protocol.html#json-shape-serialization), a `timestamp` with no format defaults to `epoch-seconds`, which is represented as a number that **may include a fractional part**. The SDK truncates that fraction, so the output is spec-under-compliant and lossy on the wire.

I hit this via `AWSSDK.BedrockAgentCore` (`CreateEvent` → Amazon Bedrock AgentCore Memory), where the timestamp is used as a high-resolution ordering key, but the root cause is in shared code generation and affects many services.

### Regression?

No — this appears to be long-standing behavior, not a recent regression.

### Expected behavior

For an `EventTimestamp` of `2025-08-25T19:19:14.123Z`, the request body should preserve the fraction:

```json
{ "eventTimestamp": 1756149554.123 }
```

### Current behavior

The fraction is dropped before the request is sent:

```json
{ "eventTimestamp": 1756149554 }
```

### Root cause

The generated marshaller `sdk/src/Services/BedrockAgentCore/Generated/Model/Internal/MarshallTransformations/CreateEventRequestMarshaller.cs` emits:

```csharp
context.Writer.WriteNumberValue(
    Convert.ToInt64(StringUtils.FromDateTimeToUnixTimestamp(publicRequest.EventTimestamp.Value)));
```

This originates from the generator template `generator/ServiceClientGeneratorLib/Generators/Marshallers/JsonRPCStructureMarshaller.tt` (`DetermineNormalJsonWriteMethod`) for any timestamp whose resolved format is `UnixTimestamp` (the correct default for JSON protocols). Two steps discard precision:

1. `StringUtils.FromDateTimeToUnixTimestamp` → `AWSSDKUtils.ConvertToUnixEpochSecondsString`, which is internally `Convert.ToInt64(...TotalSeconds).ToString()` — the fraction is already gone here.
2. The outer `Convert.ToInt64(...)` re-parses that whole-seconds string into a `long`, so `WriteNumberValue` never receives a fraction to write.

The format *selection* is correct; only the numeric conversion is lossy. Notably, the SDK already contains a fractional helper, `AWSSDKUtils.ConvertToUnixEpochSecondsDouble` (returns e.g. `1756149554.123` as a `double`), and `Utf8JsonWriter.WriteNumberValue(double)` serializes the fraction correctly — the default path just doesn't use it. The read/unmarshal path is already correct: `DateTimeUnmarshaller.UnmarshallInternal` uses `Double.TryParse` + `AddSeconds`, so it round-trips fractions when the server returns them.

The affected service model shape is a bare `"Timestamp":{"type":"timestamp"}` (no `timestampFormat`) in `bedrock-agentcore-2024-02-28.normal.json`.

### Affects both generators

This is not specific to the classic C2J generator. The newer SmithyDotNet generator reproduces the identical truncation in `generator/SmithyDotNet/SmithyDotNet.Generator/Writers/Serialization/JsonScalarMarshaller.cs` (`WriteTimestamp`), whose `epoch-seconds` branch is:

```csharp
"epoch-seconds" => $"context.Writer.WriteNumberValue(Convert.ToInt64(StringUtils.FromDateTimeToUnixTimestamp({epochValue})));",
```

(the source even comments that it is "matching C2J"). A complete fix should therefore address **both** generators — or the shared helper — otherwise migrating a service from C2J to the Smithy generator silently carries the truncation forward.

### Scope

Not service-specific. The exact `WriteNumberValue(Convert.ToInt64(StringUtils.FromDateTimeToUnixTimestamp(...)))` pattern appears in ~425 generated marshaller files across AWS services. Most consumers are unaffected because whole-second resolution is sufficient for typical timestamp fields; it becomes load-bearing for AgentCore Memory, which derives an event ordering key of the form `{zero-padded-epoch-millis}#{random-hex}` from this timestamp — flattening millis to `000` makes same-second events sort by a random suffix instead of chronologically.

### Reproduction steps

```csharp
var client = new AmazonBedrockAgentCoreClient();

// Two events in the same wall-clock second, differing only in milliseconds
await client.CreateEventAsync(new CreateEventRequest {
    MemoryId = memoryId, SessionId = sid, ActorId = sid,
    EventTimestamp = new DateTime(2025, 8, 25, 19, 19, 14, 123, DateTimeKind.Utc),
    Payload = [ /* ... */ ]
});
await client.CreateEventAsync(new CreateEventRequest {
    MemoryId = memoryId, SessionId = sid, ActorId = sid,
    EventTimestamp = new DateTime(2025, 8, 25, 19, 19, 14, 456, DateTimeKind.Utc),
    Payload = [ /* ... */ ]
});
```

Observe (via a wire capture, or `ListEvents(IncludePayloads=true)` and inspecting the returned `eventId`s): both `eventTimestamp`s serialize to `1756149554`, and both `eventId` millis components are `000`, so the two events do not preserve their write order.

A wire-level assertion is the tightest repro: marshal a `CreateEventRequest` with a fractional `EventTimestamp` and assert the JSON body contains `1756149554.123`, not `1756149554`.

### Possible solution

Two options, in increasing blast radius:

1. **Narrow:** route the epoch-seconds default through a fractional conversion (the generator already supports per-member custom transformations, e.g. `Amazon.Util.AWSSDKUtils.ConvertToUnixEpochMilliseconds`). Apply a fractional-seconds transformation so only the affected shape(s) change.
2. **Broad / spec-correct:** change the `UnixTimestamp` / `epoch-seconds` branch in both `JsonRPCStructureMarshaller.tt` and `JsonScalarMarshaller.cs` to emit `AWSSDKUtils.ConvertToUnixEpochSecondsDouble(...)` via `WriteNumberValue(double)`, making all JSON-protocol services spec-compliant. This is the correct long-term fix but changes on-wire output for many services and warrants protocol-test validation.

### Environment

- SDK: `AWSSDK.BedrockAgentCore` 4.0.23 (also reproducible in shared codegen for other JSON-protocol services)
- Observed against AgentCore Memory `CreateEvent` / `ListEvents`

### Precedent in other AWS SDKs

.NET is the outlier here — other SDKs already emit fractional epoch-seconds (or at least allow it).

**Java v2 preserves milliseconds by default** (`software.amazon.awssdk.utils.DateUtils`):

```java
/** Formats the given Instant into a Unix timestamp with millisecond decimal precision. */
public static String formatUnixTimestampInstant(Instant instant) {
    BigDecimal dateValue = BigDecimal.valueOf(instant.toEpochMilli());
    return dateValue.scaleByPowerOfTen(0 - MILLI_SECOND_PRECISION).toPlainString(); // MILLI_SECOND_PRECISION = 3
}
```

It builds the value from `toEpochMilli()` and shifts the decimal 3 places, emitting e.g. `1756149554.123`. No opt-in required — correct by default.

**Python/botocore truncates by default but exposes an opt-in** (`botocore/serialize.py`, `JSONSerializer`/`RestJSONSerializer`, `TIMESTAMP_FORMAT = 'unixtimestamp'`):

```python
def _timestamp_unixtimestamp(self, value):
    if self._timestamp_precision == TIMESTAMP_PRECISION_MILLISECOND:
        base_timestamp = calendar.timegm(value.timetuple())
        milliseconds = (value.microsecond // 1000) / 1000.0
        return base_timestamp + milliseconds        # e.g. 1756149554.123
    else:
        return int(calendar.timegm(value.timetuple()))  # default: truncated
```

Summary:

| SDK | Default unix-timestamp on wire | Sub-second opt-in? |
|---|---|---|
| Java v2 | `1756149554.123` (ms preserved) | n/a — correct by default |
| Python/botocore | `1756149554` (truncated) | Yes — `timestamp_precision='millisecond'` |
| .NET | `1756149554` (truncated) | No |

This means the .NET fix would bring it in line with Java v2 rather than introducing new behavior — mirror `formatUnixTimestampInstant` by building from `AWSSDKUtils.ConvertToUnixEpochMilliseconds` (or the existing `ConvertToUnixEpochSecondsDouble`) and emitting a decimal via `WriteNumberValue`.

peterrsongg:
## Description

For a `timestamp` shape whose resolved format is `epoch-seconds` (the default under `rest-json` /
`json-rpc` when no `timestampFormat` is set), the JSON body marshallers emitted a truncated
whole-second `Int64`:

```csharp
context.Writer.WriteNumberValue(Convert.ToInt64(StringUtils.FromDateTimeToUnixTimestamp(publicRequest.EventTimestamp.Value)));
```

Both conversions in that expression drop the fraction — `FromDateTimeToUnixTimestamp` →
`ConvertToUnixEpochSecondsString` is already whole seconds, and the outer `Convert.ToInt64` reparses
it — so `WriteNumberValue` never sees a fractional value.

This changes the `epoch-seconds` body branch in **both** generators to write the value with
millisecond precision:

```csharp
context.Writer.WriteNumberValue(Amazon.Util.AWSSDKUtils.ConvertToUnixEpochSecondsDecimal(publicRequest.EventTimestamp.Value));
```

- `generator/ServiceClientGeneratorLib/Generators/Marshallers/JsonRPCStructureMarshaller.tt` (C2J) —
  all four `epoch-seconds` sites: member, sparse member, map value, sparse map value.
- `generator/SmithyDotNet/.../Serialization/JsonScalarMarshaller.cs` (SmithyDotNet) — the
  `WriteTimestamp` `epoch-seconds` branch, so a service migrated from C2J doesn't carry the
  truncation forward.
- New `AWSSDKUtils.ConvertToUnixEpochSecondsDecimal(DateTime)` in `AWSSDK.Core`.

### Why `decimal` rather than the existing `ConvertToUnixEpochSecondsDouble`

`WriteNumberValue(double)` is **not** TFM-stable. System.Text.Json formats a `double`
shortest-round-trippable on .NET Core, but with `G17` on net472/netstandard2.0, so the same call
emits different bytes per target:

| conversion | net8.0 | net472 |
|---|---|---|
| `double` | `1756149554.123` | `1756149554.1229999` |
| `decimal` | `1756149554.123` | `1756149554.123` |

`decimal` also keeps whole seconds **byte-identical** to the pre-fix output (`1756149554`, no
trailing `.0`), which is what keeps this change invisible to the many services and callers that only
ever send whole seconds.

Header / query / label `epoch-seconds` bindings are unchanged (still whole seconds via
`StringUtils.FromDateTimeToUnixTimestamp`) — only the JSON body path is in scope here.

The read path already round-trips fractions: `DateTimeUnmarshaller.UnmarshallInternal` uses
`Double.TryParse` + `AddSeconds`.

## Motivation and Context

Fixes #4496.

Per the [restJson1 protocol](https://smithy.io/2.0/aws/protocols/aws-restjson1-protocol.html#json-shape-serialization),
a `timestamp` with no explicit format defaults to `epoch-seconds`, represented as a number that **may
include a fractional part**. Truncating it is under-compliant and lossy on the wire.

It becomes load-bearing for Bedrock AgentCore Memory `CreateEvent`, which derives an event ordering
key of the form `{zero-padded-epoch-millis}#{random-hex}` from `eventTimestamp`. Flattening the
millis to `000` makes same-second events sort by a random suffix instead of chronologically.

This brings .NET in line with the Java v2 SDK, which has always preserved milliseconds by default
(`DateUtils.formatUnixTimestampInstant` builds from `Instant#toEpochMilli` and shifts the decimal
three places). Python/botocore truncates by default but exposes `timestamp_precision='millisecond'`;
.NET was the only SDK with no way to send the fraction at all.

## Testing

Verified end-to-end against the local build with a console app that marshals a real
`CreateEventRequest` offline (no service call) and inspects the JSON body:

```
fraction .123  2025-08-25T19:19:14.1230000Z      ->  eventTimestamp: 1756149554.123
fraction .456  2025-08-25T19:19:14.4560000Z      ->  eventTimestamp: 1756149554.456
whole second   2025-08-25T19:19:14.0000000Z      ->  eventTimestamp: 1756149554
sub-ms ticks   2025-08-25T19:19:14.1234567Z      ->  eventTimestamp: 1756149554.123
local kind     2025-08-25T12:19:14.1230000-07:00 ->  eventTimestamp: 1756149554.123
epoch          1970-01-01T00:00:00.0000000Z      ->  eventTimestamp: 0
pre-epoch      1969-12-31T23:59:59.8770000Z      ->  eventTimestamp: -0.123
```

Sub-millisecond ticks round to the nearest millisecond, matching Java v2. Pre-epoch values stay
negative and correct.

Existing suites, all green on the regenerated code:

- `sdk/test/ProtocolTests` — 751/751 on net8.0. The smithy `JsonTimestamps` / `JsonLists` /
  `KitchenSink` cases all use whole seconds and are unaffected, which confirms the no-op for
  whole-second values.
- `generator/SmithyDotNet` — `dotnet test SmithyDotNet.Generator.sln`, 669/669. Codegen expectations
  in `ScalarMemberCodegenTests` / `CollectionElementCodegenTests` updated to the new emitted call.
- `generator/SmithyDotNet/skills/marshalling/SKILL.md` timestamp table updated to match.

### Dry-runs
- **DotNet Dry-run ID:** cb43dc41-bcb0-4a2e-bf7c-91e99a9ac2f9
  - [x] Pending
  - [ ] Completed successfully
  - [ ] Failed
- **PowerShell Dry-run ID:** b2467ce9-8a68-4285-ae2a-f2b37f11b0a4
  - [x] Pending
  - [ ] Completed successfully
  - [ ] Failed

Not a breaking API change — no public signature changes; `ConvertToUnixEpochSecondsDecimal` is
purely additive.

It *is* an on-wire behavior change for JSON-protocol services, so calling it out explicitly:

1. **What changed:** a body `epoch-seconds` timestamp now serializes with its millisecond fraction
   (`1756149554.123`) instead of being truncated (`1756149554`). ~425 generated marshaller files
   across services carry the pattern.
   * **Customer impact:** none for callers passing whole seconds — the emitted bytes are identical.
     Callers who pass a `DateTime` with a millisecond component will now have it reach the service
     rather than being silently dropped. Services accept a fractional epoch-seconds number per the
     protocol spec, and the value is still a JSON number in the same field.
   * **Why not narrower:** the per-member customization route (option 1 in #4496) would fix only
     AgentCore's shape and leave every other JSON-protocol service under-compliant, so the next
     service to need sub-second precision would file the same bug. The lossy step is in shared
     codegen, which is where it should be fixed.
   * **Best practices:** matches Java v2's default, and `decimal` was chosen over `double`
     specifically so the wire bytes don't vary by target framework.
   * **Testing:** see above.
2. Senior/+ engineer review requested given the cross-service wire impact.

## Types of changes

## Checklist

## License


peterrsongg:
Also checked the utf8JsonWriter claim that writeNumberValue writes a different value for double across TFMs and it was true.
```
Confirmed empirically against the real SDK — the existing double helper is not sufficient.

net8.0 — double works, identical to decimal:
fraction .123  wire: 1756149554.123    Double: 1756149554.123    Decimal: 1756149554.123
PASS: body contains 1756149554.123

net472 — double breaks:
case           marshalled wire value    ...Double                ...Decimal
fraction .123  1756149554.1229999       1756149554.1229999       1756149554.123
fraction .456  1756149554.4560001       1756149554.4560001       1756149554.456
sub-ms ticks   1756149554.1229999       1756149554.1229999       1756149554.123
whole second   1756149554               1756149554               1756149554
epoch          0                        0                        0
pre-epoch      -0.123                   -0.123                   -0.123

FAIL: body does not contain 1756149554.123

Note .456 → 1756149554.4560001 — it errs upward there, so it isn't even a consistent truncation you could reason about downstream.

Why: ConvertToUnixEpochSecondsDouble itself is fine — the bug i). System.Text.Json's TryFormatDouble calls Utf8Formatter(shortest round-trippable) only on inbox builds; on the netstandard2.0 package build it falls back to value.ToString("G17"). So the same helper emits different bytes per TFM. That's also why the existing CBOR use  — CborWriterExtensions.WriteOptimizedNumber writes a binaryIEEE float, never decimal text.

AgentCore's eventId key is {zero-padded-epoch-millis}#{hex}, so 1756149554.1229999 on .NET Framework would land in a different millisecond than on .NET Core for the identical DateTime — a worse failure than the origs now platform-dependent.
```

sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
```diff
@@ -792,6 +792,21 @@ public static double ConvertToUnixEpochSecondsDouble(DateTime dateTime)
...
+        /// Converts a DateTime to Unix epoch seconds, keeping millisecond precision
+        /// (for example 1756149554.123). Whole seconds carry no fractional part.
+        /// </summary>
+        /// <remarks>
+        /// Returns <see cref="decimal"/> rather than <see cref="double"/> so that the value
+        /// serializes to the same digits on every target framework: System.Text.Json formats
+        /// a double with "G17" on .NET Framework/netstandard2.0 (1756149554.123 renders as
+        /// 1756149554.1229999) but shortest-round-trippable on .NET Core.
+        /// </remarks>
+        public static decimal ConvertToUnixEpochSecondsDecimal(DateTime dateTime)
```
dscpinheiro, line 805: We can add a test for this new method in https://github.com/aws/aws-sdk-net/tree/main/sdk/test/UnitTests/Custom/Util and it'll run for all TFMs (seems worth to me since we don't cover this in protocol tests...)

generator/.DevConfigs/d195c828-0ced-4e30-8c3a-cb1f04d73c6a.json
```diff
@@ -0,0 +1,9 @@
+{
+  "core": {
+    "updateMinimum": true,
+    "type": "patch",
+    "changeLogMessages": [
+      "All JSON based services will send sub-second precise timestamps in the body if specified on the request."
```
dscpinheiro: What do you think of this? 

```suggestion
      "Timestamps serialized as numeric epoch-seconds in JSON request bodies now preserve millisecond precision (e.g. `1756149554.123` instead of `1756149554`) rather than truncating to whole seconds"
```

peterrsongg: I like it, updated

sdk/src/Core/Amazon.Util/AWSSDKUtils.cs
peterrsongg: added.

normj (approved):

dscpinheiro (approved):
Approving assuming the dry-runs pass.

PR #4464, opened 2026-07-10
Re-implement region fix without imds fallback

peterrsongg:
## Description

Due to https://github.com/aws/aws-sdk-net/issues/4444 and https://github.com/aws/aws-sdk-net/issues/4445. I had to rollback the fix. If a user provides a custom endpoint via AWS_ENDPOINT_URL or serviceURL and that endpoint is not an aws endpoint, the SDK will look to the `FallbackRegionFactory` for the region instead of defaulting to `us-east-1` always. 

This was also pointed out in the latest Localstack release: https://github.com/localstack-dotnet/dotnet-aspire-for-localstack/releases/tag/13.4.0

I did a test with a console app and and made sure it doesn't fallback to IMDS.

## Testing

I tested the exact scenario in the issue 
```
var lambdaConfig = new AmazonLambdaConfig
{
    ServiceURL = "https://fake-endpoint.com"
};
var creds = new BasicAWSCredentials("test", "test");
using var lambdaClient = new AmazonLambdaClient(creds, lambdaConfig);
try
{
    await lambdaClient.ListFunctionsAsync().ConfigureAwait(continueOnCapturedContext: false);

}
catch (Exception e)
{
    
}
```

### Dry-runs

- **DotNet Dry-run ID:** cf9ff13b-40e5-4ff4-ba37-5a939f13785d
  - [ ] Pending
  - [x] Completed successfully
  - [ ] Failed
- **PowerShell Dry-run ID:** ec778314-b286-4c66-bc0e-a02e90740cec
  - [ ] Pending
  - [x] Completed successfully
  - [ ] Failed

## Breaking Changes Assessment

1. Identify all breaking changes including the following details:
    * What functionality was changed?
    * How will this impact customers?
    * Why does this need to be a breaking change and what are the most notable non-breaking alternatives?
    * Are best practices being followed?
    * How have you tested this breaking change?
2. Has a senior/+ engineer been assigned to review this PR?

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] My code follows the code style of this project
- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [ ] I have read the **README** document
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed

## License

- [x] I confirm that this pull request can be released under the Apache 2 license

[...]

Issue #4445, opened 2026-06-22
Performance degradation after upgrade to AWSSDK.Core 4.0.9.4


martincostello:
### Describe the bug

After upgrading from version `4.0.24.4` of AWSSDK.S3 to `4.0.25.2` (https://github.com/martincostello/aspnetcore-opentelemetry-benchmarks/commit/21e36bd92d9e4efe25b0d70064e667f54890f3f3) I noticed a large performance regression in terms of both duration and allocation in some continuous benchmarking I have:

<img width="700" height="450" alt="Image" src="https://github.com/user-attachments/assets/8e1362d6-012e-4d30-85e0-af9a850b5b8a" />

These [benchmarks](https://github.com/martincostello/aspnetcore-opentelemetry-benchmarks/blob/74ab7d529185ce53677951e1bac7f2d718171694/perf/Benchmarks/AwsBenchmarks.cs) perform the following operation using [floci](https://github.com/floci-io/floci):

```csharp
_ = await client.GetBucketLocationAsync("benchmarks");
```

with an additional single setup call to create the bucket:

```csharp
await client.PutBucketAsync(new PutBucketRequest() { BucketName = "benchmarks" });
```

~~I imagine this is a direct consequence of the fix for #4432.~~

Actually, I'm not sure what changed caused this, but reverting to `4.0.24.5` shows things return to the performance before the upgrade:

<img width="700" height="450" alt="Image" src="https://github.com/user-attachments/assets/bab724fc-a0cf-486d-b3a1-602060fe931a" />

### Regression Issue

- [x] Select this option if this issue appears to be a regression.

### Expected Behavior

Performance is approximately the same as `4.0.24.x`.

### Current Behavior

- Duration is ~5x larger
- Allocated memory is ~1.7x larger

### Reproduction Steps

1. Clone https://github.com/martincostello/aspnetcore-opentelemetry-benchmarks/commit/21e36bd92d9e4efe25b0d70064e667f54890f3f3
2. Ensure `docker` is running
3. Run `./benchmark.ps1 -Filter "*AWS*" -Job Short` from the root of the repository

### AWS .NET SDK and/or Package version used

AWSSDK.S3@4.0.25.2

### Targeted .NET Platform

.NET 10

### Operating System and version

Ubuntu 24.04

dscpinheiro:
We're taking a look, that S3 fix was supposed to only impact a very specific scenario (non-seekable empty streams in PutObject...). 

[...]

PR #4501, opened 2026-09-08
Cache failed lookups in FallbackRegionFactory

martincostello:
## Description

Cache a failed lookup to get the fallback region if AWS region environment variables are not available to avoid repeated internal failing calls to avoid ~12KB of allocations.

## Motivation and Context

I have some benchmarks that track the performance of the OpenTelemetry .NET SDK over time, and when updating from 4.0.101 to 4.0.101.3 of the AWSSDK.S3 there was a notable regression in allocations as highlighted by the purple box in the screenshot below:

image

I asked Claude Code to investigate the regression, which found that the regression is related to the same functionality as identified in #4445, which was reintroduced with improvements by #4464.

Without these changes, the impact can be mitigated by setting `AmazonS3Config.AuthenticationRegion` ([example](https://github.com/martincostello/aspnetcore-opentelemetry-benchmarks/commit/a975affa75012738acece95dd8c6793cbe102c2e)):

image

## Testing

Existing tests, but new unit tests.

### Dry-runs

- **DotNet Dry-run ID:**
  - [ ] Pending
  - [ ] Completed successfully
  - [ ] Failed
- **PowerShell Dry-run ID:**
  - [ ] Pending
  - [ ] Completed successfully
  - [ ] Failed

See above.

## Types of changes

## Checklist

## License

### DIFF
--- a/generator/.DevConfigs/6cd71680-e75d-43e4-90c1-2986368df4b5.json
+++ b/generator/.DevConfigs/6cd71680-e75d-43e4-90c1-2986368df4b5.json
@@ -1,0 +1,9 @@
+{
+  "core": {
+    "updateMinimum": false,
+    "type": "patch",
+    "changeLogMessages": [
+      "Fixed `FallbackRegionFactory.GetRegionEndpoint` re-running its full region probe on every call when no source resolves a region."
+    ]
+  }
+}
--- a/sdk/src/Core/Amazon.Runtime/AWSRegion.cs
+++ b/sdk/src/Core/Amazon.Runtime/AWSRegion.cs
@@ -196,6 +196,8 @@ namespace Amazon.Runtime
         public static void Reset()
         {
             cachedRegion = null;
+            nonMetadataLookupFailed = false;
+            allSourcesLookupFailed = false;
             AllGenerators = new List<RegionGenerator>
             {
                 () => new AppConfigAWSRegion(),
@@ -214,6 +216,15 @@ namespace Amazon.Runtime
 
         private static AWSRegion cachedRegion;
 
+        // Remembers that a generator set was already probed and found nothing, so repeated calls
+        // (e.g. once per request, from RegionFinder, for a client whose ServiceURL never resolves
+        // to a region from any source) don't re-run the same failing - and sometimes exception
+        // throwing, disk-IO performing - probe every time. A failure of the metadata-inclusive set
+        // implies the non-metadata subset also failed; the reverse doesn't hold, since
+        // InstanceProfileAWSRegion (EC2 instance metadata) might still find a region.
+        private static bool nonMetadataLookupFailed;
+        private static bool allSourcesLookupFailed;
+
         public static RegionEndpoint GetRegionEndpoint()
         {
             return GetRegionEndpoint(true);
@@ -225,6 +236,9 @@ namespace Amazon.Runtime
             {
                 if (cachedRegion != null)
                     return cachedRegion.Region;
+
+                if (allSourcesLookupFailed || (!includeInstanceMetadata && nonMetadataLookupFailed))
+                    return null;
 
                 IEnumerable<RegionGenerator> generators
                     = includeInstanceMetadata ? AllGenerators : NonMetadataGenerators;
@@ -246,6 +260,13 @@ namespace Amazon.Runtime
 
                     if (cachedRegion != null)
                         break;
+                }
+
+                if (cachedRegion == null)
+                {
+                    nonMetadataLookupFailed = true;
+                    if (includeInstanceMetadata)
+                        allSourcesLookupFailed = true;
                 }
 
                 return cachedRegion != null ? cachedRegion.Region : null;
--- a/sdk/test/UnitTests/Custom/Runtime/Credentials/FallbackFactoryTest.cs
+++ b/sdk/test/UnitTests/Custom/Runtime/Credentials/FallbackFactoryTest.cs
@@ -164,6 +164,81 @@ namespace AWSSDK.UnitTests
         }
 
         [TestMethod]
+        public void TestFailedRegionLookupIsCachedUntilReset()
+        {
+            var newEnvVariables = new Dictionary<string, string>()
+            {
+                { EnvironmentVariableAWSRegion.ENVIRONMENT_VARIABLE_REGION, string.Empty },
+                { EnvironmentVariableAWSRegion.ENVIRONMENT_VARIABLE_DEFAULT_REGION, string.Empty },
+            };
+
+            // A profile name that doesn't exist in ProfileText, so every non-metadata source
+            // (app config, environment variable, profile) fails to resolve a region.
+            using (new FallbackFactoryTestFixture(ProfileText, "doesNotExist", newEnvVariables))
+            {
+                var region = FallbackRegionFactory.GetRegionEndpoint(false);
+                Assert.IsNull(region);
+
+                var originalRegionEnvVar = Environment.GetEnvironmentVariable(EnvironmentVariableAWSRegion.ENVIRONMENT_VARIABLE_REGION);
+                try
+                {
+                    // A region now becomes available. Before caching a failed lookup, the SDK
+                    // re-ran the full (exception-throwing) probe on every call and would have
+                    // picked this up; it must now stay cached as "not found" until Reset().
+                    Environment.SetEnvironmentVariable(EnvironmentVariableAWSRegion.ENVIRONMENT_VARIABLE_REGION, "us-west-2");
+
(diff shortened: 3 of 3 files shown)


sdk/test/UnitTests/Custom/Runtime/Credentials/FallbackFactoryTest.cs
```diff
@@ -163,6 +163,81 @@ public void TestOtherProfile()
...
+                var region = FallbackRegionFactory.GetRegionEndpoint(false);
+                Assert.IsNull(region);
+
+                var originalRegionEnvVar = Environment.GetEnvironmentVariable(EnvironmentVariableAWSRegion.ENVIRONMENT_VARIABLE_REGION);
+                try
+                {
+                    // A region now becomes available. Before caching a failed lookup, the SDK
+                    // re-ran the full (exception-throwing) probe on every call and would have
+                    // picked this up; it must now stay cached as "not found" until Reset().
+                    Environment.SetEnvironmentVariable(EnvironmentVariableAWSRegion.ENVIRONMENT_VARIABLE_REGION, "us-west-2");
```
peterrsongg, line 188: In general I think this is a good change, but instead of putting this in just a comment here I think we should make this clear in the changelog that a side effect of your change is that if a region becomes available after the result is already cached as null, the FallbackRegionFactory won't pick it up anymore until Reset() is called.

generator/.DevConfigs/6cd71680-e75d-43e4-90c1-2986368df4b5.json
```diff
@@ -0,0 +1,9 @@
+{
+  "core": {
+    "updateMinimum": false,
+    "type": "patch",
+    "changeLogMessages": [
+      "Fixed `FallbackRegionFactory.GetRegionEndpoint` re-running its full region probe on every call when no source resolves a region."
+    ]
+  }
```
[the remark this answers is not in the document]
peterrsongg: this is a valid callout

sdk/src/Core/Amazon.Runtime/AWSRegion.cs
```diff
@@ -196,6 +196,8 @@ static FallbackRegionFactory()
         public static void Reset()
         {
             cachedRegion = null;
+            nonMetadataLookupFailed = false;
+            allSourcesLookupFailed = false;
```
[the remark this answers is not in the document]
peterrsongg: seems like this is something that already existed, but I don't see the harm in wrapping it in a lock.
```


lock (_lock)
{
          cachedRegion = null;
          nonMetadataLookupFailed = false;
          allSourcesLookupFailed = false;
          AllGenerators = new List<RegionGenerator>
          {
              () => new AppConfigAWSRegion(),
              () => new EnvironmentVariableAWSRegion(),
              () => new ProfileAWSRegion(credentialProfileChain),
              () => new InstanceProfileAWSRegion()
          };
        
          NonMetadataGenerators = new List<RegionGenerator>
          {
              () => new AppConfigAWSRegion(),
              () => new EnvironmentVariableAWSRegion(),
              () => new ProfileAWSRegion(credentialProfileChain),
          };
}
```

sdk/src/Core/Amazon.Runtime/AWSRegion.cs
```diff
@@ -226,6 +237,9 @@ public static RegionEndpoint GetRegionEndpoint(bool includeInstanceMetadata)
                 if (cachedRegion != null)
                     return cachedRegion.Region;
 
+                if (allSourcesLookupFailed || (!includeInstanceMetadata && nonMetadataLookupFailed))
```
peterrsongg, line 240: What makes me a bit nervous here is that if the first region probe happens while IMDS is briefly unreachable, every subsequent request returns null for the rest of the process lifetime. Whereas before, if this happened, the process would self-heal since the probe happened every time. 

@normj what do you think? do you think that's fine?

I did have an idea in mind, where we can put a time-bound fix in `Ec2InstanceMetadata.cs` so that basically if we know IMDS is unreachable for a transient reason, we mark it as unreachable and for the next 30s (or whatever time window) we return null instantly like this PR was attempting to do. Then after that window is up we will try the reprobe again. but maybe that is overkill.

normj: I am worried that by caching the cache miss we stop the ability for self correction. 

@martincostello What is your performance testing that triggers this? Because my understanding is this should only happen if no region is explicitly set or there is no environmental region that can be found. In which case the SDK is going to fail anyway. So if I have to decide on saving a bit of allocations in a fatal error situation or leave the SDK open for the user to do some self healing I prefer having the ability for healing.

martincostello: The benchmarks use floci as an AWS emulator. They don't fail outright, but as noted in the PR description `AuthenticationRegion` has to be explicitly set to avoid the allocation overhead. It's also not a trivial overhead, the improvement is ~12KB.

peterrsongg: Right, but the PR you linked (which I wrote btw😅) adds the ~12KB allocation when a region couldn't be determined from the endpoint. So is the performance regression you're talking about when no actual AWS service call is being made? I'd assume so if the regression is being reported from the emulator.

martincostello: No, the regression appears to occur when the region can't automatically be determined (e.g. when the service endpoint is overridden, like when using an emulator like localstack or floci). The benchmarks showing the allocation are doing an S3 call to list buckets.

peterrsongg: is there repro code I can see?

martincostello: I linked to the change in the PR description that makes the allocation go away here: https://github.com/martincostello/aspnetcore-opentelemetry-benchmarks/commit/a975affa75012738acece95dd8c6793cbe102c2e

normj: Sorry I'm still confused. If the region was not detected how was the call to your AWS emulator successful? I agree that 12KB allocation is significant but this scenario should mean a production application is misconfigured and access to AWS will fail. In that scenario the 12KB allocation is insignificant to the production down issue for the misconfigured application. 

As @peterrsongg said there can also be times, although rare, that EC2's instance metadata can be slow so health checks on an application might fail for a bit when the instance first comes online because the region can't be detected. With the current reevaluation that application will become health once IMDS catches up.

martincostello: I can debug it tomorrow, but AFAIK the calls to S3 _shouldn't_ be failing, otherwise the benchmark should fail due to a thrown exception from an HTTP 500 exception bubbling out of the handler listing the bucket names.

normj: The SDK itself should have failed before even making the HTTP call because it would have needed the region to sign the request.

peterrsongg: With the change I made, if the region couldn't be determined from the endpoint, it would go to the fallbackFactory and if that failed, then return the default `us-east-1` like before. 

Seems like in order to bypass all of this, martin set the AuthenticationRegion. However, the failure case is the one that causes the allocation spike, so still curious how the allocation spike occurs and the call is successful.

martincostello: I took a quick look in the debugger, and the call to [`IAmazonS3.GetBucketLocationAsync(string)`](https://github.com/martincostello/aspnetcore-opentelemetry-benchmarks/blob/9d131d77cd56d4dbc2519d15c1051c1e0754c5ab/perf/Benchmarks/AwsBenchmarks.cs#L64) succeeds with and without `AmazonS3Config.AuthenticationRegion` being set.

normj: Okay I dug into the SDK code some more and I do still think this is more of a testing/benchmark scenario and not a real production issue.

Since your are using the emulator and manually setting the service url the code goes into different code path for determining region where for legacy reasons will default to us-east-1. https://github.com/aws/aws-sdk-net/blob/37a8573299620e862a70031ecc0ba3ee52cebd69/sdk/src/Core/Amazon.Util/Internal/RegionFinder.cs#L105

In the normal prod situation where you are hitting the real AWS and relying on the region being configured you would get the error "'No RegionEndpoint or ServiceURL configured" before the SDK even attempted to make an HTTP call.

PR #4504, opened 2026-09-10
Generate EC2 password test key at runtime

sankettangade:
## Description

Replaces the hardcoded PEM RSA private key in `PasswordTest.TestMissingLeadingZeroIssue` (`sdk/test/Services/EC2/UnitTests/Custom/PasswordTest.cs`) with a key generated at test time.

The test exists to guard one specific bug: when an RSA CRT parameter's value is shorter than its field width, it is DER-encoded without its leading zero byte, and `PemReader.ReadPrivatekey()` + `RSACryptoServiceProvider.ImportParameters` used to mishandle that and mangle the decrypted password. The original fixture key was chosen because its `InverseQ` has that leading zero. Roughly one key in forty has the property, so the fixture is generated in a loop until a qualifying key appears — swapping in an arbitrary key would silently stop testing the bug.

.NET Framework has no PEM or PKCS#1 export (`RSA.ExportRSAPrivateKey` is .NET Core 3.0+), so the generated key is encoded as a DER `RSAPrivateKey` (RFC 3447 A.1.2) by hand. That is why this is not a straight port of the equivalent v4 change.

The assertion is also tightened: the test previously only checked `Assert.IsNotNull` on the result, and now round-trips a known plaintext and asserts equality.

## Motivation and Context

An automated AppSec scan (ACAT) flagged the hardcoded key as a hardcoded credential (SIM V2361235327, Sev-3). ACAT scans every branch of the repository, so the same fixture is a finding source on the v3.7 line as well as v4.

The key was never a real credential — it was generated locally for this test, is not associated with any AWS account, EC2 instance, or key pair, and the ciphertext decrypts to a short lowercase test string rather than an EC2-generated Windows password. There is nothing to rotate. This change removes the literal so the finding has no source here.

## Testing

`dotnet test` against `AWSSDK.UnitTests.EC2.Net45.csproj`, five runs, all green:

| Target framework | Runs | Result | Duration |
| --- | --- | --- | --- |
| `net45` | 5 | Passed 1, Failed 0 | 3s / 5s / 8s / 2s / 3s |

This exercises the real v3.7 decrypt path — `ThirdParty.BouncyCastle.OpenSsl.PemReader.ReadPrivatekey()` feeding `RSACryptoServiceProvider.ImportParameters`, then `Decrypt(bytes, false)` for PKCS#1 v1.5 — which is exactly where the original bug lived. Runs were repeated deliberately so different CRT parameters end up being the short one across runs.

The hand-written DER encoder was validated separately against `RSA.ExportRSAPrivateKey()` on .NET 8 over 60 freshly generated 2048-bit keys: **byte-identical in all 60 cases**, including 2 keys that carried a short parameter.

`AWSSDK.UnitTests.EC2.Net35.csproj` could not be built locally: it references `AWSSDK.Core.Net35.csproj`, which fails with `CS0006: Metadata file 'C:\mscorlib.dll' could not be found` because the .NET Framework 3.5 targeting pack is not installed on this machine. That failure reproduces on an unmodified checkout, so it predates this change. Both EC2 test projects declare `<TargetFramework>net45</TargetFramework>`, and the file compiles and passes there, so no net35-specific language or API concern applies — but a CI build is the real confirmation.

Reviewers should weigh the runtime cost: the test went from a few milliseconds to 2-8s, because it generates 2048-bit keys until one qualifies. Dropping to 1024-bit would be roughly 10x faster but trades a hardcoded-key finding for a possible weak-key-size one, so 2048 was kept. The loop is capped at 2000 attempts and throws with a clear message rather than hanging; odds of a spurious failure are negligible.

### Dry-runs
- **DotNet Dry-run ID:**
  - [x] Pending
  - [ ] Completed successfully
  - [ ] Failed
- **PowerShell Dry-run ID:** n/a — test-only change, no SDK surface touched
  - [ ] Pending
  - [ ] Completed successfully
  - [ ] Failed

None. The change touches a single unit test; no product code, public API, or generated code is modified.

## Types of changes
- [x] Test-only change

## Checklist

## License

### DIFF
--- a/sdk/test/Services/EC2/UnitTests/Custom/PasswordTest.cs
+++ b/sdk/test/Services/EC2/UnitTests/Custom/PasswordTest.cs
@@ -1,4 +1,7 @@
 ﻿using System;
+using System.Collections.Generic;
+using System.Security.Cryptography;
+using System.Text;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
 using Amazon.EC2.Model;
@@ -8,6 +11,11 @@ namespace AWSSDK_DotNet35.UnitTests.EC2
     [TestClass]
     public class PasswordTest
     {
+        private const int KeySizeInBits = 2048;
+
+        // Around one key in forty qualifies, so the cap only exists to fail loudly instead of
+        // spinning forever if key generation ever stops behaving randomly.
+        private const int MaxKeyGenerationAttempts = 2000;
 
         // Test an issue with certain combination of private key and encrypted passwords that were causing a encoding issue while decrypting the password.
         [TestMethod]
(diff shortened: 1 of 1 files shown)


dscpinheiro:
This is the `main` branch for V3. It should not be merged to directly.

PR #4507, opened 2026-09-11
fix: Follow region redirect on cross-region HeadBucket requests

sichanyoo:
_The PR description was written with help of AI._

## Description

Fixes cross-region `HeadBucket` failing when the client's configured Region differs from the bucket's Region and the request lands on the wrong regional endpoint. In that case S3 responds with `301 MovedPermanently` carrying the correct Region in the `x-amz-bucket-region` header, but the SDK surfaced it as an `AmazonS3Exception` instead of retrying against the correct Region.

The change is intentionally scoped to `HeadBucket` only. Other S3 operations continue to surface the redirect exactly as before, so no existing behavior changes for `PutObject`/`GetObject`/etc.

There were **two** distinct problems, both fixed in `AmazonS3RetryPolicy`:

1. **Redirect status was not routed to region detection (for `HeadBucket`).**
   `SharedRetryForExceptionSync` only treated `400 BadRequest` as "inconclusive" (`return null`, which triggers `BucketRegionDetector`). A `301 MovedPermanently` (or `308 PermanentRedirect`) fell through to the base retry policy, so the region mismatch was never detected. These statuses now also `return null`, **but only when `OriginalRequest is HeadBucketRequest`**. All other operations are unaffected and still surface the redirect as an error.

2. **Correcting the signing Region alone did not move the request's endpoint.**
   On a detected mismatch the retry previously set only `AuthenticationRegion`, which re-signs the request but leaves it pointed at the **original** endpoint. In the V4 pipeline the S3 endpoint resolver runs **once, outside** the `RetryHandler` loop, so it is never re-evaluated on retry — only the signer and HTTP handler re-run. (Confirmed two ways: dumping the runtime pipeline shows `AmazonS3EndpointResolver` above `RetryHandler`, and instrumenting the resolver shows it runs **1×** for a request retried **3×**.) The request therefore kept hitting the wrong regional endpoint and S3 kept returning `301`/`400` until retries were exhausted.

   New helper `RedirectToRegion` re-resolves the operation endpoint for the corrected Region via `ClientConfig.DetermineServiceOperationEndpoint(...)` and rewrites `Request.Endpoint` directly. Signing is corrected via `AuthenticationRegion` (which `DetermineSigningRegion` reads) plus `IsSigned = false` to defer re-signing to the pipeline on retry. It also removes the stale `Host` header so the re-signed request advertises the new host — without this, TLS fails with `RemoteCertificateNameMismatch`. This mirrors the existing `AmazonS3RedirectHandler.FinalizeForRedirect` (the 307 path); the difference is that it runs in the retry policy (outer to the `Signer`), so it defers re-signing (`IsSigned = false`) rather than signing inline.

   Two robustness details:
   - `DetermineServiceOperationEndpoint` re-resolves through the full endpoint ruleset, **not** a naive region string-swap, so it preserves FIPS/dualstack/accelerate/path-style/ARN handling (verified — accelerate and access-point-ARN endpoints are correctly left unchanged).
   - If the returned Region is unknown to this SDK build, `RedirectToRegion` returns `false` and the caller defers to the base retry policy instead of throwing out of the retry path (`RegionEndpoint.GetBySystemName` guard, matching existing `BucketRegionDetector` handling).

**Files changed**
- `sdk/src/Services/S3/Custom/Internal/AmazonS3RetryPolicy.cs` — `HeadBucket`-scoped 301/308 routing, new `RedirectToRegion` helper, corrected `<returns>` doc-comment for the `null` contract. (`Runtime.` namespace qualifiers were also dropped in favor of the existing `using` directives.)
- `sdk/src/Services/S3/Custom/Internal/_async/AmazonS3RetryPolicy.cs` — call `RedirectToRegion`; added the same credentials-identity guard the `_bcl` path already had, so both retry paths behave consistently.
- `sdk/src/Services/S3/Custom/Internal/_bcl/AmazonS3RetryPolicy.cs` — call `RedirectToRegion`.
- `sdk/test/Services/S3/UnitTests/Custom/HeadBucketRegionRedirectTests.cs` — new tests.

## Motivation and Context

Reported via [aws/aws-tools-for-powershell#413](https://github.com/aws/aws-tools-for-powershell/issues/413) (`Get-S3HeadBucket` returns an error when the shell Region differs from the bucket Region); tracked as **DOTNET-8539**. `Get-S3HeadBucket` wraps `AmazonS3Client.HeadBucketAsync`, so the fix belongs in the .NET SDK.

## Testing

**Unit tests** (`AWSSDK.UnitTests.S3`, net8.0):
- `RedirectStatus_IsRoutedToRegionDetection` — 301/308 for a `HeadBucket` are routed to region detection (`return null`).
- `CrossRegionHeadBucket_RetriesAgainstCorrectRegion` — end-to-end shared async retry: on a mismatch the request is retried with the corrected signing Region (`AuthenticationRegion`) **and** the corrected endpoint host; the detected Region is cached. The endpoint assertion fails if only the signing Region were corrected (i.e., it would have caught the incomplete first attempt).
- `RedirectStatus_ForNonHeadBucketOperation_DefersToBaseRetryPolicy` — a 301/308 for a non-`HeadBucket` operation (`GetObject`) is **not** treated as inconclusive and defers to the base retry policy, locking in the `HeadBucket`-only scope.
- `NonRedirectStatus_DefersToBaseRetryPolicy` — a `404` with no region header defers to the base retry policy (not swallowed).
- Full S3 unit suite: **1500 passed, 0 failed** (net8.0).

**Real-S3 end-to-end** via PowerShell (`pwsh`) loading the patched `AWSSDK.S3.dll` and calling `HeadBucketAsync`. Test buckets were created in `us-east-1` and `us-west-2`, exercised, and deleted.

| client → bucket | before fix | after fix |
|---|:---:|:---:|
| `ap-southeast-1` → `us-east-1` bucket | ❌ FAIL (`MovedPermanently`) | ✅ SUCCESS (`BucketRegion=us-east-1`) |
| `ap-southeast-1` → `us-west-2` bucket | ❌ FAIL (`MovedPermanently`) | ✅ SUCCESS (`BucketRegion=us-west-2`) |
| `us-east-1` → `us-west-2` bucket | ✅ SUCCESS | ✅ SUCCESS |

Environment: macOS (arm64), .NET 8 SDK, PowerShell 7.6, real S3.

### Dry-runs
- **DotNet Dry-run ID:**
  - [x] Pending
  - [ ] Completed successfully
  - [ ] Failed
- **PowerShell Dry-run ID:**
  - [x] Pending
  - [ ] Completed successfully
  - [ ] Failed

Not a breaking change. The behavior change is limited to the failure path **for `HeadBucket` only**: a cross-region `HeadBucket` from a client whose Region differs from the bucket's, which previously threw `AmazonS3Exception (MovedPermanently)`, now transparently retries against the correct regional endpoint and succeeds — matching the AWS CLI and the documented `HeadBucket` contract. No public API surface changes; changes are in `Amazon.S3.Internal` retry logic. All other S3 operations are unchanged — including the existing `Test301RedirectTriggersException` integration test, where a cross-region `PutObject` still throws. Requests that already succeeded are unaffected (they never enter the region-mismatch branch), and genuine failures (e.g. `404` with no region header) still surface as before.

- A senior/+ engineer review is recommended given this touches shared S3 retry/endpoint behavior.

N/A.

## Types of changes

## Checklist

_*No authored/developer-guide docs affected; internal XML doc-comments on the retry-policy members
  were corrected and ship in AWSSDK.S3.xml._

## License

### DIFF
--- a/sdk/src/Services/S3/Custom/Internal/AmazonS3RetryPolicy.cs
+++ b/sdk/src/Services/S3/Custom/Internal/AmazonS3RetryPolicy.cs
@@ -19,6 +19,7 @@ using Amazon.Runtime.Internal;
 using Amazon.Runtime.Internal.Auth;
 using Amazon.S3.Model;
 using Amazon.S3.Util;
+using Amazon.Util;
 using System;
 using System.Collections.Generic;
 using System.Globalization;
@@ -47,11 +48,26 @@ namespace Amazon.S3.Internal
         /// </summary>
         /// <param name="executionContext"></param>
         /// <param name="exception"></param>
-        /// <returns>a value if it can be determined, or null if the IO-bound calculations need to be done</returns>
+        /// <returns>
+        /// <c>true</c> or <c>false</c> when the retry decision can be made here; <c>null</c> when the
+        /// exception looks like a bucket/region mismatch and the decision must be deferred to the
+        /// asynchronous <see cref="AmazonS3RetryPolicy.SharedRetryForExceptionAsync"/> path, which
+        /// resolves the bucket's Region (from the x-amz-bucket-region response header when present,
+        /// otherwise via a HEAD bucket network call) before deciding whether to retry.
+        /// </returns>
         public bool? RetryForExceptionSync(Runtime.IExecutionContext executionContext, Exception exception)
         {
             return SharedRetryForExceptionSync(executionContext, exception, Logger, base.RetryForException);            
         }
+        /// <summary>
+        /// Shared synchronous retry triage for S3.
+        /// Returns <c>true</c>/<c>false</c> for cases that can be decided without any further lookup.
+        /// Returns <c>null</c> to signal that the exception may be a bucket/region mismatch: the
+        /// caller (<see cref="AmazonS3RetryPolicy.SharedRetryForExceptionAsync"/>) then resolves the
+        /// bucket's Region — cheaply from the x-amz-bucket-region response header when it is present,
+        /// or via a HEAD bucket network call as a fallback — and retries against the correct Region.
+        /// <c>null</c> does not by itself imply a network call will happen.
+        /// </summary>
         internal static bool? SharedRetryForExceptionSync(Runtime.IExecutionContext executionContext, Exception exception, 
             Runtime.Internal.Util.ILogger logger,
             Func<Runtime.IExecutionContext, Exception, bool> baseRetryForException)
@@ -69,6 +85,22 @@ namespace Amazon.S3.Internal
                         // response.
                         return true;
                     }
+                }
+
+                // A bucket that lives in a different Region than the one the client is
+                // configured for can respond with a redirect status instead of a 400 when the
+                // bucket is in us-east-1 and the client is not (301 MovedPermanently), or in the
+                // rarer 308 PermanentRedirect case. These responses still carry the correct
+                // Region in the x-amz-bucket-region header, so treat them as inconclusive and let
+                // the caller detect the region mismatch and retry against the correct Region
+                // (mirroring the AWS CLI behavior). Without this, HeadBucket and other operations
+                // against a us-east-1 bucket from a non-us-east-1 client would surface the error
+                // instead of following the redirect.
+                // See https://github.com/aws/aws-tools-for-powershell/issues/413 (DOTNET-8539).
+                if (serviceException.StatusCode == HttpStatusCode.MovedPermanently ||
+                    serviceException.StatusCode == HttpStatusCode.PermanentRedirect)
+                {
+                    return null;
                 }
 
                 if (serviceException.StatusCode == HttpStatusCode.BadRequest)
@@ -123,6 +155,54 @@ namespace Amazon.S3.Internal
 
             return baseRetryForException(executionContext, exception);
         }
+
+        /// <summary>
+        /// Redirects the request to <paramref name="correctedRegion"/> after a bucket/region
+        /// mismatch is detected, so the retried request is both sent to and signed for that Region.
+        ///
+        /// The endpoint (not just <see cref="IRequest.AuthenticationRegion"/>) must be rewritten
+        /// because the S3 endpoint resolver runs once, before the retry loop, and is not
+        /// re-evaluated on retry; the same reason <see cref="Amazon.Runtime.Internal.RedirectHandler"/>
+        /// rewrites the endpoint when following a 307. Returns <c>false</c> if the Region is not
+        /// recognized by this SDK build, in which case the caller should not retry.
+        /// </summary>
+        internal static bool RedirectToRegion(Runtime.IExecutionContext executionContext, string correctedRegion)
+        {
+            RegionEndpoint correctedEndpoint;
+            try
+            {
+                correctedEndpoint = RegionEndpoint.GetBySystemName(correctedRegion);
+            }
+            catch (AmazonClientException)
+            {
+                // Unknown region on an older SDK build: leave the request untouched.
+                return false;
+            }
+
+            var requestContext = executionContext.RequestContext;
+
+            // Re-resolve the operation endpoint for the corrected Region using the pipeline's
+            // endpoint provider (preserves FIPS/dualstack/accelerate/path-style/ARN handling).
+            var parameters = new ServiceOperationEndpointParameters(requestContext.OriginalRequest, correctedEndpoint);
+            var endpoint = requestContext.ClientConfig.DetermineServiceOperationEndpoint(parameters);
+            requestContext.Request.Endpoint = new Uri(endpoint.URL);
+
+            // Drop the stale Host header so it is recomputed for the new endpoint on re-sign,
+            // mirroring AmazonS3RedirectHandler.FinalizeForRedirect (the 307 path).
+            if (requestContext.Request.Headers.ContainsKey(HeaderKeys.HostHeader))
+            {
+                requestContext.Request.Headers.Remove(HeaderKeys.HostHeader);
+            }
+
+            // Set the signing region and let the pipeline re-sign. Unlike AmazonS3RedirectHandler
+            // (which signs inline because it runs inner to the Signer), this runs in the retry
+            // policy -- outer to the Signer -- so the Signer re-runs on the retry. AlternateEndpoint
+            // is the primary signing-region signal (see AWS4Signer.DetermineSigningRegion).
+            requestContext.Request.AlternateEndpoint = correctedEndpoint;
+            requestContext.Request.AuthenticationRegion = correctedRegion;
+            requestContext.IsSigned = false;
+            return true;
+        }
     }
 
 
@@ -143,7 +223,13 @@ namespace Amazon.S3.Internal
         /// </summary>
         /// <param name="executionContext"></param>
         /// <param name="exception"></param>
-        /// <returns>a value if it can be determined, or null if the IO-bound calculations need to be done</returns>
+        /// <returns>
+        /// <c>true</c> or <c>false</c> when the retry decision can be made here; <c>null</c> when the
+        /// exception looks like a bucket/region mismatch and the decision must be deferred to the
+        /// asynchronous <see cref="AmazonS3RetryPolicy.SharedRetryForExceptionAsync"/> path, which
+        /// resolves the bucket's Region (from the x-amz-bucket-region response header when present,
+        /// otherwise via a HEAD bucket network call) before deciding whether to retry.
+        /// </returns>
         public bool? RetryForExceptionSync(Runtime.IExecutionContext executionContext, Exception exception)
         {
             return AmazonS3RetryPolicy.SharedRetryForExceptionSync(executionContext, exception, Logger, base.RetryForException);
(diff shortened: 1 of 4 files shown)


sdk/src/Services/S3/Custom/Internal/AmazonS3RetryPolicy.cs
```diff
@@ -71,6 +87,22 @@ public AmazonS3RetryPolicy(IClientConfig config) :
...
+                // bucket is in us-east-1 and the client is not (301 MovedPermanently), or in the
+                // rarer 308 PermanentRedirect case. These responses still carry the correct
+                // Region in the x-amz-bucket-region header, so treat them as inconclusive and let
+                // the caller detect the region mismatch and retry against the correct Region
+                // (mirroring the AWS CLI behavior). Without this, HeadBucket and other operations
+                // against a us-east-1 bucket from a non-us-east-1 client would surface the error
+                // instead of following the redirect.
+                // See https://github.com/aws/aws-tools-for-powershell/issues/413 (DOTNET-8539).
+                if (serviceException.StatusCode == HttpStatusCode.MovedPermanently ||
+                    serviceException.StatusCode == HttpStatusCode.PermanentRedirect)
```
dscpinheiro: A bit annoying but `HttpStatusCode.PermanentRedirect` only exists for newer versions of .NET... If you try to build (even on a non-Windows machine) it'll fail:
```bash
$ dotnet build sdk/test/UnitTests/AWSSDK.UnitTests.Core.csproj

  AWSSDK.S3.NetFramework net472 failed with 1 error(s) (9.2s)
    ./aws-sdk-net/sdk/src/Services/S3/Custom/Internal/AmazonS3RetryPolicy.cs(101,67): error CS0117: 'HttpStatusCode' does not contain a definition for 'PermanentRedirect'
```

We can do this instead (you will need to update the unit test too):
```csharp
// 308 PermanentRedirect: HttpStatusCode.PermanentRedirect is not available on
// net472/netstandard2.0, so compare the numeric value.
if (serviceException.StatusCode == HttpStatusCode.MovedPermanently || (int)serviceException.StatusCode == 308)
{
    return null;
}
```

sdk/src/Services/S3/Custom/Internal/AmazonS3RetryPolicy.cs
```diff
@@ -123,6 +155,54 @@ public AmazonS3RetryPolicy(IClientConfig config) :
...
+        internal static bool RedirectToRegion(Runtime.IExecutionContext executionContext, string correctedRegion)
+        {
+            RegionEndpoint correctedEndpoint;
+            try
+            {
+                correctedEndpoint = RegionEndpoint.GetBySystemName(correctedRegion);
+            }
+            catch (AmazonClientException)
+            {
+                // Unknown region on an older SDK build: leave the request untouched.
```
dscpinheiro: I think this comment is incorrect... `RegionEndpoint.GetBySystemName` will not throw for all unknown regions (it'll assume the `aws` partition if nothing matches but will return a value - it only throws for values that are invalid hostname labels).

If I understood this code correctly it's trying to protect against a malformed `x-amz-bucket-region` header value, right? If so, I'd mention that in the comment (and in the summary of this method).

dscpinheiro (commented):
PR looks good, a couple of minor comments and one general feedback (which is something that annoys me): AI tends to be very verbose when writing comments, and most of the time (IMO) it adds unnecessary details. 

If you ask your agent to make them more concise I think we'd still have enough detail about the problem / how we're fixing it.

sdk/src/Services/S3/Custom/Internal/AmazonS3RetryPolicy.cs
sichanyoo: Tightened the wording 👍

sdk/src/Services/S3/Custom/Internal/AmazonS3RetryPolicy.cs
sichanyoo: Changed as suggested.

sdk/src/Services/S3/Custom/Internal/AmazonS3RetryPolicy.cs
```diff
@@ -123,6 +144,45 @@ public AmazonS3RetryPolicy(IClientConfig config) :
...
             return baseRetryForException(executionContext, exception);
         }
+
+        /// <summary>
+        /// Redirects the request to <paramref name="correctedRegion"/> after a bucket/region
+        /// mismatch is detected, so the retried request is both sent to and signed for that Region.
+        /// Returns <c>false</c> when <paramref name="correctedRegion"/> (from the x-amz-bucket-region header)
+        /// is not a valid hostname component.
+        /// </summary>
+        internal static bool RedirectToRegion(Runtime.IExecutionContext executionContext, string correctedRegion)
```
peterrsongg: nit: you can drop the `Runtime` here, it's not needed since you already declared it in the using statement at the top

sdk/src/Services/S3/Custom/Internal/_async/AmazonS3RetryPolicy.cs
peterrsongg, line 55: this isn't introduced by your PR but I noticed in the _bcl version we do 
```csharp
if (executionContext.RequestContext.Identity is AWSCredentials && !(executionContext.RequestContext.Identity is AnonymousAWSCredentials))
```

can we do that here too? i think it makes sense for the two retry behaviors to be the same

sdk/src/Services/S3/Custom/Internal/AmazonS3RetryPolicy.cs
```diff
@@ -123,6 +144,45 @@ public AmazonS3RetryPolicy(IClientConfig config) :
...
+            requestContext.Request.Endpoint = new Uri(endpoint.URL);
+
+            // Drop the stale Host header so it is recomputed for the new endpoint on re-sign.
+            if (requestContext.Request.Headers.ContainsKey(HeaderKeys.HostHeader))
+            {
+                requestContext.Request.Headers.Remove(HeaderKeys.HostHeader);
+            }
+
+            // Set the signing region and let the pipeline re-sign on retry.
+            requestContext.Request.AlternateEndpoint = correctedEndpoint;
```
peterrsongg: You shouldn't have to assign AlternateEndpoint, since `DetermineSigningRegion` checks AuthenticationRegion too for signing and since you already assigned it above on lines 171 and 172

It doesn't really hurt anything, but it isn't doing anything here either

peterrsongg (changes requested):

sdk/src/Services/S3/Custom/Internal/AmazonS3RetryPolicy.cs
sichanyoo: Dropped it here, and in other places in this file that was pre-existing also.

sdk/src/Services/S3/Custom/Internal/_async/AmazonS3RetryPolicy.cs
sichanyoo: Added the check as suggested.

sdk/src/Services/S3/Custom/Internal/AmazonS3RetryPolicy.cs
sichanyoo: Removed unncessary setter.

peterrsongg (dismissed):

sdk/src/Services/S3/Custom/Internal/_async/AmazonS3RetryPolicy.cs
```diff
@@ -62,10 +65,12 @@ internal static async Task<bool> SharedRetryForExceptionAsync(Runtime.IExecution
                 }
                 else
                 {
-                    // change authentication region of request and signal the handler to sign again with the new region
-                    executionContext.RequestContext.Request.AuthenticationRegion = correctedRegion;
-                    executionContext.RequestContext.IsSigned = false;
-                    return true;
+                    // Redirect the retried request to the bucket's actual Region (endpoint + signing).
```
dscpinheiro: > Same comment applies to the BCL retry policy. 

Should we check for `executionContext.RequestContext.OriginalRequest is HeadBucketRequest` here too? The 301/308 check is gated, but the existing 400 path (`AuthorizationHeaderMalformed` / `x-amz-bucket-region`) returns `null` for every operation, so `RedirectToRegion` currently would run for other operations (such as `GetObject`) as well.

An example of a test that fails right now but it should pass:
```csharp
[TestMethod]
public async Task RegionMismatch_ForNonHeadBucketOperation_DoesNotRedirectEndpoint()
{
    var endpoint = new Uri("https://" + BucketName + ".s3.us-west-1.amazonaws.com");
    var context = CreateContext(new GetObjectRequest { BucketName = BucketName, Key = "test.txt" }, endpoint);

    await AmazonS3RetryPolicy.SharedRetryForExceptionAsync(context, CreateS3Exception(HttpStatusCode.BadRequest, BucketActualRegion), (_, __) => null, (_, __) => false);
    Assert.AreEqual(endpoint.Host, context.RequestContext.Request.Endpoint.Host, "Only HeadBucket may be redirected to the bucket's Region.");
}
```

sdk/test/Services/S3/UnitTests/Custom/HeadBucketRegionRedirectTests.cs
```diff
@@ -0,0 +1,213 @@
...
+using Amazon.S3.Util;
+using Amazon.Util;
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+
+namespace AWSSDK.UnitTests
+{
+    /// <summary>
+    /// Verifies that a cross-Region HeadBucket (and any other S3 operation) issued against a
```
dscpinheiro: This comment still says `(and any other S3 operation)`

sdk/src/Services/S3/Custom/Internal/_async/AmazonS3RetryPolicy.cs
sichanyoo: ~~Doing that for 400 path might make previously succeeding case (through retry) into error case which I don't know if we should risk and/or include in this PR.~~

Added gate as suggested.

sdk/test/Services/S3/UnitTests/Custom/HeadBucketRegionRedirectTests.cs
sichanyoo: Oops, good catch; updated it.

dscpinheiro (approved):

normj (approved):

